PR #4, opened 2018-12-13
Neutrino arrival generation

daniel-zid:
Neutrino zenith and azimuth angles can be specified.
The effective volume is stored in the hdf5 output file.


daniel-zid:
Yeah, I had already implemented that but I had a problem (I'm new to git)
and I had to rewrite all my changes. I forgot to save the angles.

Ok, I'll delete the Veff in the output.

Le jeu. 13 déc. 2018 à 19:15, Christian Glaser <notifications@github.com> a
écrit :

PR #6, opened 2018-12-14
Tau event generator

cg-laser:
add taus to the event generator.


daniel-zid:
I've written an alternative version that creates files with a fixed number of events. It seems to work just fine. Do you want me to commit it here or do I create a branch? @cg-laser

cg-laser:
With 'same number of events per file' you mean that the files can have different number of entries? This potentially leads to large variation between the requested entries per file and what will be in the file in the end... We should also think about more exotic models where we can have many vertices per event. 

I think its better to create a new branch (from this one) so that we can compare the code.

cg-laser:
@daniel-zid I had a look at your alternative code to split up the data into several files. I don't have a strong opinion about it. My code (this pull request) will lead to more similar entries in each file, while your code has the exact same number of events in each file but possibly different number of entries. Maybe your approach is the better one, it might be easier for the users if all sub files have the same number of events. Although this functionality is is mostly an intermediate step.. You have green light from me to change this branch with your proposed changes. Although I'm not sure how to merge a branch with a branch...

cg-laser:
Another thing that we need to consider is that the tau vertices might be displaced by a lot and will never trigger the detector if we only simulate one station. On the other hand, there might be first vertices outside of our simulation volume that might produce second vertices (tau decay) inside our fiducial volume. Can we come up with a smart way to include these events while keeping the statistics correct for the Veff calculation? 
The brute force way would be to have a really large simulation volume, but then we waste a lot of computing time and might need to consider the Earth curvature.

daniel-zid:
Maybe we should increase the volume for taus, save every tau we can, but
perform
the ray tracing and the calculation only for taus satisfying some
geometrical criteria?

cg-laser:
I think this pull request is ready to be merged with the master to test tau simulations in conjunction with the new ARZ module

daniel-zid:
I really like this way better! The only thing that should be changed in the
effective volume scripts would be to check the n_interaction so as not to
count twice the same event. But in this version there is no Veff scripts.
Could we make a pull rebase from master to this branch?

daniel-zid:
I was thinking that we don't store the time between the tau neutrino interactions and the tau decay. This is not too relevant in principle for the volume calculations, but it can be important for simulating the electric field and reconstructing if both bangs are seen.

Which leads me to the other issue: should I implement the tau energy losses to know the actual energy of the tau when it decays? Should I track the tau position as a function of time? It would be nice to have them, but is it a priority right now? What do you think?

cg-laser:
notauregen_flavorcomparison

@persic  did a study with ShelfMC showing the impact of simulating the tau second band with shelfMC for Moore's Bay. The effect is pretty small and only relevant at low energies. However, this might be different for the SP as there is more ice than just the 'narrow' ice shelf of 500m. Still it might not be that important at the moment. I actually saw that for ~100 events in the fiducial volume I only get 2 additional events from tau decays. But  my simulation volume might not have been large enough. 

Storing extra attributes is a little difficult in the current scheme. We can of course add a new data set with travel times, but most events will just have Nones in it... 

Energy losses seem like a good idea, depending on how difficult it is to implement them.

cg-laser:
I'll merge this branch then later today to be able to work on the Veff script. I think it should already work because we correctly keep track of the total number of events.

PR #7, opened 2018-12-14
Add ARZ semi MC signal model

cg-laser:
add time domain signal semi MC signal model. This model calculates the Askaryan signal in the time domain via integration along the charge-excess profile. The CE profile is obtained from a library of ZHS showers. 
So far we have a preliminary library of EM and HAD showers. TAU showers are coming soon. 

The module still needs to be integrated into NuRadioMC but most technical things are already solved. Two versions to calculate the vector potential exists. One is a one-to-one python transcription of the original fortran code from Jaime. The second one is a faster implementation that makes use of numpy vectorization and produced the exact same results. It is almost a factor of x1000 faster which reduces the computing time to less than 100ms which is sufficiently good for now.


daniel-zid:
Hmmm... Not necessarily. Keep in mind that the geometrical form factor and
the distance
carry also a lot of weight in the calculation, and they will be affected by
the interpolation as
well. Anyway, let's give it a try.

daniel-zid:
I've modified the fast vector potential calculations and now it works better. There is a factor named interp_factor that controls the precision of the method. The larger the factor, the more precise the method is. I am following Christian's idea of using the trapezoid rule, but with a larger number of points, and in fact, for very close distances (10 m), the trapezoid rule with an interp_factor of 100 is faster and more accurate than the integration using quad! This is because the integration range for the quad function is too wide, and as a result the quadratic adaptative algorithm yields some wild peaks. The solution would be to divide the interval in several subintervals to feed the quad function with. However, it is better to use the trapezoid rule.

We should now find a criterion for the interp_factor but it should not be too hard. 10 m away from the shower it can be equal to 100 (or maybe less) and then it should go down with the distance.

cg-laser:
great! This looks very useful to me. 
Can you provide more information on how the speed changes with interpolation factor? 
And maybe add an example (plot) on how the precision (the vector potential) changes with interpolation factor?

From current simulations we see that most vertices are relatively far away (at least a couple of hundred meters). Hence, the code should be optimized in terms of speed for those events, but we also want good accuracy for the 'near' events but these special cases wouldn't be so time critical. 

BTW: Near vertices for a surface station are difficult anyways because the interaction takes place in the firn with varying index of refraction, whereas all ZHS simulations are done assuming a constant n=1.78.

daniel-zid:
I've modified the fast vector potential calculations and now it works better. There is a factor named interp_factor that controls the precision of the method. The larger the factor, the more precise the method is. I am following Christian's idea of using the trapezoid rule, but with a larger number of points, and in fact, for very close distances (10 m), the trapezoid rule with an interp_factor of 100 is faster and more accurate than the integration using quad! This is because the integration range for the quad function is too wide, and as a result the quadratic adaptative algorithm yields some wild peaks. The solution would be to divide the interval in several subintervals to feed the quad function with. However, it is better to use the trapezoid rule.

We should now find a criterion for the interp_factor but it should not be too hard. 10 m away from the shower it can be equal to 100 (or maybe less) and then it should go down with the distance.

daniel-zid:
You can find here a comparison of the z component of the T02 file vector potential and the relative error for the interpolation. We can see that at 100 m, a factor of 5 yields an accuracy of 0.01 (1%).
[Potential_100_z.pdf](https://github.com/nu-radio/NuRadioMC/files/2703612/Potential_100_z.pdf)
[Diffs_100_z.pdf](https://github.com/nu-radio/NuRadioMC/files/2703613/Diffs_100_z.pdf)

daniel-zid:
Relevant: https://www.johndcook.com/blog/2010/12/02/three-surprises-with-the-trapezoid-rule/

daniel-zid:
What is the typical distance? I'd say that for events 100 m away, we need
additional interpolation if we want a 1% accuracy. Besides, no extra
interpolation creates some nasty harmonics. See plots for 100 m

Le ven. 21 déc. 2018 à 19:27, Christian Glaser <notifications@github.com> a
écrit :

daniel-zid:
The machines in Zeuthen are quite slow. That is why the time values may
seem a bit off to you. Basically, a factor of 1 is as fast as your previous
implementation. A factor of N is N times slower.

cg-laser:
Vertex distribution for 1e18eV neutrinos (with a 15m deep 2 sigma trigger)
vertex_distribution

Vertex distribution for 1e16.5eV neutrinos (with a 15m deep 2 sigma trigger)
vertex_distribution

Even for very low energies, almost all events are at least 100m away. So I don't think we need to worry too much about 10m. But your implementation with the interpolation is nice, so we can just increase the precision if the shower is too near. However, these events are anyway 'wrong' because the showers are simulated for different index of refraction and a changing index of refraction over the shower dimensions are not taken into account. 

The oszillation in the vector potential is interesting, do you know where it comes from?

daniel-zid:
Well, I think it is relevant because the RA_c is parameterised as an
exponential function,
and double exponentials seem to get along quite well with the trapezoid
rule. I mean, the
quad method is quite clever and with just a few points and sampled
functions you can get
an insanely good estimation of most integrals! I think that our integrand
has some places
that are quite difficult to evaluate with the quadrature method and a
simple trapezoid rule
works better. We could try as well Simpson's rule or something simple like
that to see if
speed is boosted more.

That oscillation is just a numerical error. They tend to do things like
that even when they
converge. I couldn't tell you the exact mathematical origin right now,
though...

cg-laser:
open issue: in the ARZ code the coordinate system is centered at the particle injection. We set the observer position to have the specified viewing angle with respect to (0,0,0). However, shouldn't the point of reference be the shower maximum? If the shower size is small against the observation distance this is probably negligible but we should think how to do this correct in general.

cg-laser:
open issue 2: The eR component is not exactly zero. How do we deal with it. Is this really physical? Or a consequence of the first issue? 
We can't process it any further in the detector description because the antenna response is not defined for the eR direction. 
However, it seems that the eR component is negligible wrt the to  eTheta component, so we could just ignore it.

cg-laser:
I did a few test with spurious results. `tests/T02TestARZ.py` calculates the askaryan signal once with the xmax shift and once without. I find substantial differences between the two cases. Although the distance is 5km and the xmax shift is only 17m, the resulting efield amplitude changed by 60%. This seems way too much and looks like a bug. Any thoughts? 
Or is the signal that sensitive on the viewing angle?

daniel-zid:
Not necessarily, if the approximations of ARZ are met (and they are), the
resulting electric field is independent of the coordinate system used to
describe the shower as long as the Taylor expansion is made around the
axis. You are going to get the same physical field. You would be just
shifting the variable of integration for the charge profile. The change of
angle is taken into account in the equations of ARZ. See the section where
they derive the 'near field', which is the formalism we use.

The electric field does not change, we just want to refer it to a point in
the shower to apply our antenna response for a single incoming direction.
We want to make the approximation that all the field comes from a point, so
that's why we want to use Xmax. The problem lies in this assumption and not
in the calculation of the field.

I think the user wants the electric field for a specific antenna position
and know where it comes from (viewing angle), not the other way around. The
internal theta we use for emission is irrelevant. The theta for reception
is what matters, and that we can calculate.

O 23:25, Mar., 22 de xan. de 2019, Christian Glaser <
notifications@github.com> escribiu:

daniel-zid:
Maybe the coordinate system cannot be changed like that without changing
some of the equations or the intervals for the integral. Once again, the
field should be independent of the choice of reference point. Maybe there
is some function you have to change as well. I'll take a look tomorrow

O 00:37, Mér., 23 de xan. de 2019, Christian Glaser <
notifications@github.com> escribiu:

cg-laser:
well the first thing that we do inaccurate is to do the ray tracing from 
the vertex position to the observer. However, we could just redefine the 
vertex position as the position of best point source approximation which 
would be the shower maximum. The other Askaryan modules also essentially 
calculate the emission for the vertex position (I guess).

We can't feed the observer position directly into the ARZ code because 
of the ray bending. Therefore we calculate the launch angle/viewing 
angle and then just place a virtual observer position assuming no ray 
bending for the ARZ code.

Am 22.01.19 um 16:28 schrieb daniel-zid:

cg-laser:
One place where the definition of 'theta' becomes relevant is the comparison with previous model. The following plots show a comparison of ARZ (blue solid curve) with the Alvares2000 model (frequency domain parametrization, phases set to +90deg)

04_1eev_had
04_1eev_em
04_10pev_em

1EeV EM showers create a fuzzy efield in the ARZ code, likely because of the LPM elongation and corresponding spatially displaced sub showers. I needed to use some interpolation (20x here) to get a smooth pulse for this case.

cg-laser:
I think the module is ready to merge. Two improvements should be done wrt to computing speed in the future:
- improve speedup by calculating only the relevant time bins
- automatically determine the necessary interpolation factor

cg-laser:
@daniel-zid can you approve this pull request so that I can merge it into the master branch? I already resolved all conflicts.

daniel-zid:
Sure! As soon as I get to the office.

O 00:02, Xov., 7 de feb. de 2019, Christian Glaser <notifications@github.com>
escribiu:

PR #15, opened 2019-02-07
Code for tau photonuclear losses.

daniel-zid:
I've created a code for the tau track length including photonuclear losses. The only problem I see is that it is too slow as of now. However, I have already been able to extract some physics. Using the EvtGen/test/T03_tau_length.py file, we can obtain the track length for the mean tau lifetime with and without losses (see figure). You can also see the decay energy as a function of the tau energy (it tends to saturate!). The problem is that the decay time is, of course, random, so we cannot parameterise it so easily. I have to come up with something.

figure_1

It is not ready to merge, as I said we need to think of a smart way of speeding the calculation up.


daniel-zid:
We can see here that the decay energy curve mimics the track length curve:

figure_1

cg-laser:
Interesting, looks a little like GZK losses of cosmic-rays. 

I didn't really follow how you take the energy losses into account. Maybe you could document the physics in the NuRadioMC paper draft?

What exactly do you mean with slow? So far the event generation is only a tiny tiny fraction of the NuRadioMC simulation time, i.e., event if it might be significantly slower, it still might be fast enough.

daniel-zid:
Yeah, sure. I will try to document it this afternoon.

By slow I mean that if you want to generate 1 million taus it can take
several minutes to produce a single file, I think. I haven't had the time
to check it. The integration of the energy losses for knowing the times and
the solving of the proper time equation take a lot of time. For my Gen2
analysis, I would need 50 energies x 20 zenith bands x 1 million events, so
I would need to use the cluster for generating the events only! If we
increase the fiducial volume, I would need even more than one million
events for each energy and zenith band. I don't feel too comfortable with
this, but maybe we can think of something...

daniel-zid:
Fuck, I've forgot to change the condition back! One sec

O 19:03, Ven., 8 de feb. de 2019, Christian Glaser <notifications@github.com>
escribiu:

daniel-zid:
There it is. Changed.

daniel-zid:
Done. More interactions than 2!

cg-laser:
A thought about possible speedup: 
The function E(t) is always the same right? So we can parameterize it once and just store the parameterization. A convenient method to interpolate arbitrary functions is a spline interpolation. 

And we can also precompute the tau decay length. It would be a 2D histogram 'decay length' vs. 'tau energy' where we store the decay length distribution per tau energy. 

The fast code would then just draw a tau track length from the precomputed distribution and use the E(t) function to get the tau decay energy. 

Does this sound like a reasonable approach?

daniel-zid:
I am not sure. What do you mean with E(t) is always the same? The decay energy of the tau depends on the tau initial energy and the decay time (in rest frame, to simplify things). Therefore, I think that the best way would be to have 2D hists of both the decay length and the decay energy as a function of the initial tau energy and the decay time. That should be easy. For our purposes, nearest-neighbour interpolation is enough but we can use a 2D linear interpolation as well. What format would you want me to use to save the histograms?

cg-laser:
but aren't the decay energy and time correlated? So choosing both randomly and independent of each other will lead to wrong results? I thought that maybe the function E(t) is a function that is the same for all tau energies and decay times which we can just parameterize, but maybe I'm wrong. 

I think a 2D numpy array is the best way to store the histograms.

daniel-zid:
The issue is that you need the initial energy as well, so Efinal =
Efinal(Einitial, tau_time). We cannot create a unique Efinal(tau_time),
since the energy loss is energy dependent.

O 18:46, Ven., 1 de mar. de 2019, Christian Glaser <notifications@github.com>
escribiu:

PR #17, opened 2019-02-08
Arz speedup

cg-laser:
ARZ model is currently too slow to be usable in Veff simulation. It slows down the simulation by ~ a factor of 100. 

There are two options to speed up the processing: 
1) calculate signal only for relevant time bins:
   the ARZ code internally computes only +-5 ns around the pulse peak. I wonder if this is sufficient for off-cone configurations, and if it can be further tuned for on-cone configurations. 

2) the interpolation factor of the CE profile. 
  @daniel-zid suggested a scaling of the interpolation factor with distance. I implemented a logaritmic scaling factor = 10**(3- log10(R/1m)) which gives 10x at 100m and 1x at 1km. However, it also depends on the viewing angle (see following examples):
1km away, 10x interpolation
intp_10_had19

1km away 1x (no) interpolation
intp_1_had19

we get substantial numerical noise for off-cone viewing angles. How can we solve it?


cg-laser:
the oscillation for small interpolation factors looks to me like a problem in the numerical integration that should be solvable by a more clever method. Any thoughts?

daniel-zid:
We could try Simpson's rule or another Newton-Cotes' formula. But keep in
mind that going up that ladder will get us back to quadrature methods,
which were slower (quad function). I have the impression that Simpson's
rule may work a bit better, but that means more evaluations of the
function, which will probably mean more computation time. Maybe
interpolating more is just faster?

daniel-zid:
I read on the scipy doc that you can use the function simps for Simpson's
rule and romb for Romberg's method. Could you try them for making your
plots and see if the precision or the times improve?

clark2668:
I'm following this only in cursory fashion, and I'm certainly not an expert on numerical integration, but if you want to hang on to a quadrature technique, and want it to be faster, we might consider another Cython wrapped C++ algorithm with a numerical package like GSL. Just a thought.

cg-laser:
@clark2668 do you have a concrete library in mind? 

I probably should provide some plots... This is the integrand. 
integrand

The problem is the the integrand is very peaked and the peak is not sampled good enough. Hence, I think we don't win a lot with a better numerical method if we don't interpolate the relevant part a little more. 

The result is a slight oszillation in the vector potential (left panel) which then leads to a completely wrong efield:

vector_pot_efield

I think the way to go is to not interpolate the full CE profile but only the relevant part around the peak.

clark2668:
@cg-laser, the integrand looks really well behaved, I'm surprised the interpolation has so many oscillations... What interpolation scheme are you using?

I was thinking about something like the GSL qags routine? The documentation says it has adaptive sub-interval sizes to deal with discontinuities like the cusp we're seeing in the integrand.
https://www.gnu.org/software/gsl/doc/html/integration.html#qags-adaptive-integration-with-singularities

daniel-zid:
We already used python's quad method, similar to qags, and it works fine.
The problem is that it is too slow. The trapezoid rule with interpolation
is much faster and precise for this function, but you need a sufficiently
fine interpolation.

Qags or qag will work well, but I'm not sure the wrapped qags will be
faster than the python trapz. Should we try?

O 18:59, Mar., 12 de feb. de 2019, Brian Clark <notifications@github.com>
escribiu:

clark2668:
Not sure... In the ray tracing code, even with the cython wrapping, we got back about a factor of 20 in time savings. But that's a different problem, so the win may not be as considerable.

cg-laser:
good news: I found a way to increase the accuracy without a significant loss of speed. As I wrote earlier, we only need the interpolation (aka upsampling aka finer sampling) in the region of the peak in the integrant. So I just sample the profile finer in the regions that correspond to t_obs in [-1ns,1ns]. This is controlled via a second interpolation factor 'interpolation2'. These are the speed results:

no interpolation, 10 showers = 1.20s
10x interpolation, 10 showers = 3.66s
100x interpolation, 10 showers = 29.14s
10x interpolation2, 10 showers = 1.54s
100x interpolation2, 10 showers = 2.26s

Hence, with 100x interpolation the code is just a factor of 2 slower! We can probably tune it further by time range, or by downsampling the profile in the first place (interpolation factor smaller than 1). Although the latter seem to have almost no effect:

0.1x interpolation, 1000x interpolation2, 10 showers = 2.31s
0.1x interpolation, 100x interpolation2, 10 showers = 1.22s

What is still TBD is to check what interpolation factors are required.

daniel-zid:
I've made several plots, but I think these two summarise what I want to say well.
With an interp1 of 1 and an interp2 of 10, the accuracy is better than 1% at 100 m
and of course it is better at larger distances. If the CPU time is acceptable, I suggest
interp_factor = 1 and interp_factor2 = 10 for all our purposes. If we want to save a
bit more, we can decrease the interp_factor2 so that it becomes 1 at 1 km.

What you see in the plots as interp_factor is actually interp_factor 2, the original interp_factor is always 1

By the way, the GitHub system for answering through gmail is a bit of a shit.

[Diffs_interp2_0_100_z.pdf](https://github.com/nu-radio/NuRadioMC/files/2861583/Diffs_interp2_0_100_z.pdf)
[Potential_interp2_0_100_z.pdf](https://github.com/nu-radio/NuRadioMC/files/2861584/Potential_interp2_0_100_z.pdf)

daniel-zid:
This plot has been made for the Cherenkov angle. Differences are more accused far from the Cherenkov angle. Although the precision rests the same for interp_factor2 = 10, I don't know how the oscillations are going to affect the spectrum. If we can, I would recommend to increase interp_factor2 as much as the computation time allows us.

This plot is for an angle of 65 deg, R = 100 m

[Diffs_interp2_0_101_z.pdf](https://github.com/nu-radio/NuRadioMC/files/2861604/Diffs_interp2_0_101_z.pdf)
[Potential_interp2_0_101_z.pdf](https://github.com/nu-radio/NuRadioMC/files/2861605/Potential_interp2_0_101_z.pdf)

cg-laser:
Is this on the cherenkov cone? I saw that it works fine on the Cherenkov cone but we get larger errors off-cone. 

I also think we should check the electric field, because little fluctuations in the vector potential get blown up when you take the derivative.

daniel-zid:
The first one is for the Cherenkov cone, and the second one is for theta_cherenkov + 10 deg. Yes, I should take a look at the electric fields. You've done the calculation, so what would you think it would be an acceptable interp_factor2 if interp_factor1 = 1?

cg-laser:
with intp2 =100, we can simulate 10,000 events in roughly 10h. This is managable, but a factor 20 slower than before. 

I think it would be best if we came up with some kind of parameterization of the interpolation factor (and the time interval) as a function of distance and viewing angle

cg-laser:
So far I didn't see a problem with inp2 = 100

daniel-zid:
I have included a new code that calculates the mean error between a very fine interpolation and another with smaller interpolation factors. I stop when the mean error between the two calculations in a window of 40 ns is smaller than 0.1. In practice, this translates into a precision of more than 10% for the region of interest. You can execute the T04 code with different distances (careful, it takes a looong time) to convince yourself. Here are the results for both interpolation factors as a function of the zenith angles and for several distances:
interpolation_plot
It does not seem very regular at all. Suggestions? More angles? A different kind of analysis? Can we just take the largest interp_factors just to be sure?

daniel-zid:
Narrowing the window and changing the criterion (it was too strict before), I arrive to the following result. Interp_1 can be left equal to 1, but I don't know if we should parameterise interp_2 or just set it to 30. The criterion I used is more than enough for triggering purposes, and should always give a time trace with less error than 5%. I'll commit the code in a while.
interpolation_plot_1e-3

PR #23, opened 2019-02-27
calculate improved start value for C_0

cg-laser:
This improved start value calculation for the minimization might solve the problem reported by Robert, that sometimes no solution is found.

### DIFF
--- a/NuRadioMC/SignalProp/analyticraytraycing.py
+++ b/NuRadioMC/SignalProp/analyticraytraycing.py
@@ -80,7 +80,7 @@ class ray_tracing_2D():
         transforms the fit parameter C_0 so that the likelihood looks better
         """
         return np.exp(logC0) + 1. / self.medium.n_ice
-
+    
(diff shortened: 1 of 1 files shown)


cg-laser:
actually it turned out to be not a good idea and Robert is working on a better solution.

cg-laser:
regarding the workflow, maybe it would be better if the person who initiated the pull request does the merging. Sometimes there are also more changes to come that the reviewing person is not aware of.

anelles:
I agree and disagree with this.
Yes, people should merge themselves, but don't create a pull request, if you are not done with the changes. Make a branch, put all changes in and then create a pull request. Don't request a pull on something that is not finished.
If you find out that the pull request was too early, close it again.

cg-laser:
I see the pull request pages also as a discussion forum of possible code extensions. It allows to discuss possible code changes, and it gives others a quick way to inspect the changes. 

In this particular case I discussed it with Robert offline and we didn't add our conclusions to the pull request page...

PR #30, opened 2019-03-04
add example which calculates the double station coincidence rate

cg-laser:
this is example 3 of the NuRadioMC paper


cg-laser:
there is a wiki page documenting this example in detail: https://github.com/nu-radio/NuRadioMC/wiki/Example:-Multi-station-coincidences

Adding another script which does the pre simulation might be useful. I anyway wanted to do it as I'm uncertain about the results. Currently I'm using some old simulations from 6 month ago.

anelles:
Looks much better. I suggest that you announce early on that you want to talk about this on the call on Thursday and that people should check out the code if they are interested. 
I suggest to then push this commit after the discussion.

cg-laser:
This is the station configuration I'm simulating: 
layout
These are the preliminary results I get:
coincidences

cg-laser:
Updated results with new 'pre simulation'. Results didn't really change but have more statistics. 
coincidences

A likely explanation for why a deep station is more effected (dashed line) is that we have two chances of hitting the second station because of the direct and reflected ray. For a shallow station, the difference in viewing angle for the direct and reflected ray is small. 

I also added a visualization script for the 3D geometry of the event. The following plots are for a surface station at 1e18 eV neutrino energy. 
example_01
example_02
example_03

PR #41, opened 2019-03-27
add wrapper around propagation module to make it configurable

cg-laser:


cg-laser:
this pull request adds a wrapper around the propagation class to allow for a future extension with different modules. 
It also adds a unit test that performs a full NuRadioMC simulation and checks if the results are consistent with a reference reco.

cg-laser:
the recent commits are just to prepare merging with the master branch

cg-laser:
yes it works. This commit introduces a validation test (as mentioned above). So it is easy to test any changes. Just run `./validate` in the test directory.

PR #43, opened 2019-03-27
Raytracing rl

relamber:
Start value for find_solutions() is now the C_0 value corresponding to the surface skimming angle (method get_surf_skim_angle() to find this angle was added). This starting value worked for all emitter/receiver configurations that I used, but in order to come from the angle to the corresponding C_0, a root finding method is employed, which is potentially time consuming.
A number of new methods have been implemented, mostly helper functions. Their names should give a rather clear indication what they are doing and an effort has been made to document them comprehensively.


anelles:
Has this gone dormant? The branch is in conflict now. @relamber are you still working on this?

relamber:
Hi Anna,

I am not currently working on the code, but I am using it and I think the changesI made are useful, so I think they should be considered. I have also added a few new routines, but those should not create conflicts.

Best regards,
Robert

cg-laser:
I think there are still problems with this addition. Not that it is not working but it creates some inconsistencies: 
1) Only the python part is altered, hence, the C++ implementation is not a 1 to 1 replacement anymore. We should avoid that
2) If I understood correctly, an additional minimization is performed that potentially slows down the code. 

An alternative option to address the problem that sometimes no solution is found is to don't cut the objective function by setting impossible solutions to a delta_y = infinity by return the actual value. This should make the objective function smooth and will probably result in a better finding of the minimum without additional computing costs.

anelles:
I assumed that Robert and you had worked out the details and I did the annoying part of merging manually. 
Shall we revert this again and leave this open? But then someone needs to work on this. Otherwise, it will just become stale.

cg-laser:
What about making this additional setting of the initial conditions optional (switchable via the init function) and the default is to not use it? Then we have all the nice additional functions in but don't change the default behaviour until further investigation.

relamber:
Hi Anna,

took me a while to remember the details of the code - Christian and myself discussed the matter a while ago and I made the pull request for further discussion, but then it fell into oblivion.

I have tried a number of starting values and the only one which (so far) has always found a solution when there was one is the procedure I have implemented. Yes, the procedure requires an additional minimisation algorithm which will slow things down, but I have not quantified this. I have also not quantified how often the original starting value did not find a solution when it should have.

So if our emphasis is on speed, then we should not implement my change. We can implement the starting value as an option, as Christian suggested, but in reality what will happen is that most likely everybody (except for myself) will use the code with the default options and the matter will not be investigated further. So for simplicity I suggest we just comment out my change and everybody who want to use it (i.e. me) can make the changes in the local version. But I’m fine with either way, I just would not like to be the one to implement the switch in the init function as it has been a while since I looked into the matter and I’d like to avoid creating another inconsistency.

Best regards,
Robert

cg-laser:
ok I can do that

cg-laser:
see https://github.com/nu-radio/NuRadioMC/pull/72 and continue discussion there

PR #44, opened 2019-03-28
Cross sections

anelles:
In order to avoid the usage of different cross-sections throughout the code, there is now one utility file for all. So far, all implemented cross-sections agreed. 
@daniel-zid can you check whether this works as desired. I have checked that both implementations gave the same results for a simple example. 
I think the unit tests will be available soon (see branch test), but for now, I cannot chekc.


daniel-zid:
The code seems alright to me. I'm going to create a separate file for calculating random inelasticities.

PR #53, opened 2019-04-04
Bugfix  cpp path length

clark2668:
This addresses a bug in the CPP based path length finder where it will fail for some integrals. This is addressed by adaptively increasing the size of the allowed relative error within some bound.

I also include a study of the speed of the CPP speed of the ray tracer in the `SignalProp/tests` directory.


cg-laser:
why did you put a copy (I guess) of the cpp file into the T04 directory? Can't we directly use the original file? I think we should try to avoid code replication.

clark2668:
Yeah, we could, and I agree it would be better to avoid code duplication. I was being lazy. Let me see if I can do something more intelligent...

anelles:
Just a little ping: @clark2668 have you been able to find a solution to this?

clark2668:
Hey @anelles, no, I haven't found time to address this. I might revise the commit stripping out the speed test just to get the bugfix implemented.

anelles:
Great @clark2668. I am just checking all pull-requests periodically to not have them go out of sync for too long. 
I would indeed prioritize the bug fix, if you don't have time to work on more elegant code.

clark2668:
Thanks @anelles for doing that! I think these mods are isolated enough to hopefully not cause major problems. 

And, I think this is ready to go now. I removed the speed test code for now.

clark2668:
@cg-laser--done! :)

cg-laser:
looks good to me. I didn't test it but I assume that you tested that it compiles and works.

clark2668:
The change to the path length integral was tested pretty thoroughly back when I made the bugfix. I didn't test the NAN very aggressively (it compiles and runs, etc), but I'm assuming that change is minor enough to not cause problems. I can exercise it more thoroughly if you think that's important though.

cg-laser:
I think it is fine. Thanks a lot Brian for fixing this!

PR #56, opened 2019-04-09
Update example figure

anelles:
As per my ToDo for the documentation paper. The figure is now changed to have a square format.


anelles:
This is non-critical but super annoying not to have. I will merge without review.

PR #58, opened 2019-04-16
update doc strings, remove pyrex wrapper to time domain model

cg-laser:


cg-laser:
thanks for spotting that. It seems that my editor has problems with these characters. Or it was just because I copied updated references from a paper which didn't do it correctly.

PR #59, opened 2019-04-16
update example

cg-laser:


cg-laser:
this pull request also adds a new example on how to read ari files (still need to be tested)

PR #60, opened 2019-04-19
add looping over stations

cg-laser:
This pull request modifies that output data structure (HDF5). Previously, the user needed to specify the station_id that was simulated. Now, all stations specified in the detector description are simulated automatically. This gives more flexibility and allows for the simulation of large arrays.


cg-laser:
FYI: this pull request does not take into account the absolute position of stations. This will be fixed in another pull request on a spice core example

cg-laser:
this pull request is integrated into the 'example-spice' branch, new pull request will be opened soon

PR #70, opened 2019-05-13
Fix examples

anelles:
Examples didn't work anymore. Mostly smaller fixes. 
When changing default values, please always check the examples.


anelles:
The DnR example still doesn't work, because grouping in the hdf5 file is different now.

anelles:
Major parts have already been addressed in #68 
If everything is there also, then we can close this one again. 
Can I request giving things a bit more descriptive names that it is clear from the title what pull requests are doing.

PR #76, opened 2019-05-16
pass 'a' argument to Hanson module

cg-laser:


daniel-zid:
With the current implementation, we no longer need the hanson2017.py file.

I've corrected the energy for the LPM routine, put the ZHS factor in (Christian, remove it if my interpretation is wrong). Agreement is much better for hadronic showers and electromagnetic showers seem to work ok... until the LPM factor is turned on and the code says that the shower width is 50 m. Which might as well be the case for the actual shower, but definitely NOT for the electrodynamic emission.

If anyone is going to use this model, I strongly advise against the use of the LPM elongation.

cg-laser:
from a computational point of view it would still be good to have in in Python only.

daniel-zid:
Well, sure. But writing it is going to take some time, and I don't think it's worth wasting much time when we have an implementation that works quite close to his. Or, at least, not until Jordan answers. 

In other words, if the author of the method does not find time for pointing out how it should be done correctly, why should I spend a lot of time translating the code? What advantages does this code offer over Alvarez2009?

daniel-zid:
Unless we have everything we need in hanson2017.py file. In that case, you can give it a try with the new calculation for the shower width.

cg-laser:
agreed, I just wanted to opt against removing this file, as we might need it in the future. 
And it is cumbersome for users to first compile something.

daniel-zid:
Yes, I agree. Can you adapt the hanson2017.py to see if it gives the same results from the C++ code? I see that it is quite advanced already :D

Issue #65, opened 2019-04-25
synchronization between NuRadioMC and NuRadioReco


cg-laser:
Some changes in NuRadioMC require changes in NuRadioReco, e.g. saving of additional parameters in the parameter storage. How do we synchronize these changes? 
Can we automatically check for a NuRadioReco version? But right now we don't really habe a versioning. 
Can we read out the git commit status? And give a warning or stop if NuRadioReco is not at least at a certain commit level?

anelles:
I think versioning is in principle good pratice anyway, the question is when to start this. Are we at a stable enough state to start this now?

cg-laser:
I investigated versioning a bit more. One easy way is to store the unique status of the git repo (hash of first line of `git show`). 
In addition we should start thinking about proper versioning. See (https://git-scm.com/book/en/v2/Git-Basics-Tagging) for how to tag versions. One policy could be that we always increment minor versions with each pull commit to master (including a descriptive message). In this way one could get a quick overview of what changed between two versions.

christophwelling:
I think the main problem will be how to tell if a change in NuRadioReco affects NuRadioMC and vice versa. Once we have the framework properly implemented, most commits in NuRadioReco will probably not affect NuRadioMC, so specifying the commit is probably overkill. But I can't think of anything better than manually specifying a minimum required version.
Will doing this with git commits work if someone is working on a branch? Printing a warning if versions don't match should be fine, but throwing an error would be super annoying.

anelles:
This really triggers the thought about continuous integration again. We need tests running, otherwise, this will be an endless problem.

anelles:
I will have a look at what you can go with Travis CI which github seems to promote.

cg-laser:
the git commit hash was also more to be able to reproduce a result. This hash uniquely defines the state of the repository (if not uncommited local changes were present.)

cg-laser:
It seems to me that after the time changes from Christoph, that NuRadioReco is getting pretty stable. The same for NuRadioMC. At least I can't think of any major interface/framework change. Or do you @anelles @christophwelling @daniel-zid  have anything else in mind? 

It would also be good to release a version 1.0.0 for the papers and the DESY workshop. 
I suggest the following scheme: Bump up the sub-sub version after each pull request merge. 1.0.1, 1.0.2 etc. 
[...]

PR #77, opened 2019-05-17
save git commit hash to NuRadioMC hdf5 output file

cg-laser:
addresses #65


cg-laser:
why would you check the hash somewhere? We just want to now with which version NuRadioMC ran. E.g. to understand older results. 

The other two commits are just formatting and a bug fix I introduced earlier. Needed to be on this branch to be able to test the code.

anelles:
OK, fine with me. Just storing the hash for now. I mean there is no point in storing it, if there is no instance that checks it.

PR #80, opened 2019-05-23
Ray tracing improvements

cg-laser:
This commit fixes the problem that ray tracing solutions that 'touch' the surface are not found. objective function is now well defined everywhere and smooth. This makes the extra calculation of improved start parameters superfluous. We might want to remove this functionality completely. What do you think @relamber ? Can you test this new code?


cg-laser:
p4 = 299792458.0 is from an alternative analytic function to calculate the travel time that is not used anywhere. I'm not sure if it works at all. We should probably just delete it.

PR #82, opened 2019-05-28
Implementation of different neutrino cross section

anelles:
Implement the one used by IceCube by default. 
Differences are ~10%, but we should have options.


anelles:
I thought it might be nice to be complete, also in our paper.

PR #85, opened 2019-05-30
Auto compile raytracer

cg-laser:
I added a install script that should simplify the installation of the Cpp ray tracer extension. 

I also coded that this script is executed everytime the ray tracer is loaded and the cpp version can't be found. I don't know if this is really what we want. If will certainly make things easier if it runs out-of-the box. But it might be annoying if for some reason we want to use the python version or if the compilation fails (then the user will get an error message everytime he loads the ray tracer). 

The user can still deactivate the cpp version via
```python
from NuRadioMC.SignalProp import analyticraytracing
analyticraytracing.cpp_available = False
```
but this is secret developer knowledge ;-) what do you think?

### DIFF
--- a/NuRadioMC/SignalProp/analyticraytracing.py
+++ b/NuRadioMC/SignalProp/analyticraytracing.py
@@ -23,16 +23,16 @@ try:
     print("using CPP version of ray tracer")
 except:
     print("trying to compile the CPP extension on-the-fly")
-    import subprocess
-    import os
-    subprocess.call(os.path.join(os.path.dirname(os.path.abspath(__file__)),
-                                 "install.sh"))
     try:
+        import subprocess
+        import os
+        subprocess.call(os.path.join(os.path.dirname(os.path.abspath(__file__)),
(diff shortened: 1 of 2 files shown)


cg-laser:
@anelles would it be possible to install GSL on the buildbot? Then we could also test the cpp compilation?

anelles:
@cg-laser, if you tell me the syntax for installing it on Unix? I failed in finding the apt install version of it.

cg-laser:
which Ubuntu version is running?
Can you try `sudo apt-get install libgsl-dev`
or 
https://askubuntu.com/questions/490465/install-gnu-scientific-library-gsl-on-ubuntu-14-04-via-terminal

or you would need to do it manually: https://astrointro.wordpress.com/2017/05/17/installing-gnu-scientific-library-gsl-in-ubuntu-16-04-and-compiling-codes/

anelles:
If you merge today or tonight, I can still fix things on the built-bot, otherwise you may need to wait until I am back in 10 days.

PR #98, opened 2019-06-24
Fixed greenland attenuation

daniel-zid:
The previous attenuation model for Greenland had negative values, which gave nonsensical results. This has been fixed.

Besides, when the ray tracing gives infinite values for the times or the lengths, which happens with the Greenland refractive index, we ignore this refracted solutions and care only about the direct or reflected fields.


daniel-zid:
Wooops, you're right.

Yes, I have opened an issue. I think it would be nice if the author of the integrated path (Ben?) could comment on this special case.

cg-laser:
@daniel-zid what is causing this problem? Which of the variables is zero? Did you investigate this a little further? 
A good workaround would be to switch to the numerical integration if the analytic one failed. This already implemented. 

And it would be nice to wait a little longer before merging so that I have the chance to comment. Especially if the ray tracing part is touched that I mostly coded. 
I don't like the current implementation where events are just silently skipped when the calculation was not successful. This might lead to wrong results without ever being noticed by the user.

daniel-zid:
The problem is the variable l1, that is sometimes zero, so its logarithm is -inf. I don't know what physical case this corresponds to, I could not get any sense of the equations.

I have reverted the pull request. You can approve the request and revert the changes so that we can write a better solution.

PR #100, opened 2019-06-24
Fix greenland numerical

daniel-zid:
The integral for the travel times and length is now performed numerically if the analytic expression is not well defined.

I don't know why the rebase from the other Greenland branch has not worked. Maybe it was because of the reverting of the pull request. The only different lines are the ones in the end of analyticraytracing.py


cg-laser:
it is weird that it shows so many changes... although it is comparing to the master.

cg-laser:
ok the reason was a merge conflict that I just resolved

daniel-zid:
It has worked with the simulations I've tried. But I have some more in the oven so we can wait until tomorrow just to be sure.

I think the issue is purely mathematical. For certain values of the angles and the indices, the given solution for the integral no longer applies and we need to resort to the numerical calculation. In fact, the numerical calculation seems to give sensible results. If you try to make l1 = 0, you will see that it does not correspond to any evident geometrical configuration. Or at least, I am not able to see it.

daniel-zid:
Perhaps a lru_cache in medium.py? Although the lru_cache can work as well in the analyticraytracing.py and I'd say the z_deep it's more of a property of the implementation of the ray tracing than of the medium itself. I'm happy with either way.

daniel-zid:
Alright, I've implemented an lru_cache that works. Let me know if it's fine or if I should change it.

daniel-zid:
Let me checked if it's correctly cached. If it's not, I'll move it to the medium module.

daniel-zid:
It does save the cache. The cache seems to exist for the whole class and not for each specific member. You can check it putting the following line after calling get_z_deep:

print(self.get_z_deep.cache_info())

daniel-zid:
No, I don't think it is correctly cached. And if I try to put the lru_cache in the medium class, z_deep becomes a function wrapper... Let me try something.

daniel-zid:
Alright, the new version is maybe a bit less elegant, but everything is correctly cached. You can write: print(get_z_deep.cache_info()) after calling the function to check it.

Issue #106, opened 2019-07-11
Update the testing suit


anelles:
More tests for different aspects on the code. Needed before we switch over to Python 3

anelles:
I think we are basically done. This is for reference.

PR #104, opened 2019-07-09
Adding cython and trying to fix the built-bot

anelles:

PR #107, opened 2019-07-11
new test for effective volume

cg-laser:
This test performs a full Veff simulation and verifies that the Veff prediction remains the same. All steps are tested. 

Due to the MC nature, there is a certain scatter. Simulating 50,000 events in a 4km radius zylinder at SP at 1e18 leads to the following scatter in the Veff prediction of about 10%.
veff_scatter

I implemented the verification such that the deviation is not allowed to be larger than 2 sigma which should cover 95% of the test runs. In case we get a larger deviation, the test just needs to be rerun. Do you think this is a feasible solution? 

The simulation takes ~4min on my laptop. We could simulate more events to decrease the scatter? 

@anelles we need to C++ ray tracer to work on the built bot. Can you try fixing that? It seems that the inclusion path is not set correctly, probably some googling on how to set up GSL on ubuntu will help. 

Addresses #106

### DIFF
--- a/NuRadioMC/EvtGen/generator.py
+++ b/NuRadioMC/EvtGen/generator.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, division, print_function
 import numpy as np
+import NuRadioMC
 from NuRadioMC.utilities import units
 from NuRadioMC.utilities import inelasticities
+from NuRadioMC.utilities import version
 from six import iterkeys, iteritems
 from scipy import constants
 from scipy.integrate import quad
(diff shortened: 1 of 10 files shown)


anelles:
GLS works on other branch #104 
Tests are still failing, the arrays are different. 1e-7 relative numerical difference. So not bad, but worrisome that different machines get different results. I think we need to get to the bottom of this before adding new tests.

cg-laser:
> When I run this, it fails. I think you are missing ../config.yaml if you don't want to use the default.

fixed it, can you try again?

cg-laser:
yes this is fine. I gives an error when it deviates more than 2 sigma.

PR #109, opened 2019-07-19
Nu radio proposal

daniel-zid:
Well, after messing up the EvtGen_PROPOSAL rebase (God knows why), I've created a proper new branch to put the content of the Proposal modules in. This module has been tested with the surface muons study and the tau decay length and energy tests.


daniel-zid:
Just a quick comment. items() in python3 does what iteritems() did in python2. There's no iteritems in python3 anymore.

daniel-zid:
> Why did you need an explicit conversion of the 'trigger_time' values to string? I also worked before, didn't it?

trigger_time? What do you mean?

> can you add a docstring to get_product_position, I couldn't figure out what it does.

Yes, I forgot.

> Can you add the default values for the optional parameters to the docstring of the generate_cylinder function?
I don't understand the resample option. What do I exactly specify with different integers.

    resample: integer or None
	        if integer, PROPOSAL generates a number of propagations equal to resample
	        and then reuses them. Only to be used with a single kind of lepton (muon or tau)

We use a number of muon propagations equal to the resample variable value and then we reuse them if we want to simulate more than 'resample'. For the muon study, I have set resample = 10000, so I use 10000 different propagations that are repeated. I had to do this because of the time constraints.

> surface_muons: I don't understand the doc string. Does activating this option disables neutrino generation? And muons from air showers are simulated instead and vertices are placed where high dE/dX interactions happen? Shouldn't we write a separate generator function for this?

Yes, but then many many lines would be repeated. I thought this was the most compact way.

> Can't tau neutrinos interact in the bedrock and make it to the ice to make a second interaction? I thought that this potentially increase the volume substantially. But you removed the extension of zmin.

But then we should consider bedrock propagation appropriately, not with the current code that has been prepared for ice.

> For surface_muons option: Why don't you generate muons at the surface? It seems that you're generating neutrinos with CC interaction and y=0. Shouldn't you start with a muon and the use proposal to propagate them?

We need the parent neutrinos, otherwise the code does not work. It was either modifying the rest of the code or generating an artificial neutrino. The latter was the faster solution, and I needed to have the results ready as soon as possible while I was preparing the ICRC. Maybe I should just write another generator?

> Great that you added testing scripts. Can you add a doc string at the beginning describing what they do?

Yes!

> I'm confused about the change in simulation2.py. Was there a bug? Or why did you need to change the attribute names from "energy" to "energies" (same for zenith and flavor).

Because those are the actual names! It was a really stupid bug. I also changed the parent event criterion, that was not good.

> Does the new code save all 'preceeding' events of a triggered event? Or just the initial event? Can you clarify this in the code as comment?

Only the initial neutrino. We had agreed on that, I think.

> Isn't the effective area the effective volume multiplied with the interaction length of neutrinos at a certain energy?

Yes, and in this case the interaction length is taken into account into the PROPOSAL propagation.

> Also it should be a separate function. Having the function "effective volume" return an area is confusing. Same for the export function.

I agree.

cg-laser:
Am 27.07.19 um 12:02 schrieb daniel-zid:
>
>     Why did you need an explicit conversion of the 'trigger_time'
>     values to string? I also worked before, didn't it?
>
> trigger_time? What do you mean?
>
I meant interaction_type
>
>     Can you add the default values for the optional parameters to the
>     docstring of the generate_cylinder function?
>     I don't understand the resample option. What do I exactly specify
>     with different integers.
>
> |resample: integer or None if integer, PROPOSAL generates a number of 
> propagations equal to resample and then reuses them. Only to be used 
> with a single kind of lepton (muon or tau) |
>
> We use a number of muon propagations equal to the resample variable 
> value and then we reuse them if we want to simulate more than 
> 'resample'. For the muon study, I have set resample = 10000, so I use 
> 10000 different propagations that are repeated. I had to do this 
> because of the time constraints.
>
ok can you update the doc string with this description?
>
>     surface_muons: I don't understand the doc string. Does activating
>     this option disables neutrino generation? And muons from air
>     showers are simulated instead and vertices are placed where high
>     dE/dX interactions happen? Shouldn't we write a separate generator
>     function for this?
>
> Yes, but then many many lines would be repeated. I thought this was 
> the most compact way.
>
maybe its ok, but in general one should rather write subfunction for the 
joint tasks and only expose two function for neutrinos and muons to the 
user.
>
>     Can't tau neutrinos interact in the bedrock and make it to the ice
>     to make a second interaction? I thought that this potentially
>     increase the volume substantially. But you removed the extension
>     of zmin.
>
> But then we should consider bedrock propagation appropriately, not 
> with the current code that has been prepared for ice.
>
Can proposal do that? Wouldn't it be a substantial contribution?
>
>     For surface_muons option: Why don't you generate muons at the
>     surface? It seems that you're generating neutrinos with CC
>     interaction and y=0. Shouldn't you start with a muon and the use
>     proposal to propagate them?
>
> We need the parent neutrinos, otherwise the code does not work. It was 
> either modifying the rest of the code or generating an artificial 
> neutrino. The latter was the faster solution, and I needed to have the 
> results ready as soon as possible while I was preparing the ICRC. 
> Maybe I should just write another generator?
>
I don't understand. Why does it only work for neutrinos? I think we 
should avoid these hacks, this will make the code difficult to maintain. 
We should rather encapsulate common functionality in functions. Or make 
it a class to more easily share member variables. The generate_cylinder 
function is already pretty long and thus difficult to maintain. Too many 
if/elses in there make it error prone. I opt for some refactoring.
>
>     Does the new code save all 'preceeding' events of a triggered
>     event? Or just the initial event? Can you clarify this in the code
>     as comment?
>
> Only the initial neutrino. We had agreed on that, I think.
>
That's fine! I just couldn't extract it directly from the code. Can you 
add a comment saying this?
>
>     Isn't the effective area the effective volume multiplied with the
>     interaction length of neutrinos at a certain energy?
>
> Yes, and in this case the interaction length is taken into account 
> into the PROPOSAL propagation.
>
Well but in the code you just multiply the surface area with 
sum(triggers)/N_total. So it is not the effective area.

daniel-zid:
> Am 27.07.19 um 12:02 schrieb daniel-zid:
> Why did you need an explicit conversion of the 'trigger_time' values to string? I also worked before, didn't it? trigger_time? What do you mean?
> I meant interaction_type
This is because of the different way of handling strings on python3. Better safe than sorry. I got some weird mistakes.
> Can you add the default values for the optional parameters to the docstring of the generate_cylinder function? I don't understand the resample option. What do I exactly specify with different integers. |resample: integer or None if integer, PROPOSAL generates a number of propagations equal to resample and then reuses them. Only to be used with a single kind of lepton (muon or tau) | We use a number of muon propagations equal to the resample variable value and then we reuse them if we want to simulate more than 'resample'. For the muon study, I have set resample = 10000, so I use 10000 different propagations that are repeated. I had to do this because of the time constraints.
> ok can you update the doc string with this description?
Eeuh, I think it is like that now... :/
> surface_muons: I don't understand the doc string. Does activating this option disables neutrino generation? And muons from air showers are simulated instead and vertices are placed where high dE/dX interactions happen? Shouldn't we write a separate generator function for this? Yes, but then many many lines would be repeated. I thought this was the most compact way.
> maybe its ok, but in general one should rather write subfunction for the joint tasks and only expose two function for neutrinos and muons to the user.
You're right. I am going to write a different function.
> Can't tau neutrinos interact in the bedrock and make it to the ice to make a second interaction? I thought that this potentially increase the volume substantially. But you removed the extension of zmin. But then we should consider bedrock propagation appropriately, not with the current code that has been prepared for ice.
> Can proposal do that? Wouldn't it be a substantial contribution?
I don't know if it can be used for bedrock. I am going to ask Jan. But if that's the case, I have to rewrite the modules as well.
> For surface_muons option: Why don't you generate muons at the surface? It seems that you're generating neutrinos with CC interaction and y=0. Shouldn't you start with a muon and the use proposal to propagate them? We need the parent neutrinos, otherwise the code does not work. It was either modifying the rest of the code or generating an artificial neutrino. The latter was the faster solution, and I needed to have the results ready as soon as possible while I was preparing the ICRC. Maybe I should just write another generator?
> I don't understand. Why does it only work for neutrinos? I think we should avoid these hacks, this will make the code difficult to maintain. We should rather encapsulate common functionality in functions. Or make it a class to more easily share member variables. The generate_cylinder function is already pretty long and thus difficult to maintain. Too many if/elses in there make it error prone. I opt for some refactoring.
Yes, I am going to write a new function. I don't remember if there was also a problem in the simulation2.py file, but we can sort it out.
> Does the new code save all 'preceeding' events of a triggered event? Or just the initial event? Can you clarify this in the code as comment? Only the initial neutrino. We had agreed on that, I think.
> That's fine! I just couldn't extract it directly from the code. Can you add a comment saying this?
Absolutely!
> Isn't the effective area the effective volume multiplied with the interaction length of neutrinos at a certain energy? Yes, and in this case the interaction length is taken into account into the PROPOSAL propagation.
> Well but in the code you just multiply the surface area with sum(triggers)/N_total. So it is not the effective area.
Well, it is because I don't force muons to interact! That's the difference with the neutrino calculation. I force a muon to be there but not to interact at all! That's why it is actually the effective area.
> […](#)

daniel-zid:
It seems that PROPOSAL can only handle transparent media...

daniel-zid:
With respect to the effective area calculation, think of it like this: the muon interaction length is already given by the interaction rate which is given by PROPOSAL. That's radically different with respect to what we do for neutrinos, where we assume that every neutrino interacts and then the interaction length tells us the actual rate. Here, PROPOSAL tells us the actual rate of cascade-inducing muon interactions.

I don't know if I'm being clear... Tell me if I'm not.

cg-laser:
> > Isn't the effective area the effective volume multiplied with the interaction length of neutrinos at a certain energy? Yes, and in this case the interaction length is taken into account into the PROPOSAL propagation.
> > Well but in the code you just multiply the surface area with sum(triggers)/N_total. So it is not the effective area.
> 
> Well, it is because I don't force muons to interact! That's the difference with the neutrino calculation. I force a muon to be there but not to interact at all! That's why it is actually the effective area.

Ok but this is dangerous as this is only a special case for muons and not applicable for neutrinos

daniel-zid:
> > > Isn't the effective area the effective volume multiplied with the interaction length of neutrinos at a certain energy? Yes, and in this case the interaction length is taken into account into the PROPOSAL propagation.
> > > Well but in the code you just multiply the surface area with sum(triggers)/N_total. So it is not the effective area.
> > 
> > 
> > Well, it is because I don't force muons to interact! That's the difference with the neutrino calculation. I force a muon to be there but not to interact at all! That's why it is actually the effective area.
> 
> Ok but this is dangerous as this is only a special case for muons and not applicable for neutrinos

I am aware, I am aware, but it's the most accurate way PROPOSAL can be used. How should we implement it or document it better?

cg-laser:
> I am aware, I am aware, but it's the most accurate way PROPOSAL can be 
> used. How should we implement it or document it better?
>
Make it a separate function where it is explicit that this only is for 
muon simulations (and adding the details to the doc string).

daniel-zid:
Ok, changed points:

- The code was already python2/3 compatible, now is python2/3 efficient!
- Docstring to get_product_position added
- New function generate_surface_muons, with proper treatment of muons without redundant neutrinos added.
- PROPOSAL threshold and table paths changed for speed

Now I should change the geometry to accommodate for bedrock, ice and air media and create a json file to parse the configuration for PROPOSAL. To do after the holidays. The code can be merged if the reviews are positive and we'll take care of this later.

I have still to write the NuRadioPROPOSAL documentation before merging though!

anelles:
I assume that this is work in progress again now, after you talked to the PROPOSAL developers at ICRC?

daniel-zid:
The changes have been implemented and the pull request is ready for review.
The improvements are work in progress.

O 08:41, xov., 22/08/2019, Anna Nelles <notifications@github.com> escribiu:

cg-laser:
I think you took into account my requests, so I'm ok with merging. But I didn't go through the changes again.

cg-laser:
but why do the tests fail? Is this because the branch is so old that it did not include the updated tests? @daniel-zid can you merge the master banch into this branch? Then the test should be up to date and we can check that this commit doesn't change the default evtgen behaviour?

daniel-zid:
I had merged the master branch, but there was a weird problem that we could not solve. I forced the branch to have the same test as the current master branch and now the checks work fine. When you give me a sign I can merge the pull request.

PR #111, opened 2019-07-22
Focusing

yvpan:
Added a way to calculate the focusing effect on propagation


yvpan:
@daniel-zid Sorry I didn't realized that link was protected by a password. I attached the note from dave seckel, which i referred to.
[focusingDaveSeckel.pdf](https://github.com/nu-radio/NuRadioMC/files/3428159/focusingDaveSeckel.pdf)

cg-laser:
Two more general questions? I first thought we might want to use a dZ that is about the size of the antennas ~1m. But then we might run into the problem that there might not be a solution for the dZ displaces receiver...
On the other hand, just shifting by a 1cm requires a high precision of the ray tracer. This relates to #112. @clark2668 Can you change the C++ code to have a higher precision (1e-9 instead of 1e-6) in  #112 for testing?

cg-laser:
Also does the focussing lead to interference for our wavelength? Can you make test with 1m displacement and compare the travel times? When the differences in travel times are much less than 1ns then it should be fine. Otherwise we run into interference problems and the treatment gets more complicated.

yvpan:
hi @cg-laser , thank you for comments. I fixed the issues of spectrum[1:] *= focusing and float return.
I compared the dZ of 1m, 0.1m and 0.01m. 1m usually gives travel times difference more than 1ns, while 0.1m less than 1ns. Considering a precision of 1e-6, 0.01m should be fine i guess? Or do you suggest a default dZ of 0.1m?

cg-laser:
I'm not worried about the precision of the calculation (and would keep dZ small) but the validity of the method because the different time delays will decrease the coherence (between the different signal paths).

anelles:
@yvpan and @cg-laser, this is now in conflict with the master branch and given the conversation, I am not quite sure whether the changes requested is still up to date. This needs therefore attention.

anelles:
@cg-laser, now that we have figured out what has to be done in principle, are you taking care of this branch?

cg-laser:
we should merge #158 first so that we can fix the tests here. I also don't get why the tests at some point started to fail... Maybe something in NuRadioReco

cg-laser:
did someone restarted this test? I'm pretty sure that it ran successful when I merged it. It says it ran 2days ago the last time and failed because the focusing branch does't exist anymore....

PR #112, opened 2019-07-27
new pull request to test precision of ray tracer

cg-laser:


anelles:
@cg-laser what was the status of this? The tests are failing now, we need this cleaned up, I suppose.

cg-laser:
this is because the testing is more stringent. I think this is because the precision of the C++ ray tracer is not good enough (but probably good enough for our needs). I hoped that @clark2668 will have a look at some point.

anelles:
So the precision of the ray tracer also is the issue when testing whether the .nur files are identical in #128 . If @clark2668 has no time to work on this, we should assign someone else to look into this. Otherwise, all detailed testing will keep failing.

clark2668:
Hey @anelles, thanks for the reminder. I've fallen out of touch with what the exact issue is: where is the discrepancy showing up? In the C0/C1 parameters themselves, or in the higher level variables like travel time and arrival angle?

anelles:
It seems that the ray tracer delivers slightly different results on different plattforms. When @cg-laser and I run on our Macs, we get results that differ from the built-bot (fresh and peachy native unix). I think you can best see this in #128.

clark2668:
So this isn't a discrepancy between, say, the python vs c++ version. This is a discrepancy between the c++ version on a mac vs on the build-bot... Interesting. I wonder if they're running slightly different versions of GSL. I will try to take a look at this. Is there an easy way to view the build bot answers?

anelles:
In principle the built bot just executes the tests in the code. When you click on the 'check' from travis below, you should be getting to the results. If you want more output you need to change the tests and push to the relevant branch.

anelles:
I suggest to close this pull request. It is pretty outdated and we have made additional fixes along the way. @cg-laser any objections?

PR #114, opened 2019-07-29
New tests

anelles:
Adding Veff test for built-bot


cg-laser:
you can check pyproject.toml in NuRadioMC and NuRadioReco for all required non-standard packages

anelles:
As shown here, the test fails in 5% of the time. I guess for now this is fine, as typically both the branch and the pull request is built close after each other.

anelles:
> packages

This is not enough for the built-bot. It literally has nothing, no even numpy.

anelles:
By the way, I reran the test and now it is 19% different. This is a bit cumbersome ...

cg-laser:
I thought more about @christophweeling idea to fix the seed. I think this is a good idea, although it does not work for certain cases where we change something in the event gen that alters the random sequence. For these cases we will need to life with the limited statistical precision, but the rest should result in an identical result. 

I don't have time to fix it before my vacation, so I'm already approving the pull request as it already works fine as it is, and this optimization can be done later.

anelles:
@cg-laser I don't think you actually approved this before leaving for vacation. So I will just use my priviliges to merge to move ahead.

Issue #108, opened 2019-07-12
Change how we load yaml


anelles:
The built-bot complains: 
/home/travis/build/nu-radio/NuRadioMC/NuRadioMC/simulation/simulation2.py:81: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.

PR #115, opened 2019-07-29
Fixing depreciation warning in yaml.

anelles:
Adressing #108


anelles:
See also https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

PR #118, opened 2019-08-09
Python3

christophwelling:
Implements changes necessary for NuRadioMC to run on python3.
The examples (at least those that are not broken on master) run without error and the results look reasonable.
The units test says it fails, but this is not quite true. The triggers are stored in a different order than in the reference, otherwise the results are the same. If I understand this https://github.com/h5py/h5py/issues/1029 correctly, the reason is that h5py just doesn't care about the order and we are fine otherwise.
So in principle this should be ready to merge, but we should first set up some more unit tests.

If you want to try this, make sure to check out the python3 branch from NuRadioReco and the radiotools as well.


cg-laser:
Do the dictionary iterators break python2.7? Nor sure if we want to try to keep is compatible with both. Probably there is no need for it.

christophwelling:
I assume you mean replacing attributes.iteritems() with attributes.items() in generator.py? I would have to check, but I think this should still be fine in py2. It returns a list instead of an iterator in py2, but that should be fine.

.travis.yml
```diff
@@ -15,9 +15,9 @@ before_script:
   - unzip /tmp/radiotools.zip
   - mv radiotools-master radiotools
   - export PYTHONPATH=$PWD/radiotools
-  - wget https://github.com/nu-radio/NuRadioReco/archive/master.zip -O /tmp/NuRadioReco.zip
```
anelles: Reminder to switch this to master again. In a pull request right after merging all.

NuRadioMC/utilities/attenuation.py
```diff
@@ -71,5 +71,11 @@ def get_attenuation_length(z, frequency, model):
             att_length_f[ att_length_f < min_length ] = min_length
 
         return att_length_f
+    elif(model == "MB1"): # from 10.3189/2015JoG14J214
```
anelles: @cg-laser I think the MB model would have been nicer in a new branch, but I can live with this.

NuRadioMC/simulation/simulation2.py
```diff
@@ -679,6 +685,10 @@ def _add_empty_electric_field(self, channel_id):
         self._sim_station.add_electric_field(electric_field)
 
     def _write_ouput_file(self):
+        folder = os.path.dirname(self._outputfilename)
+        if(not os.path.exists(folder)):
+            logger.warning(f"output folder {folder} does not exist, creating folder...")
```
anelles: This is the new p3 for format, I guess?

NuRadioMC/simulation/simulation2.py
```diff
@@ -141,7 +150,7 @@ def __init__(self, eventlist,
         else:
             self._bandwidth = bandwidth
         self._Vrms = (self._Tnoise * 50 * constants.k *
-                       self._bandwidth / units.Hz) ** 0.5
+                       self._bandwidth / units.Hz) ** 0.5  # from elog:1566
```
anelles: What is this comment tyring to tell me? This will be very unclear, if we move away from ARIANNA elog.

NuRadioMC/simulation/simulation2.py
```diff
@@ -16,6 +16,7 @@
 import NuRadioReco.modules.io.eventWriter
```
anelles: It might be a little late for this, but how do people feel about renaming simulation2.py to something else in the py3 release? It has been rather sad, that our main file is called soemthing like work in progress.

NuRadioMC/utilities/attenuation.py
cg-laser: I agree but I didn't know how to do it differently at the moment because we can only merge pull requests into the master branch (at least through the web interface).

NuRadioMC/simulation/simulation2.py
cg-laser: yes this is a fancy new p3 string formatting feature

NuRadioMC/simulation/simulation2.py
cg-laser: I stubled accross this equation and didn't understand where it came from, and it took me a while to figure out where it came from. I agree that the reference is not optimal but better than nothing.

NuRadioMC/simulation/simulation2.py
cg-laser: yes we should to that! Now is the right time

.travis.yml
cg-laser: shouldn't we use python 3.7?

NuRadioMC/EvtGen/generator.py
```diff
@@ -509,10 +509,11 @@ def write_events_to_hdf5(filename, data_sets, attributes, n_events_per_file=None
 #             else:
 #                 stop_index = tmp[0]
 
-        for key, value in iteritems(data_sets):
-            if (key == 'interaction_type'):
-                value = np.array(value, dtype='S')
-            fout[key] = value[start_index:stop_index]
+        for key, value in data_sets.items():
+            if value.dtype.kind == 'U':
+                fout[key] = [np.char.encode(c, 'utf8') for c in value[start_index:stop_index]]
```
cg-laser: why is this explicit encoding necessary?

NuRadioMC/simulation/simulation2.py
```diff
@@ -644,7 +650,7 @@ def _read_input_neutrino_properties(self):
         self._event_id = self._fin['event_ids'][self._iE]
         self._flavor = self._fin['flavors'][self._iE]
         self._energy = self._fin['energies'][self._iE]
-        self._inttype = self._fin['interaction_type'][self._iE]
+        self._inttype = self._fin['interaction_type'][self._iE].astype('str')
```
cg-laser: why is the explicit type conversion necessary?

christophwelling: This (and the one above) are to read in eventlists created under python2. Otherwise python3 doesn't know how to handle the strings. I did this so I didn't have to change the eventlists in the unit tests.

Issue #120, opened 2019-08-16
time resolution of analytic ray tracing


cg-laser:
The uncertainty in the travel time calculation of the analytic formula seems to be around 0.1ns. This is too large when used for DnR calculations. I think this uncertainty was introduced when the Greenland ice model was added and the `get_z_deep` function was added to the analytic ray tracer. This function automatically calculates the depth for which the ice can be assumed to be uniform (n(z) = const). Changing the condition to be stricter, increases precision. 

I changed the value in https://github.com/nu-radio/NuRadioMC/commit/ac8262c97dc0105a4d616c7c815d1fc3506234b6, but we should make a thorough test of the time resolution for different ice models by comparing to the numerical calculation.

PR #122, opened 2019-08-21
Raytracing mooresbay

cg-laser:
This pull request adds a reflective bottom layer to the analytic ray tracing as needed for Moore's Bay. 

The medium class is extended with an optional property `self.reflection = -576 * units.m` specifying the depth of the reflective layer. 

The interface remains backward compatible and by default no bottom reflections are considered (this might change though). Because both the bottom and the surface are reflective, an infinite number of solutions exist. To handle this, the user needs to specify how many bottom reflections should be considered at maximum. 4 additional solutions exist per bottom reflection. I.e., for one reflection we get 6 solutions in total, and for 2 reflections, 10 solutions in total. 

The number of reflections to consider need to be specified in the initialization of the (3D) ray tracing class. If the ice model does not specify a reflection layer, the number of reflections is set to zero and a warning is issued. 

The results are returned with two additional attributes
```
{'type': solution_types[i],
 'C0': C0s[i],
'C1': C1s[i],
'reflection': reflection[i],
'reflection_case': reflection_case[i]})
```
where 'reflection' is an integer specifying the number of bottom reflections, and `reflection_case` can be either 1 or 2. 1: rays start upwards, 2: rays start downwards. 

The 'bottom' reflections are implemented both in C++ as well as in python, but not in a C++ stand-alone version: By default, only the minimization to find the C_0 parameter, and the numerical integration to find the freq. dependent attenuation factors are used from the C++ implementation. Only these C++ functions are extended to also simulate bottom reflections. 

## technical details
a ray reflecting (multiple times) off the bottom can be decomposed into several rays with the same C_0 parameter with the start/stop point being the point of reflection off the bottom. This allows to calculate the solution for any number of reflections without additional computing time. Also the same formulas for attenuation length, distance and travel time can be used. They are just called several times for the ray segments. 

The 2D ray tracing assumes that the ray always starts upgoing, which can always be achieved via a coordinate rotation in case we have no bottom reflection. However, with a bottom reflection, we have two solutions where the ray starts upgoing, and two solutions where the ray starts downgoing. We implement the second case by first calculating a virtual point left of the start position that leads to a ray that goes though the start position and is downgoing at the start point. 

## unit tests
This pull request also adds several unit tests. All python scripts starting with "T" in the 'tests' folder are a unit test that should be added to the built bot/travis CI. 
While doing this I discovered that the analytic formula for the travel time and distance calculation has a precision problem. Because the function is called for each ray segment, the errors can add up. Without bottom reflections, the precision is ~40ps (better on average but this is the maximum deviation allowed so that the unit tests pass with 1000 events). The uncertainty also depends on where to put the `z_deep` cutoff (below this depth a straight line calculation is assumed). I optimized this value to achieve maximum precision. 
40ps might sound very small, but it becomes relevant for the high precision DnR calculations. (addresses #120 )


cg-laser:
This probably also the reason why the built bot fails now. The precision of the T and D calculations increased.

cg-laser:
There is one computing speed problem with additional ray solutions: The time delay between travel times of different rays can get really large (~10k ns). When simulating the detector response all efields from the different rays are combined into one voltage trace which gets really large. This results in slow FFT speeds. I'm not yet sure how to handle this in the best way. 

What I did for now that I check each ray for its off cone angle and skip individual rays whose angular distance is too low (we already skip the whole event if the angular distance of all solutions is too large). This eliminates most cases where we get a large time gap between arriving rays (as they typically have large viewing angle differences). 

Another way would be to somehow split these cases up into individual events. One could find all ray tracing solution and travel times of all channels first and then group them in groups of similar travel time, e.g., via a maximum allowed time gap between them. If only one station is simulated it would work fine (except for the technical problem that we dynamically increase the output array dimensions). However, I don't know how to handle different stations correctly. Because one event has several stations, but now we want to split up a single station into several events... Maybe leaving the implementation as it is is the best to do anyway. 

A downside of this is that only the first trigger is counted, because the trace is cut to the readout trace lenght. So potentially triggers at a later time are discarded. If this is physical would depend on the deadtime of a station after a trigger. 

Another option would be to add a new station for any delayed signals, i.e., a single station could trigger multiple times for the same event. "station_101_a", "station_101_b" etc. This would require a change in NuRadioReco though, because currently an event can only have one station with the same station id.

cg-laser:
I also added a fixed seed to the Veff test. It was annoying that the test failed so often

christophwelling:
I changed 3 small things:
- Unit tests moved from SignalProp folder into test folder
- Added some print outputs so one can see what stage the unit test is currently at. I also set the raytracing logging to WARNING to not drown any errors in the INFO output
- I had to increase the tolerance of the T06unit_test_C0_mooresbay test to 1.e-6. I think this just means that the numerical errors were a little larger than expected, but should be fine otherwise.

cg-laser:
thanks for the changes and the test integration!

I have one additional request but this is also not high priority: I thought the ray tracing tests could run as an independent test, so that the github page says "2 check passed". One test is the Veff tests, and one test is the ray tracing tests. This would give an easier overview of which test failed. With the staging we need to check the detailed travis page to see what went wrong. Also a staging does not make much sense (but not sure if this is being done): The ray tracing tests can all run independent and all should always be executed to directly see all possible errors. Do you know what I mean? I have no idea how much work this would be though ;-)

BTW: The precision is fine. The numerical precision of the minimizer is set to 1e-6 I think.

anelles:
Travis runs most of them independently, but I haven't found a way to 
display this on github. It does this per pull request and not per test.
Before it did branch and pull request, but that was just doubling 
compute time.

anelles:
You see it on the travis pages though. 
Screen Shot 2019-08-23 at 15 14 42

PR #123, opened 2019-08-22
add unit tests for Askaryan modules

cg-laser:
also updates doc string


cg-laser:
@anelles can you add the U01 script as a separate test? It would be great to see something like "Travis CI - Askaryan tests"

cg-laser:
yes

anelles:
@cg-laser : I cannot show them as individual tests here on the github page. Before they showed two, because I build the branch and the pull request. If it fails, you have to look at the log about why it failed.

PR #125, opened 2019-08-26
Unit tests

christophwelling:
- Adds the examples to the unit test.
- Removes the trigger checks from the single event unit test. Since hdf5 does not keep track of the order in which it stores data, the triggers may have a different order from the one in the reference and the check fails, though it is fine otherwise. Since all trigger modules are checked in NuRadioReco anyway, the simplest solution is to just not check the triggers here.


NuRadioMC/test/SingleEvents/T03validate.py
```diff
@@ -17,8 +17,7 @@
 fin1 = h5py.File(file1, 'r')
 fin2 = h5py.File(file2, 'r')
 
-attributes = [u'trigger_names',
```
anelles: I think @cg-laser is working on changing all dicts to ordered dicts. Then we should be able to test the triggers again, right?

christophwelling:
Trigger order should now be consistent and I updated the reference accordingly. Triggers are now part of the test again.

PR #126, opened 2019-08-29
Fix coincidences

daniel-zid:
Quick fix for the coincidences example


cg-laser:
why did you remove the presimulation step? This saved a huge amount of computing time! It shouldn't be removed. See also explanation of this trick in https://github.com/nu-radio/NuRadioMC/wiki/Example:-Multi-station-coincidences step 3.

anelles:
@cg-laser can you check, whether this now full-fills the changes that you requested? 
@daniel-zid please then also add the changelog.txt

cg-laser:
no its still wrong as far as I can see. One should not simulate the all stations without a pre simulation. This just burn many CPU hours and we should not advertise this in an example. 

The logic of this example should be (as the original numbering reflected):
- generate input files using the event generator
- do a presimulation (just the central station) 
- do a full simulation (with all stations) using the outputfiles of the pre simulation as input files
- analyze the final results.

daniel-zid:
Then it's still clear that I don't understand how it works. There were no input files generator on the other version of the example. If the example can still be run the way you like with the current files, would you mind changing what it's needed?

anelles:
@daniel-zid @cg-laser should we close this and move on, or is there hope of fixing the documentation in such a way that it is clear for everyone what is needed?

PR #128, opened 2019-09-03
also test writing *.nur files

cg-laser:


cg-laser:
yes maybe, although it might be difficult because newer versions might add more output parameters. So a simple 1to1 comparison will not be sufficient.

anelles:
How about we only compare the trace. That one should be identical and is likely to reveal changes that are not caught in the hdf5 files.

cg-laser:
sounds good

Am 05.09.19 um 11:45 schrieb Anna Nelles:

anelles:
I know why we never tested the .nur files. The file that is generated is 130 MB. @cg-laser this cannot be, right? It is 10 events ...

anelles:
In any case, I have the test ready, if you want to figure out whether we can generate a smaller file sizes ....

cg-laser:
can you try resampling the simulated efield to the detector sampling rate?

Am 06.09.19 um 11:38 schrieb Anna Nelles:

anelles:
It is downsampled by default. That is why the filesize confuses me.

cg-laser:
yes that's weird, we could add debug output to the serialization functions that output how much space what object occupies.

anelles:
I have now stopped the script from writing efields. This makes the file size handlable for now. 
Still 250 KB per event is big. Just assume 0.1Hz trigger of 100 stations. This is 0.8 Mio events in a day -- so a lot of MB.

anelles:
It is very interesting that the traces are not equal across the built-bot and my computer. Will be followed up.

anelles:
Ok, I will merge then like this and we can still reiterate on the non-deterministic results.

PR #129, opened 2019-09-04
use module registry to find out the bandwidth of the detector

cg-laser:
This pull request works together with the new NuRadioReco feature (same branch name) to keep track of all applied modules, and in particular to find out what filters were applied.

Test fails because the bandwidth of the example was incompatible with the used filters


cg-laser:
the bandwidth is calculated for each station/channel separately as the response of each channel could be different. However, the trigger modules do not support a per channel threshold yet. Therefore, we just use the bandwidth of the first station and channel for now.

cg-laser:
I still need to implement a second use case where the Vrms value is specified directly (instead of the noise temperature). This might be interested for the simulation of real detectors where we now the noise level from measurements. 

I suggest that the user can either specify a noise temperature or a RMS value.

NuRadioMC/simulation/simulation.py
```diff
@@ -143,12 +143,68 @@ def __init__(self, eventlist,
...
+                    if hasattr(instance, "get_filter"):
+                        filt *= instance.get_filter(ff, self._station_id, channel_id, self._det, **kwargs)
+                filt = np.abs(filt)
+                filt /= filt.max()
+                bandwidth = np.trapz(np.abs(filt)**2, ff)**0.5
+                self._bandwidth_per_channel[self._station_id][channel_id] = bandwidth
+                logger.info(f"bandwidth of station {self._station_id} channel {channel_id} is {bandwidth/units.MHz:.1f}MHz")
+        ################################
+
+#         bandwidth = self._cfg['trigger']['bandwidth']
```
anelles: Clean up needed

NuRadioMC/simulation/simulation.py
```diff
@@ -465,6 +520,19 @@ def run(self):
 
         t_total = time.time() - t_start
         outputTime = time.time() - t5
+
```
anelles: Should we move this to a utility?  This function is in principle handy for all sorts of timing.

anelles: I mean the code below, sorry if being unclear.

cg-laser:
there is essentially no overhead due to the dry run. It takes less time than a full event and typically we're simulating 10k or more events per file.

cg-laser:
Because the bandwidth of the tests was not set according to the actual bandwidth in the simulations, the test results changed. I tested if the test pass if I force the bandwidth to the old value and then 'test single events' test passed. Then I updated the reference files for the correct bandwidth. Hence, I think we can be sure that this pull request does not change the results of NuRadioMC. 

But people should check if they calculated the bandwidth correctly in their simulations! Now this extra step is no longer necessary.

NuRadioMC/simulation/simulation.py
```diff
@@ -144,14 +149,112 @@ def __init__(self, eventlist,
...
+            raise AttributeError(f"Specifying noise temperature (set to {Tnoise}) and Vrms (set to {Vrms} is not allowed.")
+        if(Tnoise is not None):
+            self._Tnoise = float(Tnoise)
+            self._Vrms = (self._Tnoise * 50 * constants.k *
+                           self._bandwidth / units.Hz) ** 0.5  # from elog:1566 and https://en.wikipedia.org/wiki/Johnson%E2%80%93Nyquist_noise (last Eq. in "noise voltage and power" section
+            logger.warning('noise temperature = {}, bandwidth = {:.2f} MHz -> Vrms = {:.2f} muV'.format(self._Tnoise, self._bandwidth / units.MHz, self._Vrms / units.V / units.micro))
+        elif(Vrms is not None):
+            self._Vrms = float(Vrms) * units.V
+            self._Tnoise = None
+#             Vrms_before_amplification = self._Vrms / amplification
```
anelles: Cleanup, please.

NuRadioMC/simulation/simulation.py
```diff
@@ -276,7 +378,7 @@ def run(self):
                     if(not r.has_solution()):
                         logger.debug("event {} and station {}, channel {} does not have any ray tracing solution ({} to {})".format(
                             self._event_id, self._station_id, channel_id, x1, x2))
-                        self._add_empty_electric_field(channel_id)
+#                         self._add_empty_electric_field(channel_id)
```
anelles: Cleanup.

NuRadioMC/simulation/simulation.py
```diff
@@ -300,7 +402,7 @@ def run(self):
                     # discard event if delta_C (angle off cherenkov cone) is too large
                     if(min(np.abs(delta_Cs)) > self._cfg['speedup']['delta_C_cut']):
                         logger.debug('delta_C too large, event unlikely to be observed, skipping event')
-                        self._add_empty_electric_field(channel_id)
+#                         self._add_empty_electric_field(channel_id)
```
anelles: Cleanup

NuRadioMC/simulation/simulation.py
```diff
@@ -423,10 +524,10 @@ def run(self):
                         # apply a simple threshold cut to speed up the simulation,
                         # application of antenna response will just decrease the
                         # signal amplitude
-                        if(np.max(np.abs(electric_field.get_trace())) > float(self._cfg['speedup']['min_efield_amplitude']) * self._Vrms):
+                        if(np.max(np.abs(electric_field.get_trace())) > float(self._cfg['speedup']['min_efield_amplitude']) * self._Vrms_efield):
                             candidate_event = True
 
-                #print("start detector simulation. time: " + str(time.time()))
```
anelles: Cleanup

NuRadioMC/utilities/Veff.py
```diff
@@ -94,6 +94,7 @@ def get_Aeff_proposal(folder, trigger_combinations={}, zenithbins=False):
     deposited = False
 
     for iF, filename in enumerate(sorted(glob.glob(os.path.join(folder, '*.hdf5')))):
+        print(f"reading {filename}")
```
anelles: Remove print statement, change to logger.

NuRadioMC/utilities/Veff.py
```diff
@@ -115,7 +116,6 @@ def get_Aeff_proposal(folder, trigger_combinations={}, zenithbins=False):
     print("Trigger names:", trigger_names)
```
anelles: See above.

NuRadioMC/utilities/Veff.py
```diff
@@ -304,7 +311,10 @@ def get_Veff(folder, trigger_combinations={}, zenithbins=False):
     prev_deposited = None
     deposited = False
 
+    if(len(glob.glob(os.path.join(folder, '*.hdf5'))) == 0):
+        raise FileNotFoundError(f"couldnt find any hdf5 file in folder {folder}")
     for iF, filename in enumerate(sorted(glob.glob(os.path.join(folder, '*.hdf5')))):
+        print(f"reading {filename}")
```
anelles: Logger, not print

cg-laser:
@anelles made the remaining cleanup

PR #133, opened 2019-09-11
change in effective volume calculation utitlity

cg-laser:
The getVeff function now returns just the effective volume which is 
simulated volume / number of simulated event * sum over triggered events (neutrino weight)

To get the Veff steradian, the user needs to multiply this number by the solid angle of the simulation which is also returned. 

Also the normalization to 'water' is removed. 

The interaction length function is moved to the cross sections script. I also changed it to use the proton mass (and not the avogadro constant) which is the prescription of Eq. 28 of 10.1103/PhysRevD.83.113009. In the previous definition also the units didn't work out but the numerical value was essentially the same.

### DIFF
--- a/NuRadioMC/examples/Sensitivities/E2_fluxes3.py
+++ b/NuRadioMC/examples/Sensitivities/E2_fluxes3.py
@@ -491,21 +491,24 @@ def get_E2_limit_figure(diffuse = True,
     plt.tight_layout()
     return fig, ax
 
-def add_limit(ax, limit_labels, E, Veff, n_stations, label, livetime=3*units.year, linestyle='-',color='r',linewidth=3,band=False):
+def add_limit(ax, limit_labels, E, Veffsr, n_stations, label, livetime=3*units.year, linestyle='-',color='r',linewidth=3,band=False):
+    """
+    add limit curve to limit plot
+    """
     E = np.array(E)
-    Veff = np.array(Veff)
+    Veffsr = np.array(Veffsr)
 
     if band:
 
         limit_lower = fluxes.get_limit_e2_flux(energy = E,
-                                         veff = Veff[0],
+                                         veff = Veffsr[0],
                                          livetime = livetime,
                                          signalEff = n_stations,
                                          energyBinsPerDecade=energyBinsPerDecade,
                                          upperLimOnEvents=2.44,
                                          nuCrsScn='ctw')
         limit_upper = fluxes.get_limit_e2_flux(energy = E,
-                                         veff = Veff[1],
+                                         veff = Veffsr[1],
                                          livetime = livetime,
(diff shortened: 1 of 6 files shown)


cg-laser:
I also refactored the getVeff function quite a bit. The current function didn't allow to parse hdf5 files with different zenith angle bins. 
Now, the function returns an array of dictionaries (on per file) where the dictionaries contain all relevant properties of the file such as Veff, energy, zenith bin, ...
For easier plotting and slicing, I added another function get_Veff_array which returns a multi-dim array for easy slicing, e.g., averaging over all zenith bins. I think this is quite useful. 
Also export/buffering of effective volumes is much easier. Just pickle/jsondump the result of the getVeff function.

cg-laser:
I addressed all requests. @daniel-zid can you also refactor the get_Aff function to the same layout of the get_Veff function? I don't have a test case so it would be difficult for me.

daniel-zid:
> I addressed all requests. @daniel-zid can you also refactor the get_Aff function to the same layout of the get_Veff function? I don't have a test case so it would be difficult for me.

I'm changing it, but all the code is going to be repeated. Should we put a switch somewhere? Otherwise we will have get_Veff_array and get_Aeff_array, export_Veff and export_Aeff, and they will do exactly the same.

daniel-zid:
Why do we request the minimum energy to be equal to the maximum energy? I think this is not useful for lots of applications. In fact, it is most of the time better to have the effective volume within an energy bin than only for a given point. In any case, none of them should be discouraged by our code.

daniel-zid:
The export method fails for json files. It looks like some keys are not valid:

TypeError: keys must be str, int, float, bool or None, not bytes_

daniel-zid:
Actually, a list of dictionaries tries to be exported and that fails with the json module, but with the yaml file we also get a weird output.

I've put the methods for Aeff, which are a copy of the Veff methods. Should we merge them? All in all, they only have one variable that is different (area vs volume).

cg-laser:
yes I guess we can merge the export functions. Do you also get the export error for Veff? I tested it and on my system it works fine. @daniel-zid can you send me a hdf5 file that fails so I can test it?

cg-laser:
the bug Daniel saw is because he ran simulation with python2 where the trigger names ended up at byte strings that can't be properly encoded. 
I unified the export function to work with Aeff and Veff. @daniel-zid @anelles ok now?

daniel-zid:
It looks alright to me!

daniel-zid:
Why do the checks fail?

cg-laser:
because I split it off the saving-module-sequence branch (and want to merge it into this branch). This branch fails here because it needs a certain NuRadioReco branch to run. I made the adjustments already on the saving-module-sequence where the tests don't fail anymore.

PR #134, opened 2019-09-18
Fix tests (master had also nur file comparison, missed that in the branch)

cg-laser:
also add Moore's Bay test


cg-laser:
I saw before that the differences for MB are slightly higher because the uncertainty doubles with every bottom reflection. The travel distance is just an analytical formula, but this equation is numerically not very stable... Not sure why. And yes, would be good to find out what causes these differences.

PR #136, opened 2019-09-18
Fixing the multiple_triggers key for multiple stations

daniel-zid:
The multiple_triggers key was not created for every station group when there were multiple stations. This pull request fixes it.


cg-laser:
maybe just clean up the code (remove the commented out code parts)

daniel-zid:
You're right.

daniel-zid:
Done.

Yes, adding a multiple station example may be a good idea.

PR #138, opened 2019-09-18
Fix multi station simulation, add test case

cg-laser:
This pull request fixes a bug when multiple stations were simulated. It was not checked if a station has any efields. Now, we check per station if we need to calculate a detector description at all. 

2 station simulation also added to test suite. Reference files stay the same because 2nd station never triggers and the comparison script anyway only compares station 101.


cg-laser:
I took the opportunity to also add a test for a simulation with noise. We can't verify the output, as it is random, but we can still check that it runs at least. 

Thereby I encountered a problem with the current way it was done. The new way is
```
if self._is_simulate_noise():
            max_freq = 0.5 / self._dt
            norm = self._get_noise_normalization(self._station.get_id())  # assuming the same noise level for all stations
            Vrms = self._Vrms / (norm / (max_freq)) ** 0.5  # normalize noise level to the bandwidth its generated for
            channelGenericNoiseAdder.run(self._evt, self._station, self._det, amplitude=Vrms, min_freq=0 * units.MHz,
                                         max_freq=max_freq, type='rayleigh')
```
This might be annoying for users. Therefore see also discussion in #139

christophwelling:
Can't we set the random seed? Then the noise should always be the same.

PR #142, opened 2019-09-20
add function to calculate depth averaged index of refraction

cg-laser:


cg-laser:
it is something that I needed, and I thought it doesn't hurt having such a convenience function in the code.

PR #145, opened 2019-09-27
implement ARZ Askaryan model via convolution

cg-laser:


cg-laser:
It can't be something explicit. The Cherenkov angle/viewing angle is not well defined. It must come out naturally

daniel-zid:
How come it's not well defined? It's the angle that makes the argument in the RAC switch signs.

bhokansonfasig:
I think we've decided that the timing is right inside the Cherenkov angle via emails, in which case I think the waveforms make sense the way they are now (without an additional factor of -1). Considering this from an argument of continuity, it makes more sense to me that the pulses would always swing upwards and then downwards on either side of the Cherenkov cone (see attached plot which I believe is the current behavior). If there were an additional factor of -1 inside of the Cherenkov angle there would be some sort of discontinuity at the Cherenkov angle, which I guess is possible but seems less physical to me.

sample_waveforms

cg-laser:
yes, your right. When I add
```
    if(dz < 0):
        convolution = np.flip(convolution, axis=1)
```
I get almost the same waveforms as in the integral approach, but there is still a small residual time shift. I think this is because of the downsampling. I did the flipping after the downsampling. I think we need to do it before, but then the downsampling/cutting needs to be adjusted.

cg-laser:
> I think we've decided that the timing is right inside the Cherenkov angle via emails, in which case I think the waveforms make sense the way they are now (without an additional factor of -1). Considering this from an argument of continuity, it makes more sense to me that the pulses would always swing upwards and then downwards on either side of the Cherenkov cone (see attached plot which I believe is the current behavior). If there were an additional factor of -1 inside of the Cherenkov angle there would be some sort of discontinuity at the Cherenkov angle, which I guess is possible but seems less physical to me.
> 
> sample_waveforms

what did you do to change the time behavior in your code?
I think the problem is this line `z_Q_vals = np.arange(n_Q) * np.abs(dz)` where you always use the absolute number of the `z_to_t` relation.

bhokansonfasig:
Yes, I'm still looking into this, but it is almost certainly something to do with either that absolute value or the `z_to_t` factor

cg-laser:
ok cool thanks! My guess is that you need to flip the convolution but then adjust the cutting, because the end it the beginning now and the other way around.

anelles:
@daniel-zid can you also post your plots concerning the convolution here, once they are ready? It is probably a good place to document the discussion about near-field effects.

daniel-zid:
Sure! These are some of the files that can be produced executing the T08_check_farfield.py file. We show on these figures the comparison between the convolution and the integral method. The red lines indicate the frequency at which the eta (or Fresnel) parameter equals 0.25 for a given distance. This parameter should be small in all our frequency range in order for the far-field convolution to work properly.

[Conv_vs_int_45.0deg.pdf](https://github.com/nu-radio/NuRadioMC/files/3804414/Conv_vs_int_45.0deg.pdf)
[Conv_vs_int_57.0deg.pdf](https://github.com/nu-radio/NuRadioMC/files/3804415/Conv_vs_int_57.0deg.pdf)
[Conv_vs_int_65.0deg.pdf](https://github.com/nu-radio/NuRadioMC/files/3804416/Conv_vs_int_65.0deg.pdf)

anelles:
Everyone, have we come to a conclusion regarding this? 
Should we add this only for far-field cases or not at all?

daniel-zid:
I think it can be useful to have the convolution as a limiting case. We don't know yet if it's possible to use the convolution or a similar method for the near field, but we should give ourselves more time to find a way.

bhokansonfasig:
I haven't had a chance to try to figure out a convolution for the near-field case in the last month. I'm hoping to get back to exploring this in January, but only time will tell if it's possible. I agree with Daniel and don't want to give up on it quite yet

cg-laser:
I would leave this PR open for now. The far field convolution is implemented properly, but the near field experimental convolution not working correctly and would need to be removed if we wanted to merge now.

cg-laser:
closing this PR. Accuracy is more important than speed. The near-field solution is relevant for many geometries and can't be neglected. 
Furthermore, speed improvements via numba made the ARZ module already much faster.

PR #147, opened 2019-10-02
logger.INFO option added when calling script

a-anker:
Seems to work, but let me know if there is anything else I should change or add like maybe a debug option too?


a-anker:
Changes have been made and it worked when I tested it again.

PR #148, opened 2019-10-02
fix ARA limit, add ARIANNA limit

cg-laser:
The ARA limit was a factor of 2 too good. The plot in the paper has a data point every half decade in energy but the scaling is for a full decade energy bin.


cg-laser:
at this point we should also clean up the module. We have several files doing the same.

PR #149, opened 2019-10-04
initialize array of  number of reflections and reflection case as int

cg-laser:
use n_reflections in get_ray_path function

small leftovers from introducing multiple reflections of the bottom


cg-laser:
hmm you're right. It doesn't crash but the multiplications converts the array into nans. 

So a better initialization is with -1 which does not make any sense. 

I want it to be integers because the ray tracing module expects this value to be an int.

anelles:
@cg-laser, now we need to update the reference files. Since -1 is no longer nan.

PR #150, opened 2019-10-08
MB attenuation length, improved ray tracing precision

cg-laser:
This pull request fixes a small bug in the depth dependence of the Moore's Bay attenuation length, leading to ~10% larger attenuation lengths. 

I also merged the focusing branch into this branch. 

I also increased the (C++) ray tracing precision from 1e-6 to 1e-9. I couldn't see any difference in speed but the results changed slightly (up the ~1e-6 as expected). However, this sometimes leads to differences in amplitudes of a few percent. I adapted the tests accordingly.


cg-laser:
waiting on focusing branch to merge first...

anelles:
The fact that the tests are failing seems to be another plattform dependent problem of the ray-tracer. The built-bot has it as Mismatch: 0.05%, the tests run fine on my system, which is I assume since I am also running MC as @cg-laser 
What do we do about this? I don't like mismatches between plattforms.

anelles:
Never mind. It seems that we broke all tests some time earlier.

cg-laser:
So this particular test fails because I updated the ray tracing precision but did not adjust the test yet. 
They probably run fine on your system because you didn't recompile the ray tracing code. This is a little annoying, I know, but the C++ extension is only compiled automatically if it doesn't exist, and there is now automated way implemented of detecting when the source code has changed. So one just needs to delete the *.so file, and then it gets recompiled automatically.

anelles:
Somehow, merging the master was a bit tricky. There will be a bunch of commits to get this branch back to work.

anelles:
So, @cg-laser with the different precision in the ray tracer, the differences between plattforms got seriously worse. We have not relative errros up to 0.009 in amplitudes. This is almost a percent and I think this is not acceptable. 
I don't know how this ray tracer works, but we need to address this.

cg-laser:
Did you merge the new reference files into this branch? 
Interestingly only the SP simulation including noise fails due to amplitude differences. Did we change anything that could effect the seeds?

cg-laser:
I fixed the tests that they actually loop through the events. Then, it shows correctly that only one trace is different. 
I can also exactly reproduce the built bot on my machine. I think this error occurs because the ray tracing precision is more accurate. The difference in C0 is only 6.24937725e-07, so below the 1e-6 precision of the setting used to generate this file. But now the precision is 1e-9. So apparently, 1e-6 is not enough for all cases and we should switch to 1e-9.

anelles:
Now we have a problem with the signal propagation test. And the South Pole tests fail on my system.

cg-laser:
> Now we have a problem with the signal propagation test. And the South Pole tests fail on my system.

did you recompile the ray tracer? What exactly is failing?

anelles:
We actually never propagated the error from the first test ... Let's see what will happen now. My mistake.

cg-laser:
the signal propagation hasn't been executed before at all. This is likely also just because of the better ray tracing precision. The reference file has not yet been updated to the new values.

anelles:
> the signal propagation hasn't been executed before at all. This is likely also just because of the better ray tracing precision. The reference file has not yet been updated to the new values.

Ok, I agree. Can you update the file? We don't seem to have the trivial file comparison here, but onlly memory with reference.

cg-laser:
I guess I can merge this branch now?

PR #155, opened 2019-11-05
Decay muons

daniel-zid:
The code has been refactorised, the units are now handled more properly and there is a class with private methods that use Proposal units, so as to avoid units problems. The muons coming from the decay of taus are now propagated and their showers stored.

### DIFF
--- a/NuRadioMC/EvtGen/.gitignore
+++ b/NuRadioMC/EvtGen/.gitignore
@@ -1,0 +1,1 @@
+config*json
--- a/NuRadioMC/EvtGen/NuRadioProposal.py
+++ b/NuRadioMC/EvtGen/NuRadioProposal.py
@@ -16,7 +16,7 @@ about are the following:
 - Brems: a bremsstrahlung photon
 - DeltaE: an ionized electron
 - EPair: an electron/positron pair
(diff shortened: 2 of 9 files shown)


cg-laser:
Ok then its fine. I was worried that we need some heavy pre calculations to be able to use proposal. 

One solution that one typically does is to have a "*.config.sample" file. The user is then supposed to make a copy to "*.config" where he adjusts the paths. The "*.config" files will not be tracked by git. This way every user can have their custom paths in the config file. 
The problem now is that I would need to adjust the path, but then I would always have an uncomitted change. If I wanted to change some other setting and commit it, it would not be possible.

daniel-zid:
I've deleted the config files and added the config.sample files, together with comments and an error message that explains the way they should be changed. Let me know if you like it.

cg-laser:
can you add the config files to a ".gitignore" file? You might need to create it first. So that the user config files do not show up in git.

daniel-zid:
I did not know one could do that. Tell me if that's what you wanted. I've included a .gitignore file for all the config*json files.

cg-laser:
looks good to me. Now the *.json files should not show up when you do a `git status .`

PR #157, opened 2019-11-14
Declination bands

daniel-zid:
NuRadioMC can now generate events for fixed declination bands and transform them to horizontal coordinates given the observer's latitude and longitude. I've decided to implement a simple function to get the zenith and azimuth from the declination, right ascension and observer's location because astropy's SkyCoord was really slow, even when operations were vectorised. I've also thought about implementing an EarthLocation class, but since the only things we need are observer latitude and longitude, this might be overkill, and for that the user can employ the equivalent astropy class if necessary.

### DIFF
--- a/NuRadioMC/EvtGen/generator.py
+++ b/NuRadioMC/EvtGen/generator.py
@@ -688,6 +688,7 @@ def get_zenith_azimuth(declination, right_ascension, latitude, longitude, GST=0)
     """
 
     check_array(declination/units.deg, -90, 90, 'Declination', 'degrees')
+    check_array(right_ascension/units.deg, -180, 180, 'Declination', 'degrees')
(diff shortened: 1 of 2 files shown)


daniel-zid:
Thanks to Cosmin I realised that we must change the way draw our arrival directions. The arrival directions must now be weighted using the projected area for the detector, which means that this pull request for declinations bands must also be changed.

daniel-zid:
The random angular distributions have to be changed in NuRadioMC so that they are now proportional to the projected area. This means that all 4pi simulations should be reevaluated, but also that a branch like this one has no purpose anymore, since it is much more complicated to draw a fair distribution in equatorial coordinates than to run a simulation in horizontal coordinates with an appropriate angular distribution and then reshuffle from horizontal to equatorial using a toy Monte Carlo. I am closing this pull request.

PR #158, opened 2019-11-18
New distribution for zeniths that includes projected area.

daniel-zid:
This new pull request implements a correct way of drawing zenith arrival directions. The function draw_zeniths returns zenith directions according to a sin(theta)*projected_cylinder_area distribution, which has a McDonald's-like M shape from 0 to 180 degrees.

Remember that the way of creating events is changed and the tests are going to fail a lot. We should address that when we all agree the code is ok.

### DIFF
--- a/NuRadioMC/utilities/Veff.py
+++ b/NuRadioMC/utilities/Veff.py
@@ -154,6 +154,10 @@ def get_Aeff_proposal(folder, trigger_combinations={}, station=101):
             fin.attrs['phimax']
         dZ = fin.attrs['zmax'] - fin.attrs['zmin']
         area = np.pi * (rmax ** 2 - rmin ** 2)
+        # The used area must be the projected area, perpendicular to the incoming
+        # flux, which leaves us with the following correction. Remember that the
+        # zenith bins must be small for the effective area to be correct.
+        proj_area = area * 0.5 * (np.abs(np.cos(thetamin))+np.abs(np.cos(thetamax)))
         V = area * dZ
         Vrms = fin.attrs['Vrms']
 
@@ -174,7 +178,7 @@ def get_Aeff_proposal(folder, trigger_combinations={}, station=101):
         else:
             for iT, trigger_name in enumerate(trigger_names):
                 triggered = np.array(fin['multiple_triggers'][:, iT], dtype=np.bool)
-                Aeff = area * np.sum(weights[triggered]) / n_events
+                Aeff = proj_area * np.sum(weights[triggered]) / n_events
                 Aeff_error = 0
                 if(np.sum(weights[triggered]) > 0):
                     Aeff_error = Aeff / np.sum(weights[triggered]) ** 0.5
@@ -228,7 +232,7 @@ def get_Aeff_proposal(folder, trigger_combinations={}, station=101):
                     mask = np.array([sol[i, values['ray_channel'], max_amps[i]] == values['ray_solution'] for i in range(len(max_amps))], dtype=np.bool)
                     triggered = triggered & mask
 
-                Aeff = area * np.sum(weights[triggered]) / n_events
+                Aeff = proj_area * np.sum(weights[triggered]) / n_events
 
                 if('efficiency' in values.keys()):
                     SNReff, eff = np.loadtxt("analysis_efficiency_{}.csv".format(values['efficiency']), delimiter=",", unpack=True)
@@ -237,7 +241,7 @@ def get_Aeff_proposal(folder, trigger_combinations={}, station=101):
                     if('efficiency_scale' in values.keys()):
                         As *= values['efficiency_scale']
                     e = get_eff(As / Vrms)
-                    Aeff = area * np.sum((weights * e)[triggered]) / n_events
+                    Aeff = proj_area * np.sum((weights * e)[triggered]) / n_events
 
                 out['Aeffs'][trigger_name] = [Aeff, Aeff / np.sum(weights[triggered]) ** 0.5, np.sum(weights[triggered])]
         Aeff_output.append(out)
--- a/NuRadioMC/utilities/Veff.py
+++ b/NuRadioMC/utilities/Veff.py
@@ -264,7 +264,7 @@ def get_Veff_water_equivalent(Veff, density_medium=0.917 * units.g / units.cm **
     return Veff * density_medium / density_water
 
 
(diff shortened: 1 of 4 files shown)


cg-laser:
Thanks a lot for working on that @daniel-zid . Can you add a plot to this pull request how the distribution looks like for typical cylinders for SP (4x2.7km) and Mooresbay (4km x 600m)?

Where does the formula come from? Can we add a reference? 

I think the best way is to simulate the zenith distribution correctly right from the beginning, but can't we just reweigh existing simulations? Wouldn't the weight depending on the zenith angle just be the ratio between the zenith angle histogram with/without weighting?

cg-laser:
I implemented a "on-the-fly" correction for existing simulations. 

The correction to the weight depends strongly on geometry. Here one example for the full zenith range. The correction to the weight would be the inverse. 
zenith_weights

I did simulations already for cos(zenith)=0.1 zenith bins. So the effect is relatively small:
zenith_weights_Veff

daniel-zid:
> Thanks a lot for working on that @daniel-zid . Can you add a plot to this pull request how the distribution looks like for typical cylinders for SP (4x2.7km) and Mooresbay (4km x 600m)?
> 
Sure
> Where does the formula come from? Can we add a reference?
> 
> I think the best way is to simulate the zenith distribution correctly right from the beginning, but can't we just reweigh existing simulations? Wouldn't the weight depending on the zenith angle just be the ratio between the zenith angle histogram with/without weighting?



> I implemented a "on-the-fly" correction for existing simulations.
> 
> The correction to the weight depends strongly on geometry. Here one example for the full zenith range. The correction to the weight would be the inverse.
> zenith_weights
> 
> I did simulations already for cos(zenith)=0.1 zenith bins. So the effect is relatively small:
> zenith_weights_Veff

Wait a second. Why do you use the inverse? The corrected weights for horizontal events should be less than one, since we were overestimating the flux. I think we should use N1/N2 and not the other way around! Keep in mind that you can also implement an analytic function for the weight, which is the projected area, as long as the average weight for the whole range is 1.

NuRadioMC/utilities/Veff.py
```diff
@@ -375,6 +377,28 @@ def get_Veff(folder, trigger_combinations={}, station=101):
...
+                bins = np.linspace(thetamin, thetamax, 100)
+                N1, bin_edges = np.histogram(zeniths1, bins=bins)
+                N2, bin_edges = np.histogram(zeniths2, bins=bins)
+                X = (bins[1:] + bins[:-1]) * 0.5
+
+                def get_weights(zeniths):
+                    weights = np.zeros_like(zeniths)
+                    for i, z in enumerate(zeniths):
+                        mask = np.argmin(np.abs(z - X))
+                        weights[i] = (N2 / N1)[mask]
```
daniel-zid: I think you should use N1/N2 here, or directly use the normalised projected area.

daniel-zid:
> Thanks a lot for working on that @daniel-zid . Can you add a plot to this pull request how the distribution looks like for typical cylinders for SP (4x2.7km) and Mooresbay (4km x 600m)?
> 
Here it is. The blue line is for the South Pole and the orange, Moore's Bay. It makes sense because the projected area at Moore's Bay at the horizon is smaller.
distribution_SP_MB

> Where does the formula come from? Can we add a reference?
> 
I thought it was simple enough not to look for a reference. I can ask Jakob if they have a reference for this kind of simulations. I checked with him that this M shape is what they use.
> I think the best way is to simulate the zenith distribution correctly right from the beginning, but can't we just reweigh existing simulations? Wouldn't the weight depending on the zenith angle just be the ratio between the zenith angle histogram with/without weighting?
That's a way of doing it. The other way is just using the projected area (not the zenith distribution, careful) correctly normalised to the angular range Delta theta, so that the mean weight is 1.

NuRadioMC/utilities/Veff.py
cg-laser: yes I think your right: For Moores Bay the projected area at the horizon is smaller than for the zenith. So with a ~sin(theta) distribution I have too many events at the horizon. Which means I need to weight them down.

NuRadioMC/utilities/Veff.py
```diff
@@ -383,25 +383,33 @@ def get_Veff(folder, trigger_combinations={}, station=101, correct_zenith_sampli
...
+
+                    def integral(theta, R, d):
+                        """
+                        integral of Aproj
+                        """
+                        return (-np.pi * R ** 2 * 0.5 * np.cos(theta) ** 2 * np.sign(np.cos(theta)) + 0.5 * 2 * R * d * (theta - np.sin(theta) * np.cos(theta)))
+
+                    # calculate the average value of Aproj within the zenith band -> int(Aproc(theta) dcostheta)/int(1, dcostheta)
+                    norm = integral(thetamax, R, d) - integral(thetamin, R, d)  # int(Aproc(theta) dcostheta)
+                    norm /= (np.cos(thetamin) - np.cos(thetamax))  # int(1, dcostheta)
```
daniel-zid: Sorry that I said it should be normalised to Delta theta. It must be integrated and normalised as a function of cos(theta), as you have done here. Thank you for implementing this.

anelles:
The tests currently fail, since the new distribution of zenith angles, leads to a deviation in V-eff. It goes from an average of 4.7 km^3 to roughly 3.5 km^3
I will update the reference numbers, but everyone should be aware of this.

cg-laser:
@anelles how did you calculate the new mean Veff value? You also reduced the sigma substantially which looks odd to me. What I did in the past, I ran the script 100x to properly calculate the mean and STD. 

We set the seed to the same value so that there should not be any statistical variation as long as the event generation is not touched. However, this test was mostly to test changes in the event generation, which means which such a small sigma parameter, the test is likely to fail even if everything is ok within statistics.

anelles:
This is literally what I did. Only that I 'just' ran it 50 times. I think you may have just had a missing zero in your sigma before. It triggers roughly  120 +/- 5 events per 1000. You don't get such a large variation in uncertainty as you had before. feel free to run this again though.

cg-laser:
sry that I didn't follow up earlier: You had the seed always set to the same value don't you? Therefore you don't get any fluctuations. The only fluctuations that come in are from the Alvarez2009 Askaryan model that has some randomness for LMP showers.

PR #162, opened 2019-12-03
remove unused arguments from get_filter call

cg-laser:
fix NuRadioMC for changes in NuRadioReco https://github.com/nu-radio/NuRadioReco/pull/163


anelles:
I think this needs a larger fix, as this now breaks example 5, where the hardwareResponseIncorporator is called, which still needs those arguments. Which way do we want to go?

cg-laser:
:-( not I remember why I implemented it like this in the first place... I think I should revert the NuRadioReco changes... otherwise the `get_filter` methods can't be used with the same syntax

PR #163, opened 2019-12-04
Robust tests

anelles:
Issue being addressed:

Minor fixes so that tests run on local machine without errors. Especially allowing to overwrite files and existing folders.


anelles:
Since nothing happend here for 13 days. Can someone review this pull request, please? It is also a good start for those of you, who have never reviewed before.

anelles:
All NuRadioMC examples have it now. By default it doesn't overwrite, just as it did before. So nothing should change for the NuRadioReco examples. But we can do it too, yes.

anelles:
In order to clean this up a bit, I will merge this, despite tests failing and then fix the tests on the try-fix-master branch.

PR #164, opened 2019-12-17
Propagate error, so we actually see things break.

anelles:
Trying to fix the master and back-track where this went wrong.


anelles:
With an accuracy of 0.05% the results now match. Please review so we can merge this into the master and move forward.

PR #167, opened 2020-01-15
place pulse into the center of the trace

cg-laser:
before the pulse was placed at 50ns into the trace. However I think it is much save to place it into the center of the trace as also done by the ARZ code.


anelles:
So the tests really fail now. @cg-laser  are you sure that you move the pulse correctly? Maximum amplitudes are totally different.

cg-laser:
Well the Alvarez models don't really have timing, so yes it's different 
than in ARZ. but in ARZ the pulse is centered around the center of the 
trace + additional time delays. Directly on the cherenkov cone it is 
centered in the trace.


Am 16.01.20 um 06:06 schrieb daniel-zid:

cg-laser:
I found the reason for the discrepancy. It's the sampling rate. The traces are downsampled to 2GHz before the maximum amplitude is calculated. Because of the different position in the trace, the pulses are differently sampled leading to differences in the maximum signal amplitude. 
When I upsample the traces to 20GHz, the files are identical.

cg-laser:
We should calculate the signal properties before we downsample the traces, but this would require some changes. No downsampling in the user code, but automatically in the `simulation.py` file. This is needed because the signal properties are calculated automatically after the user code.

cg-laser:
I updated the reference files and changed the code in general so that the amplitudes get calculated on the internal simulation sampling rate. Also the traces get cut to the same length (I added the triggerTimeAdjuster module). The also the nur files are identical after the change. 
The noise files are of course different because the pulse is at a different position when noise is added, so different interference.

PR #169, opened 2020-01-23
add option to pass a list of event ids to the main simulation class

cg-laser:
All events that are not in the provided list of event-ids will be skipped.


anelles:
Wasn't this how most of us handled it anyway?

PR #172, opened 2020-01-27
Vertex times

daniel-zid:
Now each interaction has a 'vertex_time' variable that measures the time from the first neutrino interaction.


daniel-zid:
The test fails because the reference files don't contain the new array. I have to change that.

daniel-zid:
There seems to be something wrong with the test scripts. The T02 files use as input the output reference file! What happened here?

anelles:
@daniel-zid the code runs on the built-bot now, until the key error vertex_times. Please add the exception as discussed.

NuRadioMC/EvtGen/generator.py
```diff
@@ -895,6 +897,7 @@ def generate_surface_muons(filename, n_events, Emin, Emax,
 
     data_sets["event_ids"] = np.arange(n_events) + start_event_id
     data_sets["n_interaction"] = np.ones(n_events, dtype=np.int)
+    data_sets["vertex_times"] = np.zeros(n_events, dtype=np.int)
```
cg-laser: why is this of type int? Shouldn't it be float?

NuRadioMC/EvtGen/generator.py
```diff
@@ -1222,6 +1228,7 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
     data_sets["event_ids"] = np.arange(n_events) + start_event_id
     logger.debug("generating number of interactions")
     data_sets["n_interaction"] = np.ones(n_events, dtype=np.int)
+    data_sets["vertex_times"] = np.zeros(n_events, dtype=np.int)
```
cg-laser: same here, shouldn't the type be float?

NuRadioMC/simulation/simulation.py
```diff
@@ -551,6 +554,12 @@ def run(self):
 
                         electric_field = NuRadioReco.framework.electric_field.ElectricField([channel_id], self._det.get_relative_position(self._sim_station.get_id(), channel_id))
                         electric_field.set_frequency_spectrum(np.array([eR, eTheta, ePhi]), 1. / self._dt)
+                        # Trace start time is equal to the interaction time relative to the first
+                        # interaction plus the wave travel time.
+                        if hasattr(self, '_vertex_time'):
+                            trace_start_time = self._vertex_time + T
+                        else:
+                            trace_start_time = T
                         electric_field.set_trace_start_time(T)
```
cg-laser: shouldn't this be `electric_field.set_trace_start_time(trace_start_time)` ?

NuRadioMC/EvtGen/generator.py
daniel-zid: Yes.

NuRadioMC/EvtGen/generator.py
daniel-zid: Yes, again.

NuRadioMC/simulation/simulation.py
daniel-zid: I don't know what happened to me while writing this code, honestly...

daniel-zid: I'll change everything tomorrow morning.

daniel-zid: The changes have been addressed.

cg-laser:
what about the trigger modules? The trigger time so far is relative to the trace start time. I'm not sure how to implement this properly. The trigger is a per station object. But the trace start time is per channel. However, we require that the trace start time is the same for all channels in the trigger module. But in principle we could have modules later on that change that. Or even change all trace start times in the same way but then don't adjust the trigger times correctly. 

Actually even now the triggerTimeAdjuster module is buggy for cases where multiple triggered where issued. Because the trigger time is only adjusted for the trigger was selected. 

I think it is safest if we have a global trigger time. One can always find the correct position in the trace by subtracting the trace start time from the global trigger time. What do you think? 
The required changes would be: All trigger modules need to add the trace start time to the trigger time.

daniel-zid:
I find it a bit confusing already that what it's called trace_start_time in
NuRadioMC for the electric field is not the trace_start_time for each
channel, but this kind of thing can't be helped sometimes.

We could change the trace start time for each channel, but that can cause
major disruption for a while, and I don't see any advantage to having a
different trace start from a simulation point of view. Maybe to account for
cable delays?

Anyway, I'm going to change the trigger modules using the global channel
start time.

cg-laser:
> I find it a bit confusing already that what it's called trace_start_time in NuRadioMC for the electric field is not the trace_start_time for each channel, but this kind of thing can't be helped sometimes. 

what do you mean by that?

cg-laser:
and the changelog.txt is wrong. It still shows the old version. Although the master has already the updated version (https://github.com/nu-radio/NuRadioMC/pull/173/commits/761b2ddc90ccaea668d2387c913e4e2e9f79c326)
Can you merge the master into this branch?

daniel-zid:
> > I find it a bit confusing already that what it's called trace_start_time in NuRadioMC for the electric field is not the trace_start_time for each channel, but this kind of thing can't be helped sometimes.
> 
> what do you mean by that?

Nothing, I just didn't remember that this was the electric field start time and that the efield to voltage converter in NuRadioReco created a common channel trace start time. Ignore it, I hadn't had breakfast yesterday morning.

daniel-zid:
> and the changelog.txt is wrong. It still shows the old version. Although the master has already the updated version ([761b2dd](https://github.com/nu-radio/NuRadioMC/commit/761b2ddc90ccaea668d2387c913e4e2e9f79c326))
> Can you merge the master into this branch?

Sure!

changelog.txt
```diff
@@ -22,6 +22,7 @@ new features:
 - Event generation using PROPOSAL lepton propagation available
 - fft and units utilities are removed. These should be taken from NuRadioReco from now on.
   IMPORTANT: Note that in the fft the frequency spectrum is now given per frequency instead of per bin
+- New data set array 'vertex_times' contains the time elapsed from the first interaction to the current interaction
```
cg-laser: this is at the wrong line. It should be under version 1.1.1

daniel-zid:
I've changed the log, but maybe not the way I was supposed to do it. Can you take a look before merging? Thank you!

daniel-zid:
The build has been repaired. I have to change the .travis file before merging. Everyone for merging? Should I wait?

cg-laser:
why did you need to touch the reference files in the first place? Weren't the test already working?

daniel-zid:
No, for whatever reason, they were not working from NuRadioReco. The differences were minimal and due to ray tracing, I think. Although I don't understand why.

If you want to take a look at what went wrong or want to create your own files, please go ahead. I am a bit lost.

cg-laser:
What was the error? Wasn't this just the fact that a new parameters was added? And the tests don't allow that at the moment. We should rather make the tests more robust, that only the parameters that are in the reference file are checked.

daniel-zid:
At first it was that the new data sets were not included in the NuRadioReco, but then when I modified those tests, the SingleEvents tests started failing, and those are taken from NuRadioMC. The differences were small, and I did not know where they were coming from. You can see here that after renewing the trigger tests, the SingleEvents started failing:

https://github.com/nu-radio/NuRadioReco/runs/436393535

No idea. The ray tracer is not completely deterministic, right? Maybe the seed has been changed because of the different order of operations?

daniel-zid:
No, that does not make any sense. I have no idea about what happened. Any ideas?

Look at this job: https://travis-ci.com/nu-radio/NuRadioReco/jobs/285491734
The simulation.py was not changed in any substantial way, and the reference files were the same. What's going on?

cg-laser:
The differences are huge, it's almost 100% difference. But it can't be the ray tracing. The test doesn't complain about that.

daniel-zid:
Yes, you're right. Then I don't understand anything. 0.99 of difference without changing any related test files. This is weird...

cg-laser:
I don't understand the print out 
```attribute SNRs not almost equal
Reconstruction of SNRs of event 1 does not agree with reference (relative error: 0.9998297691743911)```
which file is producing such an output? I think T03validate.py is being executed. But it never prints something like that. But I really think we should find out what causes the test to fail and not just update the references. We might miss something important. We should only update the references if we understood the difference.

**Comment by daniel-zid** [comment]
Yes, I agree

**Comment by daniel-zid** [comment]
> I don't understand the print out
> 
> ```
> Reconstruction of SNRs of event 1 does not agree with reference (relative error: 0.9998297691743911)```
> which file is producing such an output? I think T03validate.py is being executed. But it never prints something like that. But I really think we should find out what causes the test to fail and not just update the references. We might miss something important. We should only update the references if we understood the difference. 
> ```

No, it's T04validate...py

**Comment by cg-laser** [comment]
I checkout out the vertex branch of NuRadioMC and Reco at the point before updating the reference files. 
The test fails locally:
```
NuRadioMC cglaser$ python NuRadioMC/test/SingleEvents/T04validate_allmost_equal.py NuRadioMC/test/SingleEvents/1e18_output.hdf5 NuRadioMC/test/SingleEvents/1e18_output_reference.hdf5
Testing the files NuRadioMC/test/SingleEvents/1e18_output.hdf5 and NuRadioMC/test/SingleEvents/1e18_output_reference.hdf5 for (almost) equality
Reconstruction of SNRs of event 7 does not agree with reference (relative error: 0.9998413486976497)

 attribute SNRs not almost equal
NuRadioMC/test/SingleEvents/T04validate_allmost_equal.py:76: RuntimeWarning: invalid value encountered in true_divide
  max_diff = np.max(np.abs((arr1[i] - arr2[i]) / arr2[i]))
Reconstruction of maximum_amplitudes of event 7 does not agree with reference (relative error: 0.9998804518645715)

 attribute maximum_amplitudes not almost equal
Reconstruction of maximum_amplitudes_envelope of event 2 does not agree with reference (relative error: 0.09587810751436524)

 attribute maximum_amplitudes_envelope not almost equal
Reconstruction of maximum_amplitudes_envelope of event 7 does not agree with reference (relative error: 0.9998736739522912)

 attribute maximum_amplitudes_envelope not almost equal
```

However, the *nur file comparison works fine
```
Maximum difference between traces [mV] 3.3862292736083294e-17
Traces agree within 7 decimals, but not completely identical
Traces are identical
```
But this shouldn't be able to happen. If the nur files agree, then also the hdf5 files should be equal. 
The only difference is that the traces get downsampled to detector sampling rate (1GHz in this example) but the max. amplitudes get calculated on the internal resolution (5GHz).

PR #173, opened 2020-01-28
Arz2020

daniel-zid:
The new ARZ2020 model has been implemented. The new fits give almost the same result for EM showers, while for HAD showers, normalisation tends to be ~10% higher (**Note: this is not the case when including the correct energy factor for hadronic cascades in ARZ2020. This factor was hidden in ARZ2019. Both models are really similar**).


daniel-zid:
Here's a plot for some EM 1 EeV showers.
[2020vs2019_EM.pdf](https://github.com/nu-radio/NuRadioMC/files/4123831/2020vs2019_EM.pdf)

daniel-zid:
Same thing, for a HAD shower.
[2020vs2019_HAD.pdf](https://github.com/nu-radio/NuRadioMC/files/4123847/2020vs2019_HAD.pdf)

daniel-zid:
Christian, could you run the scripts you used for the NuRadioMC paper with the new model? We should be able to see that ARZ2019 and ARZ2020 differ only a bit for hadronic cascades.

NuRadioMC/SignalGen/ARZ/ARZ.py
```diff
@@ -809,69 +846,9 @@ def get_time_trace(self, shower_energy, theta, N, dt, shower_type, n_index, R,
...
-        
+
         trace2 *= self._library['meta']['R'] / R * rescaling_factor
-        
+
         if(output_mode == 'Xmax'):
             return trace2, Lmax
         return trace2
-
-
```
anelles: Why did you remove the __main__ example? This was an easy way to test things, right?

daniel-zid:
The test module and the files in the tests folder do the same, only better.
The main function was broken anyway, and the example was not really
useful...

NuRadioMC/SignalGen/ARZ/ARZ.py
daniel-zid: The test module and the files in the tests folder do the same, only better.
The main function was broken anyway, and the example was not really
useful...

daniel-zid:
I've forgotten to include another factor for hadronic showers that I think will get us closer to the previous results. I think that the 2020 model and the 2019 model are going to be indistinguishable for practical purposes.

daniel-zid:
Alright, I have included the energy factor for hadronic cascades for the ARZ2020 model. It was approximated as a constant value in the ARZ2019 model. As you can see, this makes the models almost equivalent for this case:
[2020vs2019_HAD.pdf](https://github.com/nu-radio/NuRadioMC/files/4127522/2020vs2019_HAD.pdf)

I think it's safe to assume that the 2020 and the 2019 models are almost the same for practical effects.

cg-laser:
because you restructured the code it is difficult to track the changes. Are the only real difference between the two models the different constants that are set in the `__set_model_parameter`?

And I see it correctly that the ARZ2019 test still runs fine and gives the correct result? So we can be sure that there is no bug in the restructuring? 

And did you add the bold comments later on? Then is makes sense ;-)

daniel-zid:
> because you restructured the code it is difficult to track the changes. Are the only real difference between the two models the different constants that are set in the `__set_model_parameter`?
> 
Yes, exactly.
> And I see it correctly that the ARZ2019 test still runs fine and gives the correct result? So we can be sure that there is no bug in the restructuring?
> 
That's the idea. The test is the same as before, so that means the ARZ2019 is the same as before. Changing the reference files would be cheating. But now that we know the ARZ2019 is the same and the ARZ2020 is similar, we can also include the ARZ2020 in the tests at some point.
> And did you add the bold comments later on? Then is makes sense ;-)
Yep :D

cg-laser:
ok cool, thanks for implementing it so fast and in such an elegant way. 
I agree, we know that the tests work, so we should also add the ARZ2020 model to the test before we merge this PR. 
And can you also add a note to the changelox.txt?

PR #174, opened 2020-01-29
Stops ARZ module from setting numpy.random seed by default

christophwelling:
I think having the module set a random seed by default is pretty dangerous If we split simulations up into multiple jobs, they might all easily end with the exact same event properties, noise, etc.


anelles:
I am never sure whether I understand seeds correctly. But do we then not need to set all examples to the same seed to work, if you remove the default?

christophwelling:
The examples all set a seed already, the ARZ module just overwrites it. The way it is currently done, the seed is set globally, since all numpy.random calls draw from the same object. That is exactly the problem with setting it in the ARZ module, since it may set a seed without the user actually wanting it to.
So the tests still run with a fixed seed, the ARZ SignalGen test just runs with a different one now.

anelles:
Ok. Let's give @cg-laser the chance to comment before we move further. The seed implementation was his doing, I believe.

cg-laser:
the thing is that we also want reproducibility. If we run the same simulation a second time, we should get the same result. 
Having the same initial seed for multiple jobs is not a problem because the input events are different, and we'll never have the same sequence of events triggering. 

But I agree that a more global handling of the seeds would be good. What we actually want is that the random sequence per module remains the same for reproducibility (or not if the user wants this for some reason). Now, one could add a another module with a random call and it would change the numbers of all succeeding calls. 

What we would need is some global random number class, where each module can get its own generator. And then maybe a global setting of the seed.

christophwelling:
If you put a
`import numpy.random`
`numpy.random.seed`
into your code anywhere, it will set the seed globally, because all calls to numpy.random call the same object. I actually found this issue because I noticed that I got the exact same noise every time I reran my script.
If i want this reproducibility, all I need to do is put a numpy.random.seed anwhere in my code. That is more transparent and consistent than having it in some module. For example, if I use the ARZ model, my code is reproducible (but only after the ARZ object is created), if I use any other model it is not.
Another option would be to give the ARZ class its own numpy random state:
`random = numpy.random.RandomState(0)`
and let it draw all random numbers from there. That way, the ARZ model would always be reproducible but not affect anything else.

cg-laser:
yes this is what we want. Every module should have it's own random object and we need to find a way how to set the seeds. Maybe we can do this again with some global random registry (singleton)?
I think we need to have only two options: All random seed for all modules, or fixed seed for all modules. This could be an option for the config file. Actually this is how we can implement it. Every module that uses random numbers gets an argument "seed" in its __init__ function which is either None (-> random seed) or a seed specified in the config file. 

We should start a new PR though because of the ARZ changes in the master branch.

christophwelling:
Okay, I am going to close this PR. Who wants to take this over? I've never done anything to the config files before...

cg-laser:
I can put this on my todo list

PR #175, opened 2020-01-29
Save sim config and detector

christophwelling:
Allows to store the simulation configuration and the detector description in the .nur file. Is is needed if we want to have a fancy 3D plot in the eventbrowser.


anelles:
Ok, this matches the NuRadioReco pull request, right?

christophwelling:
Yes, this goes with PR 183 in NuRadioReco, so I can't merge until someone approves that one

cg-laser:
I already approved it but you have a merge conflict that you need to resolve first.

PR #178, opened 2020-02-07
Sub events

daniel-zid:
This pull request introduces the event group ids and serves as a complement to the NuRadioReco branch of the same name.


NuRadioMC/EvtGen/generator.py
```diff
@@ -1026,6 +1026,8 @@ def get_inverse_spectrum(N, gamma):
 
     print("number of fiducial showers", len(data_sets_fiducial['flavors']))
 
+    data_sets_fiducial["event_ids"] = np.arange(len(data_sets_fiducial["event_group_ids"])) + start_event_id
```
cg-laser: was this a bug before, or why did you need to add this line?

daniel-zid: We treat the event_group_ids as what was the previous event_ids data set, since they are the same thing. event_ids represents a new dataset giving unique ids to each interaction vertex. It does not matter what number as long as it is unique. That's why I only assign event_ids to the fiducial events that are going to be simulated afterwards.

daniel-zid:
After talking to Christoph, we agreed that we needed:
1) An event_group_ids dataset with the same purpose as the previous event_ids. The previous event_ids jumped all over the place as well because some events were not in the fiducial volumes, so I did not think this would be a problem.
2) An event_ids dataset with unique tags for each event. It does not matter which one as long as it is unique, in fact one option was just to assign really long random numbers with a really small chance of repeating themselves.

I'm not sure if I follow why you think this scheme is wrong. Christoph has said that there might be a problem if event_group_ids reach larger numbers than event_ids, so I'm going to change that. The event_ids can be removed from the generator if necessary and created during simulation, but if it's going to be the same data set in the end, it does not hurt to create it upon generation, right?

daniel-zid:
Alright, now I have modified the event_group_ids data set so that all the integers are used, starting with start_event_id. They won't jump around anymore. As for the event_ids, they're just an arange starting with start_event_id and having the same number of entries as event_group_ids. With this ordering, we guarantee that the last event_id will be equal to or greater than the last event_group_id. For instance, we can have the following data sets:

event_group_ids:  [1, 1, 2, 3, 4, 5, 5]
event_ids:             [1, 2, 3, 4, 5, 6, 7]

So the events 1 and 2 are associated to the same event group, which is 1. The event 3 stands alone and belongs to the event group id number 2. And so on.

The data set event_ids can also be generated during simulation, if you prefer.

cg-laser:
I agree that technically, the event ids can be random numbers as long as they are unique, but this will be quite inconvenient. We should try to have them as consecutive as possible. And I see no reason why not to do so. The event_id_generator just needs to increment the (largest) id on every call. 

Therefore, we should not have event_ids in the NuRadioMC input files. Because we don't know the event_ids yet. Any event in the input file could be split up into two. With your scheme this would result in event_id sequences like [1, 20, 2,3,4, 21, 5, ...]. 

For the cases where we have multiple interactions per event_group_id in the input file, we have the n_interactions counter to correctly reference those events.

cg-laser:
There is another thing that we need to think through: How do we handle multiprocessing? I mean when we split up one input file into several small ones and do the simulation on many cores (the standard procedure). Each resulting nur files will likely have overlapping event ids -> we won't be able to merge the files, or analyze them jointly, which can be a nuisance. 
The only way around that is to have a different run id per file. This would require to give the hdf5 file an attribute 'run_id' or 'file_id'.

daniel-zid:
> I agree that technically, the event ids can be random numbers as long as they are unique, but this will be quite inconvenient. We should try to have them as consecutive as possible. And I see no reason why not to do so. The event_id_generator just needs to increment the (largest) id on every call.
> 
> Therefore, we should not have event_ids in the NuRadioMC input files. Because we don't know the event_ids yet. Any event in the input file could be split up into two. With your scheme this would result in event_id sequences like [1, 20, 2,3,4, 21, 5, ...].
> 

No, that's not what happens with my generator. They are always increasing and files are split according to event group ids.
> For the cases where we have multiple interactions per event_group_id in the input file, we have the n_interactions counter to correctly reference those events.

daniel-zid:
> I agree that technically, the event ids can be random numbers as long as they are unique, but this will be quite inconvenient. We should try to have them as consecutive as possible. And I see no reason why not to do so. The event_id_generator just needs to increment the (largest) id on every call.
> 
> Therefore, we should not have event_ids in the NuRadioMC input files. Because we don't know the event_ids yet. Any event in the input file could be split up into two. With your scheme this would result in event_id sequences like [1, 20, 2,3,4, 21, 5, ...].
> 

I don't understand why you say that. The event_ids in this generator are increasing and are as consecutive as they can get. Look at the explanation above. They start at start_event_id and end at start_event_id + total number of interactions...
> For the cases where we have multiple interactions per event_group_id in the input file, we have the n_interactions counter to correctly reference those events.

daniel-zid:
> > I agree that technically, the event ids can be random numbers as long as they are unique, but this will be quite inconvenient. We should try to have them as consecutive as possible. And I see no reason why not to do so. The event_id_generator just needs to increment the (largest) id on every call.
> > Therefore, we should not have event_ids in the NuRadioMC input files. Because we don't know the event_ids yet. Any event in the input file could be split up into two. With your scheme this would result in event_id sequences like [1, 20, 2,3,4, 21, 5, ...].
> 
> I don't understand why you say that. The event_ids in this generator are increasing and are as consecutive as they can get. Look at the explanation above. They start at start_event_id and end at start_event_id + total number of interactions...
> 
> > For the cases where we have multiple interactions per event_group_id in the input file, we have the n_interactions counter to correctly reference those events.

daniel-zid:
> I agree that technically, the event ids can be random numbers as long as they are unique, but this will be quite inconvenient. We should try to have them as consecutive as possible. And I see no reason why not to do so. The event_id_generator just needs to increment the (largest) id on every call.
> 
> Therefore, we should not have event_ids in the NuRadioMC input files. Because we don't know the event_ids yet. Any event in the input file could be split up into two. With your scheme this would result in event_id sequences like [1, 20, 2,3,4, 21, 5, ...].
> 
> For the cases where we have multiple interactions per event_group_id in the input file, we have the n_interactions counter to correctly reference those events.

I see what you mean, but then we have to find a way of splitting and numbering events within the simulation, and we still don't know all the cases and ways to split an event, right?

In any case, this branch's main purpose was being useful for plotting multiple-bang events. I had no intention of tackling the splitting of the events.

cg-laser:
but this goes hand in hand. 
I think we should just not specify an event_id in the input file. The event ids will be assigned during the simulation.

cg-laser:
there is yet another problem: If we split events, do we save both events in the output hdf5 file? This will lead to problems if we resimulate this event, i.e. using the output as the input file. Everytime we resimulate the event, it will double itself...

cg-laser:
this PR can be closed. We went for another solution that we implemented in the `internal-looping` branch.

PR #179, opened 2020-02-07
Random generators

cg-laser:


cg-laser:
This PR adds a global see option to the config file. The default is "1234" to have reproducible results. 
Also, each module that uses a random generator gets its own instance. 

The charge excessprofile that is used in ARZ and the shower energy is now saves as electric field parameters.

cg-laser:
keeping track of the seed is a little hacked, but I don't see any disadvantages. What do you think?

anelles:
Looks like this one also needs a fix that only parameters that are present are being compared.

cg-laser:
yes. I didn't write the test so I though maybe @christophwelling you can have a look?

christophwelling:
On the NuRadioReco pull request that you use here, we both agreed that adding these parameters to the E-field is not a good idea. So you should either remove these parameters or put them into the sim shower.

cg-laser:
agreed

cg-laser:
this PR fixes #182

anelles:
@christophwelling can you review this? You found the problem wit the random generator in the first place.

NuRadioMC/SignalGen/ARZ/ARZ.py
```diff
@@ -11,9 +11,19 @@
 import os
 import copy
 import logging
+import six
 logger = logging.getLogger("SignalGen.ARZ")
 logging.basicConfig()
 
+
+class Singleton(type):
```
christophwelling: I think it would be better to have this class in its own file in the utilities folder, preferably in NuRadioReco. It's not unlikely that it will be used somewhere else in the future, and if someone looks for it it will be hard to find here.

NuRadioMC/SignalGen/parametrizations.py
```diff
@@ -21,13 +21,16 @@ def set_log_level(level):
...
+_Alvarez2009_k_L = None
+
 
 def get_parametrizations():
     """ returns a list of all implemented parametrizations """
     return ['ZHS1992', 'Alvarez2000', 'Alvarez2009', 'Alvarez2012']
 
 
-def get_time_trace(energy, theta, N, dt, shower_type, n_index, R, model):
+def get_time_trace(energy, theta, N, dt, shower_type, n_index, R, model, seed=None, same_shower=False):
```
christophwelling: Could you add same_shower to the docstring?

NuRadioMC/simulation/simulation.py
```diff
@@ -139,6 +139,9 @@ def __init__(self, inputfilename,
                 new_cfg = merge_config(local_config, self._cfg)
                 self._cfg = new_cfg
 
+        if(self._cfg['seed'] is None):
```
christophwelling: Maybe add a comment describing what you are doing here? This could easily confuse someone

NuRadioMC/SignalGen/ARZ/ARZ.py
cg-laser: I'll change that once your NuRadioReco PR is merged that provides this utility

NuRadioMC/SignalGen/parametrizations.py
cg-laser: done

NuRadioMC/simulation/simulation.py
cg-laser: done

NuRadioMC/SignalGen/ARZ/ARZ.py
cg-laser: I added a meta class. I anyway had a corresponding NuRadioReco PR

cg-laser:
@christophwelling can you approve the PR?

PR #183, opened 2020-02-27
calculate weights for combining simulations of zenith angle bins

cg-laser:
If the full-sky effective volume is calculated from simulations of individual zenith angle bins, each bin needs to be weighted with the projected area of the cylinder (the target volume). The calculation of weights is now done within the Veff helper function.


daniel-zid:
I also had another comment. What if the different files events in different zenith angle bins, but each bin has a correctly isotropic flux? This case is not well handled, and if you use the get_weights function for that you would be somehow correcting within each bin a flux that's already isotropic in that bin. Do you see what I mean?

NuRadioMC/utilities/Veff.py
```diff
@@ -379,13 +380,15 @@ def get_Veff(folder, trigger_combinations={}, station=101, correct_zenith_sampli
         dZ = fin.attrs['zmax'] - fin.attrs['zmin']
         area = np.pi * (rmax ** 2 - rmin ** 2)
         V = area * dZ
+        # calculate the average projected area in this zenith angle bin
+        Aproj_avg = get_projected_area_cylinder_integral(thetamax, R=rmax, d=dZ) - get_projected_area_cylinder_integral(thetamin, R=rmax, d=dZ)
+        Aproj_avg /= np.cos(thetamin) - np.cos(thetamax)
+        out['Aproj'] = Aproj_avg
```
daniel-zid: Oh, I see where. Never mind.

daniel-zid:
I can't get the analytic formula to match a numerical calculation, I don't know what I'm doing wrong. Have you run any checks of the integral formula?

daniel-zid:
I've checked that the integral is correct. Sorry for the annoyance.

cg-laser:
thanks for thoroughly checking the implementation. I also compared a full-sky simulation to a 0.1 cos zenith bin simulation and I get the same results within statistical uncertainties. 

I added the bin size to the weights. Good suggestion. It should now work for unequal zenith angle bins unless there is an overlap (which one should anyway not do). 

Is it ok now?

daniel-zid:
It looks correct to me.

PR #184, opened 2020-02-27
Add prem model

cg-laser:
this PR adds improved earth attenuation models. Now the path through the Earth to the interaction point is calculated. Also more detailed PREM model is added.

### DIFF
--- a/NuRadioMC/utilities/earth_attenuation.py
+++ b/NuRadioMC/utilities/earth_attenuation.py
@@ -9,11 +9,12 @@ R_earth = 6357390 * units.m
 DensityCRUST = 2900 * units.kg / units.m ** 3
 AMU = 1.66e-27 * units.kg
 R_EARTH = 6.378140e6 * units.m
-densities = np.array([14000.0, 3400.0, 2900.0]) * units.kg / units.m ** 3 # inner layer, middle layer, outer layer
-radii = np.array([3.46e6 * units.m, R_EARTH - 4.0e4 * units.m, R_EARTH]) # average radii of boundaries between earth layers
+densities = np.array([14000.0, 3400.0, 2900.0]) * units.kg / units.m ** 3  # inner layer, middle layer, outer layer
+radii = np.array([3.46e6 * units.m, R_EARTH - 4.0e4 * units.m, R_EARTH])  # average radii of boundaries between earth layers
 
 
-def get_weight(theta_nu, pnu, flavors, mode='simple',cross_section_type = 'ghandi'):
+def get_weight(theta_nu, pnu, flavors, mode='simple', cross_section_type='ctw',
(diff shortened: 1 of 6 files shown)


NuRadioMC/utilities/earth_attenuation.py
```diff
@@ -23,19 +24,35 @@ def get_weight(theta_nu, pnu, flavors, mode='simple',cross_section_type = 'ghand
...
+    vertex_position: 3-dim array or None (default)
+        the position of the neutrino interaction
     """
     if(mode == 'simple'):
-        return get_simple_weight(theta_nu, pnu,cross_section_type = cross_section_type)
+        return get_simple_weight(theta_nu, pnu, cross_section_type=cross_section_type)
     elif (mode == "core_mantle_crust"):
-        return get_core_mantle_crust_weight(theta_nu, pnu, flavors,cross_section_type = cross_section_type)
+        return get_core_mantle_crust_weight(theta_nu, pnu, flavors, cross_section_type=cross_section_type)
+    elif (mode == "core_mantle_crust2"):
```
anelles: CAn we use something a bit more descriptive than just adding a '2'?

NuRadioMC/utilities/earth_attenuation.py
```diff
@@ -53,10 +70,11 @@ def get_simple_weight(theta_nu, pnu, cross_section_type = 'ghandi'):
...
         return np.ones_like(theta_nu)
     else:  # coming from below
-        sigma = cross_sections.get_nu_cross_section(pnu, flavors = 0, cross_section_type = cross_section_type)
-        d = - 2 * R_earth * np.cos(theta_nu)
+        sigma = cross_sections.get_nu_cross_section(pnu, flavors=0, cross_section_type=cross_section_type)
+        d = -2 * R_earth * np.cos(theta_nu)
         return np.exp(-d * sigma * DensityCRUST / AMU)
 
+
 def get_core_mantle_crust_weight(theta_nu, pnu, flavors, cross_section_type='ctw'):
```
anelles: If I believe our own documentation core_mantle crust 1 and 2 should be identical since they are both supposed to copy ARAsim. This will just cause confusion futher down the road. What are the differences apart from Ben's code structure?

NuRadioMC/utilities/earth_attenuation.py
cg-laser: do you have a suggestion? I couldn't find a better naming.
We could rename the old one to 'core_mantle_crust_simple' but I was trying to avoid that for backward compability

cg-laser:
I added documentation and renamed the old CMC model to core_mantle_crust_simple

cg-laser:
@anelles ok now?

PR #185, opened 2020-03-03
Improves plotting script of input files

cg-laser:


daniel-zid:
The changes are minimal, so there is no risk in merging this. I would only like to know what the number of bins has been changed to 2 for the zenith distribution, and to 5 for the azimuth distribution.

cg-laser:
the input files contain hundres of thousands of events, so the statistics are good enough to use a smaller binning.

PR #186, opened 2020-03-03
unforced event generator

cg-laser:
This PR adds an unforced event generator for the purpose of testing our "forced interaction" event generation procedure.

### DIFF
--- a/NuRadioMC/EvtGen/generate_unforced.py
+++ b/NuRadioMC/EvtGen/generate_unforced.py
@@ -36,8 +36,8 @@ print(f"cylinder r = {r_cylinder/units.km:.1f}km, h = {h_cylinder/units.km:.1f}k
 
 phimin = 0
 phimax = 360 * units.deg
-thetamin = 90 * units.deg
-thetamax = 92 * units.deg
+thetamin = 0 * units.deg
+thetamax = 180 * units.deg
 
 
 def perp(a) :
@@ -477,7 +477,7 @@ for j, i in enumerate(np.arange(n_events, dtype=np.int)[mask]):
     is_in_cylinder = points_in_cylinder(pt1, pt2, r_cylinder, Xint)
     mask_int[i] = is_in_cylinder
     if(is_in_cylinder):
-        logger.info(f"event {i}, interaction point ({Xint[0]:.1f}, {Xint[1]:.1f}, {Xint[2]-R_earth:.1f}), in cylinder {is_in_cylinder}")
+        logger.debug(f"event {i}, interaction point ({Xint[0]:.1f}, {Xint[1]:.1f}, {Xint[2]-R_earth:.1f}), in cylinder {is_in_cylinder}")
         data_sets['xx'].append(Xint[0])
         data_sets['yy'].append(Xint[1])
         data_sets['zz'].append(Xint[2] - R_earth)
--- a/NuRadioMC/EvtGen/generate_unforced.py
+++ b/NuRadioMC/EvtGen/generate_unforced.py
@@ -13,6 +13,7 @@ import pickle
 import os
 import time
 import logging
+np.random.seed(10)  # just for testing
 
 logging.basicConfig(level=logging.WARNING)
 logger = logging.getLogger(__name__)
@@ -25,7 +26,7 @@ earth = earth_attenuation.PREM()
 h_cylinder = 2.7 * units.km
 pt1 = np.array([0, 0, R_earth])
 pt2 = np.array([0, 0, R_earth - h_cylinder])
-r_cylinder = 5 * units.km
+r_cylinder = 10 * units.km
 
 # calculate maximum width of projected area
 theta_max = np.arctan(h_cylinder / 2 / r_cylinder)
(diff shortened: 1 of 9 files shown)


daniel-zid:
Thank you for this necessary check. Could you also revert the generate_eventlist_cylinder to the old way of generating zeniths? The surface muons method should be alright.

I can also take care of that if you want.

cg-laser:
The following plots show a comparison between the zenith distribution from the unforced event generator in comparison to the forced event generation where each interaction is weighted by the probability of the neutrino reaching the simulation volume. I performed this comparison for different neutrino energies (relevant for Earth attenuation) and cylinder volumes (-> should not have any effect). For both generators I used the PREM Earth model and I calculated the weights by integrating from the interaction vertex to the point where the neutrino enters the Earth. 

The statistics of the unforced event generator is smaller and the errorbars present sqrt(N) where N is the number of events in this bin (but then rescaled to give a total sum of 1 over all bins). 

The two distributions generally agree although the bins around the horizon shows a small systematic difference. The unforced event generator predicts less events in those bins. I'm not sure about the origin of this effect. 

forced_vs_unforced_17 0_5km
forced_vs_unforced_17 0_10km
forced_vs_unforced_18 0_5km
forced_vs_unforced_18 0_10km
forced_vs_unforced_19 0_5km
forced_vs_unforced_19 0_10km_50mstep

cg-laser:
I also changed the forced event generator back to to a flat cos(zenith) distribution

cg-laser:
A possible cause of this difference is that the weight calculation ignores the x/y position of the vertex. It just uses the depth. This leads to an effective small change in zenith angle.

daniel-zid:
> A possible cause of this difference is that the weight calculation ignores the x/y position of the vertex. It just uses the depth. This leads to an effective small change in zenith angle.

That would make sense. The implicit assumption is that (x,y) is (0,0), right? Would it be enough to change the x0 variable in earth.slant_depth to the radius? The zenith angle does not change upon rotations in azimuth, so it should be correct.

daniel-zid:
I've read the unforced generator part and everything looks sensible to me. I have to say that the code is a bit difficult to follow and there are lots of debug code that subsist there, although this is not critical. After we try to change the weight geometry for the forced generator and check it's fine, I'm approving this pull request.

cg-laser:
@daniel-zid  I already changed the forced generator. And the plots I presented above are a comparison between the unforced generator and the forced generator of this branch. 

I will clean up the code a bit more, but I thought that the 3D debug plots might be useful in the future. I was quite an effort to create them so I don't want to remove them completely.

bhokansonfasig:
> That would make sense. The implicit assumption is that (x,y) is (0,0), right? Would it be enough to change the x0 variable in earth.slant_depth to the radial coordinate rho? The zenith angle does not change upon rotations in azimuth, so it should be correct.

I think that's right. Setting `x0` to the radial coordinate should work. The only additional change is for the `angle==0` case which then needs to be
```
x1 = x0
z1 = - np.sqrt(self.earth_radius**2 - x0**2)
```

I would be a little surprised if this changed the results much, but still it would make things more accurate and might as well be added

bhokansonfasig:
Actually, I take back my previous comment. If we don't assume (x,y) = (0,0) then the entry point into the earth is dependent on the azimuth angle as well as the zenith angle. It should still be possible to calculate, but not a simple change to the existing code.

cg-laser:
can't we just do the calculation in a 2D plane. Hence, setting x = (x**2 + y**2)**0.5? Then, your suggested method should work, shouldn't it? 
Plus, we could correct the depth, as we are slightly closer to the surface.

cg-laser:
I guess, alternatively, we could also transform the zenith angle.

bhokansonfasig:
It turns out the doing the calculation of a line intersecting a sphere wasn't too bad (with references available). I just finished implementing it in PyREx, see the latest two commits. I think the problem with doing it in a 2D plane as suggested is that there isn't symmetry in the two sides of the Earth anymore, so it would matter which side you came in from (maybe easier to see if you can draw it out for yourself). And in fact, there are a range of intersection points between those two in the 3D problem depending on the azimuth angle.

Your second suggestion may work, but makes less sense to me than just solving the 3D problem.

cg-laser:
I also implemented your method into NuRadioMC, but is has very very little effect. In the above histograms I don't see any difference. I guess this is somewhat expected as 5km is really small compared to the radius of the Earth. The downside of the new method is that it seems to be quite a bit slower, although I didn't really compare it to the old method.

daniel-zid:
> I also implemented your method into NuRadioMC, but is has very very little effect. In the above histograms I don't see any difference. I guess this is somewhat expected as 5km is really small compared to the radius of the Earth. The downside of the new method is that it seems to be quite a bit slower, although I didn't really compare it to the old method.

What I don't understand is that this should be important for almost horizontal neutrinos, where the radial coordinate is comparable to the chord length traversed before interacting. And you don't get any difference, right?

Where else could the forced/unforced difference come from?

bhokansonfasig:
> The downside of the new method is that it seems to be quite a bit slower, although I didn't really compare it to the old method.

Ah, thanks for pointing this out. I was trying to be clever to reduce some code duplication, but it's much faster without that trick. Fixed in the latest PyREx commit so that now it's back to its old speed.

cg-laser:
@daniel-zid good to merge?

daniel-zid:
By all means! Go ahead.

Any new thoughts on the horizon discrepancy?

PR #189, opened 2020-03-06
add saveguard against too many ray tracing solutions (

cg-laser:
can rarely happen because of numerical imprecisions. Sometimes the same solution is found twice. This leads to a crash of NuRadioMC


daniel-zid:
I agree that is better not to have any solutions than to cause a crash, but I also think that maybe the whole ray tracing module has to be examined with care once again. It is the module that causes most of the problems in NuRadioMC.

cg-laser:
this crash happens very rarely. I only noticed it because I ran 1M events per file, and I also was still on the ray tracer version with the 1e-6 precision. We now changed it to 1e-9

PR #190, opened 2020-03-09
Add split utility

cg-laser:
adds a new utility script to split up 1 hdf5 file into multiple smaller ones. This is useful for a two step simulation process. First, a simulation is run with a simple and fast Askaryan and trigger module. The output hdf5 files can be merged together and now split up again into smaller files with the optimal number of events (such that a single job needs a few hours). 

This PR replaces the current "split hdf5 files" function. It anyway didn't work anymore because of the "station group" layout. So I don't think it was used by anyone.


NuRadioMC/EvtGen/generator.py
```diff
@@ -1578,14 +1579,78 @@ def split_hdf5_input_file(input_filename, output_filename, number_of_events_per_
...
+    total_number_of_events = attributes['n_events']
+
+    n_files = math.ceil(n_events / number_of_events_per_file)
+    for iFile in range(n_files):
+        filename2 = output_filename + ".part{:04}".format(iFile)
+        logger.debug(f"saving file {iFile} with {number_of_events_per_file} events to {filename2}")
+        fout = h5py.File(filename2, 'w')
+        for key, value in attributes.items():
+            fout.attrs[key] = value
+        fout.attrs['total_number_of_events'] = total_number_of_events / n_files
```
daniel-zid: Is this correct? Isn't this a float version of fout.attrs['n_events']?

NuRadioMC/EvtGen/generator.py
```diff
@@ -1578,14 +1579,78 @@ def split_hdf5_input_file(input_filename, output_filename, number_of_events_per_
...
     for key, value in iteritems(fin.attrs):
         attributes[key] = value
-
+#     logger.info(f"setting number of events from {attributes['n_events']} to the actual number of events in this file {len(data_sets['event_ids'])}")
     fin.close()
 
-    write_events_to_hdf5(output_filename, data_sets, attributes, n_events_per_file=number_of_events_per_file)
+    n_events = len(data['event_ids'])
+    logger.info("saving {} events in total".format(n_events))
+    total_number_of_events = attributes['n_events']
```
daniel-zid: total_number_of_events and n_events are not equivalent because using the Proposal modules we have non-unique event_ids. If we split according to n_events, we might end up having interactions from the same parent neutrino in two different files. This is not a problem as long as the correct event_ids and n_interaction are conserved _and_ the files are merged again in the end, but it's something to be careful with.

NuRadioMC/EvtGen/generator.py
cg-laser: yes, after splitting files, the original total number of event of this file might not be an int.

cg-laser:
yes I should move this functionality to the utility folder. 
I'm not concerned about having the same event ids in the same file because this splitting is for cases where you already have a full simulation that you want to split up further for resimulation purposes.

PR #191, opened 2020-03-10
Adding configuration options for Greenland in Proposal

daniel-zid:
Simple modification to include a Greenland configuration file.


daniel-zid:
I've added some lines in the generator and the Proposal interface. Let me know if that's what you wanted.

cg-laser:
looks good. I guess we should also merge the evt-gen branch soon to have the zenith simulation correct again.

PR #194, opened 2020-04-07
Save all nur

daniel-zid:
Now the user can make save_all_nur: True/False in the config file to specify whether they want to save all the events on the nur file.


anelles:
@cg-laser in all your merging efforts. We should merge this too, no?

anelles:
@christophwelling and @cg-laser  I just discussed with Daniel that there is the worry that this doesn't work, because we will create lots of empty .nur files, if there is no trigger. Do you have an example handy with which you could test this?

christophwelling:
That's not a problem. The eventWriter only creates a new file the first time it saves an event. So if you never give it one, it will do nothing. I tested it by setting the trigger threshold in the Veff example ridiculously high and no .nur file was created.

anelles:
Excellent, then I will just  merge this, as it makes sense and does not cause any issues.

PR #195, opened 2020-04-08
ARZ time

daniel-zid:
This module implements a time shift if the electric field model used is one of the ARZ models. This is to ensure that for a first interaction, the time at which the signal from the vertex is seen is precisely T, the time calculated by the ray tracing module from vertex to observer. For subsequent interactions, the time middle of the trace is the vertex time (vertex_time) plus the propagation time T, and the middle of the trace still contains the signal emitted from the vertex.

If all the electric fields have the same sampling rate and number of samples, this is just a global shift for all events and observers, but I think this is something that makes the code more consistent.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -6,6 +6,7 @@ version 1.1.2 -
 new features:
 - Veff utility can now handle extended bins
 - New tutorial and example for the webinar
+- trace start time for the electric field models adjusted such that global time of pulse position corresponds to propagation time
 
 bugfixes:
 
@@ -23,8 +24,7 @@ new features
   to happen at the surface)
 - detector description is saved to nur output files
 - new handling of random numbers. Each module has its own random generator instance. Global seed can be controlled via 
-  config file setting. 
-- trace start time for the electric field models adjusted such that global time of pulse position corresponds to propagation time
+  config file setting.
 
 bugfixes:
 - ARZxxxx and Alvarez2009 Askaryan modules now use the same (random) shower per event. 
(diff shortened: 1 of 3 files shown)


changelog.txt
```diff
@@ -22,6 +22,7 @@ new features
 - detector description is saved to nur output files
 - new handling of random numbers. Each module has its own random generator instance. Global seed can be controlled via 
   config file setting. 
+- More accurate timing for the ARZ models
```
cg-laser, line 25: the timing is not more accurate, you just redefine the trace start time. I'd rather name the change "trace start time adjusted such that global time of pulse position corresponds to propagation time"

daniel-zid: Sure, you're right. Changed.

daniel-zid:
> shouldn't this be applied for all Askaryan modules? All SignalGen modules place the pulse in the center of the trace (+- any additional physical time delays).

It could be applied to all Askaryan modules, but the parameterisations never intended to be accurate when it comes to time. To begin with, they don't have phases and approximate the shower as being far away, so their extension is negligible. It is true that if we shift all the modules the timing should be also more consistent for the parameterisations, but I would advise against using the parameterisations for timing studies.

cg-laser:
you intend to have the time of the pulse maximum the time equal to the full propagation time. So we should do it for all Askaryan modules. Just because they are anyway less accurate is not an argument to make them more wrong. So I suggest to change it for all modules.

daniel-zid:
The shift is now performed for all modules. We should be ready to go.

daniel-zid:
Keep in mind that the trigger times comparison fails now due to the new definition.

daniel-zid:
For that I would need to modify NuRadioReco, right?

cg-laser:
yes, this is a little cumbersome. Actually I think you can just merge this PR, and then make a PR in NuRadioReco that fixes the test. Because only the trigger time comparison fails.

PR #196, opened 2020-04-08
Veff bins

daniel-zid:
So far, the standard effective volume utility could only handle point bins, that is, an error appeared if the bin was extended. I think that point bins should be discouraged, that's why I'm giving the possibility of using extended bins if the user makes point_bins=False. I'm keeping the point bins as the default choice.

### DIFF
--- a/NuRadioMC/utilities/Veff.py
+++ b/NuRadioMC/utilities/Veff.py
@@ -119,7 +119,9 @@ def get_Aeff_proposal(folder, trigger_combinations={}, station=101):
     for iF, filename in enumerate(sorted(glob.glob(os.path.join(folder, '*.hdf5')))):
         fin = h5py.File(filename, 'r')
         out = {}
-        E = fin.attrs['Emin']
+        Emin = fin.attrs['Emin']
+        Emax = fin.attrs['Emax']
+        E = 10 ** ( 0.5 * (np.log10(Emin) + np.log10(Emax)) )
         out['energy'] = E
 
         weights = np.array(fin['weights'])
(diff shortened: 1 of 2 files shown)


daniel-zid:
> see my comment in the webinar PR. I think simulating single energies is actually better, but there is no harm in adding this feature. I originally implemented this exception as a safeguard against accidentally processing files of different discrete energies

I agree with you in a sense - the only problem I see with your reasoning is that when calculating like that, you're only as good as your interpolation. But if you calculate the effective volume of the whole bin, you already get the average effective volume for that bin which can directly be used for integrating in log(E). Do you agree?

daniel-zid:
And while it's true that the average is going to be dominated by the high-energy events, that is true regardless of the way you perform the calculation. The average effective volume in any bin will be dominated by the higher energies, but that is not a problem! The average will always be the average. And keep in mind that the low-energy events that do not trigger contribute to the denominator anyway!

daniel-zid:
And if you want to convolve any quantity with the average effective volumes calculated for log bins, you only need to know the average value of said quantity in the same bin and then multiply it by the average of the effective volume. This approach is mathematically rigorous. Using the interpolation will yield almost the same thing, but you need either more points or a well-tested interpolation function.

cg-laser:
I need to think about it, but you are probably right that you get the correct results for an integration (if you want to calculate the event numbers). But you might end up with a different optical impression when comparing results obtained otherwise. 
But this is testable. Did you ever calculated the Veff binned and for the central energy? How much does the result differ?

daniel-zid:
> I need to think about it, but you are probably right that you get the correct results for an integration (if you want to calculate the event numbers). But you might end up with a different optical impression when comparing results obtained otherwise.
> But this is testable. Did you ever calculated the Veff binned and for the central energy? How much does the result differ?

I could check, but that is really dependent on bin width and in any case what I want with extended bins is not to approximate the effective volume for the centre energy, but just to calculate the average. I think the physical meaning is different and the user should be the one to decide if they want point or extended bins.

cg-laser:
I see it differently. This whole discussion only makes sense if there is a difference. Of course, we can always defer something to the user, but I think it is better if we find it our ourselves and then can make a good recommendation. And implement the functionality that is useful.

daniel-zid:
Both are useful, it just depends on the context. I don't see how giving the option of treating effective volumes as point values and averages is a liability.

So, can I merge this or should I change my webinar example?

cg-laser:
yes you can merge it. 

I just mean that it would be useful to study the difference between these two approaches.

PR #197, opened 2020-04-10
Webinar

daniel-zid:
New tutorial with lots of comments for the webinar. I have included another pull request in this one, since it is needed for the example.

Please comment and/or approve this pull request before the seminar at 17:00. Thanks!


NuRadioMC/examples/06_webinar/W02RunSimulation.py
```diff
@@ -0,0 +1,206 @@
...
+
+        """
+        First we convolve the electric field with the antenna pattern to obtain
+        the voltage at the antenna terminals. This is done by the efieldtoVoltageConverter.
+        """
+        efieldToVoltageConverter.run(self._evt, self._station, self._det)
+        """
+        Our simulation uses the default sampling rate of 5 GHz, or 5 GS/s, or
+        equivalently, a time step of 0.2 ns. Such a high resolution is not common
+        at all in radio experiments, where the sampling rates tend to lie around
```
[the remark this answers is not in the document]
daniel-zid: I agree, I was talking about experimental data. We must have 5 GHz or more when simulating, but detectors tend to have sampling rates of 1 GS/s, am I wrong?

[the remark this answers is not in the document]
daniel-zid: The point is, do we want to study from a theoretical point of view what the signal is, or do we want to approximate what an experiment would see? I think the tutorial leaves clear how to do both - just resample or not, as the user sees fit.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
```diff
@@ -0,0 +1,206 @@
...
+        the gigahertz. So, if we want our simulation as realistic as possible,
+        we must resample to lower sampling rates. Let us a assume we have a
+        sampling rate of 2 GHz. Then, we can use the channelResampler to downsample.
+        """
+        new_sampling_rate = 2 * units.GHz
```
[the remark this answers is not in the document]
daniel-zid: Changed.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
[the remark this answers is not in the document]
daniel-zid: I have added a couple of comments. Feel free to modify them if they're not as clear as they should be.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
```diff
@@ -0,0 +1,206 @@
...
+        channelResampler.run(self._evt, self._station, self._det, sampling_rate=new_sampling_rate)
+
+        """
+        If our config file has specified 'noise: True', this steering file will add
```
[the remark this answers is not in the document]
daniel-zid: I've added some comments. Feel free to improve them.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
```diff
@@ -0,0 +1,206 @@
...
+        noise to the simulation. Keep in mind that adding noise can cause some
+        events to trigger on noise, while they should not be triggering at all.
+        This is a typical problem with detectors. The solution would be to find
+        a threshold to trigger on as many signals as possible while keeping the
+        noise trigger rate as low as possible. This can be studied setting
+        'signal: zerosignal: True' in the yaml config file. The detector will
+        try to trigger on noise only and that will give an estimate on the noise
+        trigger rate and how many events are not triggering on signal.
+        """
+        if bool(self._cfg['noise']):
```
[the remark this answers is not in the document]
daniel-zid: Changed

NuRadioMC/examples/06_webinar/W05ElectricFields.py
```diff
@@ -0,0 +1,129 @@
...
+plt.xlabel('Frequency [MHz]')
+plt.ylabel(r'Electric field transform [$\mu$V m$^{-1}$ MHz$^{-1}$]')
+plt.show()
+
+"""
+If the user wants to use the ARZ model and specify the distance and viewing angle with
+respect to the shower, they will have to import the ARZ class from the SignalGen.ARZ.ARZ.py
+module and use the get_time_trace function with the arguments:
+
+shift_for_xmax=True, theta_reference='Xmax'
```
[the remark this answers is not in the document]
daniel-zid: I'm not sure what you mean here. The default arguments are the right ones when calculate the field for the simulation, I agree. But the W05 file is here to illustrate how to use the SignalGen modules by themselves, and when dealing with the ARZ module, it is most of the time more convenient to use the Xmax as reference. I don't think it hurts to teach how to use the option, anyway.

NuRadioMC/examples/06_webinar/config.yaml
```diff
@@ -0,0 +1,28 @@
+# We will add noise to our signals. Keep in mind that this will cause some events
```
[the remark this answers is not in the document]
daniel-zid: Changed.

NuRadioMC/examples/06_webinar/W05ElectricFields.py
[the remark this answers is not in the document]
daniel-zid: And when used stand-alone, these two options do not 'overcorrect' twice Xmax. shift_for_xmax=True refers the observer distance and angle to the Xmax, while theta_reference='Xmax' refers the line-of-sight or radial vector to the observer-Xmax vector, which is what you want most of the time with pure electrodynamics calculations.

Once again, this has nothing to do with the simulations - it's only for illustrating electric field calculations.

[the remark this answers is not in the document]
daniel-zid: Take a look at lines 292-294 in the ARZ module. The reference radial line is always the observer-Xmax line. The problem is that if this assumes that theta has to be given in the 'X0' or vertex reference frame. If the user wants to specify the angle with respect to the Xmax, they have to choose theta_reference='Xmax'. Then, the code transforms this angle to the vertex or 'X0' reference frame (lines 280-283), and then it transforms it back to the 'Xmax' reference frame.

In other words, using shift_for_xmax=True but theta_reference=False makes the amplitude of the electric field referred to the distance an angles with respect to the Xmax but the polarization is calculated with the angle with respect to X0. That is inconsistent.

So, either shift_for_xmax=False and theta_reference='X0', as it is correctly done for simulations, or shift_for_xmax=True and theta_reference='Xmax'. Any other option is inconsistent. In fact, I would suggest to change these to options to a single one.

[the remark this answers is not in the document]
daniel-zid: In fact, I think the polarisation in the code is inconsistent if shift_for_Xmax=True, regardless of the theta_reference. I'm going to modify the code, let me know what you think.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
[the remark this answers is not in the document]
cg-laser: the trigger is done analog for most detectors. E.g. the high/low trigger in ARIANNA. So it does not suffer from discrete sampling.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
```diff
@@ -85,10 +85,12 @@ def _detector_simulation(self):
...
         the gigahertz. So, if we want our simulation as realistic as possible,
-        we must resample to lower sampling rates. Let us a assume we have a
-        sampling rate of 2 GHz. Then, we can use the channelResampler to downsample.
+        we must resample to lower sampling rates, to the actual sampling rate of
+        our analog-to-digital converter. We have specified in our detector.json
+        file an adc_sampling_frequency of 2 GHz, which can be accessed using the
+        property _sampling_rate_detector.
         """
-        new_sampling_rate = 2 * units.GHz
+        new_sampling_rate = self._sampling_rate_detector
```
cg-laser, line 96: for most detectors this is still a bad choice. Rather use the internal sampling rate of the simulation

NuRadioMC/examples/06_webinar/W02RunSimulation.py
```diff
@@ -82,9 +82,12 @@ def _detector_simulation(self):
...
         """
         Our simulation uses the default sampling rate of 5 GHz, or 5 GS/s, or
-        equivalently, a time step of 0.2 ns. Such a high resolution is not common
+        equivalently, a time step of 0.2 ns. Such a high resolution, while needed
+        during simulations to capture all the details of the radio wave, is not common
         at all in radio experiments, where the sampling rates tend to lie around
-        the gigahertz. So, if we want our simulation as realistic as possible,
+        the gigahertz. If we want our simulation to be a theoretical study, we
+        can use 5 GHz as the electric field sampling rate. However, if we want
+        to simulate what an actual radio detector would see,
```
cg-laser, line 90: not correct. None of the detectors digitizes the trace and then runs a amplitude trigger.

NuRadioMC/SignalGen/ARZ/ARZ.py
```diff
@@ -277,20 +274,17 @@ def get_time_trace(self, shower_energy, theta, N, dt, shower_type, n_index, R, s
...
                                             self._interp_factor, self._interp_factor2, shift_for_xmax)
         trace = -np.diff(vp, axis=0) / dt
 #         trace = -np.gradient(vp, axis=0) / dt
 
         # use viewing angle relative to shower maximum for rotation into spherical coordinate system (that reduced eR component)
-        thetaprime = theta_to_thetaprime(theta, xmax, R)
+        if shift_for_xmax:
+            thetaprime = theta
+        else:
+            thetaprime = theta_to_thetaprime(theta, xmax, R)
```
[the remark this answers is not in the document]
cg-laser: Can we move this change into its own PR? This is a rather important change that we need to review properly. It shouldn't be hidden (or quickly approved) through this webinar branch.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
daniel-zid: Well, yes. But a 1 or 2 GS/s trace is still going to be closer to what the ADC sees than a 5 GS/s trace, right?

NuRadioMC/examples/06_webinar/W02RunSimulation.py
daniel-zid: Why? The simulation runs with 5 GHz. No current detector does that...

cg-laser: an analog trigger has no sampling problem. It always sees the full signal amplitude, i.e. what we get with infinite upsampling.

cg-laser: and before saving the nur files, the traces get downsampled to detector sampling rate automatically.

cg-laser: so to resemble the detector situation, you need to run the trigger on an upsampled trace

daniel-zid: That is if the trigger is analog. For the phased array, for instance, the trigger has to be digital. I can remove the resampling for the example - my only intention was being pedagogical and teach how to use a feature I always needed.

cg-laser: Ok I see your point. If this is described in the documentation/comments choosing the detector sampling rate is fine too I guess. But doesn't the phased array use an internal upsampling and anyway runs on a very high sampling rate of 3.7GHz?

NuRadioMC/examples/06_webinar/W02RunSimulation.py
```diff
@@ -84,16 +84,15 @@ def _detector_simulation(self):
...
+        at all in radio experiments after the wave has been digitised, with
+        sampling rates around the gigahertz. However, we are going to suppose
+        that our trigger is analog, so it sees a continuous waveform. We can upsample
+        our trace from 5 GS/s to 10 GS/s to illustrate the use of the channel
+        resampler. Keep in mind that if the trigger were digital and we needed
+        a sampling rate of, for instance, 2 GHz, the same channelResampler
+        module can also perform a downsampling.
         """
-        new_sampling_rate = self._sampling_rate_detector
+        new_sampling_rate = 2 * self._sampling_rate_detector
```
cg-laser, line 95: sorry to bother again. But I think we should either use the `_sampling_rate_detector` if this is really the correct choice, or use the `internal sampling rate` from the config file i.e. `1/self._dt`. Otherwise this config property is useless.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
daniel-zid: Not exactly. The ideal phased array for RNO had 3 GS/s, but that is not implementable due to power budget restrictions. What I'm studying right now with the alias branch is the possibility of using low sampling rates and then upsampling to 1 GS/s with a FIR via the FPGA. It kind of looks ok, but 1 GS/s doesn't give a lot of angles to work with, since the delays are more limited than with 3 GS/s, so the effective volume is going to drop.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
daniel-zid: I don't agree. Using 1/self._dt to resample is the same as doing nothing, isn't it? The field is already sampled at 1/self._dt

daniel-zid: I don't mind using the _sampling_rate_detector. I just thought you didn't want to. Anyway, might as well remove all the resampling business if it's so controversial.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
cg-laser: shouldn't this up/downsampling be directly implemented into the trigger module? Otherwise the trigger performance will depend on the users choice of internal sampling rate etc. So a simulation of the PA is error prone.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
cg-laser: no its not the same. The efieldtovoltageconverter upsamples the traces to 20GHz to have a good time precision when adjusting the efield for time delays. So downsampling afterwards is recommended for computing speed.

daniel-zid: I've introduced a comment on how to downsample if it's necessary. See if you're ok with that.

NuRadioMC/examples/06_webinar/W02RunSimulation.py
daniel-zid: The resampling needed to be done in the steering file so far if there was no ADC conversion, but that's no longer the case with the new ADC module, everything is read from the detector file. I don't find that cumbersome to modify the channel sampling rate in the steering file no more than adding noise and filters manually in the same file, but that's my opinion.

daniel-zid: And I wouldn't say it's error-prone. The original phased array module works as it should with every sampling rate, you just need to give the right one at some point, whether in the detector file or the steering file.

cg-laser: To agree with your documentation, you should add 
`channelResampler.run(self._evt, self._station, self._det, sampling_rate=1/self._dt)`
This is how the internal logic with the internal sampling rate works. Or actually better set `new_sampling_rate` to `1/self._dt` so that it is consistent with the adding noise part. 

Cool great that it is incorporated in the ADC module.

daniel-zid: But that's not what I would want to do with a digital trigger. 1/self._dt is 5 GHz, and I would like to show the user how to downsample to 2 GHz. What am I not getting here?

cg-laser: But you are simulating the high/low trigger which is an analog trigger. So to simulate it properly, you must oversample the trace. If you call this module with the detector sampling rate you get wrong results (too few triggered events).

daniel-zid: Changed. Take a look when you can.

PR #198, opened 2020-04-13
ARZ polarisation

daniel-zid:
I have simplified the treatment of the polarisation in the ARZ module. The module intended to use as polarisation vector the line between observer and Xmax. So, the input viewing angle could be referred either to the vertex or the shower maximum. The problem is that the option shift_for_xmax=True would interpret the input viewing angle as the angle with respect to Xmax for the electric field calculation, but if theta_reference='X0', the same angle would be interpreted as with respect to the vertex, transformed to the Xmax frame and then this wrong angle would be used for defining the radial polarisation vector.

With the new treatment, either the input angle and distance are referred to the shower maximum (with the option shift_for_xmax=True) and the polarisation vector is calculated with the input angle referred to the shower maximum, or the input angle and distance are referred to the shower vertex (with the option shift_for_xmax=False), and the polarisation vector is calculated with the input angle referred to the shower vertex. Keep in mind that in both cases take as the radial polarisation vector the line from observer to shower maximum.

To put it in code terms, if shift_for_xmax=True, then the input theta is already thetaprime and it should not be transformed. If shift_for_xmax=False, the input theta must be transformed to the prime reference frame.


NuRadioMC/SignalGen/ARZ/ARZ.py
```diff
@@ -32,17 +32,37 @@
...
+    ----------
+    thetaprime: float
+        Angle formed between the shower axis and the shower maximum-observer line
+    xmax: float
+        The shower maximum depth in amount of matter traversed
+    R_prime: float
+        The distance between shower maximum and observer
     """
     L = xmax / rho
-    return thetaprime - np.arcsin((L * np.sin(np.pi - thetaprime)) / R)
```
cg-laser, line 40: what was wrong with the previous equation?

PR #199, opened 2020-04-13
only store attenuation output if 'attenuate_ice' is True

iplaisier:
set if statement for attenuation output. 
attn is not defined when attenuate_ice is False in config file.


daniel-zid:
So, this is just choosing to print the attenuation properties of the medium. Since you said output, I thought it was something that was stored.

Your modification looks nice to me.

daniel-zid:
I supposed you ran into an error when trying to remove attenuation, right?

iplaisier:
Indeed, then attn is not defined

PR #201, opened 2020-04-15
Add library_dirs to Extension call in cython cpp wrapper

clark2668:
setup.py for the cpp ray tracing wrapper now explicitly lists
the library_dirs to include. Otherwise, in some cases,
the compiler doesn't seem to be able to find -lgsl and -lgslcblas
during the linking stage.

Issue being addressed:
In some cases when trying to compile the gsl wrapper, compilation failed at linking stage because library_dirs were not fully specified.

Did you remember to: 
* include doc strings in all commits ?
* edit the changelog.txt file describing your changes?

### DIFF
--- a/NuRadioMC/SignalProp/CPPAnalyticRayTracing/setup.py
+++ b/NuRadioMC/SignalProp/CPPAnalyticRayTracing/setup.py
@@ -13,8 +13,8 @@ except:
 
 extensions = [
     Extension('wrapper', ['wrapper.pyx'],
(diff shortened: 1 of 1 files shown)


clark2668:
For example, on the MSU HPCC, this fixes errors like this:

/path/to/anaconda/compiler_compat/ld: cannot find -lgsl
/path/to/anaconda/compiler_compat/ld: cannot find -lgslcblas
collect2: error: ld returned 1 exit status
error: command 'g++' failed with exit status 1

clark2668:
sure, no problem!

PR #202, opened 2020-04-16
New proposal

daniel-zid:
I have adapted NuRadioMC to support the new version of Proposal. Thanks to Brian, I discovered that NuRadioMC was not compatible with the latest versions of Proposal, that have a brand new Python API and support download via pip. The new code seems to work fine and it has actually been simplified. That being said, we should be careful because the physics have changed. To begin with, when the electron-positron pair cross section is turned on (it was off in the examples I saw before), the pair production goes through the roof at high energies. It becomes one order of magnitude larger than the photonuclear interaction.

[numbers_tau.pdf](https://github.com/nu-radio/NuRadioMC/files/4487623/numbers_tau.pdf)

In any case, you can see that the rest of the interactions are roughly the same that we had before. Which means that the total cross sections haven't changed that much, if we ignore pair production. So I decided to study the shower spectra as a further check, with the electron-positron cross section turned off. I found out that instead of falling with shower energy, they rise. I'm completely lost.

[shower_spectra.pdf](https://github.com/nu-radio/NuRadioMC/files/4487659/shower_spectra.pdf)

I've also included a new test for generating events with proposal.

I don't know if it's a good move to merge this pull request to the master without understanding what has changed, but on the other hand our users have to download an old Proposal version and install it manually. So...


daniel-zid:
After talking to Jan, I learned that the definition of energy has changed. Now:

- sec.energy is the energy of the parent particle after emitting the secondary particle
- sec.parent_particle_energy is the energy of the parent particle

So, to get the old sec.energy we need to do: (sec.parent_particle_energy - sec.energy). This gives nice results as you can see:

[shower_spectra.pdf](https://github.com/nu-radio/NuRadioMC/files/4488637/shower_spectra.pdf)

However, it still looks slightly different to what we had before. It's not a critical difference but it may change some physics results. For comparison:

[shower_spectra.pdf](https://github.com/nu-radio/NuRadioMC/files/4488639/shower_spectra.pdf)

Now we should only clarify whether the electron-positron pair model should be used. In a sense that would be great, since it could increase the effective volume a lot, although with LPM-elongated showers.

clark2668:
Thanks Daniel! Should be exciting to see what the e+- production does to the veff (if the KKP model is the right one to use).

daniel-zid:
Well, as Anna has pointed out, the new histograms are not quite right. Low-energy taus produce showers with more energy than the taus themselves.

clark2668:
Oh shoot, good point... but it looks like only for the <=1 EeV taus? strange...

anelles:
I will hold on reviewing until we figured out that proposal actually works as they intended it to ...

daniel-zid:
This the new shower spectra once the decay particles are treated as they should. In the new version of Proposal, secondary interactions and secondary particles have a different treatment for their energy. The curves are not exactly what we had in the paper even though they've been obtained with the same configuration file, so I suppose there's been some physics change somewhere. The difference is not that important, though.

[shower_spectra.pdf](https://github.com/nu-radio/NuRadioMC/files/4527195/shower_spectra.pdf)

cg-laser:
great! Thanks for adjusting everything to the PROPOSAL changes. Can you make a plot with old vs. new? 

We still have the unresolved pair production interaction channel don't we? Or is this new version now ready to go?

daniel-zid:
> great! Thanks for adjusting everything to the PROPOSAL changes. Can you make a plot with old vs. new?
> 
> We still have the unresolved pair production interaction channel don't we? Or is this new version now ready to go?

We need to include pair production and see the difference with the old effective volume. That being said, I cannot produce input files as I could before with the old version because the cluster runs out of memory with the new version. That could either mean that the new version handles memory differently or that the old version Jan and I had installed in the cluster had some really weird behaviour. I'm going to try to reproduce the old effective volumes for muons with the new code and then if we understand the results I'll turn pair production on and redo all the figures for the paper.

daniel-zid:
This plot shows a comparison between the old and the new versions, run on my laptop. The new results follow the colour legend and the old results are plotted with dashed lines. You can see that the difference is negligible and probably due to statistics.

[proposal_versions_shower_spectra.pdf](https://github.com/nu-radio/NuRadioMC/files/4528646/proposal_versions_shower_spectra.pdf)

anelles:
Great, this looks identical to me!

daniel-zid:
> Great, this looks identical to me!

Too identical, let me recheck just in case.

daniel-zid:
Yep, I knew the difference was more than that. I had to properly restart the Jupyter kernel. Here it is. The difference is not that much, but it's more than noticeable. The configuration files and the wrapping functions are exactly the same for both versions.

[proposal_versions_shower_spectra.pdf](https://github.com/nu-radio/NuRadioMC/files/4528718/proposal_versions_shower_spectra.pdf)

daniel-zid:
The difference probably comes from the bremsstrahlung differences, that's why muons are more affected, although the physics are the same. Maybe it was a bug correction?

clark2668:
sorry, just for clarity, what are all of the pink lines?

daniel-zid:
> sorry, just for clarity, what are all of the pink lines?

The same as the nearest coloured lines, but calculated with the old Proposal version.

clark2668:
> > sorry, just for clarity, what are all of the pink lines?
> 
> The same as the nearest coloured lines, but calculated with the old Proposal version.

Okay great, thanks!

FWIW, I looked at the latest two commits, and at L489 might be worth clarifying that the `> 1000000000` is about checking (I think) if the secondary is a proposal interaction type (which has codes running from 1000000001-1000000011). That was not obvious to me the first time I looked at proposal...

daniel-zid:
> > > sorry, just for clarity, what are all of the pink lines?
> > 
> > 
> > The same as the nearest coloured lines, but calculated with the old Proposal version.
> 
> Okay great, thanks!
> 
> FWIW, I looked at the latest two commits, and at L489 might be worth clarifying that the `> 1000000000` is about checking (I think) if the secondary is a proposal interaction type (which has codes running from 1000000001-1000000011). That was not obvious to me the first time I looked at proposal...

This is relatively new and one of the things that confused me the most when trying to adapt the code. I've added a couple of comments.

clark2668:
Great, thanks very much!

cg-laser:
> > sorry, just for clarity, what are all of the pink lines?
> 
> The same as the nearest coloured lines, but calculated with the old Proposal version.
Do I see it correctly that the new version always predicts slightly more showers per lepton?

daniel-zid:
So, thanks to Jan and company, it seems that we have a fix for the stalling problem that Christian and I came across. It can be found on the new Proposal branch called fix_sector_bugs. The downside is that it cannot be installed with pip until it is pulled to the master and propagated to the pip version.

I would say that as soon as they update their master branch and pip version, we're ready to evaluate and pull the present branch.

daniel-zid:
After some hesitation, I've changed the way the Proposal functions are called in the generator. The function get_secondaries_array took a list of the properties of several leptons as an input and returned a 2D array with the secondary particles produced by all of these leptons. This is quite practical, but when the list is too long memory requirements go through the roof. So I'm keeping the function as it is because it's useful for debugging and plotting the results of lepton propagation, but I've changed the generator file so that it calls the get_secondaries_array function with a list of just one lepton, so there's no need to create a huge array.

In principle calling the function with a single lepton should take less memory and that's why I've chosen it. However, this seems to be system-dependent. On my laptop, for instance, this is what I get when simulating 50000 muon neutrinos with the old code (creating the large 2D array):
input_array

While this happens when I use a loop and call the function once per lepton (new code):
input_loop

So I think my laptop is handling memory in a clever way without me realising, but perhaps that's not the case for every computer we can find out there? I don't see any difference on my laptop, so might as well use the safer version just in case.

@cg-laser could you try to create input files on the UCI with this new version of the branch and monitor the memory, please? For some reason the DESY cluster crashes with either version, so I'm doing everything locally. Thanks!

daniel-zid:
I have calculated the effective volume for muon neutrinos (without pair production) using the new version of the code. This is the result.
[mu_volumes_ratio.pdf](https://github.com/nu-radio/NuRadioMC/files/4579645/mu_volumes_ratio.pdf)

And this is what we had with the old Proposal version - less effective volume. I'm assuming these changes come from some kind of review of the cross sections used. In any case, the new version predict more effective volume and I still have to turn pair production on, which will hopefully give us a little bit more.
[mu_volumes_ratio.pdf](https://github.com/nu-radio/NuRadioMC/files/4579655/mu_volumes_ratio.pdf)

Would you mind reviewing the new code? I think everything is set and ready to go. Thanks!

daniel-zid:
Apparently there was a problem with the merging and all the commits are laid down even though I used rebase. Strange. Apologies for that.

NuRadioMC/EvtGen/generator.py
```diff
@@ -1060,6 +1041,11 @@ def get_inverse_spectrum(N, gamma):
 
     print("number of fiducial showers", len(data_sets_fiducial['flavors']))
 
+    if len(data_sets_fiducial['event_ids']) == 0:
+        for key, value in data_sets.items():
+            data_sets_fiducial[key] = np.array( [data_sets[key][0]] )
+        data_sets_fiducial['flavors'] = np.array( [14] )
```
cg-laser, line 1047: can you explain these lines? It seems as if you were adding one muon event in case there are no events in the fiducial volume. This doesn't seem right.

daniel-zid: The hdf5 writing function crashes when data_sets_fiducial is empty, and this make my scripts crash. For the muons, we do want files that are empty, they're just telling you that there are no events for that particular try. The easiest option was to add a muon neutrino event (that is not a muon, so it can be spotted easily), with zero shower energy (zero inelasticity), and at the top of the atmosphere so it wouldn't trigger anyway.

If you prefer a different solution just tell me.

daniel-zid:
> I'm not sure if I can reasonable review these changes. There are a lot of changes in the proposal interface and I'm not familiar with proposal but I guess if you thoroughly tested it it will be ok.
> 
> I have one general suggestion though. Shall we wait with merging until we fully redone the analysis (also with pair production)? And until I had time to test it on the UCI cluster? These changes should be independent of other pull requests, so there is no urgent need to merge them. And the potential memory leak when generating input files still worries me.

Sure. I have no complaints if no one else except you and me are using this branch right now. But it can be a problem if some new people try to use the current master with the current version of Proposal. I can assure you that the changes in the interface I've made have been the fewest I had to do, and the new test plots confirm that results are almost the same.

If you're worried about the memory leaks, we could leave the master as it is and enforce one of the Proposal 5.x.x versions, but that's going to complicate life for our users. I still have no clue as to why it works on my laptop but fails on the DESY cluster.

NuRadioMC/EvtGen/generator.py
cg-laser: this seems pretty hacked. Although is should work fine... I don't know a better way at the moment. Do you need to write a file? Or would not writing a file at all be an option? Then I would just implement this saveguard in the writer function. 

Can you add least add a comment to the source code explaining what you do and why?

daniel-zid: Let's assume you don't write a file. Then you are biasing your effective areas because you're only taking files with shower-inducing events and ignoring the ones that don't, artificially increasing the effective area. However, it is not that critical, since at these energies the effective areas are usually zero.

I am going to add a comment right now.

cg-laser:
Shall we merge this PR now that I was able to successfully use it at UCI's hpc cluster?

PR #205, opened 2020-05-07
Specifies tinydb version

christophwelling:
Sets tinydb version to 3.15.2


daniel-zid:
It looks good. Do you need to change the .toml file for this one or not?

PR #208, opened 2020-05-19
change internal looping in NuRadioMC

cg-laser:
This PR introduces a major change in how the internal looping of NuRadioMC is handled. We change the input files to specify each shower, and the simulation will first loop over all showers of one event(group).  This will allow to resimulate events with the ARZ model where the shower realization is random.


daniel-zid:
It looks alright to me. We should now change the input generation files in the 06_webinar example and remove the functions in inelasticities.py that calculated the tau branching.

cg-laser:
Presentation documenting the current status and changes to the structure:
[20200611_NuRadioMC_looping.pdf](https://github.com/nu-radio/NuRadioMC/files/4766754/20200611_NuRadioMC_looping.pdf)

cg-laser:
@anelles @daniel-zid @christophwelling after our zoom discussion yesterday I improved the "saving of triggers" keeping in mind Daniels usecase. 
On station level I only mark the showers as triggered that contributed to a event that triggered that station
On detector level (i.e. considering all stations), I only mark the showers as triggered that contributed to an event that triggered any station.

Thus, you can quickly see which showers triggered, and as long as their (time) separation is large enough, they correspond to individual events and Daniel you can do your analysis as before. You can also still use the `vertex_time` to check if they overlap or not. 
However, if the two showers are too close together they will be simulated as a single event, i.e. you don't really know if both signals from the two showers were big enough to trigger individually. I'm thinking of adding an option to force the simulation to simulate every shower as an independent event. Then, you can rerun the triggered events with this setting to see which showers triggered independently. BTW: If you just simulate a single antenna per station then you can also just set the time separation to a really small value and every shower will be simulated independently. 
This is a crude way of doing it though. With the new `SimChannels` you know the signals for every shower, channel, and ray tracing solution. So you can just check if both showers have a reasonable signal. And you can lower the required threshold for the second shower. 

Let me also explain a bit better why we need the scheme I propose:
- interference between showers can only be implemented that way. We need to simulate multiple showers as a single event. 
- we want different ray tracing solution to be simulated as separated events to be able to calculate the rate of events where the D + R signals will be triggered as independent events
- a station with shallow and (very) deep components will have substantial time differences between the surface and downhole. Thus, for the same event we can get two independent triggers for the surface and deep antennas. 

And in general: We follow the principle to simulate as experiment like as possible. And the new implementation is exactly doing that.

daniel-zid:
> @anelles @daniel-zid @christophwelling after our zoom discussion yesterday I improved the "saving of triggers" keeping in mind Daniels usecase.
> On station level I only mark the showers as triggered that contributed to a event that triggered that station

This still unnecessarily complicates the analysis of the different effective volume channels for a whole detector. With the current master, the global triggers return everything that is needed for effective volume calculation just by applying some masks. With the current implementation, if I want to calculate the decay effective volume, I have to look at every station, check if there are decays, and then I should be careful not to count the same event N times by comparing the results of every station. Isn't it possible to have a global 'triggered' data set like the one we had before? What are the advantages of the new one, if we really want to get as close to the detector as we can? If the new one is really desirable, why can't we have both?

> On detector level (i.e. considering all stations), I only mark the showers as triggered that contributed to an event that triggered any station.
> 
> Thus, you can quickly see which showers triggered, and as long as their (time) separation is large enough, they correspond to individual events and Daniel you can do your analysis as before. You can also still use the `vertex_time` to check if they overlap or not.
> However, if the two showers are too close together they will be simulated as a single event, i.e. you don't really know if both signals from the two showers were big enough to trigger individually. I'm thinking of adding an option to force the simulation to simulate every shower as an independent event. Then, you can rerun the triggered events with this setting to see which showers triggered independently. BTW: If you just simulate a single antenna per station then you can also just set the time separation to a really small value and every shower will be simulated independently.
> This is a crude way of doing it though. With the new `SimChannels` you know the signals for every shower, channel, and ray tracing solution. So you can just check if both showers have a reasonable signal. And you can lower the required threshold for the second shower.
> 
> Let me also explain a bit better why we need the scheme I propose:
> 
> * interference between showers can only be implemented that way. We need to simulate multiple showers as a single event.
> * we want different ray tracing solution to be simulated as separated events to be able to calculate the rate of events where the D + R signals will be triggered as independent events
> * a station with shallow and (very) deep components will have substantial time differences between the surface and downhole. Thus, for the same event we can get two independent triggers for the surface and deep antennas.
> 
> And in general: We follow the principle to simulate as experiment like as possible. And the new implementation is exactly doing that.

cg-laser:
@daniel-zid it is possible to create such an array quickly with just a few lines and without any significant computing cost. I post a full working example below. You can still use the `n_interactions` data set to select certain interactions. 
The only additional thing to consider are electron cc interactions where both EM and HAD shower might have triggered. You can filter them out by 
```python
uids, number_of_occurences = numpy.unique(fin['event_group_ids'][triggered & fin['n_interaction' == 1],  return_counts=True)
n_triggered = len(uids)
```

The global `triggered` data set still works as before with the exception of showers that are so close to each other so that they interfere with each other. 


```python
import h5py
import numpy as np

# fin = h5py.File("myfile.hdf5")
n_showers = 3
fin = {}
fin["station_101"] = {}
fin["station_102"] = {}
fin["station_103"] = {}
fin["station_101"]["max_amp_shower_and_ray"] = np.zeros((n_showers, 1, 2))
fin["station_102"]["max_amp_shower_and_ray"] = np.zeros((n_showers, 1, 2))
fin["station_103"]["max_amp_shower_and_ray"] = np.zeros((n_showers, 1, 2))
fin["station_101"]["max_amp_shower_and_ray"][0, 0, 0] = 20
fin["station_101"]["max_amp_shower_and_ray"][2, 0, 1] = 16
fin["station_103"]["max_amp_shower_and_ray"][2, 0, 1] = 30
station_ids = [101, 102, 103]

Vrms = 10
threshold = 1.5 * Vrms
individual_shower_triggers = np.zeros(n_showers, dtype=np.bool)
for station_id in station_ids:
    station = fin[f'station_{station_id:d}']
    amplitudes = station['max_amp_shower_and_ray']
    individual_shower_triggers |= np.any(amplitudes > threshold, axis=(1, 2))
```

cg-laser:
actually it is even easier with the new helper function. You can deselect both nu_e cc triggers by
```python
from NuRadioMC.utilities.Veff import remove_duplicate_triggers
triggered = remove_duplicate_triggers(triggered)
```

christophwelling:
So, what exactly is the purpose of the SimChannels? I don't quite get how they differ from the normal channels.

cg-laser:
They store the voltage trace for each shower and ray tracing solution, i.e. each SimEfield gets one corresponding voltage trace. This is needed to calculate the signal times for each channel correctly. And I think it is quite handy as we can now visualize the signal contribution from every shower/ray tracing solution on voltage level

daniel-zid:
Just another comment. Why is the vertex time the configurable parameter when what it actually matters is the arrival time? I understand that the vertex time doesn't depend on the observer and therefore is an easier criterion, but keep in mind that near the Cherenkov angle and when the showers are far away, compression effects will make the pulses arrive at the same time, even if they have been emitted with a difference of 100 ns or more.

cg-laser:
Did I say it depended on the vertex time? If so it was wrong. As you correctly point out it depends on the arrival time, or rather at the signal time at the trigger electronics. And it is implemented like this.

cg-laser:
After a morning of bug hunting I found out why the Moore's Bay test for single events was failing. The test reports a difference of 1.3% in maximum amplitude for one channel of one of the 40ish events. All other quantities, including the maximum amplitude of the hilbert envelope were the same. So quite puzzling. 
The reason for this difference was a limited time resolution of the efieldToVoltageConverter. This module combines multiple electric fields (e.g. from different ray tracing solutions) of one channel to a combined voltage trace. By default the traces are upsamples to 10GHz -> 0.1ns time binning. The propagation time has small differences between different systems because of differences in the numerical integration, especially for Moores Bay where one can have multiple reflections (the travel time is calculated for each segment, thus the errors accumulate). 
On my system I get
```
DEBUG:NuRadioReco.efieldToVoltageConverter:smallest trace start time 7320.9, largest trace time 8033.8 -> n_samples = 7128 713ns)
DEBUG:NuRadioReco.efieldToVoltageConverter:channel id 0
DEBUG:NuRadioReco.efieldToVoltageConverter:channel 0, start time 7542.8754 = bin 2220, ray solution reflected
DEBUG:NuRadioReco.efieldToVoltageConverter:channel 0, start time 7549.2901 = bin 2284, ray solution reflected
```
whereas on travis I get
```
DEBUG:NuRadioReco.efieldToVoltageConverter:smallest trace start time 7320.9, largest trace time 8033.8 -> n_samples = 7128 713ns)
DEBUG:NuRadioReco.efieldToVoltageConverter:channel id 0
DEBUG:NuRadioReco.efieldToVoltageConverter:channel 0, start time 7542.8744 = bin 2219, ray solution reflected
DEBUG:NuRadioReco.efieldToVoltageConverter:channel 0, start time 7549.2901 = bin 2284, ray solution reflected
```
Although the difference in propagation time is just 1ps, the time difference fell so unlucky onto a bin boundary that it resulted in a different start bin. This then led to a difference in amplitude because of a slight difference in interference. 
My system
```
INFO:channelSignalReconstructor:event 122.0 station 101 channel 0 max amp = 9.86338e-06 max amp env 1.00225e-05
```
Travis:
```
INFO:channelSignalReconstructor:event 122.0 station 101 channel 0 max amp = 9.99958e-06 max amp env 1.00243e-05
```

After increasing the time resolution to 0.01ns, the problem is solved:
My system
```
DEBUG:NuRadioReco.efieldToVoltageConverter:smallest trace start time 7320.9, largest trace time 8033.8 -> n_samples = 71284 713ns)
DEBUG:NuRadioReco.efieldToVoltageConverter:channel id 0
DEBUG:NuRadioReco.efieldToVoltageConverter:channel 0, start time 7542.8754 = bin 22195, ray solution reflected
DEBUG:NuRadioReco.efieldToVoltageConverter:channel 0, start time 7549.2901 = bin 22837, ray solution reflected
INFO:channelSignalReconstructor:event 122.0 station 101 channel 0 max amp = 1.00064e-05 max amp env 1.00235e-05
```
travis:
```
DEBUG:NuRadioReco.efieldToVoltageConverter:smallest trace start time 7320.9, largest trace time 8033.8 -> n_samples = 71284 713ns)
DEBUG:NuRadioReco.efieldToVoltageConverter:channel id 0
DEBUG:NuRadioReco.efieldToVoltageConverter:channel 0, start time 7542.8744 = bin 22195, ray solution reflected
DEBUG:NuRadioReco.efieldToVoltageConverter:channel 0, start time 7549.2901 = bin 22837, ray solution reflected
INFO:channelSignalReconstructor:event 122.0 station 101 channel 0 max amp = 1.00064e-05 max amp env 1.00235e-05
```
The only disadvantage of increasing the time resolution is computing speed. I haven't benchmarked it yet though.

cg-laser:
I did a quick benchmark on ~2500 showers at SP
With 0.1ns time resolution
```
WARNING:sim:Timing of NuRadioReco modules
efieldToVoltageConverter: 8s 50.4%
channelResampler: 0s 3.8%
channelBandPassFilter: 2s 15.4%
channelBandPassFilter: 2s 15.4%
triggerSimulator: 1s 11.2%
triggerSimulator: 0s 1.8%
triggerSimulator: 0s 1.8%
triggerTimeAdjuster: 0s 0.0%

WARNING:sim:2455 events processed in 1m43s = 42.15ms/event (131.1% input, 74.8% ray tracing, 1.5% askaryan, 6.4% detector simulation, 0.2% output)
```
with 0.01ns I get
```
efieldToVoltageConverter: 1m29s 82.6%
channelResampler: 10s 10.0%
channelBandPassFilter: 2s 2.5%
channelBandPassFilter: 2s 2.5%
triggerSimulator: 1s 1.9%
triggerSimulator: 0s 0.3%
triggerSimulator: 0s 0.3%
triggerTimeAdjuster: 0s 0.0%

WARNING:sim:2455 events processed in 3m14s = 79.14ms/event (136.8% input, 40.0% ray tracing, 0.8% askaryan, 9.8% detector simulation, 0.1% output)
```
A finer time resolution leads to a substantial increase in computing time. We likely don't need this level of precision in normal simulations. I made the parameter configurable via the config file. I suggest to leave the default value as it is and only increase the precision in the tests.

anelles:
Now it seems like the built just stalls. Is this due to the increase in accuracy? If yes, then this doesn't solve the problem.

cg-laser:
The free travis computing resources are quite slow. It ran on my computer in just 4min. The reason why it it so slow is a very long efield trace because two ray tracing solutions with 7micro s time separation triggered. This happens often at Moores Bay because of the multiple reflections. But with the new event splitting we can circumvent it. Not it runs much faster. 
However, I still got a 0.15% difference in the maximum amplitude. All other quantities are fine, also the maximum amplitude of the envelope. I increased now the demanded accuracy only for the maximum_amplitude to 1%. I hope that is ok for everyone. 

I also added another test that checks if a simulation can be reproduced with the ARZ model (a new feature of this PR).

anelles:
I have lost the overview a bit. Is this project now complete? (Also asking for the corresponing NuRadioReco PR).

NuRadioMC/utilities/Veff.py
```diff
@@ -1,60 +1,168 @@
...
+    triggered = np.array(triggered)
+    uids, unique_mask, inv_mask, counts = np.unique(gids, return_index=True, return_inverse=True, return_counts=True)
+    for gid in uids[counts > 1]:
+        mask = gids == gid
+        if(np.sum(triggered[mask]) > 1):
+            idx = np.arange(len(triggered), dtype=np.int)[mask][triggered[mask] == True][1:]
+            triggered[idx] = False
+    return triggered
 
+# def get_triggered(fin):
```
anelles, line 48: Why is this commented out and not removed? Do we anticpiate needed it later?

NuRadioMC/test/examples/test_examples.sh
```diff
@@ -4,16 +4,26 @@ mkdir -p output
...
+rm 1e19_n1e3.hdf5
+rm Veff.pdf
+rm limits.pdf
+rm output.nur
+rm -r output
 
-cd ../02_DnR
-python3 E01detector_simulation.py event_input/1e19_n1e3comparison1.hdf5 detector/string_to_100m.json config.yaml output.hdf5
-python3 A01analyse_simulation.py output.hdf5
+#cd ../02_DnR
```
anelles, line 14: Why do we stop testing DnR?

NuRadioMC/test/Veff/1e18eV/T03check_output.py
```diff
@@ -11,24 +11,21 @@
 
 # the event generation has a fixed seed and I switched to Alvarez2000 (also no randomness)
 # thus, the Veff has no statistical scatter
-Veff_mean = 5.45361
```
anelles, line 14: How did this value change?

NuRadioMC/test/SingleEvents/test_build.sh
```diff
@@ -8,5 +8,5 @@ NuRadioMC/test/SingleEvents/T04validate_allmost_equal.py NuRadioMC/test/SingleEv
 
 NuRadioMC/test/SingleEvents/T05validate_nur_file.py NuRadioMC/test/SingleEvents/1e18_output.nur NuRadioMC/test/SingleEvents/1e18_output_reference.nur
 
-NuRadioMC/test/SingleEvents/T02RunSimulation.py NuRadioMC/test/SingleEvents/1e18_output_noise_reference.hdf5 NuRadioMC/test/SingleEvents/surface_station_1GHz.json NuRadioMC/test/SingleEvents/config_noise.yaml NuRadioMC/test/SingleEvents/1e18_output_noise.hdf5
```
anelles, line 11: Does this change make sense? Using a different input file all of a sudden?

NuRadioMC/test/SingleEvents/T04validate_allmost_equal.py
```diff
@@ -179,7 +186,21 @@ def test_almost_equal_keys(keys, fin1=fin1, fin2=fin2, error=error):
...
+# for this vairble for now.
+keys = [u'maximum_amplitudes_envelope']
+error = test_almost_equal_station_keys(keys, fin1=fin1, fin2=fin2, error=error, accuracy=0.001)
+
+# test maximimum amplitude separately because it might differ slightly because of differences in the interferene between signals
+keys = [u'maximum_amplitudes']
+error = test_almost_equal_station_keys(keys, fin1=fin1, fin2=fin2, error=error, accuracy=0.01)
+
 if error == -1:
+    from NuRadioMC.utilities.dump_hdf5 import dump
```
anelles, line 199: I think a bit of additional debug output would be nice. Like a sentence what this is, rather than file1 and file2 only.

NuRadioMC/test/SingleEvents/T03validate.py
```diff
@@ -82,7 +82,7 @@
         print(e)
         error = -1
 
-keys2 = [u'SNRs',
```
anelles, line 85: Why don't we test the SNRs anymore?

NuRadioMC/examples/Sensitivities/E2_fluxes3.py
```diff
@@ -303,7 +303,10 @@ def get_E2_limit_figure(diffuse=True,
                         show_grand_200k=False,
                         show_radar=False,
                         show_Heinze=True,
-                        show_TA=False):
+                        show_TA=False,
+                        show_RNOG=False,
+                        show_IceCubeGen2=False,
+                        shower_Auger=True):
 
     # Limit E2 Plot
```
anelles, line 311: Not really a comment to the code: But does this belong in this PR?

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1515,7 +1515,7 @@ class ray_tracing:
                       3: 'reflected'}
 
     def __init__(self, x1, x2, medium, attenuation_model="SP1", log_level=logging.WARNING,
-                 n_frequencies_integration=6,
+                 n_frequencies_integration=100,
```
anelles, line 1518: Why do we have a new standard and why is it so much larger?

NuRadioMC/SignalGen/parametrizations.py
```diff
@@ -68,13 +69,21 @@ def get_time_trace(energy, theta, N, dt, shower_type, n_index, R, model, seed=No
         if False, for each request a new random shower realization is choosen.
         if True, the shower from the last request of the same shower type is used. This is needed to get the Askaryan
         signal for both ray tracing solutions from the same shower.
+    k_L: None or float
+        the k_L parameter for EM showers of the Alvarez2009 model. If a this parameter is provided, this value is used
+        and the parameter will not be drawn from a random distribution. 
+        This setting overrides the `same_shower` setting
+    full_output: bool (default False)
```
anelles, line 76: The default is actually True. And some more docs would be nice what 'additiona' output is returned?

anelles (changes requested):
Not sure that I have caught everything, but with such a huge project, I think we alos need a second reviewer.

NuRadioMC/SignalProp/analyticraytracing.py
cg-laser: the smaller number can lead to uncertainties. Better a precise result that might be a bit slower than a wrong result that is fast.

NuRadioMC/examples/Sensitivities/E2_fluxes3.py
cg-laser: I was working on this branch all the time. It might have slipped into it (or maybe it was even a separte PR that was merged into this branch)

NuRadioMC/test/SingleEvents/T03validate.py
cg-laser: I removed SNRs from the output file as this is just a duplication of data. The same information is contained in amplitude. Just divide the amplitude by the fixed RMS noise and one gets the SNR.

NuRadioMC/test/SingleEvents/test_build.sh
cg-laser: Yes, this is actually the better choice. Before I used the noiseless reference to produce the noise input. Which I then used as input again. This was possible because we couldn't reproduce runs so well but with the improvements in the random generators this is no longer required.

NuRadioMC/test/examples/test_examples.sh
cg-laser: The example does not work anymore because the hdf5 structure changed. Before it was implemented in a complicated way that involved the execution of a custom module. With the new structure this analysis is easier but it would still be some work to re implement this analysis. As we don't really need this at the moment, I rather wanted to remove the example. I should also remove it from the repo completely and not just the testing

NuRadioMC/utilities/Veff.py
cg-laser: this logic is not used anymore. Now the function above is used.  I just didn't delete it yet completely as this code snippet might be useful in the future. I'll remove it

NuRadioMC/test/Veff/1e18eV/T03check_output.py
cg-laser: see https://github.com/nu-radio/NuRadioMC/commit/fec51da0f0998bbb5dfa2d05a46cbe9f63ea5c01#diff-6ee18f45f38c958db1d8604849a200ee
It only changed within the statistical precision. However we test much more stringent which is possible because we have a fixed input file. Thus the same events should always trigger. The new structure (one line per shower instead of event) required a new input file which solved the random distribution of the events. 
The story is even longer (and should be documented by commit messages). I was able to reproduce the exact same Veff with the new structure. But then I needed to change the order in which the different variables were generated in the event generator. I think before it was first zenith angles and then vertex positions. Now it is the other way around which changed the random distribution of the events.

NuRadioMC/test/SingleEvents/T04validate_allmost_equal.py
cg-laser: ok good point

NuRadioMC/SignalGen/parametrizations.py
cg-laser: good catch. I changed it and documented the option

NuRadioMC/EvtGen/generator.py
```diff
@@ -934,123 +823,116 @@ def generate_surface_muons(filename, n_events, Emin, Emax,
...
+        data_sets["azimuths"] = np.random.uniform(phimin, phimax, n_events_batch)
+        # zenith directions are distruted as sin(theta) (to make the distribution istotropic) * cos(theta) (to account for the projection onto the surface)
+        data_sets["zeniths"] = np.arcsin(np.random.uniform(np.sin(thetamin) ** 2, np.sin(thetamax) ** 2, n_events_batch) ** 0.5)
 
-    import time
-    init_time = time.time()
-    # Initialising data_sets_fiducial with empty values
-    for key, value in data_sets.items():
-        data_sets_fiducial[key] = []
+        data_sets["event_group_ids"] = np.arange(i_batch * max_n_events_batch, i_batch * max_n_events_batch + n_events_batch) + start_event_id
```
christophwelling, line 845: Won't this generate floats instead of integers?

NuRadioMC/simulation/simulation.py
```diff
@@ -27,14 +32,29 @@
...
 import datetime
 import logging
 from six import iteritems
 import yaml
 import os
 import collections
-# import confuse
-logger = logging.getLogger("sim")
+
+STATUS = 31
```
christophwelling, line 43: What does this variable mean?

christophwelling (approved):
Looks good, as far as I can see. Though I have to admit that I don't understand every detail of the simulation

NuRadioMC/simulation/simulation.py
cg-laser: I improved the output. We had a lot of into messages via logger.warning which confused people. So I added a new custom "status" message `logger.status("")` which is just a big higher than warning but below info.

NuRadioMC/EvtGen/generator.py
cg-laser: good catch. I changed it

anelles (approved):

PR #209, opened 2020-05-22
Speedup distances

daniel-zid:
I've implemented a distance cut as a function of shower energy, and it seems to work really well. The time needed to simulate large arrays with Proposal is reduced by a factor of more than 10. In order to do so, I've implemented a fit to the maximum distances as a function of shower energy taken from a previous simulation for a 10x10 1.5 sigma dipole array in Greenland. I've used a cover factor of 1.5, or 50%, just to be completely sure that we're not losing events. We should study if this cut is good enough for other media, like South Pole, that have a larger attenuation length.

[fit_max_distances.pdf](https://github.com/nu-radio/NuRadioMC/files/4666751/fit_max_distances.pdf)

### DIFF
--- a/NuRadioMC/simulation/config_default.yaml
+++ b/NuRadioMC/simulation/config_default.yaml
@@ -13,9 +13,9 @@ speedup:
   redo_raytracing: False  # redo ray tracing even if previous calculated ray tracing solutions are present
   min_efield_amplitude: 2  # the minimum signal amplitude of the efield as a factor of the noise RMS. If the value is smaller, no detector simulation is performed. As the vector effecive length of antennas is typically less than 1, this cut does not introduce any bias as long as the value is smaller than the trigger threshold.
   amp_per_ray_solution: True  # if False, the maximum aplitude for each ray tracing solution is not calculated
-  distance_cut: True # if True, a cut for the vertex-observer distance as a function of shower energy is applied (log10(max_dist) = intercept + slope * log10(shower_energy))
+  distance_cut: True # if True, a cut for the vertex-observer distance as a function of shower energy is applied (log10(max_dist / m) = intercept + slope * log10(shower_energy / eV))
   distance_cut_intercept: -12.31 # intercept for the maximum distance cut
-  distance_cut_slope: 0.9542 # slope for the maximum distance cut 
+  distance_cut_slope: 0.9542 # slope for the maximum distance cut
 
 propagation:
   module: analytic
--- a/NuRadioMC/simulation/config_default.yaml
+++ b/NuRadioMC/simulation/config_default.yaml
@@ -13,7 +13,9 @@ speedup:
   redo_raytracing: False  # redo ray tracing even if previous calculated ray tracing solutions are present
   min_efield_amplitude: 2  # the minimum signal amplitude of the efield as a factor of the noise RMS. If the value is smaller, no detector simulation is performed. As the vector effecive length of antennas is typically less than 1, this cut does not introduce any bias as long as the value is smaller than the trigger threshold.
   amp_per_ray_solution: True  # if False, the maximum aplitude for each ray tracing solution is not calculated
-  distance_cut: True # if True, an energy-dependent cut for the vertex-observer distance is applied
+  distance_cut: True # if True, a cut for the vertex-observer distance as a function of shower energy is applied (log10(max_dist) = intercept + slope * log10(shower_energy))
+  distance_cut_intercept: -12.31 # intercept for the maximum distance cut
+  distance_cut_slope: 0.9542 # slope for the maximum distance cut 
 
 propagation:
   module: analytic
--- a/NuRadioMC/simulation/config_default.yaml
+++ b/NuRadioMC/simulation/config_default.yaml
@@ -13,7 +13,7 @@ speedup:
   redo_raytracing: False  # redo ray tracing even if previous calculated ray tracing solutions are present
   min_efield_amplitude: 2  # the minimum signal amplitude of the efield as a factor of the noise RMS. If the value is smaller, no detector simulation is performed. As the vector effecive length of antennas is typically less than 1, this cut does not introduce any bias as long as the value is smaller than the trigger threshold.
   amp_per_ray_solution: True  # if False, the maximum aplitude for each ray tracing solution is not calculated
-  distance_cut: True # if True, a cut for the vertex-observer distance as a function of shower energy is applied (log10(max_dist / m) = intercept + slope * log10(shower_energy / eV))
+  distance_cut: False # if True, a cut for the vertex-observer distance as a function of shower energy is applied (log10(max_dist / m) = intercept + slope * log10(shower_energy / eV))
   distance_cut_intercept: -12.14 # intercept for the maximum distance cut
   distance_cut_slope: 0.9542 # slope for the maximum distance cut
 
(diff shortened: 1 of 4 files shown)


NuRadioMC/simulation/simulation.py
```diff
@@ -426,9 +438,15 @@ def run(self):
                             temp_reflection_case = sg_pre['ray_tracing_reflection_case'][self._iE][channel_id]
                         r.set_solution(sg_pre['ray_tracing_C0'][self._iE][channel_id], sg_pre['ray_tracing_C1'][self._iE][channel_id],
                                        sg_pre['ray_tracing_solution_type'][self._iE][channel_id], temp_reflection, temp_reflection_case)
-                    else:
+                    elif find_solutions:
```
cg-laser, line 441: the logic is implemented in a non ideal way (at least to me): Why don't you just add a `continue` to line 431? and also the debug output? 

Then you don't need to touch the rest of the file

daniel-zid: No reason except that I found it easier to read like this, and I find that the simulation module is already too dense, so a bit of code for dummies shouldn't be too bad. I can change it if you like.

daniel-zid:
> I suggest to make the cut better configurable at the expense of making it slightly worse. The relevant shower energies are the lowest ones where the dependence between max. distance and shower energy is almost linear. This is because the early part is independent of the attenuation length and dominated by the 1/R scaling.

But do you want to keep the one we have right now AND adding a configurable one? I think that could work.
> I would allow the user to specify the value A and B of the function log10(max. D) = A * log10(E) + B.

Right, that would be nice.
> If really needed (but I doubt that it will lead to a lot of speed improvements) add a second straight line cut to describe the high energy behaviour which is dominated by attenuation length.

cg-laser:
I opt for replacing the current one. This fit is only valid for Greenland for the very specific thresholds and geometries. I don't think we should hardcode this in NuRadioMC. 

Can you also post the histograms that go into the figure? Or make a 2D histrogram (xaxis: energy, yaxis distance per event, z-axis: number of entries).

daniel-zid:
> I opt for replacing the current one. This fit is only valid for Greenland for the very specific thresholds and geometries. I don't think we should hardcode this in NuRadioMC.

Seems reasonable.
> Can you also post the histograms that go into the figure? Or make a 2D histrogram (xaxis: energy, yaxis distance per event, z-axis: number of entries).

I'm going to upload them all if you like, I think it's better for future reference to have a lot of 1D histograms than a single 2D colour histogram.

daniel-zid:
I've changed the cut to a linear function in log space and used as default parameters the ones that give the red curve in this figure:

[fit_max_distances.pdf](https://github.com/nu-radio/NuRadioMC/files/4667372/fit_max_distances.pdf)

NuRadioMC/simulation/config_default.yaml
```diff
@@ -13,6 +13,9 @@ speedup:
   redo_raytracing: False  # redo ray tracing even if previous calculated ray tracing solutions are present
   min_efield_amplitude: 2  # the minimum signal amplitude of the efield as a factor of the noise RMS. If the value is smaller, no detector simulation is performed. As the vector effecive length of antennas is typically less than 1, this cut does not introduce any bias as long as the value is smaller than the trigger threshold.
   amp_per_ray_solution: True  # if False, the maximum aplitude for each ray tracing solution is not calculated
+  distance_cut: True # if True, a cut for the vertex-observer distance as a function of shower energy is applied (log10(max_dist) = intercept + slope * log10(shower_energy))
```
cg-laser, line 16: better make the default False. And can you add units to the equation? I think it should be log10(max_dist / m) = intercept + slope * log10(shower_energy / eV)

NuRadioMC/simulation/simulation.py
cg-laser: yes please change it. This cut should just be a single if block

NuRadioMC/simulation/simulation.py
```diff
@@ -1084,6 +1102,35 @@ def _get_em_had_fraction(self, inelasticity, inttype, flavor):
                 fem = 0
         return fem, fhad
 
+    def _get_distance_cut(self, shower_energy):
```
cg-laser, line 1105: just a small thing.  I think it would be better if the function took the slope and intercept as parameters. Then it will be clear where the parameters came from without looking into the function.

cg-laser:
> I'm going to upload them all if you like, I think it's better for future reference to have a lot of 1D histograms than a single 2D colour histogram.

Yes please upload them. I don't have a strong opinion on wether it should be a 1 D or 2D plot. If it is 1 2D plot you can just post it here. 
I would like to get a sense of the distribution. E.e. is the maximum value from the plot above already a exeptionally large distance.

NuRadioMC/simulation/config_default.yaml
daniel-zid: That's right, I've forgotten.

NuRadioMC/simulation/simulation.py
daniel-zid: Then it doesn't have to be even a method of the simulation class, it can be out of its scope.

daniel-zid:
Here are some of the histograms that can be used for justifying the fit. I can upload them all if you want, but they're a bunch.

[distances_to_vertex_00.pdf](https://github.com/nu-radio/NuRadioMC/files/4671307/distances_to_vertex_00.pdf)
[distances_to_vertex_01.pdf](https://github.com/nu-radio/NuRadioMC/files/4671308/distances_to_vertex_01.pdf)
[distances_to_vertex_02.pdf](https://github.com/nu-radio/NuRadioMC/files/4671309/distances_to_vertex_02.pdf)
[distances_to_vertex_03.pdf](https://github.com/nu-radio/NuRadioMC/files/4671310/distances_to_vertex_03.pdf)
[distances_to_vertex_04.pdf](https://github.com/nu-radio/NuRadioMC/files/4671311/distances_to_vertex_04.pdf)
[distances_to_vertex_05.pdf](https://github.com/nu-radio/NuRadioMC/files/4671312/distances_to_vertex_05.pdf)
[distances_to_vertex_06.pdf](https://github.com/nu-radio/NuRadioMC/files/4671313/distances_to_vertex_06.pdf)
[distances_to_vertex_07.pdf](https://github.com/nu-radio/NuRadioMC/files/4671314/distances_to_vertex_07.pdf)
[distances_to_vertex_08.pdf](https://github.com/nu-radio/NuRadioMC/files/4671316/distances_to_vertex_08.pdf)
[distances_to_vertex_11.pdf](https://github.com/nu-radio/NuRadioMC/files/4671318/distances_to_vertex_11.pdf)
[distances_to_vertex_14.pdf](https://github.com/nu-radio/NuRadioMC/files/4671317/distances_to_vertex_14.pdf)
[distances_to_vertex_17.pdf](https://github.com/nu-radio/NuRadioMC/files/4671319/distances_to_vertex_17.pdf)

daniel-zid:
Actually, this rebinning suggests that if we want to be a bit more conservative, we can use the same slope and a different intercept so that the cut at 1 PeV is 150 m. I don't think at these energies it matters that much, though.

[fit_max_distances.pdf](https://github.com/nu-radio/NuRadioMC/files/4671348/fit_max_distances.pdf)

NuRadioMC/simulation/config_default.yaml
daniel-zid: Done.

daniel-zid:
Changes have been addressed. Let me know what you think.

The Arianna server is down again. Yay!

NuRadioMC/simulation/config_default.yaml
```diff
@@ -13,6 +13,9 @@ speedup:
   redo_raytracing: False  # redo ray tracing even if previous calculated ray tracing solutions are present
   min_efield_amplitude: 2  # the minimum signal amplitude of the efield as a factor of the noise RMS. If the value is smaller, no detector simulation is performed. As the vector effecive length of antennas is typically less than 1, this cut does not introduce any bias as long as the value is smaller than the trigger threshold.
   amp_per_ray_solution: True  # if False, the maximum aplitude for each ray tracing solution is not calculated
+  distance_cut: False # if True, a cut for the vertex-observer distance as a function of shower energy is applied (log10(max_dist / m) = intercept + slope * log10(shower_energy / eV))
+  distance_cut_intercept: -12.14 # intercept for the maximum distance cut
```
cg-laser, line 17: can you add a comment from where these cuts were derived

daniel-zid: Done

PR #210, opened 2020-06-02
ARZ angular cut

daniel-zid:
An angular cut when using the ARZ model has been implemented. When the observer is near the axis or near 90 degrees from the shower axis, the integrals used for the ARZ method become unstable due to the loss of coherence. The result is that standard integration methods yield large peaks due to numerical errors and our trapezoid rule interpolation also fails, while the field for these regions should be much closer to zero than the fields at less than 10 degrees from the Cherenkov angle. The solution is implementing a cut - more than 20 degrees away from the Cherenkov angle, the model returns a null trace.


daniel-zid:
Is this fine or do you want it to be configurable in other way?

PR #211, opened 2020-06-02
Naming of antenna pattern

cg-laser:
the new default is to directly use the name of the antenna pattern as specified in the database. All example detector simulations are changed


christophwelling:
@cg-laser The NuRadioReco master is currently broken. I assume this PR is supposed to fix it? I did a dummy commit to see if tests here still pass. If they do we should merge soon.

cg-laser:
yes sorry about that, it's this cross dependency... I tested it last night and it should work. I'll wait until the tests you just triggered pass again and then I'll merge.

PR #214, opened 2020-06-11
Fix codes

daniel-zid:
There was a bug concerning the particle codes. The muon neutrinos (14 and -14) were converted to muons (13 and -13), and the tau neutrinos (16 and -16) were converted to muons (15 and -15) when using Proposal, which means that the first interaction showers were ignored during simulation. This is caused by copying by reference due to a mistake instead of copying by value. This pull request solves it.

Incidentally, even with these wrong files, the effective volume due to secondary showers with pair production did not seem to be so far away from the first interaction volume, which means that turning pair production on may give a substantial increase in sensitivity.

### DIFF
--- a/NuRadioMC/EvtGen/generator.py
+++ b/NuRadioMC/EvtGen/generator.py
@@ -15,6 +15,7 @@ from scipy.interpolate import RectBivariateSpline
 import h5py
 import os
 import math
+import copy
 import logging
 logger = logging.getLogger("EventGen")
 logging.basicConfig()
@@ -1364,7 +1365,7 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
         mask_leptons = mask_leptons & mask_theta & mask_phi
 
         E_all_leptons = (1 - data_sets["inelasticity"]) * data_sets["energies"]
-        lepton_codes = data_sets["flavors"][:]
+        lepton_codes = copy.copy(data_sets["flavors"])
         lepton_codes[lepton_codes == 14] = 13
         lepton_codes[lepton_codes == -14] = -13
         lepton_codes[lepton_codes == 16] = 15
--- a/NuRadioMC/EvtGen/generator.py
+++ b/NuRadioMC/EvtGen/generator.py
@@ -1364,7 +1364,7 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
         mask_leptons = mask_leptons & mask_theta & mask_phi
 
         E_all_leptons = (1 - data_sets["inelasticity"]) * data_sets["energies"]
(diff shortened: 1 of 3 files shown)


NuRadioMC/EvtGen/generator.py
```diff
@@ -1364,7 +1364,7 @@ def J(E):
         mask_leptons = mask_leptons & mask_theta & mask_phi
 
         E_all_leptons = (1 - data_sets["inelasticity"]) * data_sets["energies"]
-        lepton_codes = data_sets["flavors"]
+        lepton_codes = data_sets["flavors"][:]
```
cg-laser, line 1367: what is this doing? 
If you want to be sure to copy something I'd suggest to use
```python
import copy
lepton_codes = copy.copy(data_sets["flavors"])
```

NuRadioMC/simulation/simulation.py
```diff
@@ -1115,11 +1121,6 @@ def _get_em_had_fraction(self, inelasticity, inttype, flavor):
         elif(inttype == 'em'):
             fem = 1
             fhad = 0
-        elif(np.abs(flavor) == 15):
```
cg-laser, line 1118: doesn't these lines mean that for taus it was already simulated correctly despite the bug in particle codes?

NuRadioMC/EvtGen/generator.py
daniel-zid: This is the native pythonic way of copying an array-like structure by value instead of reference. I can also use numpy.copy

NuRadioMC/simulation/simulation.py
daniel-zid: No, because the tau_e and tau_had inttypes were related to the old continuous slowing down approximation for taus, so the result was that fem and fhad were zero. The flow of this function wasn't prepared for these kind of faulty input files.

NuRadioMC/EvtGen/generator.py
daniel-zid: So, to elaborate, the brackets take a slice and return the slice by value. A slice of the type [:] is equivalent to [None:None], which is the same thing as returning a slice that encompasses the whole array.

cg-laser: From https://stackoverflow.com/questions/2612802/how-to-clone-or-copy-a-list:
"Alex Martelli's opinion (at least back in 2007) about this is, that it is a weird syntax and it does not make sense to use it ever. ;) (In his opinion, the next one is more readable)."
I agree with that, it makes the code more obscure. Rather implement it explicitly using the copy module.

daniel-zid: Changed.

PR #216, opened 2020-06-22
Test for atmospheric muon effective area

daniel-zid:
I have included a new test for calculating the atmospheric muon effective area. This will make future changes to the Proposal modules easier to test.

### DIFF
--- a/NuRadioMC/test/atmospheric_Aeff/1e18eV/T03check_output.py
+++ b/NuRadioMC/test/atmospheric_Aeff/1e18eV/T03check_output.py
@@ -15,6 +15,8 @@ import os
 # Proposal gives a different result for the same seed depending on the system,
 # so now we don't fix the seed and check that the calculated effective area is
 # within a number of standard of the previous calculation.
+# The following numbers have been obtained with 5e3 input events. The mean
+# and uncertainty have been calculated using the function calculate_aeff.
 Aeff_mean = 0.406902 * units.km2
 Aeff_sigma = 0.045211 * units.km2
 
--- a/NuRadioMC/test/atmospheric_Aeff/1e18eV/T03check_output.py
+++ b/NuRadioMC/test/atmospheric_Aeff/1e18eV/T03check_output.py
@@ -10,10 +10,13 @@ import os
 # Reference values from previous run, have to be updated, if code changes
 ###########################
 
-# the event generation has a fixed seed and the Askaryan model is Alvarez2000,
-# so as a result the effective area is no statistical scatter.
-Aeff_mean = 0.175822
-Aeff_sigma = 0.0001
+# the event generation used to have a fixed seed. The Askaryan model is Alvarez2000,
+# so as a result the effective area should have no statistical scatter. However,
+# Proposal gives a different result for the same seed depending on the system,
+# so now we don't fix the seed and check that the calculated effective area is
+# within a number of standard of the previous calculation.
+Aeff_mean = 0.406902 * units.km2
+Aeff_sigma = 0.045211 * units.km2
 
 path = os.path.dirname(os.path.abspath(__file__))
 fin = h5py.File(os.path.join(path, "output.hdf5"), 'r')
@@ -27,7 +30,6 @@ def calculate_aeff(fin):
 
     n_triggered = np.sum(weights[triggered])
 
-    V = None
     rmin = fin.attrs['rmin']
     rmax = fin.attrs['rmax']
     geometrical_area = np.pi * ( rmax - rmin ) ** 2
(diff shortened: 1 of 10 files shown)


NuRadioMC/test/atmospheric_Aeff/1e18eV/run.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+python T01generate_event_list.py
+python T02RunSimulation.py 1e18_full.hdf5  ../dipole_100m.json ../config.yaml output.hdf5
+python T03check_output.py
+rm tables/*txt
```
cg-laser, line 6: can you add deleting the other outputfiles as well?

NuRadioMC/test/atmospheric_Aeff/config.yaml
```diff
@@ -0,0 +1,15 @@
...
+  minimum_weight_cut: 1.e-5
+  delta_C_cut: 0.698  # 40 degree
+  redo_raytracing: True  # redo ray tracing even if previous calculated ray tracing solutions are present
+  distance_cut: True # Skip ray tracing if vertex distance is greater than a distance cut that is a function of shower energy.
+propagation:
+  ice_model: southpole_2015
+signal:
+  model: Alvarez2000
+trigger:
+  noise_temperature: 300  # in Kelvin
```
cg-laser, line 13: to increase statistics you could also lower the noise temperature. But this is equivalent to just lowering the threshold. As we don't add noise, maybe this would be clever to reduce computing time and increase precision. But then the test is testing unrealistically low signals... so I'm not sure if it will be a good choice in the end

NuRadioMC/test/atmospheric_Aeff/1e18eV/test_build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+NuRadioMC/test/atmospheric_Aeff/1e18eV/T01generate_event_list.py
+
+NuRadioMC/test/atmospheric_Aeff/1e18eV/T02RunSimulation.py 1e18_full.hdf5  ../dipole_100m.json ../config.yaml output.hdf5 output.nur
+
+NuRadioMC/test/atmospheric_Aeff/1e18eV/T03check_output.py
+
+rm NuRadioMC/test/atmospheric_Aeff/1e18eV/tables/*txt
```
cg-laser, line 10: why is there a run and a test_build script? I think we only need one.

NuRadioMC/test/atmospheric_Aeff/1e18eV/T03check_output.py
```diff
@@ -0,0 +1,69 @@
...
+# Reference values from previous run, have to be updated, if code changes
+###########################
+
+# the event generation used to have a fixed seed. The Askaryan model is Alvarez2000,
+# so as a result the effective area should have no statistical scatter. However,
+# Proposal gives a different result for the same seed depending on the system,
+# so now we don't fix the seed and check that the calculated effective area is
+# within a number of standard of the previous calculation.
+Aeff_mean = 0.406902 * units.km2
+Aeff_sigma = 0.045211 * units.km2
```
cg-laser, line 19: can you add as a comment the info how you obtained this number? E.g. how often did you run the script (to get a sense of the statistical precision of the sigma value). 
Actually, how do you obtain it? You did not override the seed. Shouldn't you get always the same result then?

NuRadioMC/test/atmospheric_Aeff/1e18eV/test_build.sh
daniel-zid: No, we don't. That's what you can find on the Veff example.

NuRadioMC/test/atmospheric_Aeff/1e18eV/T03check_output.py
daniel-zid: I don't override the seed because as I explain on the comment, the results obtained with Proposal depend on the system. I just ran the example for 5e3 events and took the Poissonian estimator for the uncertainty.

daniel-zid: And a fixed seed using Proposal doesn't mean anything if the input properties of the muons are different - the results are going to be different each time if e.g. the positions change.

daniel-zid: If we want a more realistic number I would need to run a large simulation, but keep in mind that our test results might be biased because we are restricted to low statistics (a few thousand events).

NuRadioMC/test/atmospheric_Aeff/config.yaml
daniel-zid: Another way is to remove pair production.

NuRadioMC/test/atmospheric_Aeff/1e18eV/run.sh
daniel-zid: Done.

NuRadioMC/test/atmospheric_Aeff/1e18eV/test_build.sh
daniel-zid: Changed.

NuRadioMC/test/atmospheric_Aeff/1e18eV/T03check_output.py
cg-laser: clever idea. So you just took sqrt(N_triggered)*Aeff  as the uncertainty?

daniel-zid: Right, with the sqrt() in the denominator. There are some subtleties because it's not the same estimator (the sample standard deviation does not make any assumption on the distribution) and that can give problems for low statistics, but it seems to work for 5e3 events. I'm going to remove pair production and see if I can get more statistics easily.

PR #218, opened 2020-06-25
Removing geometry cut

daniel-zid:
I'm removing the geometry cut used with Proposal because it didn't work properly. This is a quick fix that should be changed to a proper way of speeding up the code by removing particles that will not reach the cylinder.

### DIFF
--- a/NuRadioMC/EvtGen/generator.py
+++ b/NuRadioMC/EvtGen/generator.py
@@ -962,27 +962,6 @@ def generate_surface_muons(filename, n_events, Emin, Emax,
     for key, value in data_sets.items():
         data_sets_fiducial[key] = []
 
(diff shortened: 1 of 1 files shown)


anelles:
I am happy to take this out for now. @daniel-zid doesn't have forever to spend on this. But a solution would of course be great.

daniel-zid:
> why did you remove these lines?

Look closely, some are for the surface muons and some for the cylinder. The documentation was outdated - I don't use spectrum anymore when generating muons and the parameters of the full volume where not calculated as the documentation said, both for the muons and the neutrinos.

daniel-zid:
I'm closing this pull request because the branches fix-lepton-mask and the new internal looping will make this PR obsolete.

PR #219, opened 2020-06-30
add option for cube simulation volume

cg-laser:
code cleanup, functions were modularized to remove code replication between the area and volume generator. 

A new option is added to allow the simulation of a cube as simulation volume. 

Also the depth is decreased by the tau decay length it secondary interactions are enabled. 

The geometry cut for Aeff is removed


NuRadioMC/EvtGen/generator.py
```diff
@@ -460,9 +371,134 @@ def J(E):
...
+            else:
+                full_rmax = tau_95_length + attributes['fiducial_rmax']
+            if('full_zmax' in volume):
+                zmax = volume['full_zmax']
+            else:
+                zmax = attributes['fiducial_zmax'] / 3.
+            if('full_zmin' in volume):
+                zmin = volume['full_zmin']
+            else:
+                zmin = attributes['fiducial_zmin'] - tau_95_length  # we have a minus sign here because the zmin coordinate is negative
```
daniel-zid, line 420: This is actually not a great choice. The tau range is ~ 100 km at ~ 100 EeV, and you need to keep in mind that the range in rock is reduced. This would simulate an enormous cylinder at high energies, but we know upgoing events at these energies are completely negligible.

NuRadioMC/EvtGen/generator.py
```diff
@@ -460,9 +371,134 @@ def J(E):
...
+            # increase the total number of events such that we end up with the same number of events in the fiducial volume
+            n_events = n_events * int((rmax / attributes['fiducial_rmax']) ** 2 * zmin / attributes['fiducial_zmin'])
```
daniel-zid, line 422: Shouldn't we account for the vertical length differently here?

daniel-zid: namely, zmin/fiducial_zmin is not the same as (zmax-zmin)/(fiducial_zmax-fiducial_zmin)

NuRadioMC/EvtGen/generator.py
```diff
@@ -489,23 +525,50 @@ def generate_surface_muons(filename, n_events, Emin, Emax,
...
+                * fiducial_rmax: float
+                    upper r coordinate of fiducial volume (the fiducial volume needs to be chosen large enough such that no events outside of it will trigger)
+                * fiducial_zmin: float
+                    lower z coordinate of fiducial volume (the fiducial volume needs to be chosen large enough such that no events outside of it will trigger)
+                * fiducial_zmax: float
+                    upper z coordinate of fiducial volume (the fiducial volume needs to be chosen large enough such that no events outside of it will trigger)
+                * full_rmin: float (optional)
+                    lower r coordinate of simulated volume (if None it is set to 1/3 of the fiducial volume, if second vertices are not activated it is set to the fiducial volume)
+                * full_rmax: float (optional)
+                    upper r coordinate of simulated volume (if None it is set to 5x the fiducial volume, if second vertices are not activated it is set to the fiducial volume)
```
daniel-zid, line 542: Shouldn't this documentation be updated?

NuRadioMC/EvtGen/generator.py
```diff
@@ -784,23 +789,50 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
...
+            * fiducial_rmax: float
+                upper r coordinate of fiducial volume (the fiducial volume needs to be chosen large enough such that no events outside of it will trigger)
+            * fiducial_zmin: float
+                lower z coordinate of fiducial volume (the fiducial volume needs to be chosen large enough such that no events outside of it will trigger)
+            * fiducial_zmax: float
+                upper z coordinate of fiducial volume (the fiducial volume needs to be chosen large enough such that no events outside of it will trigger)
+            * full_rmin: float (optional)
+                lower r coordinate of simulated volume (if None it is set to 1/3 of the fiducial volume, if second vertices are not activated it is set to the fiducial volume)
+            * full_rmax: float (optional)
+                upper r coordinate of simulated volume (if None it is set to 5x the fiducial volume, if second vertices are not activated it is set to the fiducial volume)
```
daniel-zid, line 806: These doc strings must be changed as well.

NuRadioMC/EvtGen/generator.py
cg-laser: @daniel-zid what do you suggest to improve it?

NuRadioMC/EvtGen/generator.py
cg-laser: good catch. I changed it.

NuRadioMC/EvtGen/generator.py
cg-laser: done

NuRadioMC/EvtGen/generator.py
cg-laser: done

NuRadioMC/EvtGen/generator.py
daniel-zid: Rock is ~ 3 times denser than ice, although radiation length is only 26.54 g/cm^2 vs 36.08 in ice. Maybe use the radiation length as correction factor? But of course that depends on the configuration file used with Proposal, so I don't have an easy solution unless we scan the config file. 

This increase in volume can be computationally expensive unless we implement a proper cut, and there's not much to be gained from it except at low energies. And even so...

PR #222, opened 2020-07-09
Signal visualization rotation

christophwelling:
The antenna rotation was displayed wrong in one of the plots. While on it, I changed the layout a bit to avoid dead space


anelles (approved):
Clean-up is always good.

PR #223, opened 2020-07-21
Fix lepton mask

daniel-zid:
This pull request implements a cut in azimuth directions for generating secondary particles.

### DIFF
--- a/NuRadioMC/EvtGen/NuRadioProposal.py
+++ b/NuRadioMC/EvtGen/NuRadioProposal.py
@@ -2,6 +2,7 @@ import proposal as pp
 import numpy as np
 from NuRadioReco.utilities import units
 import os
+import six
 
 """
 This module takes care of the PROPOSAL implementation. Some important things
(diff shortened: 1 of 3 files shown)


NuRadioMC/EvtGen/NuRadioProposal.py
```diff
@@ -224,6 +225,30 @@ def is_shower_primary(particle):
...
+    with open(config_file_path, 'r') as f:
+
+        config_file = json.load(f)
+
+    path_to_tables = config_file['global']['interpolation']['path_to_tables']
+    path_to_tables_readonly = config_file['global']['interpolation']['path_to_tables_readonly']
+
+    if not os.path.isdir(path_to_tables):
+
+        error_msg  = "'path_to_tables' in {} points to a non-existing directory. ".format(config_file_path)
```
cg-laser, line 243: could you also print the currently set path, same below

daniel-zid: You're right, that would be better.

PR #224, opened 2020-07-23
Fix veff secondaries

daniel-zid:
The current master counts the secondary interactions incorrectly when calculating effective volumes. Every interaction counts as a detected event, but this shouldn't be the case - a maximum of one triggered interaction should be counted for every parent neutrino. This pull request fixes the problem.

### DIFF
--- a/NuRadioMC/utilities/Veff.py
+++ b/NuRadioMC/utilities/Veff.py
@@ -1,6 +1,7 @@
 import numpy as np
 import h5py
 from scipy import interpolate
+from scipy.interpolate import interp1d
 import glob
 from six import iteritems
 import json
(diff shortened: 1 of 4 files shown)


NuRadioMC/utilities/Veff.py
```diff
@@ -125,8 +130,7 @@ def get_Aeff_proposal(folder, trigger_combinations={}, station=101):
         out['energy'] = E
 
         weights = np.array(fin['weights'])
-        # triggered = np.array(fin['triggered'])
-        triggered = get_triggered(fin)
```
cg-laser, line 129: shouldn't we keep the call to the `get_triggered` function here?

daniel-zid: That 'triggered' is only used to check if there are triggering events in the file, so it's faster to take the fin['triggered'] array, and the effect is the same.

daniel-zid:
> looks good. One could also add the uncertainty estimation using FC to the Veff function. But I think this can also be a feature request for later.

Sure, I can do that. It's just a handful of lines.

PR #230, opened 2020-08-12
Noise temperature

cg-laser:
adds option to use noise temperature from detector description. This option is activated by setting Tnoise to "detector" in the config file

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -306,14 +306,22 @@ class simulation():
         if(Tnoise is not None and Vrms is not None):
             raise AttributeError(f"Specifying noise temperature (set to {Tnoise}) and Vrms (set to {Vrms} is not allowed.")
         if(Tnoise is not None):
-            self._Tnoise = float(Tnoise)
+            if(Tnoise == "detector"):
+                self._Tnoise = None  # the noise temperature is defined in the detector description
+            else:
+                self._Tnoise = float(Tnoise)
             self._Vrms_per_channel = {}
             for station_id in self._bandwidth_per_channel:
                 self._Vrms_per_channel[station_id] = {}
                 for channel_id in self._bandwidth_per_channel[station_id]:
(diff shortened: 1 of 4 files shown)


NuRadioMC/simulation/simulation.py
```diff
@@ -306,14 +306,22 @@ def __init__(self, inputfilename,
         if(Tnoise is not None and Vrms is not None):
             raise AttributeError(f"Specifying noise temperature (set to {Tnoise}) and Vrms (set to {Vrms} is not allowed.")
         if(Tnoise is not None):
-            self._Tnoise = float(Tnoise)
+            if(Tnoise == "detector"):
+                self._Tnoise = None  # the noise temperature is defined in the detector description
+            else:
+                self._Tnoise = float(Tnoise)
```
anelles, line 312: I just noticed here -- sorry. But TNoise is not according to our coding convention, no? should it not be noise_temp or something?

anelles (changes requested):
Sorry to only notice this here, but any time as good as this to clean up the convention.

NuRadioMC/simulation/simulation.py
cg-laser: hmm this is how it is named in the config file. I would not change it as this will break all existing simulation settings. I could rename it just internally...

anelles:
I was not sure what you make of your comment here now. Was this a 'yes' or a 'no' on the renaming?

NuRadioMC/simulation/simulation.py
anelles: Is this a 'no' @cg-laser ?

cg-laser:
I wanted to say no, but I didn't really have a good argument, so I changed it. Can you approve it now?

anelles (approved):

cg-laser:
test is failing because of change of seed handling in NuRadioReco. This will be fixed in the internal-looping PR, so I go ahead and merge this branch into internal-looping.

PR #235, opened 2020-09-03
Refactor veff

cg-laser:
the veff and aeff for surface muons functions are combined as they did 99% the same stuff.


NuRadioMC/utilities/Veff.py
```diff
@@ -468,38 +222,48 @@ def get_Veff_single(filename, trigger_names, trigger_names_dict, trigger_combina
         fin.attrs['phimin']
```
christophwelling, line 222: It looks like this should be phimin = fin.attrs['phimin']
Same in line 224

NuRadioMC/utilities/Veff.py
```diff
@@ -468,38 +222,48 @@ def get_Veff_single(filename, trigger_names, trigger_names_dict, trigger_combina
...
         fin.attrs['phimax']
-    V = fin.attrs['volume']
+    if(veff_aeff == "veff"):
+        volume_proj_area = fin.attrs['volume']
+    elif(veff_aeff == "aeff_surface_muons"):
+        area = fin.attrs['area']
+        # The used area must be the projected area, perpendicular to the incoming
+        # flux, which leaves us with the following correction. Remember that the
+        # zenith bins must be small for the effective area to be correct.
+        volume_proj_area = area * 0.5 * (np.abs(np.cos(thetamin)) + np.abs(np.cos(thetamax)))
```
christophwelling, line 232: Shouldn't this be np.abs(...) - np.abs(...)  ?

NuRadioMC/utilities/Veff.py
```diff
@@ -557,7 +321,13 @@ def get_Veff_single(filename, trigger_names, trigger_names_dict, trigger_combina
                     triggered = triggered & (np.array(fin[f'station_{station:d}/ray_tracing_reflection'])[..., max_amps, 0][:, 0] == values['n_reflections'])
 
             triggered = remove_duplicate_triggers(triggered, fin['event_group_ids'])
-            Veff = V * np.sum(weights[triggered]) / n_events
+            Veff = volume_proj_area * np.sum(weights[triggered]) / n_events
```
christophwelling, line 324: I guess this depends on how exactly effective volume is defined, but do we have to divide by number of events or by the sum of all weights? I.e. is absorption in the earth considered part of the effective volume? Just asking to be sure

NuRadioMC/utilities/Veff.py
```diff
@@ -610,29 +380,29 @@ def get_Veff_single(filename, trigger_names, trigger_names_dict, trigger_combina
...
-            Veff_low = V * FC_low / n_events
-            Veff_high = V * FC_high / n_events
-            out['Veffs'][trigger_name] = [Veff, Vefferror, np.sum(weights[triggered]), Veff_low, Veff_high]
+            out[veff_aeff][trigger_name] = [Veff, Vefferror, np.sum(weights[triggered]), Veff_low, Veff_high]
     return out
 
 
 def tmp(args):
-    return get_Veff_single(*args)
+    return get_Veff_Aeff_single(*args)
```
christophwelling, line 396: Why is this function necessary? Can't you just call get_Veff_Aeff_single directly?

christophwelling (changes requested):
I think there are some minor errors, other than that it looks good.

NuRadioMC/utilities/Veff.py
cg-laser: no, this calculates the average zenith angle. So it is 1/2*(A+B)

christophwelling: Okay, I thought it was supposed to be an integral

NuRadioMC/utilities/Veff.py
cg-laser: you always divide by the number of events. If Earth attenuation is considered the nominator is the sum of the triggered events times their weight.

NuRadioMC/utilities/Veff.py
cg-laser: yes, thanks

NuRadioMC/utilities/Veff.py
cg-laser: there is a tiny difference. tmp unpacks the arguments with the `*` operator

christophwelling (approved):

PR #238, opened 2020-09-09
Direct line raytracer

christophwelling:
We added a simple direct line raytracer that just propagates the signal in a straight line. This is in preparation for adding more complex propagation modules like RadioPropa in the future.

To do this we had to make some changes to make it easier to add new raytracer modules:
- Any parameters that are specific to a given raytracer and methofds that use them have been moved from simulation.py into the raytracer module.
- Applying propagation effects to the electric field is handled by the apply_propagation_effects method of the raytracer module. It is the propagation module's job to know which effects to apply, though settings can still be specified in the config file.
- The propagation module needs to provide a function that returns the maximum number of possible raytracing solutions so the output data structure can be created. We might want to consider handling this differently in the future, since this number could be hard to figure out once more complicated propagation effects are considered, but it works for now
- Start and end points of the raytracing are no longer specified in the init method but in a dedicated method. This means that instead of creating a new raytracer object every time, only a single one is created and the start and end points are changed as needed.


NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1909,10 +1966,10 @@ def get_focusing(self, iS, dz, limit=2.):
             vetPos = copy.copy(self.__X1)
             recPos = copy.copy(self.__X2)
             recPos1 = np.array([self.__X2[0], self.__X2[1], self.__X2[2] + dz])
-        if(not hasattr(self, "_r1")):
-            self._r1 = ray_tracing(vetPos, recPos1, self.__medium, self.__attenuation_model, logging.WARNING,
-                             self.__n_frequencies_integration, self.__n_reflections)
-            self._r1.find_solutions()
+        self._r1 = ray_tracing(self.__medium, self.__attenuation_model, logging.WARNING,
```
cg-laser, line 1969: there was a reason for buffering results. I don't remember exactly what it was but I suspect that this slows down the code.

NuRadioMC/simulation/simulation.py
```diff
@@ -558,53 +572,38 @@ def run(self):
...
                         # loop through all ray tracing solution
-                        for iS in range(r.get_number_of_solutions()):
-                            sg['ray_tracing_C0'][iSh, channel_id, iS] = r.get_results()[iS]['C0']
-                            sg['ray_tracing_C1'][iSh, channel_id, iS] = r.get_results()[iS]['C1']
-                            sg['ray_tracing_reflection'][iSh, channel_id, iS] = r.get_results()[iS]['reflection']
-                            sg['ray_tracing_reflection_case'][iSh, channel_id, iS] = r.get_results()[iS]['reflection_case']
-                            sg['ray_tracing_solution_type'][iSh, channel_id, iS] = r.get_solution_type(iS)
-                            self._launch_vector = r.get_launch_vector(iS)
+                        for iS in range(self._raytracer.get_number_of_solutions()):
+                            self._raytracer.write_raytracing_output(sg, iSh, channel_id, iS)
```
cg-laser, line 590: this interface should be different too. The ray tracer should return a dict with the parameters that define the solution. simulation.py should take case how it is written into the hdf5 file

NuRadioMC/simulation/simulation.py
```diff
@@ -1225,24 +1165,19 @@ def _create_meta_output_datastructures(self):
...
-        sg['ray_tracing_reflection_case'] = np.ones((n_showers, n_antennas, nS), dtype=np.int) * -1
-        sg['ray_tracing_solution_type'] = np.ones((n_showers, n_antennas, nS), dtype=np.int) * -1
         sg['polarization'] = np.zeros((n_showers, n_antennas, nS, 3)) * np.nan
         sg['travel_times'] = np.zeros((n_showers, n_antennas, nS)) * np.nan
         sg['travel_distances'] = np.zeros((n_showers, n_antennas, nS)) * np.nan
-        sg['focusing_factor'] = np.ones((n_showers, n_antennas, nS))
         if(self._cfg['speedup']['amp_per_ray_solution']):
             sg['max_amp_shower_and_ray'] = np.zeros((n_showers, n_antennas, nS))
             sg['time_shower_and_ray'] = np.zeros((n_showers, n_antennas, nS))
+        self._raytracer.create_output_data_structure(sg, n_showers, n_antennas)
```
cg-laser, line 1180: the ray tracer should return a list of parameters. But not write them directly into the hdf5 file.

NuRadioMC/simulation/simulation.py
```diff
@@ -754,6 +695,7 @@ def run(self):
                             if(iS is None):
                                 a = 1 / 0
                             electric_field.set_frequency_spectrum(np.array([eR, eTheta, ePhi]), 1. / self._dt)
+                            electric_field = self._raytracer.apply_propagation_effects(electric_field, iS)
```
cg-laser, line 698: I think it would be better that this function takes the relevant config object. For many cases this might not be needed from the start. Also for certain scenarios one would like to get different propagation effects for the same ray tracing solution. This would not be possible with this interface.

cg-laser (changes requested):
sorry for being slow on providing feedback. This is a major change in the interface, so we should review it properly. 

It is great to see that you take a first step in further modularizing the signal propagation component. Many changes are good, but I don't think the current solution is optimal yet. 

A lot of simulation.py specifics are put into the ray tracing class which is against the modularity of the software. E.g. that the ray tracer has a shower and channel id now. It does not need to know that. There are many other cases the ray tracer is used for (e.g. to calculate things for calibration measurements) where this is not applicable. 

I think it is ok/good to put more of the calculation of the propagation effects into the ray tracer but would make the interface different. The calculation should not be tied to a NuRadioReco object, and one should also not need a detector object to use the ray tracer. Instead, the function can just return an array in the frequency domain with the attenuation factors for the requested frequencies.

NuRadioMC/simulation/simulation.py
christophwelling: Then how should we handle initial values? Does that also need to be specified or can we just initialize them all as nan or something like that?

NuRadioMC/simulation/simulation.py
christophwelling: I think it is better to avoid passing the config as a function parameter. If it is stored as a class parameter, it is available for any class method. So if we want to make one of the functions configurable for one raytracer, we don't have to change the interface to do so.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1940,3 +1998,79 @@ def get_ray_path(self, iS):
...
+    def create_output_data_structure(self, dictionary, n_showers, n_antennas):
+        nS = self.get_number_of_raytracing_solutions()
+        dictionary['ray_tracing_C0'] = np.zeros((n_showers, n_antennas, nS)) * np.nan
+        dictionary['ray_tracing_C1'] = np.zeros((n_showers, n_antennas, nS)) * np.nan
+        dictionary['focusing_factor'] = np.ones((n_showers, n_antennas, nS))
+        dictionary['ray_tracing_reflection'] = np.ones((n_showers, n_antennas, nS), dtype=np.int) * -1
+        dictionary['ray_tracing_reflection_case'] = np.ones((n_showers, n_antennas, nS), dtype=np.int) * -1
+        dictionary['ray_tracing_solution_type'] = np.ones((n_showers, n_antennas, nS), dtype=np.int) * -1
+
+    def get_ray_tracing_perfomed(self, station_dictionary, station_id):
```
christophwelling, line 2011: I guess this is what you meant with the raytracer now having a station ID? I don't see another solution here, since the ray_tracing_C0 property is specific to this raytracer, so we should keep it out of simulation.py, but we need to know the station for which we need to check for the solution

NuRadioMC/SignalProp/analyticraytracing.py
christophwelling: Okay, I added the buffering back in

NuRadioMC/simulation/simulation.py
cg-laser: I would initialize everything as nan. Then it is clear if a value was filled of was just left empty

NuRadioMC/simulation/simulation.py
cg-laser: but this would mean calculating different attenuation options for the same ray become quite cumbersome. 
The interface change would be backward compatible. the function would just get an additional `config` attribute that could have a default setting. 
I agree, individual function arguments might limit us. but what about passing a config dict to the function?

NuRadioMC/SignalProp/analyticraytracing.py
cg-laser: If you move the filling of hdf5 files (or the data structure in general) to simulation.py (how it was previously) you can just check if you have an entry in your data for this station id. If you do have it, initialize the ray tracer accordingly.

christophwelling: But just having an entry for the station ID doesn't necessarily mean that that the raytracing was actually performed, right? The entry for the station is created at the start and filled with nan, so we would need to check if the parameters for the raytracing solution are set. The problem is that the C_0 and C_1 parameters are specific to the analytic raytracer, so we should check for them there.
How about if simulation.py passes only the entry for the station (station_dictionary[station_id]) to the raytracer and it checks if the parameters for its solution are set? That way the raytracer doesn't need to know about the station IDs and the simulation.py doesn't need to know anything specific about the raytracer.

NuRadioMC/simulation/simulation.py
christophwelling: This wouldn't quite solve the problem. If we want to make another method of another raytracer configurable, we need to change the method call in simulation.py to pass the config as well. So unless we change the methods in all other raytracers to accept a config, we will get the "unrecognized parameter" error if they are used.

NuRadioMC/simulation/simulation.py
christophwelling: Done. I added a parameter to specify the number of entries, in case we have multi-dimensional parameters

NuRadioMC/simulation/simulation.py
christophwelling: Done

NuRadioMC/SignalProp/analyticraytracing.py
cg-laser: maybe I'm missing something here, but wouldn't you need to first read the ray tracer solution values from the hdf5 file and initialize the ray tracer with these values? So you always need to check the hdf5 file first. 
If you have a nan in the hdf5 file, then this is a correct information, because it means that no ray tracing solution was found for this parameter.

NuRadioMC/simulation/simulation.py
cg-laser: no, the config dictionary that you would pass to the init function of the ray tracer, would just be passed to the function directly. If the dict contains an additional parameter that the function does not use, it won't complain.

christophwelling: Yes, I am talking about passing the config to a function that did not accept it before. Let's say I want to make one of the other functions configurable in another raytracer. I would have to change that function in the raytracer to accept the config and change the simulation.py to pass the config to that function. If I don't change this function in all other raytracers, they will throw an error, since they now get an additional paramter they don't know.
As I said, I'm okay with implementing it this way, since I don't work on the raytracers much. But this does make the raytracers somewhat less modular

NuRadioMC/SignalProp/analyticraytracing.py
christophwelling: I don't think this is ever set in the raytracer. The all the entries in the input hdf are copied into the output data structure, but that is simply done by looping over its keys, so we don't need to hardcode the key names there. If get_raytracing_performed returns True, that means the raytracing solution is already in the output dict, so we don't need to do anything else.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1584,23 +1627,37 @@ class describing the index-of-refraction profile
         self.__logger.debug("X2 - X1 = {}, X1r = {}, X2r = {}".format(self.__X2 - self.__X1, X1r, X2r))
         self.__x1 = np.array([X1r[0], X1r[2]])
         self.__x2 = np.array([X2r[0], X2r[2]])
-
         self.__logger.debug("2D points {} {}".format(self.__x1, self.__x2))
-        self.__r2d = ray_tracing_2D(self.__medium, self.__attenuation_model, log_level=log_level,
-                                    n_frequencies_integration=self.__n_frequencies_integration)
 
-    def set_solution(self, C0s, C1s, solution_types, reflection=None, reflection_case=None):
+    def set_solution(self, raytracing_results, i_shower, channel_id):
```
cg-laser, line 1632: is there a way to make the interface better? The `i_shower` and `channel_id` paramters are just used to index the `ray_tracing_results` dict. The ray tracer can also be used independent of a MC simulation. Can't you just pass a dict with the properties of this specific shower/channel combination?

NuRadioMC/simulation/simulation.py
```diff
@@ -458,7 +467,7 @@ def run(self):
 
                 ray_tracing_performed = False
                 if(self._station_id in self._fin_stations):
-                    ray_tracing_performed = ('ray_tracing_C0' in self._fin_stations[self._station_id]) and (self._was_pre_simulated)
+                    ray_tracing_performed = self._raytracer.get_raytracing_perfomed() and (self._was_pre_simulated)
```
cg-laser, line 470: I think this doesn't work yet. Instead you'd need to loop through the ray tracer parameters and check if they are in the hdf5 file.

cg-laser (changes requested):
sorry for the delay in reviewing. I added a few more inline comments. 

About passing a general config object to the init method... you listed the advantages. The disadvantage is that one can't easily get e.g. the propagation effects for different settings. E.g. with/without focusing. Instead, one would need to initiate a new ray tracing class. In NuRadioMC this is not needed. And I'm currently also not using this feature somewhere else. But it might be handy for certain analyses. This is why I think it is less modular for a user. I guess one could always just call the individual functions and puzzle the propagation effects together manually. 
But I agree, from a coding standpoint, just passing a general config object to the ray tracer class gives maximum flexibility. I guess I'm good with it. 

I'm thinking at a different use case: For a given receiver position and receive angle, calculate the ray path. For a given vertex distance the source position could be calculated. Thus, we don't have a start/stop point. Does the structure allow for this flexibility? I think yes, one would just need a different function. Instead of `set_start_and_end_point` one would use a different function like `set_start_and_receive_angle`.

cg-laser:
@christophwelling I think what we can just do is to add a `get/set_config` function to the ray tracer. This allows to reset configs on-the-fly to e.g. retrieve propagation effects with other configs without redoing the full ray tracing calculation.

christophwelling:
That sounds like a good idea.

NuRadioMC/SignalProp/analyticraytracing.py
christophwelling: I don't think there is an easy way to do this. raytracing_results is a dictionary and simulation.py does not know which properties the raytracer needs. The only solution I can think of is to loop through all keys and copy the values into a new dictionary, but that would probably be slow.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1940,3 +1998,83 @@ def get_ray_path(self, iS):
...
+            phase_shift = (i_reflections * self.__medium.reflection_phase_shift) % (2 * np.pi)
+            # we assume that both efield components are equally affected
+            spec[1] *= reflection_coefficient * np.exp(1j * phase_shift)
+            spec[2] *= reflection_coefficient * np.exp(1j * phase_shift)
+            self.__logger.debug(
+                f"ray is reflecting {i_reflections:d} times at the bottom -> reducing the signal by a factor of {reflection_coefficient:.2f}")
+
+        # apply the focusing effect
+        if self.__config['propagation']['focusing']:
+            dZRec = -0.01 * units.m
```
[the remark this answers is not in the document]
christophwelling: Done

NuRadioMC/simulation/simulation.py
cg-laser: I think we resolved this now with the get/set config method?

NuRadioMC/SignalProp/analyticraytracing.py
cg-laser: but you know which properties the ray tracer needs by calling the corresponding function of the ray tracer `get_output_parameters`.

christophwelling: So we check if all the parameters from get_output_parameters are set? I'm okay with that, but it could be a problem if we have some parameter that is not set for every solution.

cg-laser: we could also just check the first parameter to avoid this problem. The first parameter must then be the parameter that is always defined.

NuRadioMC/simulation/simulation.py
```diff
@@ -790,11 +733,9 @@ def run(self):
                             electric_field.set_trace_start_time(trace_start_time)
                             electric_field[efp.azimuth] = azimuth
                             electric_field[efp.zenith] = zenith
-                            electric_field[efp.ray_path_type] = self._prop.solution_types[r.get_solution_type(iS)]
+                            electric_field[efp.ray_path_type] = self._prop.solution_types[self._raytracer.get_solution_type(iS)]
                             electric_field[efp.nu_vertex_distance] = sg['travel_distances'][iSh, channel_id, iS]
                             electric_field[efp.nu_viewing_angle] = viewing_angles[iS]
-                            electric_field[efp.reflection_coefficient_theta] = r_theta
```
cg-laser, line 796: I just notice another thing. Can we save all ray tracing parameters to the efield? At the moment we don't know which efield corresponds to which ray tracing solution. 
We could just save the new "propagation output dict" as efield parameter.

NuRadioMC/simulation/simulation.py
christophwelling: Done

NuRadioMC/simulation/simulation.py
christophwelling: We can do that, but it's a separate issue, so let's do it in another pull request. This one is already complex enough

NuRadioMC/SignalProp/analyticraytracing.py
christophwelling: Done

christophwelling:
I think I addressed all the points now. The only thing I did not change is the way the propagation effect is applied. I think passing the electric field object gives us more flexibility in the long run if we implement more complicated propagation simulations.

christophwelling:
The tests now run on this PR again and they pass. Are there any objections to merging this now?

cg-laser:
Can we switch to the new development branch setup first? 
The reason is because we use the current master (with version tag) for the gen2 production. If we merge it now to the master as well, we won't be able to add potential hotfixes to the release version. The tests look good, so probably now really a huge risk, but I'm hesitant to merge a large change into the production version withing a large sim production and the upcoming gen2 review.

christophwelling:
Sure, I assumed we wanted to first merge all PRs and then change the branch setup, but I'm fine with doing it the other way around.

cg-laser (approved):

PR #240, opened 2020-09-10
Reference for fourier shift

christophwelling:
Updates references for test to match the new fourier shift change on NuRadioReco


cg-laser (approved):
we checked before that the differences are small (<1%) which is an expected change

PR #241, opened 2020-09-15
Fix hdf5 merge

clark2668:
Issue being addressed: merge did not generate unique `event_group_id`

The hdf5 merger didn't seem to overwrite the non-unique `event_group_ids` elements when merging hdf5 files together. In particular, the duplicated element in `current_uegids` didn't get overwritten with the new id before `current_uegids` was appended to the running list stored in `unique_uegids`. This meant that an ID (call it X) could accidentally be included twice in	the `unique_uegids` array, and so the intersection of `current_uegids` and `unique_uegids` can return X (the duplicated element) even when `current_uegids` is empty, an this will cause a crash in the `new_egid = max(unique_uegids.max(), current_uegids.max()) + 1` line. I think this was happening because the dummy array (`current_egids`) which *does* get overwritten by the new unique ID never got put back into the `current_uegids` array.

I also restored a debugging message which I think would have caught this.

And added some docs.

Did you remember to: 
* include doc strings in all commits ? - N/A
* edit the changelog.txt file describing your changes? - yes, bump to version 1.2.1

### DIFF
--- a/NuRadioMC/utilities/merge_hdf5.py
+++ b/NuRadioMC/utilities/merge_hdf5.py
@@ -101,11 +101,11 @@ def merge2(filenames, output_filename):
                 # also change the event_group_id arrays of the station groups
                 if f in non_empty_filenames:
                     for key in groups[f]:  # loop through all groups
-                        print(f"group {key}, keys = {groups[f][key].keys()}")
-                        g_egids = groups[f][key]['event_group_ids']
-                        mask_g = gid == g_egids
-                        if(np.sum(mask_g)):  # station might not have this event group id, so skip stations where this egid is not present
-                            g_egids[mask_g] = new_egid
+                        if 'event_group_ids' in groups[f][key]:
+                            g_egids = groups[f][key]['event_group_ids']
+                            mask_g = gid == g_egids
+                            if(np.sum(mask_g)):  # station might not have this event group id, so skip stations where this egid is not present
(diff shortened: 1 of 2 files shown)


cg-laser:
@clark2668 thanks for finding this bug. There was one mistake in your new logic, you actually never change the event group ids. The array `data[f]['event_group_ids']` needs to be changed. It was probably a bis obscure as `current_egids = data[f]['event_group_ids']` makes `current_egids` a reference to the dict. 
I also forgot to change the events group arrays on station level. I added this too. Can you check if it works for you now?

clark2668:
Sure @cg-laser. That `current_egids = data[f]['event_group_ids']` makes a reference was lost on me, thanks.

So, when I run it now, I get the following error:

  File "merge_hdf5.py", line 265, in <module>
    merge2(input_args[i]['filenames'], input_args[i]['output_filename'])
  File "merge_hdf5.py", line 104, in merge2
    g_egids = groups[f][key]['event_group_ids']
KeyError: 'event_group_ids'

Somehow there's a KeyError? The first "key" is "station_1", so I guess that's missing the 'event_group_ids' element?

clark2668:
This latest commit seems to do the trick, the merge goes through now. It seems Christian needed to add a check for if the `event_group_ids` are in the keys at all.

I will leave the PR open for another day for additional comments!

cg-laser (approved):

PR #243, opened 2020-09-25
speedup with weight cut

cg-laser:
move weight cut up to speed up simulation. Now an event group is skiped right away if the neutrino weight is too small.

As a consequence input files for with events below the horizon can be simulate extremely fast .

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -411,15 +411,6 @@ class simulation():
                 continue
             event_indices = np.atleast_1d(np.squeeze(np.argwhere(self._fin['event_group_ids'] == event_group_id)))
 
-            # these quantities get computed to apply the distance cut as a function of shower energies
-            # the shower energies of closeby showers will be added as they can constructively interfere
-            if self._cfg['speedup']['distance_cut']:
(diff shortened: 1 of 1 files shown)


clark2668:
Hey Christian, I think this speed up makes a lot sense. As a question, why did it changes from `self._shower_index` to `iE_mother`. Or rather, it seems like `self._mout['weights'] was always filled by mother index, so was it doing something wrong before when it was checking the weights array by shower index and not by mother index?

cg-laser:
this is just something that I overlooked when restructuring the code. Everything was correct before but just slower. Before, the weight was checked for every shower at a later stage in the simulation. However, the weight is only determined by the mother particle (the incident neutrino), so we can reject a whole event group right away. There is no need to iterate over the showers.

clark2668 (approved):
Okay--I guess I was confused why the old weights dict seemed to be indexed by neutrino but accessed by shower, while the current one is index by neutrino and accessed by neutrino (which seems correct). But there must be a subtlety evading me!

cg-laser:
yeah so what was done is that the weight of the initial neutrino was saved as the weight for each shower. A consequence of the linear structure of the hdf5 files.

PR #244, opened 2020-10-02
Speedup distance cut

cg-laser:
stations that are too far away from any shower of an event group are skipped more efficiently. This results in more than a factor of 2 speed improvements for cases where many stations are simulated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,9 @@ before_script:
   - unzip /tmp/radiotools.zip
   - mv radiotools-master radiotools
   - export PYTHONPATH=$PWD/radiotools
-  - wget https://github.com/nu-radio/NuRadioReco/archive/fourier_shift.zip -O /tmp/NuRadioReco.zip
+  - wget https://github.com/nu-radio/NuRadioReco/archive/master.zip -O /tmp/NuRadioReco.zip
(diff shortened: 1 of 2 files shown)


clark2668 (commented):
Hi Christian! Thanks for this--I think this mostly looks good!! Is there an accidental re-use of the `vertex_positions` variable? (L422 and L452). It just might save you declaring it twice.

Also, should the `+100` safety margin on the distance cut just be put into the distance cut function itself?

Also also, perhaps this will get fixed once you do the merge, but I don't see the work from the `speedup-weight-calculation` PR (https://github.com/nu-radio/NuRadioMC/commit/f620f31e6b1717d41a084fa3ab61c482ac4217d5) in here? Maybe this branch was stale?

cg-laser:
thanks Brian.
I now declare the vertex_positions only once. 

The 100m safety margin is because of a possible extent of the station. I only calculate the distance to the barycenter of the station. In the later cut where I loop over each channel, the distance to each antenna is used and this extra 100m is not used. 

I also merged the current master into this branch.

clark2668:
Great on the vertex_positions variable and the merge from the master!

And ahh, I see. Makes sense to leave the 100m in. I'll click approve!

clark2668 (approved):

PR #245, opened 2020-10-08
make event group ids in final hdf5 file consecutive

cg-laser:
this PR makes event group ids consecutive. This simplifies to make sure the event group ids are unique when creating many input files. 
Only relevant if secondary interactions are simulated.


clark2668 (commented):
Okay, i think this needs a tweak to use numpy arrays in order for the mask to work (I don't think it works on lists?). So I made that change. Otherwise I think this looks okay. Just as food for thought: should it be an option? I'm trying to imagine if there's ever a case where you don't want the event_group_ids reordered...

cg-laser:
I can't think of any case where we need it. These additional event never get saved. 
Thanks for fixing the mask

clark2668:
No problem. And okay, then I think this is ready.

PR #246, opened 2020-10-27
add option for simulating noiseless channels

cg-laser:
adds the feature to disable adding noise to certain channels if the detector description contains the field `noiseless`. Useful to add single antenna proxies for a PA to a full PA simulation.


clark2668 (approved):
Hi Christian, I think this looks okay! I assume this is accompanied by a change in NuRadioReco to the det class and to the channelGenericNoiseAdder module?

cg-laser:
I ran all tests locally and they pass

PR #247, opened 2020-11-04
H5py fix

christophwelling:
Fixes the problems caused by the new h5py version.

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -1099,10 +1099,11 @@ class simulation():
                 self._fin_stations[key] = {}
                 for key2, value2 in iteritems(value):
                     self._fin_stations[key][key2] = np.array(value2)
-            if type(value[0]) == bytes:
-                self._fin[key] = np.array(value).astype('U')
             else:
-                self._fin[key] = np.array(value)
+                if type(value[0]) == bytes:
+                    self._fin[key] = np.array(value).astype('U')
+                else:
+                    self._fin[key] = np.array(value)
         for key, value in iteritems(fin.attrs):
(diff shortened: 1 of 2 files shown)


NuRadioMC/simulation/simulation.py
```diff
@@ -1397,7 +1401,11 @@ def _write_ouput_file(self):
             if(key.startswith("station_")):
                 continue
             if(not key in fout.keys()):  # only save data sets that havn't been recomputed and saved already
-                fout[key] = np.array(self._fin[key])[saved]
+                if self._fin[key].dtype.char == 'U':
+                    fout[key] = np.array(self._fin[key].astype('S'))[saved]
```
cg-laser, line 1405: why is this explicit casting required?

christophwelling: h5py does not the numpy string type. So the best way I could come up with to solve this is to cast it back into a bytestring

cg-laser: In the event generator we do it like this
```
            if value.dtype.kind == 'U':
                fout[key] = np.array(value, dtype=h5py.string_dtype(encoding='utf-8'))[start_index:stop_index]
            else:
                fout[key] = value[start_index:stop_index]
```
this might be the betters solution(?). We want the strings to be unicode I guess.

christophwelling: Good point.

christophwelling:
Can we merge this so the tests work again?

anelles (approved):
Fine with me. I think you agreed on the string discussion now.

cg-laser (approved):

Issue #89, opened 2019-06-06
unify interface to flux models, clean up flux plots


cg-laser:
There are actually two but related issues: 
1) The examples/Sensitivties folder contains three versions of the same script E2_fluxes 1-3. It seems that they are just newer/updated versions of the same functionality.
2) we need an unified interface to flux predictions (GZK models, icecube measurements, limits etc.) This is currently reimplemented in the different flux limits plotting scripts as well as in the event generator #88

anelles:
Even though this has been assigned to me a long while ago, nothing happened. Still in progress.

PR #249, opened 2020-11-24
Adding script to make multimessenger figure from RNO-G Concept Paper.

anelles:
For documentation purposes and for everyone to have access: the multi-messenger figure. 
Disclaimer: I am sure that the linter will not like the current version.


anelles:
It is not quite #89 (sorry, @cg-laser this ship may have sailed), but at least I updated this so that everyone has access to the newer figures ...

shallmann (approved):
Thanks for adding those. I think it would be great to have the sensitivity/flux tables merged to develop.
I trust you that the posted json/txt file numbers are correct ;).
Two minor items which could be modified:
1/ do we need special characters in the file names, like + and =?
2/ maybe it is helpful to add DOI/arXiv references for the tabulated flux values if you still have it?

anelles:
I assume we can still merge this? Or will this now break all the work that @lpyras is doing?

lpyras:
we can merge this :)

PR #251, opened 2020-12-09
fix noise normalization

cg-laser:
the normalization for the genericNoiseAdder module was not calculated correctly. This was an oversight from the `internal-looping` refactoring. Originally we implemented that only the amplifieres/filters before the noise adder module were counted. Because no noise adder module was called in the presimulation to determine the bandwidth of each channel, the normalization was never set and was assumed to always be one. This led to smaller noise levels, in particular if the Nquist frequency is much higher than the detector bandwidth. 
The Vrms calculation and the corresponding thresholds were not effected but the noise that was added had a too small amplitude. 

In this PR I simplify this calculation. We always add noise before the filter, so we can just use the bandwidth of the signal chain for the noise normalization. More complicated scenarios (like adding noise at several stages) would require the absolute amount of noise added at each stage and couldn't be automated anyways. (It is possible to do that by just defining a custom detector simulation).


cg-laser:
I ran all test locally and they pass, so I think this PR would be ready to be merged. 
I added a simple script to execute all tests locally. I think this is a good workaround until we get new travis credits.

cg-laser:
the only test that is not working is the phased array example in NuRadioReco because the `get_noise_normalization` function does not exist anymore. This needs to be fixed in NuRadioReco but we can already merge this branch.

cg-laser:
> the only test that is not working is the phased array example in NuRadioReco because the `get_noise_normalization` function does not exist anymore. This needs to be fixed in NuRadioReco but we can already merge this branch.

it is actually already fixed on the refactor-pa branch

PR #253, opened 2020-12-11
add uncertainties of cross section

cg-laser:
add parameterization of uncertainties to CST cross sections.

It is a draft because I still need to add a saveguard against too low energies where the method fails numerically


shallmann (commented):
Does this safeguard work, if an array of energy is passed?

cg-laser:
good point. I did not consider that.

shallmann (approved):
looks good to me

PR #257, opened 2020-12-23
add quadrant check to apply correct volume extension

### LINKED ISSUE #254: proposal volume extension
Volume extension by the 95% quantile of the tau-length seems in my opinion to be wrongly implemented:

https://github.com/nu-radio/NuRadioMC/blob/8d335c0dd0e6453ad70d9e1604eb0ea52a1442fe/NuRadioMC/EvtGen/generator.py#L438
https://github.com/nu-radio/NuRadioMC/blob/8d335c0dd0e6453ad70d9e1604eb0ea52a1442fe/NuRadioMC/EvtGen/generator.py#L497

If I choose a zenith range from 0 to pi, the max_horizontal_distance = np.abs(max(0, 0)*fiducial_zmin = 0
as a consequence the horizontal distance is not extended.

What is intended is to have a max_in_thetarange, i.e. some sampling like
```
dvals = np.linspace(attributes['thetamin'], attributes['thetamax'], 1000)
max_horizontal_dist = np.abs(max(np.abs(np.tan(dvals)))*volume['fiducial_zmin'])
```
would be a dirty fix.. any better ideas?

shallmann:
Avoids improper extension of sim volume by checking thetaminmax quadrant, closes #254 

* edit the changelog.txt file describing your changes? No

### DIFF
--- a/NuRadioMC/EvtGen/generator.py
+++ b/NuRadioMC/EvtGen/generator.py
@@ -443,7 +443,12 @@ def set_volume_attributes(volume, proposal, attributes):
                 rmax = volume['full_rmax']
             else:
                 tau_95_length = get_tau_95_length(attributes['Emax'])
-                max_horizontal_dist = np.abs(max(np.abs(np.tan(attributes['thetamin'])), np.abs(np.tan(attributes['thetamax']))) * volume['fiducial_zmin'])  # calculates the maximum horizontal distance through the ice
+                # check if thetamax/thetamin are in same quadrant
+                is_same_quadrant = np.sign(np.cos(attributes['thetamin']))==np.sign(np.cos(attributes['thetamax']))
+                if is_same_quadrant:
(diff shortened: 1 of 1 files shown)


anelles (approved):
Looks good to me. Like I commented in the issue, I would like to see whether it makes a large difference. Just to be sure how much we are impacting previous results.

cg-laser (changes requested):
this implementation does not work. It will lead to infinite volumes whenever the zenith bin ends at 90deg, which is a very typical choice. 
Also returning infinity does not make practical sense, because it is only infinity if the Earth curvature is neglected. 
A useful implementation needs to limit the maximum extent to a reasonable number.

shallmann:
I did not change anything at the point you criticised...
the maximum distance is limited below by
d_extent = min(tau_95_length, max_horizontal_dist)
as before.

But ok, let's drop the change and add just a comment

shallmann:
@cg-laser I added the comment on the geometrical infinity

anelles (approved):

anelles:
@cg-laser I think this one is waiting on your approval

shallmann:
@cg-laser I can finally merge this right?

cg-laser:
we will make a new PR into develop and close this one unmerged.

PR #260, opened 2021-01-05
Use an energy that doesn't trigger warnings in the cross section file.

anelles:
I tried fixing the pulser example, by at least suppressing the warning for having an energy (which is not used) of 1 eV. 

The example still doesn't run, but due to an incompatibility with NuRadioReco, I assume. 

$ python A03reconstruct_sim.py someoutput.hdf5 
using CPP version of ray tracer
distance = 653.38
WARNING:NuRadioReco.detector:loading detector description from /Users/anelles/ARIANNA/Reconstruction/NuRadioReco/NuRadioReco/detector/ARIANNA/arianna_detector_db.json
ERROR:NuRadioReco.NuRadioRecoio:data file not readable. File has version 11052129732745.2586 but current version is 2.2


anelles:
I started this PR so that we fix the example in the end.

PR #266, opened 2021-01-26
add downsampling of sim channels

cg-laser:
Sim channels were not automatically downsampled to detector resolution which increases output file size of the nur files by quite a bit


christophwelling (approved):
Looks good, but shouldn't the change be added to the changelog?

PR #269, opened 2021-03-24
Removes use of NuRadioReco versions and git hashes

christophwelling:
Now that NuRadioReco is no longer its own repo, storing its versions and git hashes won't work.
This removes all the calls to them. Am I forgetting any other place where they are used?


cg-laser (changes requested):
looks good! Can you remove the old linter tests? This is now already integrated in the github actions tests. 
And when you merge it, it also needs to be merged into develop

cg-laser (approved):

PR #272, opened 2021-03-30
Hotfix/prepare2.0release

cg-laser:
update readme, changelog etc. for pip release


cg-laser:
@christophwelling did I forgot any dependencies? I thought about leaving matplotlib out, but we might have an import somewhere where it is used for debug plots.

christophwelling:
I think you forgot pickle, but other than that it looks fine.

christophwelling (approved):

cg-laser:
> I think you forgot pickle, but other than that it looks fine.

It actually turned out that pickle is a default package. therefore the pip deployment failed ;-)

PR #273, opened 2021-04-08
Prevent error if station time is not set

christophwelling:
The parse_header method in NuRadioRecoio throws an error if it finds a station whose time is not set. This PR fixes that.


lpyras (approved):
Thank you for doing this. It works for my case.

PR #280, opened 2021-04-13
fix delay_trace method.

cg-laser:
I don't understand the old way of cutting the trace. This seems to remove the time delay again. 
I made this cutting feature optional.

### DIFF
--- a/NuRadioReco/utilities/trace_utilities.py
+++ b/NuRadioReco/utilities/trace_utilities.py
@@ -111,7 +111,7 @@ def get_electric_field_energy_fluence(electric_field_trace, times, signal_window
     return f_signal * dt * conversion_factor_integrated_signal
 
 
-def upsampling_fir(trace, original_sampling_frequency, int_factor=2, ntaps=2**7):
+def upsampling_fir(trace, original_sampling_frequency, int_factor=2, ntaps=2 ** 7):
     """
     This function performs an upsampling by inserting a number of zeroes
     between samples and then applying a finite impulse response (FIR) filter.
@@ -224,7 +224,7 @@ def apply_butterworth(spectrum, frequencies, passband, order=8):
     return filtered_spectrum
 
 
-def delay_trace(trace, sampling_frequency, time_delay, delayed_samples):
+def delay_trace(trace, sampling_frequency, time_delay, delayed_samples=None):
     """
     Delays a trace by transforming it to frequency and multiplying by phases.
     Since this method is cyclic, the trace has to be cropped. It only accepts
@@ -240,8 +240,9 @@ def delay_trace(trace, sampling_frequency, time_delay, delayed_samples):
         Sampling rate for the trace
     time_delay: float
         Time delay used for transforming the trace. Must be positive or 0
-    delayed_samples: integer
+    delayed_samples: integer or None
         Number of samples that the delayed trace must contain
+        if None: the trace is not cut
 
     Returns
     -------
@@ -264,7 +265,8 @@ def delay_trace(trace, sampling_frequency, time_delay, delayed_samples):
 
     init_sample = int(time_delay * sampling_frequency) + 1
 
-    delayed_trace = delayed_trace[init_sample:None]
-    delayed_trace = delayed_trace[:delayed_samples]
+    if delayed_samples is not None:
+        delayed_trace = delayed_trace[init_sample:None]
+        delayed_trace = delayed_trace[:delayed_samples]
 
     return delayed_trace


daniel-zid:
Hi Christian,

The Fourier-based delay can do two things if left unchecked:
1) The end part of the trace can end up in the beginning, due to its cyclic nature. As a consequence, we can end up with acausal behaviour.
2) The delay is not, in general, an integer number of sampling periods.
Because of these two reasons, I decided to give the possibility of choosing the initial delayed sample that is relevant for the study. Of course, your new way of writing it works as well, but be aware of the cyclic nature of the method and remember that some artifacts and appear in the ends of the trace.

cg-laser:
Hey Daniel, cool that you're still around to look at these things. 
Can you explain a bit more what the original implementation did? Because it seems that it didn't do any time shift at all. 

```python
n_samples = 512
dt = 1 * units.ns
n_samples = 512
import numpy as np
trace = np.zeros(n_samples)
trace[255] = 1
t2 = trace_utilities.delay_trace(trace, 1/dt, 10*units.ns, n_samples)
np.argmax(t2)
np.argmax(trace)
```

daniel-zid:
What's the code you've posted from? I have never seen it.

I implemented the delay utility function for the ADC module. So, for that module what was relevant was a fractional delay. That is, we know that the ADC random shift is going to shift our trace by an amount of time that is between 0 and t_s, with t_s the sampling period. That is going to shift the trace between 0 and 1 sample, so to speak (a fractional amount). So, I implemented a Fourier-based delay, but discarded the first samples that had information about the end of the trace. What you see is something that looks like the original trace, but two important things have changed:

1) A fractional delay has been applied.
2) Although the trace looks the same, the times it refers to are not! For instance, the beginning of the trace would change from 0 ns to, let's say, 0.12 ns. In the case of the ADC the trace moves around due to the ADC jitter, and that is equivalent to having a trace with a different time frame.

Think of it this way. Let us say you want to shift the trace back two samples. Well, you would end up with the two end samples in the beginning. This makes no sense, so you discard them, and you end up with the same trace as in the beginning, only with the two samples in the end gone. Of course, this does not make much sense when the delay is an integer number of samples as we can just change our time frame, but what if the delay is fractional? Then it makes sense to discard bins that are physically rubbish.

Of course, if you want to actually shift the trace putting the end of the trace in the initial bins, what you're doing is correct. But I would be very careful with those first bins. Equivalently, if you're shifting backwards in time, be careful with the last bins.

daniel-zid:
> Hey Daniel, cool that you're still around to look at these things.
> Can you explain a bit more what the original implementation did? Because it seems that it didn't do any time shift at all.
> 
> ```python
> from NuRadioReco.utilities import trace_utilities
> n_samples = 512
> from NuRadioReco.utilities import units
> dt = 1 * units.ns
> n_samples = 512
> import numpy as np
> trace = np.zeros(n_samples)
> trace[255] = 1
> t2 = trace_utilities.delay_trace(trace, 1/dt, 10*units.ns, n_samples)
To see what I mean, change the delay from 10*units.ns to 0.5*units.ns.
> np.argmax(t2)
> np.argmax(trace)
> ```

cg-laser:
Thanks Daniel for the explanation. For delays smaller than the bin size that makes sense. But do I see it correctly that if you shift the trace by more than `dt` one also needs to keep track of the change in trace start time? 

The use case for me at the moment is to shift a trace by an arbitrary amount that is larger than the `dt` but not a multiple of `dt`. For example `dt=1ns` but I want to shift the trace by `10.21ns`. 

Maybe we should split this functionality up into two functions and be explicit in the docstring what it does exactly. For you rmethod, wouldn't it make sense to also return the change in trace start time?

daniel-zid:
Hi Christian,


O 09:25, mér., 14/04/2021, Christian Glaser ***@***.***>
escribiu:

> Thanks Daniel for the explanation. For delays smaller than the bin size
> that makes sense. But do I see it correctly that if you shift the trace by
> more than dt one also needs to keep track of the change in trace start
> time?
>
Yes, you're right. For ADCs, it is not necessary to know that change since
the jitter is random and the information is lost.

> The use case for me at the moment is to shift a trace by an arbitrary
> amount that is larger than the dt but not a multiple of dt. For example
> dt=1ns but I want to shift the trace by 10.21ns.
>
> Maybe we should split this functionality up into two functions and be
> explicit in the docstring what it does exactly.
>
That's a good solution.

> For you rmethod, wouldn't it make sense to also return the change in trace
> start time?
>
It's not really necessary but it can be returned for checking purposes.

christophwelling (approved):

PR #284, opened 2021-04-20
Added neutrino direction reconstruction module

ggaswint:
Adds a new module for neutrino direction reconstruction along with an example for how to run it.


cg-laser (dismissed):

anelles:
@Ilse can you maybe make sure that this makes sense and then merge it? 
Geoff is no longer in science, so I don't think he will ever merge this.

iplaisier (approved):

PR #285, opened 2021-04-22
extended the Veff calculation to smaller theta ranges wrt. sim file

shallmann:
The effective volume/area calculation up to now took the thetamin/max from the input file.
By using an additional optional argument for single files (`bounds_theta` in `get_Veff_Aeff_single()`) this range can now be limited to a smaller subrange.
In `get_Veff_Aeff`, where this function can be called for a (directory of) file(s), a corresponding optional argument `oversampling_theta=1` is introduced, allowing to calculate iteratively the Veff/Aeff for `oversampling_theta` subranges of theta.

Since the `get_Veff_Aeff_single()` is just called several times, this is not the most efficient way to do this, but the one with minor modification to the existing functions.

This function is helpful for coarse binning in theta, and even more if only one file with 4pi space angle has been generated

This shows the result with 5fold oversampling in theta:
[effective_area_oversampling.pdf](https://github.com/nu-radio/NuRadioMC/files/6359741/effective_area_oversampling.pdf)

(almost) minimum working example:
```
from NuRadioMC.utilities import cross_sections, Veff
import numpy as np

# assume some hdf5 files are in this directory...
directory="./test_veff"

v = []
cz = []

# first without oversampling, i.e. identical to the previous version
veff_dataarr = Veff.get_Veff_Aeff(directory)
for veff_data in veff_dataarr:
    cosz = (np.cos(veff_data['thetamin'])+np.cos(veff_data['thetamax'])) / 2.
    veff = veff_data['veff']['dipole_2.0sigma'][0] #['PA_4channel_100Hz'][0]
    v.append(veff)
    cz.append(cosz)

# now with oversampling
oversampling_theta = 5
v_fine = []
cz_fine = []
veff_dataarr = Veff.get_Veff_Aeff(directory, oversampling_theta=oversampling_theta)
for veff_data in veff_dataarr:
    cosz = (np.cos(veff_data['thetamin'])+np.cos(veff_data['thetamax'])) / 2.
    veff = veff_data['veff']['dipole_2.0sigma'][0] #['PA_4channel_100Hz'][0]
    v_fine.append(veff)
    cz_fine.append(cosz)

# convert effective volume to effective area
a = np.array(v)/cross_sections.get_interaction_length(1e18)
a_fine = np.array(v_fine)/cross_sections.get_interaction_length(1e18)

# do the plotting...
plt.plot(cz,a/units.km**2, "o")
plt.plot(cz_fine,a_fine/units.km**2, "*")

plt.xlabel("cos(zenith)")
plt.ylabel(r"Aeff [km$^2$]")
plt.savefig("effective_area_oversampling.pdf")
plt.show()
```

### DIFF
--- a/NuRadioMC/utilities/Veff.py
+++ b/NuRadioMC/utilities/Veff.py
@@ -209,17 +209,18 @@ def get_Veff_Aeff_single(filename, trigger_names, trigger_names_dict, trigger_co
     if('thetamax' in fin.attrs):
         thetamax = fin.attrs['thetamax']
 
-    # restrict the theta range, if requested
     theta_width_file = abs(np.cos(thetamin) - np.cos(thetamax))
+    # restrict the theta range, if requested
     if min(bounds_theta) > thetamin:
-        logger.info("restricting thetamin from sim file from {} to {}".format(thetamin, min(bounds_theta)))
+        logger.info("restricting thetamin from {} to {}".format(thetamin, min(bounds_theta)))
         thetamin = min(bounds_theta)
     if max(bounds_theta) < thetamax:
-        logger.info("restricting thetamax from sim file from {} to {}".format(thetamax, max(bounds_theta)))
+        logger.info("restricting thetamax from {} to {}".format(thetamax, max(bounds_theta)))
         thetamax = max(bounds_theta)
-    # assumes isotropic event generation in cos(theta) band
+    # The restriction assumes isotropic event generation in cos(theta) band
     theta_fraction = abs(np.cos(thetamin) - np.cos(thetamax))/theta_width_file
     if theta_fraction<1:
+        # adjust n_events to account for solid angle fraction in the requested theta range
         n_events *= theta_fraction
 
     if('phimin' in fin.attrs):
@@ -273,9 +274,11 @@ def get_Veff_Aeff_single(filename, trigger_names, trigger_names_dict, trigger_co
         logger.warning(f"file {filename} has no triggering events. Using trigger names from a different file: {trigger_names}")
 
     weights = np.array(fin['weights'])
-    # if zenith range is restricted, multiply with zenith mask inside bounds
+    # if theta range is restricted, select events in that range
     if theta_fraction < 1:
+        # generate boolean mask for events in fin inside selected theta range
         mask_theta = np.array((fin['zeniths']>thetamin) & (fin['zeniths']<thetamax))
+        # multiply events with mask, events outside are zero-weighted
         weights *= mask_theta
 
     if(triggered.size == 0):
--- a/NuRadioMC/utilities/Veff.py
+++ b/NuRadioMC/utilities/Veff.py
@@ -140,7 +140,7 @@ def get_Veff_water_equivalent(Veff, density_medium=0.917 * units.g / units.cm **
     return Veff * density_medium / density_water
 
 
-def get_Veff_Aeff_single(filename, trigger_names, trigger_names_dict, trigger_combinations, deposited, station, veff_aeff="veff"):
+def get_Veff_Aeff_single(filename, trigger_names, trigger_names_dict, trigger_combinations, deposited, station, veff_aeff="veff", bounds_theta=[0, np.pi]):
     """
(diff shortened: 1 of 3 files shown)


shallmann:
@lpyras I added you, since you are closest to this (not meant to build pressure to xcheck effective areas @cg-laser)

The proposal v7 is causing the tests to fail still, this is not related to my changes...

NuRadioMC/utilities/Veff.py
```diff
@@ -204,6 +208,20 @@ def get_Veff_Aeff_single(filename, trigger_names, trigger_names_dict, trigger_co
         thetamin = fin.attrs['thetamin']
     if('thetamax' in fin.attrs):
         thetamax = fin.attrs['thetamax']
+
+    # restrict the theta range, if requested
+    theta_width_file = abs(np.cos(thetamin) - np.cos(thetamax))
+    if min(bounds_theta) > thetamin:
+        logger.info("restricting thetamin from sim file from {} to {}".format(thetamin, min(bounds_theta)))
+        thetamin = min(bounds_theta)
```
lpyras, line 216: So here you change the thetamin variable from the value stored in the file, to the value specified in the argument right? Is it okay to use the same variable name here, or should it be something like thetamin_attr before? But maybe its fine and Im just overcritical ;)

NuRadioMC/utilities/Veff.py
```diff
@@ -256,6 +273,11 @@ def get_Veff_Aeff_single(filename, trigger_names, trigger_names_dict, trigger_co
         logger.warning(f"file {filename} has no triggering events. Using trigger names from a different file: {trigger_names}")
 
     weights = np.array(fin['weights'])
+    # if zenith range is restricted, multiply with zenith mask inside bounds
+    if theta_fraction < 1:
+        mask_theta = np.array((fin['zeniths']>thetamin) & (fin['zeniths']<thetamax))
```
lpyras, line 278: Zenith refers here to the key of a file, right? Maybe add a comment, that this is on event level

lpyras (approved):
Thanks for adding this option of smaller binning. I

NuRadioMC/utilities/Veff.py
shallmann: ok, I don't think thetamin_attr makes it more readable, since this is the default value. Added a bit more verbose comments though

NuRadioMC/utilities/Veff.py
shallmann: done!

shallmann:
I accidentially added some WIP stuff from another branch and forced back to the initial commit (@lpyras you therefore klick the approve button again). I added documentation comments concerning Lilly's two review comments. The tests work when forcing to pip install the proposal v6.

I think this is ready to merge now.

lpyras (approved):
Approved :)

cg-laser:
I'll have a look later today, please wait for the merge until then,

cg-laser (commented):
This is a useful feature. 
Did you think about a different interface where the user can specify intervals? Instead of an upsampling factor? 
Maybe the implementation is also good as it is because the current implementation allows to treat simulations done in small zenith bins to be resampled even stronger.

shallmann:
@cg-laser for most purposes oversampling should be the most straightforward option for running on a directory of files.
For the additional interface I see some complications with the current brute force implementation (I tried to avoid large modifications to the `get_Veff_Aeff_single` function):
in `Veff.get_Veff_Aeff(directory, oversampling_theta)`
- the passed boundaries need to be looped through to find according files and run with the adjusted bounds
- you would even need to catch the case when the chosen bin edges don't align with the ones chosen in simulations

Anything more specific is probably better solved by running `get_Veff_Aeff_single` with appropriate theta bounds?

cg-laser (approved):

cg-laser (changes requested):
please also update the changelog.txt

cg-laser:
sry one more thing: this resampling only works if the underlying zenith distribution was uniform on the sky. Can you add this as a comment to the oversampling parameter?

cg-laser (approved):

PR #288, opened 2021-04-30
Fix cpp raysolver

### LINKED ISSUE #286: number of solutions for southpole_2015 model unexpectedly zero
I'm using the analytic ray tracer for the southpole 2015 ice model to find the trajectories for many different starting positions. When scanning over these positions I expect to cross over to the shadow zone, however at some positions within the non-shadow zone I get zero solutions from `get_number_of_solutions()`. Attached you find the minimal example code for a specific position that fails. The image shows the issue on a larger scale. Purple dots indicate `get_number_of_solutions()=2`, while the blanks represent `get_number_of_solutions()!=2`. The red circled region highlight the blanks.

```
from NuRadioMC.SignalProp import propagation
import numpy as np

prop = propagation.get_propagation_module('analytic')
ref_index_model = 'southpole_2015'
ice = medium.get_ice_model(ref_index_model)
attenuation_model = 'SP1'

start = np.array([-90, 0, -10])*units.m
end = np.array([0, 0,-15])*units.m

rays = prop(ice, attenuation_model,
                    n_frequencies_integration=25,
                    n_reflections=0)
rays.set_start_and_end_point(start, end)
rays.find_solutions()
number_of_solutions = rays.get_number_of_solutions()
number_of_raytracing_solutions = rays.get_number_of_solutions()
```

issue

I don't expect the transition between the non-shadow and shadow zone to have any of these sudden blank spots.

I found that though `get_number_of_solutions()=0` at these blank spots `get_number_of_raytracing_solutions()=2`. However, `get_path()` has no entry. What's the difference between the two ?

cg-laser:
this PR fixes #286 where the Cpp implementation does not find a solution. Thanks for @clark2668 for fixing it. 
The solution is to try several initial values. For most cases this should not increase the computing time as the minumum should be found by the first trial.

### DIFF
--- a/NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp
+++ b/NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp
@@ -617,7 +617,7 @@ vector <vector <double> > find_solutions(double x1[2], double x2[2], double n_ic
 	// revealed this case where only checking -1 didn't get us close enough
 	// for the method (which is admittedly a *polishing* algorithm) to find the root.
 
-	for (double x_guess_start = 0; x_guess_start>-3; x_guess_start-=1){
+	for (double x_guess_start = -1; x_guess_start>-3; x_guess_start-=1){
 		if(found_root_1) break;
 		double x_guess = x_guess_start;
 		sfdf = gsl_root_fdfsolver_alloc(Tfdf);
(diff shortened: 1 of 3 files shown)


cg-laser:
we should also merge it as a hotfix into master. I can do that when the PR is approved.

jakob2508 (approved):
I tested the fix for the situation in issue #286. It all works fine now.

clark2668 (approved):
Awesome, thanks for opening the pull request! If you could handle the hotfix also, that'd be great.

PR #289, opened 2021-05-04
add default config to analytic ray tracer.

cg-laser:
for stand alone use cases of the ray tracer, the `get_raytracing_output` failed because it accessed config parameters that weren't set. I suggest that each RT module is initialized with a default config which I implement via this pull request. 

Another thing that we could fix here is that the input argument of `set_solution` is not the same as the return value of `get_results` or `get_raytracing_output` which is really annoying. I think this was overlooked when implementing the RT changes. Do you have an idea to solve it @christophwelling ?


cg-laser:
@boboeyen @christophwelling can you have a look at this?

christophwelling (dismissed):

christophwelling (approved):

PR #290, opened 2021-05-04
update event skeleton to new input event file structure

cg-laser:
the event generator skeleton was not updated yet to the new hdf5 file structure. This is fixed with the PR.

### DIFF
--- a/NuRadioMC/EvtGen/generator_skeleton.py
+++ b/NuRadioMC/EvtGen/generator_skeleton.py
@@ -105,7 +105,7 @@ def generate_my_events(filename, n_events):
     data_sets["interaction_type"] = np.full(n_events, "NC", dtype='S2')
     # The inelasiticiy, i.e. the fraction of the neutrino energy that is transferred into the hadronic shower.
     # This parameter is not used but saved into the output file for information purposes.
-    data_sets["inelasticity"] = np.ones(n_events) * 0.5
+    data_sets["inelasticity"] = np.ones(n_events)
 
     # write events to file
     write_events_to_hdf5(filename, data_sets, attributes)
--- a/NuRadioMC/EvtGen/generator_skeleton.py
+++ b/NuRadioMC/EvtGen/generator_skeleton.py
@@ -28,6 +28,9 @@ def generate_my_events(filename, n_events):
     attributes['start_event_id'] = 0
     # define the fiducial simulation volume. Instead of specifying fiducial_rmin and fiducial_rmin one can also specify
     # fiducial_xmin, fiducial_xmax, fiducial_ymin and fiducial_ymax
+    # the concept of the diducial volume is described in the NuRadioMC paper. In short: only interactions in this smaller
+    # fiducial volume are saved. This is useful for the simulation of secondary interactions. For this dummy example
+    # the fiduial volume is the same as the full volume.
     attributes['fiducial_rmin'] = 0
     attributes['fiducial_rmax'] = 1 * units.km
     attributes['fiducial_zmin'] = 0 * units.m
--- a/NuRadioMC/EvtGen/generator_skeleton.py
+++ b/NuRadioMC/EvtGen/generator_skeleton.py
@@ -15,6 +15,7 @@ logging.basicConfig()
 VERSION_MAJOR = 1
 VERSION_MINOR = 1
 
+
 def generate_my_events(filename, n_events):
     """
     Event generator skeleton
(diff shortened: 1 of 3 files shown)


cg-laser:
I also added that the weights can be specified already in the evt gen

NuRadioMC/simulation/simulation.py
```diff
@@ -432,7 +432,10 @@ def run(self):
             t1 = time.time()
             iE_mother = event_indices[0]
             x_int_mother = np.array([self._fin['xx'][iE_mother], self._fin['yy'][iE_mother], self._fin['zz'][iE_mother]])
-            self._mout['weights'][event_indices] = get_weight(self._fin['zeniths'][iE_mother],
+            if("weights" in self._fin):
+                self._mout['weights'] = self._fin["weights"]
+            else:
```
shallmann, line 437: In the past I have looked at the attrs of a hdf5 file to see which weights are in the file. Is it still guaranteed that the  information written in `self._cfg['weights']['weight_mode']` and `self._cfg['weights']['cross_section_type']` will be correct?

Would be annoying if one needs to reverse engineer and verify what has really been run by re-calculating event weights by hand... ;)

NuRadioMC/EvtGen/generator_skeleton.py
```diff
@@ -30,37 +31,72 @@ def generate_my_events(filename, n_events):
...
 
-    data_sets["interaction_type"] = np.ones(n_events, dtype='S2')
-    data_sets["inelasticity"] = np.ones(n_events) *  0.5
+    # optionally one can also directly set the event weight here (useful if particles other than neutrinos, or calibration
+    # setups are simulated
+    # data_sets["weights"] = np.ones(n_events)
 
+    # the interaction type. For neutrino interactions is can be either CC or NC. This parameter is not used but passed
+    # to the output file for information purposes.
+    data_sets["interaction_type"] = np.ones(n_events, dtype='S2')
```
shallmann, line 96: generally for the skeleton, np.ones/np.zeros are perfectly fine. For the 2 letter strings, maybe it is more intuitive to fill it with one of the actual values?

i.e. `np.ones(n_events, dtype='S2')` -> `np.full(n_events, 'CC', dtype='S2')`

shallmann (changes requested):
hesitant with approval concerning my comment on the attributes, want to hear your opinion on this first.

NuRadioMC/EvtGen/generator_skeleton.py
cg-laser: good point, will change that.

shallmann: @cg-laser another comment:

I guess fixing this resolves the future unimplemented warning when generating a terminal dump of the file:

/Users/shallmann/miniconda/lib/python3.8/site-packages/tables/attributeset.py:308: DataTypeWarning: Unsupported type for attribute 'deposited' in node '/'. Offending HDF5 class: 8
  value = self._g_getattr(self._v_node, name)
/ (RootGroup) ''
/Users/shallmann/miniconda/lib/python3.8/site-packages/tables/group.py:1201: UserWarning: problems loading leaf ``/interaction_type``::

  variable length strings are not supported yet

The leaf will become an ``UnImplemented`` node.
  warnings.warn(
/Users/shallmann/miniconda/lib/python3.8/site-packages/tables/group.py:1201: UserWarning: problems loading leaf ``/shower_type``::

  variable length strings are not supported yet

The leaf will become an ``UnImplemented`` node.

shallmann: but do we actually only ever store CC or NC there? If I remember correctly there was also a "tau_had", "tau_em" at some place stored? Or was that in another field?

cg-laser: at the moment we don't use it, but we might in the future. The thing is, this field is not used at all in the simulation but you still want to know it for post analyses. In the secondary shower generation, one could use this field to save what type of interaction happend (and I think this is already being done). What matters for the simulation is the "shower_type" which is so far only "em" or "had". But in principle there could be more channels, e.g. a tau decay is slightly differnt.

cg-laser:
@christophwelling does this look good?

NuRadioMC/EvtGen/generator_skeleton.py
```diff
@@ -30,37 +24,91 @@ def generate_my_events(filename, n_events):
...
+    # the direction of the shower
     data_sets["azimuths"] = np.ones(n_events)
     data_sets["zeniths"] = np.ones(n_events)
+    # the position of the shower
     data_sets["xx"] = np.ones(n_events)
     data_sets["yy"] = np.ones(n_events)
     data_sets["zz"] = np.ones(n_events)
-    data_sets["event_ids"] = np.arange(n_events)
+    # the shower energy
+    data_sets["shower_energies"] = np.ones(n_events) * 1 * units.EeV
```
christophwelling, line 73: Why are showers initialized with 1EeV?

christophwelling (dismissed):
Looks alright, but I don't understand why some of the values are specified with specific values. I guess it's just as an example? Then a docstring explaining that would be good.

NuRadioMC/EvtGen/generator_skeleton.py
shallmann: this is a valid point. We probably shouldn't suggest to set shower energies manually at all? Is it less error prone, to provide external functions for generating an energy spectrum from the attributes?... like a powerlaw between Emin, Emax (ok we don't store an index (yet)). (don't know if this unnecessarily overcomplicate things for us, the optical telescopes typically use powerlaws between emin/max in the generator)

NuRadioMC/EvtGen/generator_skeleton.py
shallmann: ok, so at least for the shower type we should make sure the string length is long enough, like `dtype='S7'`?

NuRadioMC/EvtGen/generator_skeleton.py
cg-laser: this skeleton is intended for non standard use cases. So manually setting shower energies is something that people need to know how to do

NuRadioMC/simulation/simulation.py
shallmann: looks better to me

shallmann (approved):
The change to use non-standard weights looks good to me now. I have no further objections.

Acutally I have one small comment: Although it is meant as a custom dummy script and the user is responsible for filling the simulations with useful data, I wouldn't put neutrino energy AND shower energy to 1 EeV and set the Bjorken y to 0.5 which is obviously wrong!

cg-laser:
we use this custom event generator for example to generate simulations for air-shower cores hitting the surface. There we need the option to directly set the shower energies. There are quite some non-neutrino use cases where inelasticity and total energy are not used.

shallmann (approved):
fine with me now

PR #293, opened 2021-05-12
Uproot readout

shallmann:
Using uproot to read ARIANNA data
`pip install uproot awkward` is more convenient than installing pyroot and making this work with existing ARIANNA software.

Some branches are ill-defined in the root file, which makes the readout a bit unconvenient (typically custom definition of the uproot interpretation of branches should not be necessary, but they are for times, trigger mask and the commissioning windows), but I think we have to (and can?) live with it.

Files generated with uproot and the traditional ROOT ARIANNA reader were checked for stn 32, run 126 to be identical.
Except: One parameter (l1 suppression value) is missing, due to a (likely) uproot bug, which makes it tough to guess the content of the trigger config branch.

If we can provide a single-event (or small with only the ConfigTree) root file it is likely that they can fix this bug in their software. Cf. my discussion with the uproot developer here: https://github.com/scikit-hep/uproot4/discussions/354 for details

I also added some additional ARIANNA parameters in the parameters.py which are read out.

Q:
* should uproot, awkward be added to the toml as required packages?
* in /noise/ there is a noiseImporterUproot.py script which is an uproot version of the .nur noise importer, this file will likely change in future (to suit best the needs of Simon for the classifier. I committed the current version, but should we for now remove it from the pull request or leave it in there?


NuRadioReco/modules/io/noise/noiseImporterUproot.py
```diff
@@ -0,0 +1,195 @@
...
+        self.dtms = np.concatenate(dtms)
+        self.run_number = np.concatenate(run)
+        self.nevts = len(self.data)
+
+    def _read_temperature_curve(self, nf, interpolation_times=None):
+        """
+        Imports the temperature vs. posix time from a noise file (nf)
+
+        The temperature is not taken per event, so the posix_time and temperature value is returned
+        if interpolation_times is specified, interpolate to these posix time stamps
```
anelles, line 124: This doc string doesn't quite make it clear what format the interpolation_times should have.

NuRadioReco/modules/io/noise/noiseImporterUproot.py
```diff
@@ -0,0 +1,195 @@
...
+    For different stations, new noise files need to be used.
+    Collect forced triggers from any type of station to use for analysis.
+    A seizable fraction of data is recommended for accuracy.
+
+    The noise will be random.
+    This module therefore might produce non-reproducible results on a single event basis,
+    if run several times.
+    """
+
+    def begin(self, noise_files):
```
anelles, line 62: Please document in what format "noise_files" is expected.

NuRadioReco/modules/io/noise/noiseImporterUproot.py
```diff
@@ -0,0 +1,195 @@
...
+        "l1":       2**3,
+        # not_used: 2**4,
+        "l1_scaledown":     2**5,
+        "l1_enabled" :      2**6,
+        "exceeding_buffer": 2**7
+        }
+
+class noiseImporter:
+    """
+    Imports recorded noise from ARIANNA station.
```
anelles, line 49: Should we rename this to ARIANNA noise importer or something? Otherwise, it seems to be generic and I don't assume it works with RNO-G data for example.

NuRadioReco/modules/io/snowshovel/AriUtils_dicts.py
```diff
@@ -0,0 +1,90 @@
+from NuRadioReco.modules.base.module import setup_logger
+logger = setup_logger("AriUtils_dicts")
+
+# script containing dictionaries defined in snowshovel software
+# copied from snowshovel/scripts/online/AriUtils.py
```
anelles, line 5: Can you add a link to the snowshovel software (even if it is hidden behind a password)

NuRadioReco/modules/io/snowshovel/dacs2014.py
```diff
@@ -0,0 +1,268 @@
+def enum(**enums):
```
anelles, line 1: Is this also copied from somewhere?

NuRadioReco/modules/io/snowshovel/AriUtils_dicts.py
anelles: I actually think it is ok to have it all public, but full disclosure that we didn't come up with this alone is probably good.

anelles (changes requested):
A couple of nit-picky points.

shallmann:
@anelles @cg-laser Finally picking this up. I think the only issue was referencing the origin of the DAQ components I put in `NuRadioReco/modules/io/snowshovel/AriUtils_dicts.py` and `NuRadioReco/modules/io/snowshovel/dacs2014.py`

I took them from the disabled ARIANNA svn server: 
```
Working Copy Root Path: /Users/shallmann/software/snowshovel
URL: https://arianna.ps.uci.edu/svn/repos/snowShovel/trunk
Relative URL: ^/snowShovel/trunk
Repository Root: https://arianna.ps.uci.edu/svn/repos
Repository UUID: e5861a0a-4bf6-4d27-ae03-149b890d7f82
Revision: 2673
Node Kind: directory
Schedule: normal
Last Changed Author: persic
Last Changed Rev: 2673
Last Changed Date: 2019-07-30 20:38:36 +0200 (Di, 30 Jul 2019)
```
Is there a better way to get this information? Is the svn still available somewhere else to refer from?

anelles:
@shallmann I think it is fine without reference. Especially now that the server is down (and won't be revived any time soon). A fraction of the code was written by me and the other half by people who already left science. AFAIK, this software is not being used anymore.
Just reference to the ARIANNA collaboration, you can name Corey Reed and Chris Persichilli who have done lots of the coding you are using.

NuRadioReco/modules/io/noise/noiseImporterUproot.py
cg-laser: @shallmann I think this one is still open, or more general to write doc strings for the functions.

cg-laser:
Thanks Steffen. I had a look at the code and it seems that doc strings for the relevant functions with which the user interfaces are still missing, i.e. just [noiseImporterUproot.py](https://github.com/nu-radio/NuRadioMC/pull/293/files#diff-613f3c59633169336c0601445148ccd0a60516da5434d258aac09645c5f4e8cb)

anelles:
@shallmann Is this ever used or now too outdated to look at this? It has been two years, it may be time to close this for good.

PR #294, opened 2021-05-14
Mc info in nur

shallmann:
NuRadioReco events contain additional info via a Particle class (arbitrary list of particles can be added) and generator_info:

simulation.py is modified accordingly to
* copy generated hdf5 files's attributes to output structure
* use Particle class in mc read in already and write to final particle

Note: In future, we might 
* extend functionality of simulation.py to read in additional mc particles in addition to primary (e.g. the tau in the nutau interaction and subsequent decay in proposal). In an Event, one can already now ask for a particle's daughter showers/particles and the parent / primary
* allow for re-simulation? or just extraction of the "old" hdf5 format from .nur?

Suggestions if this draft is acceptable or needs changes?


NuRadioMC/simulation/simulation.py
```diff
@@ -521,10 +537,11 @@ def run(self):
...
                                     100 * (total_time - total_time_sum) / total_time))
 
-                    # read all quantities from hdf5 file and store them in local variables
-                    self._read_input_neutrino_properties()
+
                     logger.debug(f"simulating shower {self._shower_index}: {self._shower_type} with E = {self._shower_energy/units.eV:.2g}eV")
-                    x1 = np.array([self._x, self._y, self._z])  # the interaction point
+                    #  should be of the shower, therefore we need to update the _read_neutrino_properties to the current shower
+                    self._read_input_particle_properties()
+                    x1 = self.input_particle[simp.vertex]  # the interaction point
```
cg-laser, line 544: it is not clear to me where input_particle points to? Is this now the shower? Then particle is maybe a bad name?

NuRadioMC/simulation/simulation.py
```diff
@@ -813,6 +830,15 @@ def run(self):
                         tmp = tmp[:-2] + " ns"
                         logger.info(f"creating event {iEvent} of event group {self._event_group_id} ranging rom {iStart} to {iStop} with indices {indices} corresponding to signal times of {tmp}")
                     self._evt = NuRadioReco.framework.event.Event(self._event_group_id, iEvent)  # create new event
+
+                    # add MC particles that belong to this (sub) event to event structure
+                    # add only primary and parent?
+                    # or all particles? usually not so many in our case...
+                    self._evt.add_particle(self.primary)
```
cg-laser, line 837: yes we want the primary particle, and all parents I think. but for everything we do right now, there is only one or at most two parents. For muon or tau emission it would be "neutrino" -> "tau" -> "shower". For the latter case, how would we get this information? 
Also, don't you need to also add the current particle (which is then actually a shower)?

NuRadioMC/simulation/simulation.py
```diff
@@ -1044,6 +1070,13 @@ def _read_input_hdf5(self):
                     self._fin[key] = np.array(value)
         for key, value in iteritems(fin.attrs):
             self._fin_attrs[key] = value
+
+        # copy information over to generator attributes
```
cg-laser, line 1074: it took me a while to understand what you do here. Can you improve the comment? Something like "store all relevant attributes of the input file in a dictionary"

NuRadioMC/simulation/simulation.py
```diff
@@ -1206,23 +1239,34 @@ def _create_station_output_structure(self, n_showers, n_antennas):
...
+        self.input_particle[simp.energy] = self._fin['energies'][self._shower_index]
+        self.input_particle[simp.interaction_type] = self._fin['interaction_type'][self._shower_index]
+        self.input_particle[simp.inelasticity] = self._fin['inelasticity'][self._shower_index]
+        self.input_particle[simp.vertex] = np.array([self._fin['xx'][self._shower_index],
+                                                  self._fin['yy'][self._shower_index],
+                                                  self._fin['zz'][self._shower_index]])
+        self.input_particle[simp.zenith] = self._fin['zeniths'][self._shower_index]
+        self.input_particle[simp.azimuth] = self._fin['azimuths'][self._shower_index]
+        self.input_particle[simp.inelasticity] = self._fin['inelasticity'][self._shower_index]
+        self.input_particle[simp.parent_id] = None # primary does not have a parent
```
cg-laser, line 1257: why is it always a primary? Don't you need to consult `n_interations` to see if it is the primary?

NuRadioReco/framework/event.py
```diff
@@ -21,6 +22,8 @@ def __init__(self, run_number, event_id):
         self.__radio_showers = collections.OrderedDict()
         self.__sim_showers = collections.OrderedDict()
         self.__event_time = 0
+        self.__particles = collections.OrderedDict()
+        self._generator_info = {}
```
cg-laser, line 26: add a comment that describes what this dict contains.

NuRadioReco/framework/event.py
```diff
@@ -120,6 +143,85 @@ def get_station_ids(self):
...
+    def add_particle(self, particle):
+        """
+        Adds a MC particle to the event
+
+        Parameters
+        ------------------------
+        particle: Particle object
+            The MC particle to be added to the event
+        """
+        if particle.get_id() in self.__particles:
```
cg-laser, line 155: maybe also check if `particle` is of type `Particle`?

NuRadioReco/framework/event.py
```diff
@@ -120,6 +143,85 @@ def get_station_ids(self):
...
+        Returns true if at least one MC particle is stored in the event
+
+        If particle_id is given, it checks if this particular MC particle exists
+        """
+        if particle_id is None:
+            return len(self.__particles) > 0
+        
+        return particle_id in self.__particles.keys()
+
+    def get_interaction_products(self, parent_particle, showers = True, particles = True):
```
cg-laser, line 210: add doc string

cg-laser (commented):
Hey Steffen, thanks for starting this (and also fixing a lot of typos ;-))
I added a couple of comments in-line. Some things I also don't understand. Let's discuss in the next MC dev call.

NuRadioMC/simulation/simulation.py
shallmann: right, naming is still bad, will change that, just thought the `_read_neutrino_properties` before was even worse

NuRadioMC/simulation/simulation.py
shallmann: of course

NuRadioMC/simulation/simulation.py
shallmann: yes you are right. should test with a proposal file to see it is handled correctly, the n_interactions is a good suggestion

NuRadioReco/framework/event.py
shallmann: ok

NuRadioMC/simulation/simulation.py
shallmann: as far as i can see this point maybe needs more discussion...
* Do we want to store all showers as a particle also, in principle, or does this just duplicate info in the sim showers?
  Note: I already added a parent_id to the showers, so that one can look at the sim showers and look for their parents, via the 
   event.get_parents(particle_or_shower) or via get_interaction_products(self, parent_particle, showers = True, particles = True) return the daughter particles/showers/both. Maybe you have a suggestion for a more clever way?
* In principle one should store all the particles, but I would maybe argue that one should only ever store particles passed by the generator? I.e. I would not generate outgoing leptons with (1-y) within simulation.py

* I propose: For every event group id, read neutrino properties from entry 0, then In the input shower/particle loop in simulation.py check, if the line corresponds to a shower or particle
* if particle, add particle to the list and continue to next item; if shower use the current looping over stations in simulation.py

shallmann: One more thing I add here so that I remember: maybe in the particle instead of `flavor` it would be better to store the PDG code...Makes more sense for e/mu/tau leptons, and possibly a custom code for showers could be defined.

anelles (approved):
Looks good to me. Then again I don't use it all the time, so not sure whether it results in any issues in the work-flow. But I assume that the author is the best person to judge that ;)

cg-laser (approved):
looks good. Only documentation is missing. @shallmann can you add a brief description of the new particle class to the documentation (in a separate PR or still into the unified-doc PR)?

fschlueter:
hey @shallmann, I stumbled across the `particleParameters` dedicated for cosmic rays. Why would we need them? They seem redundant and `cr_energy_em` is rather a property of the shower than the particle.

PR #296, opened 2021-05-17
ARZ shower realization

cg-laser:
if the first RT solution is too far off the Cherenkov cone, the ARZ model will return an empty trace and also set the shower realization to None. However, NuRadioMC will reuse the shower realization of the first RT solution. The module will select then a new random realization, but in the hdf5 file (and also nur file) the shower realization of all RT solution is set to None. This PR fixes that by always generating a random shower realization.


shallmann:
This change looks good to me.

shallmann (approved):
This looks good to me.

shallmann (approved):

PR #299, opened 2021-05-28
Hpol converter

iplaisier:
correcting for small bug int he Hpol converter, and adding encoding option


cg-laser (approved):
do I see it correctly that the previous version just produced NaNs?

PR #300, opened 2021-06-01
Feature/noise generator pa

cg-laser:
this PR implements a thermal noise generator that fulfills a phased array trigger. It works but it way too slow to be usable. The zip archive contains test code. Maybe someone else has a smart idea how to speed up the calculation

[testcode.zip](https://github.com/nu-radio/NuRadioMC/files/6575340/testcode.zip)


shallmann:
I dared to separate out the generation/phasing/triggering part in the noise simulation to see where the bottleneck is.
In the existing functions I just added timing/debug variables and made the traces and phased_traces member variables.

Timing before optimisation:
20000 events, 1279.58, threshold = 3085.00
**GENERATION: 10.4237, PHASING: 13.5671, TRIGGER: 7.6844**

Timing after optimisation:
20000 events, 1262.50, threshold = 3085.00
**GENERATION: 10.4024, PHASING: 3.6494, TRIGGER: 1.2973**

i.e. with only small modifications I get a **speedup of ~ x4 in the phasing part and ~ x6 in the triggering part**
Modifications are:
- when summing up phased traces, slicing the arrays seems much faster than rolling and adding
- in the triggering I am rebinning the data into smaller subsets of length step_length and summing subsequent steps up to window length (at least this is after short inspection what I think the other function does? Does this match reality?

shallmann:
Small addendum for completeness:
I tried also rolling via np.take, but the performance is always worse than slicing and scales worse with trace lengt
image
h

cg-laser:
awesome! This might already be fast enough for generating triggered noise with sufficient statistics.

shallmann:
Ok, if not sufficient yet, the generation part can probably be easily be made faster by separating out everything that involves unnecessary re-calculation in each single trace.
I.e. set/calculate the known quantities once and only redo the random generation

shallmann:
just finished to make the trigger behaviour match previous implementation at the edges.

removing parameter re-calculation in each trace helps already a wee bit...
20000, 1435.00, threshold = 3085.00
GENERATION: 7.2010, PHASING: 3.5399, TRIGGER: 1.4277

ZackfromEarth:
@cg-laser @shallmann I fixed the sampling rate bug where the test was failing. Should we merge this before the release? I'd like to use it for my big noise simulation in the coming days, let me know if I should change anything else / attempt to speed things up more.

NuRadioReco/utilities/noise.py
```diff
@@ -100,3 +203,391 @@ def generate_noise(self):
...
+            dt_phasing += time.process_time() - tstart
+
+            tstart = time.process_time()
+            is_triggered = self.__triggering()
+            dt_triggering += time.process_time() - tstart
+            if is_triggered:
+                return self._traces, self._phased_traces
+
+
+    def generate_noise2(self, debug=False):
```
cg-laser, line 529: why is this function in here? Maybe a left over from the other noise class? If so, then just remove this function.

cg-laser (commented):
agree, merging before the release might be nice. I don't see any problems. 

Why are there changed to readme.md? There shouldn't be any.

cg-laser:
I fixed it by merging in develop

ZackfromEarth (approved):

shallmann:
@ZackfromEarth @cg-laser IMO the merging was a bit quick... I would have cleaned up the `generate_noise` functions where (I think Christian?) put there two alternative implementations for testing the `generate_noise` and `generate_noise2`...
I just added the `generate_noise_optimised` demonstrating how by changing some functions the code can be made faster. I think it is maybe too much to keep all three functions without adding any comments on speed/preferred usage/etc.

cg-laser:
@shallmann thanks for raising it. This was totally on me.

PR #303, opened 2021-06-17
allow to pass ice object directly to simulation class

cg-laser:
if different ice models are checked, it is useful to be able to just pass a custom ice model to the simulation method.


boboeyen:
This seems all fine to me

boboeyen (approved):

PR #306, opened 2021-07-12
Fix/firn model doc

boboeyen:
fixing some issues:
some ice model documentation, the bulk ice value of the new firn model.
also increased accuracy in radiopropa tracer as it was necessary 
and handling of not initialised shower axis in radiopropa tracer


shallmann (approved):
Nothing to object. Using the same refractive index rounding makes totally sense. Well written docstrings! ;)

boboeyen:
Some more info:

I changed the bulk refractve index from 1.775 to 1.78 for the firn model to make it coherent with the rounding of the value as the simple model. They both should have the same index for bulk ice since only the top layer was adapted, but I was not aware of that. Otherwise you get strange effects in timing occur just by rounding the value differently.

I removed the disconituity layer of the firn model since I mixed up non-differentiable by non-continuous. There is no discontinuity.

PR #307, opened 2021-07-15
add publications to Readme

cg-laser:
I thought it would be a good idea to better document new features by their corresponding publications and adapted the readme file accordingly.


anelles (approved):
I think this is roughly complete. We only have to remember to keep updating (hopefully) :)

PR #308, opened 2021-07-21
Add getter to BaseShower class which returns the showers axis. Reduce…

fschlueter:
…s boiler code the use has to write

Hi guys,
I just want to check if you like such code. Not sure if it follows the design/idea of the framework? What do you think? Other getters which would would make life easier is a getter for a radiotools.coordinatesystems.cstrafo...
One could also think of adding this to a inherited class..

Cheers


cg-laser (changes requested):
this is totally fine. Such helper functions are really useful. But please add a docstring where I would also mention that this axis is the opposite to the propagation direction of the shower to avoid any confusion.

PR #309, opened 2021-07-21
Fix sim_station timing. Add beginning of sim efield trace as trace_st…

fschlueter:
…art_time.

This is needed to conserve the correct signal arrival times of the sim. signals, right?


cg-laser (approved):
Yes, I think this information was so far missing.

fschlueter:
Hi @cg-laser,
I guess we have to fix the check first before progressing with this. I had a look why this is failing (apparently the code can not handle negative traces start times) but so far could not figure out how to solve this. I am also wondering if the relative arrival times of the pulses are conserved across different stations? This is something which you have not cared about so far focusing on signal station events correct? 
I can have a look into this (after my holidays) but I am afraid that I am not very familiar with the module. Maybe you or another expert can have a look?
Cheers

fschlueter:
Because, as much as I understand the code, the time traces is padded. Therefore the original trace is "fill" into a larger trace containing only zeros. Where to fill depends on the trace start time / bin size. As this start time can now be < 0 the padding does not work because of a shape violation. I am not quite sure, At this point I guess it is more efficient if an expert has a look.
Cheers

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -126,13 +126,14 @@ def run(self, evt, station, det):
                     times_min.append(t0)
                     times_max.append(t0 + electric_field.get_number_of_samples() / electric_field.get_sampling_rate())
                     self.logger.debug("trace start time {}, cab_delty {}, tracelength {}".format(electric_field.get_trace_start_time(), cab_delay, electric_field.get_number_of_samples() / electric_field.get_sampling_rate()))
-        times_min = np.array(times_min)
-        times_max = np.array(times_max)
-        if times_min.min() < 0:
-            times_min -= times_min.min()
-            times_max -= times_min.min()
```
shallmann, line 133: I think this causes the problem, since after shifting times_min in l 132, the times_max is not shifted any more with this implementation. Also the padding is not included here, so that should come before as well

cg-laser (commented):
I think it looks good. The only thing that I'm worried about is that the `start_bin` might become negative but you added a check for it.

shallmann:
@cg-laser on the contrary. The shifting did produce negative start bins in the previous implementation. This should not occur any more with the changes, but the check is anyway there for safety

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -187,10 +185,25 @@ def run(self, evt, station, det):
...
+                    if start_bin < 0:
+                        # ensure new efield does not extend beyond start of trace although this should not happen
+                        self.logger.warning("electric field trace extends beyond the beginning of the trace and will be cut.")
+                        tr = np.atleast_2d(tr)[:,-start_bin:]
+                        start_bin = 0
+                    new_trace[:, start_bin:stop_bin] = tr
                 trace_object = NuRadioReco.framework.base_trace.BaseTrace()
                 trace_object.set_trace(new_trace, 1. / time_resolution)
-                trace_object.set_trace_start_time(np.min(times_min) - cab_delay)
+                trace_object.set_trace_start_time(times_min.min() - cab_delay)
```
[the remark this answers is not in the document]
shallmann: @cg-laser (private comm): "should be correct. cable delay is to be considered in forward simulation, since digitizer sees the pulse indeed later than it actually arrives."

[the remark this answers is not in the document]
cg-laser: true, this line is not needed anywhere

cg-laser (approved):

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -187,10 +185,24 @@ def run(self, evt, station, det):
...
+                        tr = np.atleast_2d(tr)[:,:stop_bin-start_bin]
+                    if start_bin < 0:
+                        # ensure new efield does not extend beyond start of trace although this should not happen
+                        self.logger.warning("electric field trace extends beyond the beginning of the trace and will be cut.")
+                        tr = np.atleast_2d(tr)[:,-start_bin:]
+                        start_bin = 0
+                    new_trace[:, start_bin:stop_bin] = tr
                 trace_object = NuRadioReco.framework.base_trace.BaseTrace()
                 trace_object.set_trace(new_trace, 1. / time_resolution)
-                trace_object.set_trace_start_time(np.min(times_min) - cab_delay)
```
christophwelling, line 193: Are you sure you can do this? The trace start time needs to be set somewhere I think.

christophwelling: Okay, the trace start time seems to only be used for debug plots. So I think this should be fine. Did you check if this messes up the debug plots?

christophwelling (approved):
I think this should be fine. The only thing I am not sure about is not setting the trace start time any more, but this should only affect the debug plots.

NuRadioReco/modules/efieldToVoltageConverter.py
shallmann: Yes, few lines below it is set for the channel (and everything that was put here is irrelevant). Therefore this is never used.

PR #313, opened 2021-09-06
Cluster runner

cg-laser:
This PR adds a new utility class to run NuRadioMC on a cluster using a full node submission, i.e., a full node with X cores is reserved.     
As NuRadioMC is a single-core process, this class will start multiple NuRadioMC simulations and will distribute them on the X cores. As soon as one job is finished, a new one is started so that all X cores are kept busy. 
The job ends after a fixed time, or after enough triggers are acquired, or after X crashes occurred.


shallmann (approved):
I did not find anything suspicious. Now it is probably our task to find out how to request full nodes in the job submission at our computing center...

PR #315, opened 2021-09-10
speed up the remove_duplicate_events function 

cg-laser:
to not double count events with multiple showers, duplicate events need to be removed. The current implementation loops over all group event ids and selects only one trigger. The loop slows down the computation a lot. 
Then new implementation speeds things up for the cases of just one or two showers (nu_e CC).

### DIFF
--- a/NuRadioMC/utilities/Veff.py
+++ b/NuRadioMC/utilities/Veff.py
@@ -24,7 +24,7 @@ def remove_duplicate_triggers(triggered, gids):
 
     The hdf5 file contains a line per shower. One event can contain many showers, i.e. if we count all triggeres
     from all showers we overestimate the effective volume. This function modifies the triggered array such
-    that it contains not more than one True value for each event group. 
+    that it contains not more than one True value for each event group.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


shallmann:
I propose instead to do something like this

```
import pandas
def remove_triggers_pandas(trig, gids):
    # replace gid==0 with magic number (just inside this function)
    gids_shifted = np.copy(gids)
    gids_shifted[gids==0] = -123456
    
    # triggered gids, 0 if not triggered
    trig_gids = trig*gids_shifted
    #print(np.count_nonzero(trig_gids))
    
    series = pd.Series(trig_gids)
    is_duplicate = series.duplicated().values
    result = trig_gids*~is_duplicate
    #print(np.count_nonzero(result))
    return result != 0
```

shallmann:
Screenshot 2021-09-10 at 14 30 54
... attached a speed comparison of the two methods

cg-laser:
thanks Steffen. I think this looks pretty good. You could add an additional check if the id `-123456` is not present in the data array. This makes my implementation for the special case probably obsolete? 

The only downside is that we would add pandas as a dependency.

shallmann:
yes the special case should be obsolete then.

This should be also nearly as fast and without pandas:

```
def remove_triggers_pandas(trig, gids):
    # replace gid==0 with magic number (just inside this function)
    #gids_shifted = np.copy(gids)
    gids_shifted = gids + 0.5
    #gids_shifted[gids==0] = -123456
    
    # triggered gids, 0 if not triggered
    trig_gids = trig*gids_shifted
    #print(np.count_nonzero(trig_gids))
    
    unique_values, unique_indices = np.unique(trig_gids, return_index=True)
    is_not_duplicate = np.zeros_like(trig, dtype=bool)
    np.put(is_not_duplicate, unique_indices, True)
    result = trig_gids*is_not_duplicate
    #print(np.count_nonzero(result))
    return result != 0
```

shallmann:
Note above: Since gids are integers, shifting everything by 0.5 does not require checking individual values
image

cg-laser:
great! can you push the changes?

shallmann:
yes

shallmann (approved):
Maybe now you should have a look if you approve to these changes before merging @cg-laser

cg-laser:
great thanks a lot. A very elegant solution!

PR #316, opened 2021-09-16
adapt example to new interface

cg-laser:
and cosmetic...


boboeyen (approved):
seems fine to me

clark2668 (approved):
Also looks good to me, thanks for fixing it!

PR #317, opened 2021-09-23
add option to simulate emitters

cg-laser:
This pull requests will add the option to simulate emitter with arbitrary polarization to NuRadioMC. This is useful to simulate calibration measurments. 
Ilya summarized the problem with the current code very well and I quote him below:

> Hi Christian,
> While discussing implementation details between me and Pawan, we came to realization of a fundamental difficulty of implementing a radio emitter as simply another plug-and-play option alongside with various askaryan shower models in NuRadioMC.
> 
>   Let me run by you my understanding of how simulation happens:
> 
> =========
> 
> 1. The electric field emitted by a source is assumed to be radial. All models in askaryan.py return a 1D E-field vector sampled in time, which is taken to be the “theta” component at this point. This is seen in here, in the get_time_trace function.
> 
> https://github.com/nu-radio/NuRadioMC/blob/857991c8e06db00196e222606e99654cc76b2f90/NuRadioMC/SignalGen/askaryan.py#L14-L17
> 
> In Pawan’s code, we have get_time_trace implemented for a radio emitter of chosen design that is fed a chosen V_src(time) pattern. We produce E(time) trace, where E is the theta polarization component (p-mode) while the phi polarization component is (s-mode) is ignored for the time being.
> 
> 2. What is actually used for event simulation is the E-field as a function of frequency, the get_frequency_spectrum of askaryan.py:
> 
> https://github.com/nu-radio/NuRadioMC/blob/857991c8e06db00196e222606e99654cc76b2f90/NuRadioMC/SignalGen/askaryan.py#L122
> 
> which simply calls the get_time_trace and makes an fft spectrum out of that.
> 
> 3. The get_time_trace is used from the simulation.py file:
> 
> https://github.com/nu-radio/NuRadioMC/blob/857991c8e06db00196e222606e99654cc76b2f90/NuRadioMC/simulation/simulation.py#L698
> 
> At this point, the electric field is 1D array E(f) representing nominally “theta” polarization. 
> 
> 4. Then, the electric field vector, originally nominally “theta”, gets rotated appropriate to the geometry of the given event in this line:
> 
> https://github.com/nu-radio/NuRadioMC/blob/857991c8e06db00196e222606e99654cc76b2f90/NuRadioMC/simulation/simulation.py#L726
> 
> which, as far as I understand, rotates the original “theta” vector appropriately so that the E-field vector is in the plane that goes through the neutrino line-of-flight, the receiver antenna location, and is directed away from the neutrino line-of-flight in that plane (assuming I am reading the code correctly). Thus, the E-field vector now has both theta and phi transverse polarizations that depend on the neutrino line-of-flight and receiver position.
> 
> ==========
> 
> This implementation, assuming I am right, is presenting us with difficulties. What we would ideally do is we would generate both theta and phi components of the electric field for a generic radio emitter, and we would disable the rotation of the polarization vector for such an emitter that assumes radial polarization direction. As far as I can see, it would be non-trivial, with get_time_trace and get_frequency_spectrum returning 2D array covering both transverse polarizations (with the second polarization being zero for neutrino showers), and then simulation.py handing that appropriately, rotating or not rotating the E-field vector depending on whether the source is a shower or a generic emitter. What solutions do you see given your knowledge?
> 
> On the shorter term, it is still possible to work with VPol emitters, it seems to me, setting the dummy “neutrino” direction to point anywhere as long as it lies in a vertical plane that goes through the receiver and the emitter locations. For this to work for a station with multiple receiver locations, one would need to point the dummy neutrino direction straight up or straight down. This would assume that pick-up on the other polarization is negligible.
> 
> However, an HPol transmitter, seems to me, would be not possible to simulate this way, both because with the difficulty to define the “dummy” neutrino direction and the fact that HPol (I think) emits both non-negligible theta and phi components to start with.
> 
> Please let me know if we are misunderstanding something or if you think non-trivial changes to askaryan.py and simulation.py that would accumulate handling two transverse polarizations without assuming radial direction with respect to a shower axis could be possible.
> 
> Thanks,
> Ilya

I think the best option is to add another config parameter that specifies if the signal source is a shower or an emitter (with the potential to add more classes in the future). The main difference is that the emitter model directly returns a 3-dim electric field. I created a new file called "emitter.py" which is similar to "askaryan.py" and has the same interface but returns a 3d vector.  This file can then store a variety of different emitters. Please have a look at the code and tell me what you think. 

I have a few questions myself
* should we better integrate emitter.py into askrayan.py? This would help to keep the two interfaces in sync but it would be messier. Another kwarg would then specify if a 1d or 3d trace is returned. 
* should we make the emitters themselves modular? I.e. emitter.py would only contain the voltage output of the pulser. Then we specify the emitting antenna and orientation and NuRadioMC automatically calculates the resulting efield. This can be achieved by a new NuRadioReco module that convolves the voltage with the antenna response. I think this would be the nicest solution. The exact signal chain (potentially there is another bandpass filter or amp in the setup), can be specified equivalent to the `def _detector_simulation(self):` method, lets call it `def _emitter_simulation(self)`. The only thing I'm not sure about is where to specify the emitter antenna configuration. At first maybe just manually in `_emitter_simulation`?

### DIFF
--- a/NuRadioMC/SignalGen/emitter.py
+++ b/NuRadioMC/SignalGen/emitter.py
@@ -11,7 +11,7 @@ def set_log_level(level):
     par.set_log_level(level)
 
 
-def get_time_trace(energy, theta, N, dt, shower_type, n_index, R, model, full_output=False, **kwargs):
+def get_time_trace(amplitude, N, dt, model, full_output=False, **kwargs):
     """
     returns the electric field of an emitter
 
@@ -53,16 +53,13 @@ def get_time_trace(energy, theta, N, dt, shower_type, n_index, R, model, full_ou
         only available if `full_output` enabled
 
     """
-    shower_type = shower_type.upper()  # some convenience to be sloppy with upper/lower case
     trace = None
     additional_output = {}
-    if(energy == 0):
+    if(amplitude == 0):
         trace = np.zeros(3, N)
     if(model == 'spherical'):
-        amplitude = 1. * energy / R
-        trace = np.zeros(3, N)
-        trace[1, N // 2] = amplitude / 0.5 ** 2  # equal amplitude in eTheta (p-polarization) and ePhi (s-polarization)
-        trace[2, N // 2] = amplitude / 0.5 ** 2
+        trace = np.zeros(N)
+        trace[N // 2] = amplitude
     else:
         raise NotImplementedError("model {} unknown".format(model))
     if(full_output):
@@ -71,7 +68,7 @@ def get_time_trace(energy, theta, N, dt, shower_type, n_index, R, model, full_ou
         return trace
 
 
-def get_frequency_spectrum(energy, theta, N, dt, shower_type, n_index, R, model, full_output=False, **kwargs):
+def get_frequency_spectrum(amplitude, N, dt, model, full_output=False, **kwargs):
     """
     returns the complex amplitudes of the frequency spectrum of an emitter
 
@@ -107,7 +104,7 @@ def get_frequency_spectrum(energy, theta, N, dt, shower_type, n_index, R, model,
         only available if `full_output` enabled
 
     """
-    tmp = get_time_trace(energy, theta, N, dt, shower_type, n_index, R, model, full_output=full_output, **kwargs)
+    tmp = get_time_trace(amplitude, N, dt, model, full_output=full_output, **kwargs)
     if(full_output):
         return fft.time2freq(tmp[0], 1 / dt), tmp[1]
     else:
--- a/NuRadioMC/SignalGen/emitter.py
+++ b/NuRadioMC/SignalGen/emitter.py
@@ -1,0 +1,114 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+from NuRadioReco.utilities import units, fft
+from NuRadioMC.SignalGen import parametrizations as par
+import logging
+logger = logging.getLogger("SignalGen.askaryan")
+
+
+def set_log_level(level):
+    logger.setLevel(level)
+    par.set_log_level(level)
+
+
+def get_time_trace(energy, theta, N, dt, shower_type, n_index, R, model, full_output=False, **kwargs):
+    """
+    returns the electric field of an emitter
+
+    We implement only the time-domain solution and obtain the frequency spectrum
+    via FFT (with the standard normalization of NuRadioMC). This approach assures
+    that the units are interpreted correctly. In the time domain, the amplitudes
+    are well defined and not details about fourier transform normalizations needs
+    to be known by the user.
+
+    Parameters
+    ----------
+    energy : float
+        energy of the shower
+    theta: float
+        viewangle: angle between shower axis (neutrino direction) and the line
+        of sight between interaction and detector
+    N : int
+        number of samples in the time domain
+    dt: float
+        time bin width, i.e. the inverse of the sampling rate
+    shower_type: string (default "HAD")
+        type of shower, either "HAD" (hadronic), "EM" (electromagnetic) or "TAU" (tau lepton induced)
+        note that TAU showers are currently only implemented in the ARZ2019 model
+    n_index: float
+        index of refraction at interaction vertex
+    R: float
+        distance from vertex to observer
+    model: string
+        specifies the signal model
+        * spherical: a simple signal model of a spherical delta pulse emitter
+    full_output: bool (default False)
+        if True, askaryan modules can return additional output
+
+    Returns
+    -------
+    time trace: 2d array, shape (3, N)
+        the amplitudes for each time bin
+    additional information: dict
+        only available if `full_output` enabled
+
+    """
+    shower_type = shower_type.upper()  # some convenience to be sloppy with upper/lower case
+    trace = None
+    additional_output = {}
+    if(energy == 0):
+        trace = np.zeros(3, N)
+    if(model == 'spherical'):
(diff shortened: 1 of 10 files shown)


ikrav:
This looks nice and simple.

I am more of a C++ person. In C++ I would want to create an abstract base class Emitter with derived classes NeutrinoShowerEmitter and PulserEmitter, with expectation of CosmicRayShowerCoreEmitter or more. The base class would define the interface. Then I would define a base class virtual function for get_time_trace and overload it in all derived classes with all functions returning 3d trace (passing maybe the launch vector into the Emitter). But this is python, and I do not know good practices in python well enough. I've heard that python does have abstract classes and interfaces since recently. I do agree that it would be nice to enforce a common interface without having multiple places to edit in case of changes.

Perhaps we should just give this implementation a go.

NuRadioMC/SignalGen/emitter.py
```diff
@@ -0,0 +1,114 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+from NuRadioReco.utilities import units, fft
+from NuRadioMC.SignalGen import parametrizations as par
```
ikrav, line 4: Is this line needed?

NuRadioMC/SignalGen/emitter.py
```diff
@@ -0,0 +1,114 @@
...
+import logging
+logger = logging.getLogger("SignalGen.askaryan")
+
+
+def set_log_level(level):
+    logger.setLevel(level)
+    par.set_log_level(level)
```
ikrav, line 11: Is this line needed?

ikrav (commented):
Overall looks good to me, just two lines in the emitter.py appear to me unneeded.

NuRadioMC/SignalGen/emitter.py
cg-laser: good point. These lines should be removed.

cg-laser:
I agree, in C++ we would write a base class to ensure the same interface. As far as I know, this won't work in python because you always overwrite the functions any way you want. Let's see what others think @shallmann @christophwelling ?

ikrav:
Christian, there is one thing that Pawan noticed. Presently, in simulation.py the function call of get_frequency_spectrum() for the emitter case does not contain the field "model" while the function implementation in the emitter.py does have it.
The call:
   https://github.com/nu-radio/NuRadioMC/blob/feature/emitter/NuRadioMC/simulation/simulation.py#L733
the implementation:
   https://github.com/nu-radio/NuRadioMC/blob/feature/emitter/NuRadioMC/SignalGen/emitter.py#L74
Generally, we need to pass into the emitter module some number of parameters, at the very least:
   - the V_src(t) pattern (such as HVSP, IDL pulsers, CW train, etc), which could be the "model"
   - the amplitude of the V_src(t), which could be the existing "energy" field
   - the transmitting antenna type (such as bicone_v8)
 Possibly other parameters would also be needed in the future. Should we at least add the missing "model" field and additionally introduce "antenna_type"?

     I also wonder if it makes sense to define in the yaml config for event generation a section "emitter" with fields "model", "antenna_type", "amplitude", etc. Then do in simulation.py something like 
    emitterConifig = self._cfg['emitter']
where emitterConfig woud only contain the emitter description fields from the yaml config. The emitterConfig can then be passed into the get_frequency_spectrum and then into the get_time_trace in the emitter.py. (I do not know if this can be done exactly as above, or if there are any reasons why it shouldn't be done). The advantage of this would be in the fact that there will be no need to change siulation.py or even emitter.py while a specific implementation of an emitter is being developed. The changes would be confined to the config.yaml and the separate .py file for a specific emitter implementation (I expect we won't have explicit calculations of anything in the emitter.py itself, but will have that in a separate .py file/class).

christophwelling (approved):
Looks good to me.

About writing a mother class for the emittters: I think the python way to do it would be to write an emitter class that has the get_trace() and get_frequency_spectrum(), which define the interface. Internally, these methods would call _get_trace() and _get_frequency(), which would be overridden by the child classes.

In this case, that seems a bit like overkill, so I am fine with leaving it as it is for now.

cg-laser:
@ikrav I started implementing what we discussed on Thursday. I created a new test for it so you directly have a working example which I also added to our automated testing routine. So far the emitter code (covolving the voltage output with the emitting antenna response ) is just in the main simulation.py code. We should probably refactor this later and put it into a separate function. But for now, we can also just leave it there to test it. You still need to implement the correct way of convolving the antenna. Let me know what you think of the new interface. 

One thing that is not very nice is that the input file still needs to contain fields like `shower_energy` although it is not used anywhere. I'll think about how we can clean up the code. It does not impact the functionality of the code, just its beauty ;-)

ikrav:
@cg-laser This looks good to me in general from just looking in the code, Pawan and I should run it to get a better feel for it. My immediate comment was, as you point out, that having actual calculations with VEL and stuff in simulation.py is not appropriate, but we can start with this and move it out later. Having a test setup that you implemented is very helpful.

pgiri4:
Thanks, Christian.  tried this one for spherical and square signals and looks good for now. We can try loading VEL outside the simulation.py . Once I find it working with all models then I will let you know .














[https://ucomm.unl.edu/images/email-signature/Nebraska_N_RGB_small_on_white.gif]

Pawan Giri

Graduate Teaching Asst

University of Nebraska–Lincoln

Physics & Astronomy

JH-253

________________________________

cg-laser:
This PR was continued on #342

PR #318, opened 2021-10-04
Update ray trace minimizer with smarter initial conditions

clark2668:
This updates the C++ ray tracer, specifically to choose the initial conditions more cleverly based on where the emitter is placed relative to the receiver. It also scans for minima more carefully if the vertex is nearby.


cg-laser:
the test fails because the maximum amplitude per ray tracing solution is not the same anymore. The differences seems to be quite large (several percent) which I find suprising. Especially because the test does not trigger other discrepancies in other parameters. @boboeyen @christophwelling any thoughts?

boboeyen:
Oh, hmm, I don't have any experience with the tests, but I'll have a look to see whether I can spot something

anelles:
This seems pretty old already. @cg-laser do you need to update the files again before the testsare fixed?

sjoerd-bouma:
Finally made time to look at this. The issue here is exactly the same as in #450 - we do not sort our ray tracing solutions consistently, so with different initial conditions the order they are returned in can change. My suggestion would be to fix that first (i.e. merge #450), then this should also be possible to be merged without further issues.

clark2668:
Hey @sjoerd-bouma --good catch! That sounds like a good plan to me.

As an aside, if you want to add some docs to the test script, that would help. I was totally unable to debug it.

cg-laser:
I checked the tests. Two are still failing. The `C0` test at the South Pole with 
```
Not equal to tolerance rtol=1e-07, atol=0
Mismatched elements: 4 / 2000 (0.2%)
Max absolute difference: 8.58962252e-07
Max relative difference: 1.32[44](https://github.com/nu-radio/NuRadioMC/actions/runs/4196723284/jobs/7279944369#step:14:45)8028e-07
 x: array([[0.      , 0.      ],
       [2.817528, 2.832811],
       [0.      , 0.      ],...
 y: array([[0.      , 0.      ],
       [2.817528, 2.832811],
       [0.      , 0.      ],...
```
This is just a tiny difference and ok, we can just update the reference file. 

However, also the Veff test with noise and phased array fails (8% deviation in Veff). This still might be just one event but I'm not sure why this happened. Anyone any idea? Maybe the random sequence was effected causing different noise realizations?

sjoerd-bouma:
@cg-laser Here's a theory - the noise is added only when a ray tracing solution is found. The first event that triggers on the new branch has a vertex at [128, -63, -194] m (trigger at -200m), where on `develop` no ray tracing solution is found. On the new branch, a solution is found, so noise is added to this event, but that means that the random noise realization for the next event is 'shifted along by one' on `raytracer_ic` compared to `develop`. 
- Event group ids triggering on `raytracer_ic`, but not on `develop`:
`[3974 4393 4671 4927 5475 6004 6632 6949 7080 8058 8175 8302 8379 8854
 9495 9931]` (NB: 3974 is ~horizontal and has no ray tracing solutions on `develop`)
- Event group ids triggering on `develop`, but not on `raytracer_ic`:
`[5636 9091 9113 9261 9792]`

So all the difference is after the first event for which the raytracing fails on `develop`. If we're happy with this explanation, I will update these reference files too. Tong has already checked yesterday that this branch successfully finds the same solutions as the Python version - maybe it would be nice to explicitly include some of these challenging cases in a test in the future.

cg-laser:
thanks Sjoerd for checking this in detail. I'm satisfied with your explanation and this that it makes sense. Please go ahead and update the reference files. 

And I agree, updating the test to include these challenging trajectories would be good. In principle, we could do it in this PR as we're anyway updating the reference files, but I leave this to you. 

Thanks Tong for directly checking the new implementation.

sjoerd-bouma:
I'm afraid I'm going to put a test for tricky vertices on the longer-term to-do list, I think it's worthwhile merging this in any case. Thanks Brian and sorry for the long wait!

cg-laser (approved):

clark2668:
No problem! Sorry this turned out to be such a thorny issue, and thanks everyone for working on it so diligently!

sjoerd-bouma:
Sorry, I was too slow to merge this so @fschlueter created a tiny merge conflict... now fixed. Can someone approve again?

anelles (approved):
Tiny change from previous approval. Works for me.

PR #319, opened 2021-10-05
Birefringence

nilsphysics:
I added a new ice model to medium.py. It combines the southpole_2015 model with data from Jordan et al. (https://arxiv.org/abs/1910.01471). This can be used to model birefringence of the radio signal.

A second smoother interpolation was added and the read in of the interpolation files was changed to the __init__ function.


cg-laser:
@nilsphysics can you add the plots of the data and your interpolation?

nilsphysics:
> @nilsphysics can you add the plots of the data and your interpolation?

This is the data from the paper:
Screenshot from 2021-10-05 14-13-11

This is my interpolation to the data (with added constant values from 0m to 140m and 1800m to 2500m):
interpolation3

This is how the model looks like zoomed in on the interesting region:
IceModel

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1744,7 +1749,362 @@ def get_path(self, iS, n_points=1000):
         dP = path_2d - np.array([self.__X1[0], 0, self.__X1[2]])
         MM = np.matmul(self.__R.T, dP.T)
         path = MM.T + self.__X1
-        return path
+        return path        
+                    
+    def get_effective_index(self, n, direction, nx, ny, nz):
+        """
+        objective function to find the possible values of 'n' into the direction of 'k'
```
anelles, line 1756: Could you add a bit more doc string. I am not sure that I follow what "possible values" mean.

anelles: Also please describe the output (format and meaning)

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1744,7 +1749,362 @@ def get_path(self, iS, n_points=1000):
...
+        sz = direction[2]
+        A = (nx**2 - n**2)*(ny**2 - n**2)*(nz**2 - n**2)
+        B = n**2 *( sx**2 *(ny**2 - n**2)*(nz**2 - n**2) + sy**2 *(nx**2 - n**2)*(nz**2 - n**2) + sz**2 *(nx**2 - n**2)*(ny**2 - n**2))
+
+        return A + B
+
+    def get_polarization(self, n, direction, nx, ny, nz, rounding=20):
+        
+        """
+        returns the polarization of a wave for the direction of propagation
```
anelles, line 1784: Please elaborate how the polarization is defined and what format the output is given.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1744,7 +1749,362 @@ def get_path(self, iS, n_points=1000):
...
+                if(np.isclose(n, nx)):
+                    return np.array([1, 0, 0])
+                elif(np.isclose(n, ny)):
+                    return np.array([0, 1, 0])
+        # now treat the cases where the direction is perpendicular to just one of the principal axes
+            elif(np.isclose(np.dot(direction, np.array([1, 0, 0])), 0)):  # direction perpendicular to x-axis
+                if(np.isclose(n, nx)):
+                    return np.array([1, 0, 0])
+                else:
+                    print(f"ERROR: propagation direction perpendicular to x-axis but n not equal to n_x {n}")
```
anelles, line 1827: Please use the logger as it is done everywhere else. Then the user can specify which level of warning or Error they want to see.

anelles: This is for ALL print statements. Print statements should NOT be used in any module.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1744,7 +1749,362 @@ def get_path(self, iS, n_points=1000):
...
+                print(n, direction, nx, ny, nz)
+                return [0, 0, 0]
+        else:
+            p = np.array([direction[0] / (n ** 2 - nx ** 2), direction[1] / (n ** 2 - ny ** 2), direction[2] / (n ** 2 - nz ** 2)])
+            return p / np.linalg.norm(p)
+
+    def get_3d_trace(self, source, antenna, acc = 1000):
+        
+        """
+        returns the trace of the two possible solutions provided by get_path
```
anelles, line 1853: Please document the format of the output

NuRadioMC/SignalProp/analyticraytracing.py
anelles: I have not commented in detail on all other print statements. Please fix all of them.

NuRadioMC/SignalProp/analyticraytracing.py
anelles: I also will not comment on future functions. I am sensing a theme that this is missing in many cases.

NuRadioMC/utilities/birefringence_example/IceModel_new.py
```diff
@@ -0,0 +1,29 @@
...
+
+plt.title('Refractive index model for birefringence')
+plt.xlabel('depth [m]')
+plt.ylabel('Refractive index')
+#plt.legend()
+#plt.xlim(100, 200)
+#plt.ylim(1.77, 1.782)
+#plt.grid()
+plt.show()
+
```
anelles, line 28: Please provide some documentation what this is supposed to do. The filename is not very instructive. I assume this is a debug script?

anelles (changes requested):
Thanks for contributing this. Could you please fix a couple of comments with regard to documentation of the new model and in particular with using the logging functionality rather than print statements.

NuRadioMC/utilities/medium.py
```diff
@@ -44,6 +45,127 @@ def get_ice_model_radiopropa(self):
             return ice
 """
 
+
+
+class birefringence_index:
```
boboeyen, line 50: The class shoudl inherit from at least 'IceModel' to make sure it is seen as an icemodel in the code. Also, I think the name should be more specific since it is a very specific ice model (birefringence of southpole 2015 using Jordan et.al. results)

boboeyen (commented):
Some general remarks if we would like to include this in the develop branch.
1. I think we should think about how to structure the medium and for example where to store the numerical data you use for the interpolation. Just so we prpare a bit for the futher
2. Now you have adapted the analytical raytracer but I don't see how to use it in the same fashion as other raytracers. Especialy because the find_solutions function is not adapted and thus not aware of the birefringence (to my understanding). I think it would be better to make it a seperate module based on the analytic ray tracer. Also, by using the birefringence, it is not analytically solvable anymore, so it could be confusing if it where the same module.

anelles:
I think this looks good now. @nilsphysics could you clean up the merging conflict?

NuRadioMC/utilities/medium.py
```diff
@@ -65,6 +330,26 @@ def __init__(self):
             z_0 = 77.*units.meter, 
             delta_n = 0.423,
             )
+        
+class southpole_2015_up(medium_base.IceModelSimple):
```
boboeyen, line 334: Do you need these somewhere?

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1545,7 +1558,10 @@ class describing the index-of-refraction profile
         # make sure that arrays are floats
         self.__logger = logging.getLogger('ray_tracing')
         self.__logger.setLevel(log_level)
+        
         self.__medium = medium
+        self.__birefringence_model = birefringence_model
```
boboeyen, line 1563: To me it looks weird that you would have an ice model and a birefringence model. I think the information should be contained in 1 object (a medium which is or is not a birefringence model) because otherwise you could mix up the greenland model with a southpole birefringence model.

boboeyen (commented):
I think the structure is already much better. Thanks for all your time and contribution! If you have questions are you're not sure how to proceed, I am glad to help you restructure things a bit further :)

NuRadioMC/utilities/birefringence_models/IceModelB.py
```diff
@@ -0,0 +1,242 @@
...
+    
+    plt.title('refractive index for birefringence')
+    plt.xlabel('depth [m]')
+    plt.ylabel('refractive index')
+    plt.legend()
+    #plt.xlim(-1750, -140)
+    #plt.ylim(1.778, 1.782)
+    plt.grid()
+    plt.show()
+
```
anelles, line 202: Please remove all those empty lines of whitespace.

NuRadioMC/utilities/birefringence_models/IceModelD.py
```diff
@@ -0,0 +1,198 @@
...
+n3 = np.sqrt(e_p + e_d * e3)
+depth = dep
+
+
+print(np.mean(n1))
+print(np.mean(n2))
+print(np.mean(n3))
+
+interpolation = np.array([np.mean(n1), np.mean(n2), np.mean(n3)])
+
```
anelles, line 107: Same as above. Please use debug output properly.

NuRadioMC/utilities/birefringence_models/IceModelE.py
```diff
@@ -0,0 +1,273 @@
...
+tck1 = f1._eval_args
+f1_rec = interpolate.UnivariateSpline._from_tck(tck1)
+
+tck2 = f2._eval_args
+f2_rec = interpolate.UnivariateSpline._from_tck(tck2)
+
+tck3 = f3._eval_args
+f3_rec = interpolate.UnivariateSpline._from_tck(tck3)
+
+
```
anelles, line 170: Debug again.

anelles (changes requested):
I think changing all the print statements to logger messages is still outstanding here. 

Also, many of @boboeyen 's comments are still valid a should be addressed. You can also choose not to change something, but then please write it up here. 

The model_description files seems empty, this needs fixing.

cg-laser:
Thanks for the comments. This PR still has a few todos before it should be merged, in particular, better integration into the main NuRadioMC simulation routine. The work will be continued after the summer break.

NuRadioMC/utilities/medium.py
nilsphysics: No, this was only for cross check reasons, I deleted it.

NuRadioMC/SignalProp/analyticraytracing.py
nilsphysics: I reworked the structure of the ice model such that the birefringence ice model can be used as a regular medium in the ray_tracer class. No need for a second ice model. Let me know if the overall structure of the model is what you had in mind.

NuRadioMC/utilities/medium.py
```diff
@@ -44,6 +47,268 @@ def get_ice_model_radiopropa(self):
...
+        nz = self.m.get_index_of_refraction(z) + self.f3_rec(-z[2]) - self.comp 
+                
+        return nx, ny, nz
+
+
+        
+    
+    
+    
+class birefringence_index_B:
```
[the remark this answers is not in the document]
nilsphysics: I tried to put the basic functionality into the base_medium class but the ice models don't all have the same structure so I'm still working on finding the best solution.

nilsphysics:
I think the birefringence module is ready for merging now. Birefringence was included in both the analytical raytracer and the radiopropa raytracer. 

The analytical version includes: 
(1) Single pulse propagation for special pulses such as the SPICE drop or calibration pulses. 
(2) Full integration into the NuRadioMC pipeline to generate large amounts of events where the propagation includes birefringence. 
(3) A speed up version using radiopropa to handle the matrix multiplication in C++. 

The radiopropa version includes: 
(1) Full integration into the NuRadioMC pipeline to generate large amounts of events where the propagation includes birefringence. 
(2) The option to combine non-exponential density functions with the birefringence effect. 

I tried to speed up the calculations as much as possible but the birefringence calculations remain the bottleneck in a full simulation. However, it's a working, usable version that could be interesting to use for others. Feel free to let me know if I should change anything else.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+        solution_type: list:    [sky_e, sky_o, l_path, N0, N1]
+        meaning:                [0] - sky_0 (numpy.array) - polarization vectors of the ordinary ray along the trace
+                                [1] - sky_1 (numpy.array) - polarization vectors of the extraordinary ray along the trace
+                                [2] - l_path (numpy.array) - length of the path increment
+                                [3] - N0 (numpy.array) - effective refractive index vectors of the ordinary ray along the trace
+                                [4] - N1 (numpy.array) - effective refractive index vectors of the extraordinary ray along the trace
+        """
+        
+        ice_n = self.__medium
+        p = self.get_3d_trace(source, antenna, acc)
```
cg-laser, line 1974: you already called this function in the `get_time_delays` function. Hence, you are repeating the ray tracing twice. You don't need to do that. The ray tracing was anyway done already. There is also no need of passing `source` and `antenna` to the function. This information already exists as member variables `_X1` and `_X2`. You can assume that the user already called the function `set_start_and_end_point(source, antenna)`

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+        
+        h_th = signal.hilbert(end_theta)
+        h_ph = signal.hilbert(end_phi)
+        
+        if style == 'single pulse':
+            t_delay = TT[h_th == max(h_th)] - TT[h_ph == max(h_ph)]
+            return(start_theta, start_phi, end_theta, end_phi, TT, t_delay)
+        
+        elif style == 'simulation pulse':
+            el_field = np.vstack((f_spectrum.get_frequency_spectrum()[0], t_theta.get_frequency_spectrum(), t_phi.get_frequency_spectrum()))
```
cg-laser, line 2192: confusing that you return the first value of `f_spectrum`. Better write `np.zeros(n_samples)` or similar.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -2083,6 +2555,23 @@ def apply_propagation_effects(self, efield, i_solution):
         if self._config['propagation']['focusing']:
             focusing = self.get_focusing(i_solution, limit=float(self._config['propagation']['focusing_limit']))
             spec[1:] *= focusing
+        
+        # apply the birefringence effect
+        if self._config['propagation']['birefringence']:
+        
+            try:
```
cg-laser, line 2562: bad style. If RadioPropa should be used or not should be a config flag.

NuRadioMC/SignalProp/radioproparaytracing.py
```diff
@@ -381,6 +392,111 @@ def delta(ray_dir,shower_dir):
     def set_minimizer_tolerance(self, xtol=1e-3*units.deg, ztol=1e-3*units.meter):
         self.__xtol = xtol
         self.__ztol = ztol
+
+    def raytracer_birefringence(self, launch_v, spec, s_rate):
+        """
+        Uses RadioPropa propagate an electric field through the correct solution of the ray path. To use add to the config.yaml file:
+        propagation:
```
cg-laser, line 399: indentation seems wrong

NuRadioMC/SignalProp/radioproparaytracing.py
```diff
@@ -381,6 +392,111 @@ def delta(ray_dir,shower_dir):
...
+        Parameters
+        ----------
+        launch_v: 3d array
+            launch vector calculated by a previous analytical or numerical ray tracing algorithm
+        
+        spec: numpy array
+            frequency spectrum of the electric field
+
+        s_rate: float
+            sampling rate of the electric field
```
cg-laser, line 413: return value not documented.

NuRadioMC/SignalProp/radioproparaytracing.py
```diff
@@ -381,6 +392,111 @@ def delta(ray_dir,shower_dir):
...
+        module: radiopropa
+        ice_model: southpole_2015
+        birefringence: True
+
+        Parameters
+        ----------
+        launch_v: 3d array
+            launch vector calculated by a previous analytical or numerical ray tracing algorithm
+        
+        spec: numpy array
```
cg-laser, line 409: Accept a BaseTrace object as input. Then you don't need to ask for the sampling rate.

NuRadioMC/SignalProp/radioproparaytracing.py
```diff
@@ -381,6 +392,111 @@ def delta(ray_dir,shower_dir):
     def set_minimizer_tolerance(self, xtol=1e-3*units.deg, ztol=1e-3*units.meter):
         self.__xtol = xtol
         self.__ztol = ztol
+
+    def raytracer_birefringence(self, launch_v, spec, s_rate):
+        """
+        Uses RadioPropa propagate an electric field through the correct solution of the ray path. To use add to the config.yaml file:
```
cg-laser, line 398: Expand doc string to say what the function does.

NuRadioMC/SignalProp/radioproparaytracing.py
```diff
@@ -381,6 +392,111 @@ def delta(ray_dir,shower_dir):
...
+        try:
+            X1 = self._X1 * (radiopropa.meter/units.meter)
+            X2 = self._X2 * (radiopropa.meter/units.meter)
+        except TypeError: 
+            self.__logger.error('NoneType: start or endpoint not initialized')
+            raise TypeError('NoneType: start or endpoint not initialized')
+
+        v = (self._X2 - self._X1)
+        u = copy.deepcopy(v)
+        u[2] = 0
```
cg-laser, line 424: this seems wrong. Don't you rather need to rotate the coordinates so that the propagation aligns in one plane?

NuRadioMC/SignalProp/radioproparaytracing.py
```diff
@@ -381,6 +392,111 @@ def delta(ray_dir,shower_dir):
...
+            launch vector calculated by a previous analytical or numerical ray tracing algorithm
+        
+        spec: numpy array
+            frequency spectrum of the electric field
+
+        s_rate: float
+            sampling rate of the electric field
+        """
+        try:
+            X1 = self._X1 * (radiopropa.meter/units.meter)
```
cg-laser, line 416: better check this directly using `if self._X1 is None or self._X2 is None: ....`

NuRadioMC/utilities/birefringence_models/IceModel_interpolation.py
```diff
@@ -0,0 +1,187 @@
...
+#import scipy.optimize
+import matplotlib.pyplot as plt
+import time
+from NuRadioMC.SignalProp import analyticraytracing as ray
+from NuRadioMC.utilities import medium
+from NuRadioMC.utilities.medium import southpole_2015
+from NuRadioReco.utilities import units
+import logging
+
+logging.basicConfig(level=logging.INFO)
```
cg-laser, line 15: add a doc string. I assume that this is never actually used and is just your script to generate the spline tables for the ice models?

NuRadioMC/utilities/medium.py
```diff
@@ -80,6 +84,36 @@ def __init__(self):
             )
 
 
+
+class ARA_2022(medium_base.IceModelSimple):
+    def __init__(self):
+        # define model parameters (SPICE 2015/southpole)
```
cg-laser, line 90: update doc string with correct reference to the paper.

NuRadioMC/utilities/medium_base.py
```diff
@@ -375,6 +376,192 @@ def _compute_default_ice_model_radiopropa(self):
...
+        self.bir_model = bir_model
+
+        self.f1 = interpolate.UnivariateSpline._from_tck(self.bir_model[0])
+        self.f2 = interpolate.UnivariateSpline._from_tck(self.bir_model[1])
+        self.f3 = interpolate.UnivariateSpline._from_tck(self.bir_model[2])
+
+        self.comp = self.exp_model.get_index_of_refraction(np.array([0, 0, -2500]))
+
+
+    def get_index_of_refraction(self, position):
```
cg-laser, line 433: this function is already implemented in the base class from which you inherit.

NuRadioMC/utilities/medium_base.py
```diff
@@ -375,6 +376,192 @@ def _compute_default_ice_model_radiopropa(self):
...
+        position:  3dim np.array
+                    point
+
+        Returns
+        -------
+        n:  list [nx, ny, nz]
+            index of refraction for every direction
+        """
+
+        nx = self.exp_model.get_index_of_refraction(position) + self.f1(-position[2]) - self.comp
```
cg-laser, line 478: some additional explanation would be good for why `self.comp` needs to be subtracted. I guess because the birefringence n(z) profile is parameterized assuming a constant density resulting in n=1.78?

cg-laser (changes requested):
Hey Nils, great work getting the birefringence implementation ready for review! This will be an important new feature of NuRadioMC. 

I went through all code changes in detail and provided inline comments. A couple of things need to be cleaned up before we can merge this. I also think there could still be quite some speed improvements.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
                                                  reflection=result['reflection'],
                                                  reflection_case=result['reflection_case'])
         path_2d = np.array([xx, np.zeros_like(xx), zz]).T
+
         dP = path_2d - np.array([self._X1[0], 0, self._X1[2]])
         MM = np.matmul(self._R.T, dP.T)
         path = MM.T + self._X1
         return path
 
+    def get_effective_index_fast(self, nx, ny, nz, direction):
```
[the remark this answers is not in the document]
nilsphysics: Done

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+
+        n1 = np.sqrt((-2*nx**2*ny**2*nz**2)/(ny**2*nz**2*(-1 + sx**2) + nx**2*(nz**2*(-1 + sy**2) + ny**2*(-1 + sz**2)) - np.sqrt(4*nx**2*ny**2*nz**2*(nz**2*(-1 + sx**2 + sy**2) + ny**2*(-1 + sx**2 + sz**2) + nx**2*(-1 + sy**2 + sz**2)) + (ny**2*nz**2*(-1 + sx**2) + nx**2*(nz**2*(-1 + sy**2) + ny**2*(-1 + sz**2)))**2)))
+        n2 = np.sqrt((-2*nx**2*ny**2*nz**2)/(ny**2*nz**2*(-1 + sx**2) + nx**2*(nz**2*(-1 + sy**2) + ny**2*(-1 + sz**2)) + np.sqrt(4*nx**2*ny**2*nz**2*(nz**2*(-1 + sx**2 + sy**2) + ny**2*(-1 + sx**2 + sz**2) + nx**2*(-1 + sy**2 + sz**2)) + (ny**2*nz**2*(-1 + sx**2) + nx**2*(nz**2*(-1 + sy**2) + ny**2*(-1 + sz**2)))**2)))
+
+
+
+
+        return(n1, n2)
+
+    def get_polarization(self, n, direction, nx, ny, nz, rounding=1e-08):
```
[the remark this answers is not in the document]
nilsphysics: Done

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+        direction: numpy.array
+            propagation direction of the wave
+        nx: float
+            the index of refraction into the x direction
+        ny: float
+            the index of refraction into the y direction
+        nz: float
+            the index of refraciton into the z direction
+        rounding: int
+            the precision of the comparison between n and nx, ny, nz for the special cases
```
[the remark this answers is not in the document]
nilsphysics: Done

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+
+        elif (np.isclose(n, nz, atol=rounding)):
+            p = np.array([0, 0, 1])
+
+        else:
+            p = np.array([direction[0] / (n ** 2 - nx ** 2), direction[1] / (n ** 2 - ny ** 2), direction[2] / (n ** 2 - nz ** 2)])
+
+        return p / np.linalg.norm(p)
+
+    def get_3d_trace(self, source, antenna, acc=1000):
```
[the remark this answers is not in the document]
nilsphysics: Done, the information is passed directly through the ray tracing solution.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+        c = speed_of_light * units.m / units.ns
+
+        t_0 = []
+        t_1 = []
+
+        for i in range(acc-1):
+
+            if path_type == 0:
+
+                n1 = ice_n.get_birefringence_index_of_refraction(p[0, i])[0]
```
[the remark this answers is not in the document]
nilsphysics: Unnecessary after the code was rewritten.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+
+            s = self.get_effective_index_fast(n1, n2, n3, direction)
+
+            n_0 = s[0]
+            n_1 = s[1]
+
+            t_0.append(len_diff * n_0 / c)
+            t_1.append(len_diff * n_1 / c)
+
+            l = open("ana_l.txt","a+")
```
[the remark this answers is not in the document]
nilsphysics: Done.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+                n2 = ice_n.get_birefringence_index_of_refraction(p[0, i])[1]
+                n3 = ice_n.get_birefringence_index_of_refraction(p[0, i])[2]
+
+                dx = p[0, i+1, 0] - p[0, i, 0]
+                dy = p[0, i+1, 1] - p[0, i, 1]
+                dz = p[0, i+1, 2] - p[0, i, 2]
+
+            elif path_type == 1:
+
+                n1 = ice_n.get_birefringence_index_of_refraction(p[1, i])[0]
```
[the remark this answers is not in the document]
nilsphysics: Done.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+                n2 = ice_n.get_birefringence_index_of_refraction(p[1, i])[1]
+                n3 = ice_n.get_birefringence_index_of_refraction(p[1, i])[2]
+
+                dx = p[1, i+1, 0] - p[1, i, 0]
```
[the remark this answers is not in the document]
nilsphysics: Done.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+            direction = np.array([dx, dy, dz])
+            len_diff = np.linalg.norm(direction)
+            direction = direction / len_diff
+
+            s = self.get_effective_index_fast(n1, n2, n3, direction)
+
+            n_0 = s[0]
+            n_1 = s[1]
+
+            t_0.append(len_diff * n_0 / c)
```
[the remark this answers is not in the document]
nilsphysics: Done.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+
+        ice_n = self.__medium
+
+        p = self.get_3d_trace(source, antenna, acc)
+        c = speed_of_light * units.m / units.ns
+
+        t_0 = []
+        t_1 = []
+
+        for i in range(acc-1):
```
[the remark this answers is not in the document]
nilsphysics: Unnecessary after the code was rewritten.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+        sky_0 = []
+        sky_1 = []
+
+        p_0 = []
+        p_1 = []
+
+        N0 = []
+        N1 = []
+
+        for i in range(acc-1):
```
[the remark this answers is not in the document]
nilsphysics: Unnecessary after the code was rewritten.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+        p_1 = np.array(p_1)
+
+        N0 = np.array(N0)
+        N1 = np.array(N1)
+
+        l_path = np.array(l_path)
+
+        return(sky_0, sky_1, l_path, N0, N1)
+
+    def get_pulse_trace_fast(self, source, antenna, pulse, path_type=0, acc=1000, samp_rate = 0.01):
```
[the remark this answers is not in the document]
nilsphysics: That's what I did in the beginning but as `apply_propagation_effects` does a spectrum to Efield conversion anyways, passing a BaseTrace object would add an extra step of converting back and forth.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+
+        t_slow = base_trace.BaseTrace()
+        t_fast = base_trace.BaseTrace()
+
+        t_theta = base_trace.BaseTrace()
+        t_phi = base_trace.BaseTrace()
+
+        f_spectrum = base_trace.BaseTrace()
+
+        if type(pulse) == str:
```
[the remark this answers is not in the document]
nilsphysics: Done.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+            
+            t_theta.set_trace(etheta, sampling_rate=samp_rate)
+            t_phi.set_trace(ephi, sampling_rate=samp_rate)
+
+        else:
+            self.__logger.error("error: wrong data type")
+
+        TT = t_theta.get_times()
+
+        shift = TT[t_theta.get_trace() == max(t_theta.get_trace())]
```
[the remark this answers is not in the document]
nilsphysics: Unnecessary after the code was rewritten.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+
+        t_theta.set_trace_start_time(-shift)
```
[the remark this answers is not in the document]
nilsphysics: The idea was to shift the pulse maximum to zero so the time delay calculations would be easier later, I removed it in the new code.

NuRadioMC/SignalProp/analyticraytracing.py
nilsphysics: Done.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+                time_shift = diff[i]
+            
+            th = t_theta.get_frequency_spectrum()
+            ph = t_phi.get_frequency_spectrum()
+            t_slow.set_frequency_spectrum(th * R[0, 0] + ph * R[0, 1], sampling_rate=1 / dt)
+            t_fast.set_frequency_spectrum(th * R[1, 0] + ph * R[1, 1], sampling_rate=1 / dt)
+            
+            t_fast.apply_time_shift(time_shift)
+            
+            Rinv = np.linalg.inv(R)
```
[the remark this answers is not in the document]
nilsphysics: Done.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+            
+            fa = t_fast.get_frequency_spectrum()
+            sl = t_slow.get_frequency_spectrum()
+            t_theta.set_frequency_spectrum(sl * Rinv[0, 0] + fa * Rinv[0, 1], sampling_rate=1 / dt)
+            t_phi.set_frequency_spectrum(sl * Rinv[1, 0] + fa * Rinv[1, 1], sampling_rate=1 / dt)
```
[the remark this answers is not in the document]
nilsphysics: One trace has to be set/got as in order to properly apply the time shift. The rest was implemented as commented.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1718,11 +1750,448 @@ def get_path(self, iS, n_points=1000):
...
+        
+        end_theta = t_theta.get_trace()
+        end_phi = t_phi.get_trace()
+        
+        h_th = signal.hilbert(end_theta)
+        h_ph = signal.hilbert(end_phi)
+        
+        if style == 'single pulse':
```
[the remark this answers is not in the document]
nilsphysics: Done.

NuRadioMC/SignalProp/analyticraytracing.py
nilsphysics: Unnecessary after the code was rewritten.

NuRadioMC/SignalProp/analyticraytracing.py
nilsphysics: Done.

NuRadioMC/SignalProp/radioproparaytracing.py
nilsphysics: Here I'm not sure what you mean. The indentation of the function itself or of the documentation? Both look normal to me.

NuRadioMC/SignalProp/radioproparaytracing.py
nilsphysics: Done.

NuRadioMC/SignalProp/radioproparaytracing.py
nilsphysics: Same as in the analytical case. `apply_propagation_effects ` does a spectrum to Efield conversion anyways, passing a BaseTrace object would add an extra step of converting back and forth.

NuRadioMC/SignalProp/radioproparaytracing.py
nilsphysics: Done.

NuRadioMC/SignalProp/radioproparaytracing.py
nilsphysics: This was taken directly from the `raytracer_iterative` function. It seems like it is only used to define the boundary behind the observer. In our case this should not be of great importance.

NuRadioMC/SignalProp/radioproparaytracing.py
nilsphysics: Done.

NuRadioMC/utilities/birefringence_models/IceModel_interpolation.py
nilsphysics: Correct, this file is only a side in case someone wants to create their own spline table. I added a doc string.

NuRadioMC/utilities/medium.py
nilsphysics: Done.

NuRadioMC/utilities/medium_base.py
```diff
@@ -375,6 +376,192 @@ def _compute_default_ice_model_radiopropa(self):
...
+
+        Parameters
+        ----------
+        exp_model:    pointer
+                        exponential parametrisation of the density effect on the refractive index
+        bir_model:    str
+                        choice for the birefringence model, 'southpole_A', 'southpole_B', ...
+        n_ice:  float, dimensionless
+                refractive index of the deep bulk ice
+        delta_n:  float, NuRadio length units
```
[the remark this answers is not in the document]
nilsphysics: Unnecessary after the code was rewritten.

NuRadioMC/utilities/medium_base.py
nilsphysics: Done.

NuRadioMC/utilities/medium_base.py
nilsphysics: Correct, added a comment in the doc string of the `analyticalraytracer`, where this subtraction is done now.

NuRadioMC/utilities/medium_base.py
```diff
@@ -375,6 +376,192 @@ def _compute_default_ice_model_radiopropa(self):
...
+            index of refraction for every direction
+        """
+
+        nx = self.exp_model.get_index_of_refraction(position) + self.f1(-position[2]) - self.comp
+        ny = self.exp_model.get_index_of_refraction(position) + self.f2(-position[2]) - self.comp
+        nz = self.exp_model.get_index_of_refraction(position) + self.f3(-position[2]) - self.comp
+
+        return nx, ny, nz
+
+    def get_average_index_of_refraction(self, position1, position2):
```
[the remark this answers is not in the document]
nilsphysics: Done.

NuRadioMC/utilities/medium_base.py
```diff
@@ -375,6 +376,192 @@ def _compute_default_ice_model_radiopropa(self):
...
+        -------
+        n_nabla:    (3,) np.array
+                    gradient of index of refraction at the point
+        """
+        gradient = np.array([0., 0., 0.])
+        if (position[2] - self.z_air_boundary) <=0:
+            gradient[2] = -self.delta_n / self.z_0 * np.exp((position[2] - self.z_shift) / self.z_0)
+        return gradient
+
+    def get_ice_model_radiopropa(self):
```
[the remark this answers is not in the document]
nilsphysics: Done.

NuRadioMC/SignalProp/analyticraytracing.py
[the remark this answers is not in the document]
nilsphysics: Actually no, for geometries which are parallel to the ice flow the matrix is less and less orthonormal, so the inverse has to be applied. This step is also not a bottle neck.

nilsphysics:
I think I finished all requested improvements. The analytical module got faster and more readable and I added the following example scripts to play with birefringence:

01 simple pulse propagation: Propagate a pulse with birefringence with different propagation modules and ice models.
02 path info: Get all quantities related to birefringence along the propagation path in a single dictionary. (Dielectric tensor, effective refractive indices, polarization eigenvectors, time delays, path trajectory).
03 ARA-A5: Example script for the Spice drop measurement at the ARA A5 station related to birefringence.
04 ARIANNA: Example script for the Spice drop measurement at the ARIANNA station related to birefringence.
05 Veff: Comparison of the effective volume with and without birefringence.

I think this could be a good point to merge the branch. There are follow up topics I'm working on (including the Greenland ice-model, including different ice models in the radiopropa ray tracer) but maybe these are topics for another pull request.

Feel free to make more comments or suggestions.

nilsphysics:
I added the ice model from (https://arxiv.org/pdf/2212.10285.pdf) it should be fully usable in the analyticalraytracer.

interpolation_greenland
icemodel_greenland_zoom
icemodel_greenland

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1833,11 +1855,385 @@ def get_path(self, iS, n_points=1000):
...
+            [2] - ePhi      - final frequency spectrum of the phi component
+        """
+
+        t_fast = base_trace.BaseTrace()
+
+        ice_n = self.__medium
+        ice_birefringence = medium.get_ice_model('birefringence_medium')
+        ice_birefringence.__init__(bire_model)
+
+        acc = int(self.get_path_length(i_solution))
```
cg-laser, line 2091: implicit assumption that the base unit is meters. Fix this by adding a `/units.m`

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1833,11 +1855,385 @@ def get_path(self, iS, n_points=1000):
...
+
+            R = np.matrix([[a, b], [c, d]])
+
+            birefringent_base = R * pulse[1:] 
+
+            t_fast.set_frequency_spectrum(birefringent_base[1], sampling_rate=samp_rate)
+            t_fast.apply_time_shift(t_1 - t_0)
+            birefringent_base[1] = t_fast.get_frequency_spectrum()
+            
+            Rinv = np.linalg.inv(R)
```
cg-laser, line 2126: I think this can be sped up by realizing that the matrix is orthogonal and, therefore, the inverse is equal to the transpose of the matrix. I don't know how np.linalg.inv is implemented but it likely uses a numerical approach which should be much slower than a simple transpose operation.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -2198,6 +2597,51 @@ def apply_propagation_effects(self, efield, i_solution):
         if self._config['propagation']['focusing']:
             focusing = self.get_focusing(i_solution, limit=float(self._config['propagation']['focusing_limit']))
             spec[1:] *= focusing
+        
+        # apply the birefringence effect
+        if self._config['propagation']['birefringence']:
+
```
cg-laser, line 2603: I would remove the blank lines

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -2198,6 +2597,51 @@ def apply_propagation_effects(self, efield, i_solution):
         if self._config['propagation']['focusing']:
             focusing = self.get_focusing(i_solution, limit=float(self._config['propagation']['focusing_limit']))
             spec[1:] *= focusing
+        
+        # apply the birefringence effect
+        if self._config['propagation']['birefringence']:
+            
+            if "birefringence_model" in self._config['propagation']:
```
cg-laser, line 2604: I don't understand this if statement. Shouldn't there always be a default birefringence model be set? If the user hasn't specified it, then it the one from the `default_config.yaml` should be taken. 
This code implicitly sets a default birefringence model. If a birefringence model is defined in `default_config.py`, then it is not necessary to check if it exists in the config and this code can be simplified.

NuRadioMC/SignalProp/examples/birefringence_examples/01_simple_propagation.py
```diff
@@ -0,0 +1,125 @@
...
+from NuRadioMC.utilities import medium
+from NuRadioReco.utilities import units
+import NuRadioReco.framework.electric_field
+import matplotlib.pyplot as plt
+import numpy as np
+
+from NuRadioMC.SignalProp import radioproparaytracing
+from NuRadioMC.SignalProp import analyticraytracing
+
+###-----------------------------------------
```
cg-laser, line 11: use doc string, i.e. `"""`

NuRadioMC/SignalProp/examples/birefringence_examples/01_simple_propagation.py
```diff
@@ -0,0 +1,125 @@
...
+config = {'propagation': {}}
+config['propagation']['attenuate_ice'] = True
+config['propagation']['focusing_limit'] = 2
+config['propagation']['focusing'] = False
+
+# ---------------   raytracing:            analyticraytracing.py ---------------------
+# ---------------   e-field propagation:   analyticraytracing.py ---------------------
+# ---------------   biefringence model:    'southpole_A' ---------------------
+
+config['propagation']['birefringence'] = True
```
cg-laser, line 73: better always set the birefringence model explicitly as below.

NuRadioMC/SignalProp/examples/birefringence_examples/02_path_info.py
```diff
@@ -0,0 +1,118 @@
+from NuRadioMC.SignalProp import propagation
+from NuRadioMC.utilities import medium
+from NuRadioReco.utilities import units
+import NuRadioReco.framework.electric_field
+import matplotlib.pyplot as plt
+import numpy as np
+
+from NuRadioMC.SignalProp import analyticraytracing
+
+###-----------------------------------------
```
cg-laser, line 10: use doc string `"""`

NuRadioMC/simulation/config_default.yaml
```diff
@@ -30,6 +30,9 @@ propagation:
   focusing: False  # if True apply the focusing effect.
   focusing_limit: 2  # the maximum amplification factor of the focusing correction
   n_reflections: 0  # the maximum number of reflections off a reflective layer at the bottom of the ice layer
+  birefringence: False  # if True apply birefringence effects
+  birefringence_propagation: 'analytical'  # can also be 'numerical' - Does not have to be set, default value is 'analytical'
+  birefringence_model: 'A'  # can also be 'B', 'C', 'D', 'E' - Does not have to be set, default value is 'A'
```
cg-laser, line 35: can you add a link to the documentation, or in general where to find out what A B C etc. mean

NuRadioMC/test/SignalProp/run_signal_test.sh
```diff
@@ -5,6 +5,7 @@ python3 T02test_analytic_D_T.py
 python3 T04MooresBay.py
 python3 T05unit_test_C0_SP.py
 python3 T06unit_test_C0_mooresbay.py
+python3 T07test_birefringence.py
```
cg-laser, line 8: also add the execution of all birefringence examples to this script.

cg-laser: the examples don't need to be validated, but running them ensures that they keep up to date

NuRadioMC/utilities/birefringence_models/IceModel_interpolation_greenland.py
```diff
@@ -0,0 +1,200 @@
+######
+
+#This file is meant to give a guide for creating spline tables that can be used for the birefringence propagation functions.
```
cg-laser, line 3: use doc string `"""`

NuRadioMC/utilities/birefringence_models/IceModel_interpolation_southpole.py
```diff
@@ -0,0 +1,186 @@
+######
+
```
cg-laser, line 2: use doc strings

NuRadioMC/SignalProp/examples/birefringence_examples/04_ARIANNA_SPice.py
```diff
@@ -0,0 +1,148 @@
+from NuRadioMC.utilities import medium
+from NuRadioReco.utilities import units
+import NuRadioReco.framework.electric_field
+import matplotlib.pyplot as plt
+import numpy as np
+from scipy import signal
+from NuRadioMC.SignalProp import analyticraytracing
+
+###-----------------------------------------
```
cg-laser, line 9: use doc string

NuRadioMC/SignalProp/examples/birefringence_examples/05_Veff_comparison.py
```diff
@@ -0,0 +1,85 @@
+from __future__ import absolute_import, division, print_function
```
cg-laser, line 1: how long does it take to execute this example?

cg-laser (changes requested):
Good progress! I found a few more things that I commented "in-line". 

The SPICE core pulse files are huge!!! It must be able to save them in a more compressed way. 

A few things can be further improved:
- write documentation for "sphinx homepage"
- add the spice pulses as pulse model to  `SignalGen/emitter.py`
- add another example of a pulse simulation that used NuRadioMC to run the full simulation. See `examples/05...`


I list a few areas of future improvement so that we don't forget about it:
- the `get_pulse_propagation_birefringence` function should implement a user-definable adaptive propagation. A criteria could be the (rate of) change in the birefringence parameters (N1, N2).

NuRadioMC/SignalProp/analyticraytracing.py
nilsphysics: I wasn't sure about the best way to implement this so it's quite possible this can be done in a better way. The idea was to have a running system even if the user only sets `birefringence = True` . However, as you say the code can be simplified a lot if we assume there will always be a  `birefringence_model` and a `birefringence_propagation` set in the config file because I added them to `default_config.yaml`.

cg-laser: you can safely assume that if you added it to the default_config.yaml

NuRadioMC/SignalProp/analyticraytracing.py
nilsphysics: Done

NuRadioMC/SignalProp/analyticraytracing.py
nilsphysics: Done, updated special cases.

NuRadioMC/SignalProp/analyticraytracing.py
nilsphysics: Done

NuRadioMC/SignalProp/analyticraytracing.py
nilsphysics: Done

NuRadioMC/SignalProp/examples/birefringence_examples/01_simple_propagation.py
nilsphysics: Done

NuRadioMC/SignalProp/examples/birefringence_examples/01_simple_propagation.py
nilsphysics: Done

NuRadioMC/SignalProp/examples/birefringence_examples/02_path_info.py
nilsphysics: Done

NuRadioMC/SignalProp/examples/birefringence_examples/04_ARIANNA_SPice.py
nilsphysics: Done

NuRadioMC/utilities/birefringence_models/IceModel_interpolation_southpole.py
nilsphysics: Done

NuRadioMC/utilities/birefringence_models/IceModel_interpolation_greenland.py
nilsphysics: Done

NuRadioMC/SignalProp/examples/birefringence_examples/05_Veff_comparison.py
nilsphysics: about 90s on my laptop

NuRadioMC/test/SignalProp/run_signal_test.sh
nilsphysics: Some of the examples use radiopropa but the unit tests fail when trying to import it, is there a good way to solve this?

cg-laser: We could add s separate test class that installs RadioPropa... but I'm not sure if this is worth it.

NuRadioMC/SignalGen/emitter.py
```diff
@@ -130,8 +130,7 @@ def get_time_trace(amplitude, N, dt, model, full_output=False, **kwargs):
         trace[1, N // 2] = (1.0 - kwargs.get("polarization", 0.5)) ** 0.5 * amplitude
         trace[2, N // 2] = kwargs.get("polarization", 0.5) ** 0.5 * amplitude
     elif(model == "efield_idl1_spice"):
-        launch_zenith, _ = hp.cartesian_to_spherical(*kwargs["launch_vector"])
-        iN = None
+        launch_zenith, _ = np.rad2deg(hp.cartesian_to_spherical(*kwargs["launch_vector"]))
```
cg-laser, line 133: this seems like a bug. The default units should be used always. I you really need to convert to degree, didive by `units.deg`. The `launch_angle` is already converted to `degree` when the file is opened.

nilsphysics: fixed

NuRadioMC/test/SignalProp/run_signal_test.sh
nilsphysics: fixed

NuRadioMC/simulation/config_default.yaml
nilsphysics: done

cg-laser (approved):

NuRadioMC/SignalGen/emitter.py
```diff
@@ -35,30 +44,42 @@ def get_time_trace(amplitude, N, dt, model, full_output=False, **kwargs):
...
+            trace = np.zeros(N)
     if(model == 'delta_pulse'):  # this takes delta signal as input voltage
         trace = np.zeros(N)
         trace[N // 2] = amplitude
     elif(model == 'cw'):  # generates a sine wave of given frequency
         time = np.linspace(-(N / 2) * dt, ((N - 1) - N / 2) * dt, N)
-        trace = amplitude * np.sin(2 * np.pi * emitter_frequency * time)
+        trace = amplitude * np.sin(2 * np.pi * kwargs["emitter_frequency"] * time)
     elif(model == 'square' or model == 'tone_burst'):  # generates a rectangular or tone_burst signal of given width and frequency
+        half_width = kwargs.get("half_width")
```
sjoerd-bouma, line 82: Consider raising a more helpful error if a kwarg is expected but not specified. Also, document the kwargs in the docstring, either under Parameters or Other Parameters.

NuRadioMC/SignalGen/emitter.py
```diff
@@ -69,10 +90,10 @@ def get_time_trace(amplitude, N, dt, model, full_output=False, **kwargs):
         if(model == 'square'):
             trace = voltage
         else:
-            trace = voltage * np.sin(2 * np.pi * emitter_frequency * time)
+            trace = voltage * np.sin(2 * np.pi * kwargs["emitter_frequency"] * time)
     elif(model == 'gaussian'):  # generates gaussian pulse where half_width represents the half width at half maximum
         time = np.linspace(-(N / 2) * dt, ((N - 1) - N / 2) * dt, N)
-        sigma = half_width / (np.sqrt(2 * np.log(2)))
+        sigma = kwargs["half_width"] / (np.sqrt(2 * np.log(2)))
```
sjoerd-bouma, line 96: see above

NuRadioMC/SignalGen/emitter.py
```diff
@@ -105,6 +126,85 @@ def get_time_trace(amplitude, N, dt, model, full_output=False, **kwargs):
         trace = amplitude * trace / np.max(np.abs(trace))  # trace now has dimension of amplitude given from event generation file
         peak_amplitude_index_new = np.where(np.abs(trace) == np.max(np.abs(trace)))[0][0]
         trace = np.roll(trace, int(N / 2) - peak_amplitude_index_new)  # this rolls the array(trace) to keep peak amplitude at center
+    elif(model == 'efield_delta_pulse'):  # this takes delta signal as input voltage
+        trace = np.zeros((3, N))
+        trace[1, N // 2] = (1.0 - kwargs.get("polarization", 0.5)) ** 0.5 * amplitude
+        trace[2, N // 2] = kwargs.get("polarization", 0.5) ** 0.5 * amplitude
```
sjoerd-bouma, line 132: see above

sjoerd-bouma (commented):
First of all, thanks a lot for this huge contribution which will be great to have in NuRadioMC! It looks like it's been a lot of work... I don't have the time right now to test everything in detail, so (fortunately or unfortunately) most of my comments are mostly aesthetic / documentation related - I imagine between Christian and yourself you've tested everything extensively. It might be worth adding (one of) the birefringence tests to the unit tests as well?

With regards to the level of my comments - I nitpick a lot, so I can usually live with not everything being implemented. On the other hand, some of your docstrings **will** not render properly online (blame rst, not me...), which means either the documentation will stay messy forever or someone has to fix it eventually. So all the fighting you do with sphinx now is greatly appreciated by someone (possibly me) in the future!

NuRadioMC/utilities/birefringence_models/model_description
```diff
@@ -0,0 +1,25 @@
+The Ice data stored in eps.csv was taken from: Voigt, D. E. (2017) "c-Axis Fabric of the South Pole Ice Core, SPC14" U.S. Antarctic Program (USAP) Data Center. doi: https://doi.org/10.15784/601057. The calculation of how to convert the data into refractive indices was taken from: Jordan, T., Besson, D., Kravchenko, I., Latif, U., Madison, B., Nokikov, A., & Shultz, A. (2020). Modeling ice birefringence and oblique radio wave propagation for neutrino detection at the South Pole. Annals of Glaciology, 61(81), 84-91. doi:10.1017/aog.2020.18.
```
fschlueter, line 1: Has this file an file extension (its not shown to me). If not can you add one?

fschlueter:
Hi, you are adding quite a number of binary "data" files with this PR. My preliminary understanding is that they are either files which contain data to run examples/tests or files containing ice models for the birefringence simulations (are those files containing splines which could be created on the fly?). I have not checked how large the files are, is there any big file (couple of MBs?). Maybe it is worth thinking of a different solution (e.g., downloading on the fly) for a couple of those files?

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1,20 +1,31 @@
 from __future__ import absolute_import, division, print_function
 import numpy as np
 import copy
-from scipy import optimize, integrate, interpolate
+from scipy import optimize, integrate, interpolate, signal
 import scipy.constants
 from operator import itemgetter
 import NuRadioReco.utilities.geometryUtilities
+#from torch.fx.experimental.fx2trt.converters.acc_ops_converters import acc_ops_dequantize
```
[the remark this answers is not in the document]
nilsphysics: fixed

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1829,6 +1841,11 @@ class describing the index-of-refraction profile
         self.__logger = logging.getLogger('ray_tracing_analytic')
         self.__logger.setLevel(log_level)
 
+        
+        self.__medium = medium
+        
+
+
```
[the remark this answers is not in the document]
nilsphysics: fixed

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1972,11 +1991,407 @@ def get_path(self, iS, n_points=1000):
...
+            the index of refraction in the x-direction
+        ny: float
+            the index of refraction in the y-direction
+        nz: float
+            the index of refraction in the z-direction
+
+        Returns
+        -------
+        output format: numpy.array: np.array([px, py, pz])
+        meaning:       normalized e-field vector in cartesian coordinates
```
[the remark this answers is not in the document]
nilsphysics: fixed

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1972,11 +1991,407 @@ def get_path(self, iS, n_points=1000):
...
+        the index of refraction in the x-direction
+        ny: float
+        the index of refraction in the y-direction
+        nz: float
+        the index of refraction in the z-direction
+
+        Returns
+        -------
+        output format: np.array([n1, n2])
+        meaning: effective refractive indices
```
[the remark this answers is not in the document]
nilsphysics: fixed

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1972,11 +1991,407 @@ def get_path(self, iS, n_points=1000):
...
+            the index of refraction in the x-direction
+        ny: float
+            the index of refraction in the y-direction
+        nz: float
+            the index of refraction in the z-direction
+
+        Returns
+        -------
+        output format: numpy.array: np.array([[p_radial_1, p_theta_1, p_phi_1], [p_radial_2, p_theta_2, p_phi_2]])
+        meaning:       normalized e-field vector in spherical coordinates for both birefringence solutions
```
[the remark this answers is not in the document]
nilsphysics: fixed

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1972,11 +1991,407 @@ def get_path(self, iS, n_points=1000):
...
+            azimuth angle of the propagation direction
+        polarization: np.array([px, py, pz])
+            normalized e-field vector in cartesian coordinates
+
+        Returns
+        -------
+        output format: numpy.array
+            np.array([p_radial_1, p_theta_1, p_phi_1])
+        meaning       
+            normalized e-field vector in spherical coordinates
```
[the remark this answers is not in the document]
nilsphysics: fixed

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1972,11 +1991,407 @@ def get_path(self, iS, n_points=1000):
...
+        The important properties include effective refractive indices, polarization eigenvectors, and incremental time delays
+
+        Parameters
+        ----------
+
+        i_solution: int
+        choose which ray-tracing solution should be propagated
+        bire_model: string
+        choose the interpolation to fit the measured refractive index data
+        options include (A, B, C, D, E) description can be found under: NuRadioMC/NuRadioMC/utilities/birefringence_models/model_description
```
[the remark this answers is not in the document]
nilsphysics: fixed

nilsphysics:
Hi Sjoerd and Felix,

thanks for reviewing the pull request and leaving comments! I'll try to fix all the details in my next commits.

@fschlueter Regarding the data files, they contain the spline parameters for a specific model. When needed the spline function is the created using these parameters on the fly. This allows the files to be quite small (~1kB).

NuRadioMC/SignalGen/emitter.py
```diff
@@ -35,30 +44,42 @@ def get_time_trace(amplitude, N, dt, model, full_output=False, **kwargs):
         * idl1 & hvsp1 : these are the waveforms generated in KU lab and stored in hdf5 files
         * gaussian : represents a gaussian pulse where sigma is defined through the half width at half maximum
         * ARA02-calPulser : a new normalized voltage signal which depicts the original CalPulser shape used in ARA-02
+        * efield_idl1_spice: direct measurement of the efield from the idl1 pulser and its antenna as used in the SPICE
```
cg-laser, line 47: please clarify in the comment that the original in-air measurement was corrected to in-ice emission with n=1.78 according to f = f_air / n (see appendix 1.4 of the NuRadioReco paper)https://link.springer.com/article/10.1140/epjc/s10052-019-6971-5#appendices
and double-check that this was the case.

cg-laser:
I agree, adding birefringence to unit tests would be good. I suggest running a neutrino simulation (without noise and fixed input hdf5 file with 5 events that trigger, a deterministic Askaryan model) and comparing the resulting voltage traces for equality (almost equal).

nilsphysics:
There is a unit test for birefringence checking that the voltage traces are calculated correctly for different geometries.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1972,11 +1991,407 @@ def get_path(self, iS, n_points=1000):
...
+            ['refractive_index_x']:         np.array - refractive index for the x-direction
+            ['refractive_index_y']:         np.array - refractive index for the y-direction
+            ['refractive_index_z']:         np.array - refractive index for the z-direction
+            ['first_refractive_index']:     np.array - effective refractive index of the first birefringent state along the full path
+            ['second_refractive_index']:    np.array - effective refractive index of the second birefringent state along the full path
+            ['first_polarization_vector']:  np.array - polarization vector of the first birefringent state in spherical coordinates along the full path
+            ['second_polarization_vector']: np.array - polarization vector of the second birefringent state in spherical coordinates along the full path
+            ['first_time_delay']:           np.array - incremental time delays of the first birefringent state along the full path
+            ['second_time_delay']:          np.array - incremental time delays of the second birefringent state along the full path
```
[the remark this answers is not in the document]
nilsphysics: fixed

NuRadioReco/framework/event.py
```diff
@@ -416,6 +419,69 @@ def has_sim_shower(self, shower_id=None):
...
+    def get_first_sim_emitter(self, ids=None):
+        """
+        Returns only the first sim emitter stored in the event. Useful in cases
+        when there is only one emitter in the event.
+
+        Parameters
+        ----------
+        ids: list of integers
+            A list of station IDs. The first emitter that is associated with
+            all stations in the list is returned
```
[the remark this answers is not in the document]
nilsphysics: fixed

NuRadioMC/SignalGen/emitter.py
nilsphysics: fixed

NuRadioMC/SignalGen/emitter.py
nilsphysics: fixed

NuRadioMC/SignalGen/emitter.py
nilsphysics: fixed

cg-laser:
continued in #650

PR #333, opened 2021-10-20
Cr flux

lpyras:
I added data and functions to get the cr flux. So far it is placed in the utilities folder.


cg-laser (dismissed):
looks great. Thanks

NuRadioReco/utilities/cr_flux.py
```diff
@@ -0,0 +1,261 @@
...
+
+    Returns
+    -------
+    scipy integration of data in NuRadio base units (which are 1/(eV m^2 sr ns))
+    """
+    if type in ['data_auger_19', 'data_TA_21']:
+        E, J, J_err_low, J_err_up = get_measured_data(type)
+        flux_interpolation = get_flux_interpolation(type)
+        integrated_flux = quad(flux_interpolation, log10e_min, log10e_max,
+                               limit=2 * E.shape[0], points=E)
```
shallmann, line 168: ... can you check if this is still correct (sorry for making you go from dE->dlogE in the interpolation, but:
dlogE/dE = 1/E, meaning if you change the integral from integrating over E to integrating over logE, you will have to likely multiply by E, because dE = E*dlogE, no?

NuRadioReco/utilities/cr_flux.py
```diff
@@ -0,0 +1,261 @@
...
+        upper zenith for zenith range in deg
+    a_eff:
+        effective area of detector
+    type: string
+        choose between auger_17, auger_19 and TA_19
+    Returns
+    -------
+    eventrate
+    """
+    projected_area = 0.5 * (np.cos(zenith_min * units.deg) + np.cos(zenith_max * units.deg))
```
shallmann, line 206: not sure if this is a safe way. if you are calculating for a large range in zenith (the default) you are just taking the middle.... whereas if you calculate over many small bins and sum, you likely will not get the same result.

shallmann: or will you? But you need to convince me first ;)

NuRadioReco/utilities/cr_flux.py
```diff
@@ -0,0 +1,261 @@
...
+
+    dOmega = 2 * np.pi * ((1 - np.cos(zenith_max * units.deg)) - (1 - np.cos(zenith_min * units.deg)))
+
+    integrated_flux = get_flux_per_energy_bin(log10e_min, log10e_max, type)
+
+    return integrated_flux * projected_area * dOmega * a_eff
+
+
+def plot_measured_spectrum(ax=None, scale=2.7, type='data_auger_19',
+                           base_units=True):
```
shallmann, line 216: I would by default set it to false. Plotting CR/astrophysical fluxes in units of ns is not often useful.

NuRadioReco/utilities/data/TA_combined_spectrum_ICRC_2021.txt
```diff
@@ -0,0 +1,62 @@
+#
```
shallmann, line 1: :) good to have the CR tables in 👍🏽

shallmann (commented):
1. Can you please check the integral over logE if there is a factor of E missing or not?
2. Can you please check if calculating an average projection over a large zenith range is accurate?

shallmann (changes requested):
please address the two comments I had before merging

NuRadioReco/utilities/cr_flux.py
lpyras: I change that :)

NuRadioReco/utilities/cr_flux.py
lpyras: Thanks for spotting this, I simplified the function so that only the differential for an energy and a zenith is given.

NuRadioReco/utilities/cr_flux.py
lpyras: Nice catch! I will remove the option to get integration from data points since we have the analytic spectrum anyways.

shallmann (approved):
fine with me now. Thanks!

lpyras:
I added auger and TA cr spectrum measurements and the analytic parametrization. Also, cr_flux.py includes some helpful functions to access the flux

PR #337, opened 2021-11-08
Detector Browser markers

christophwelling:
Updates markers specifying the different antenna types


shallmann (approved):
IMO this is good to merge

PR #339, opened 2021-11-09
add option to correct for cable delays in channelAddCableDelay

iplaisier:
Adapt channelAddCableDelay module such that it can also be used to correct for cable delays for data.


anelles (approved):
Looks ok to me. I guess it makes things a bit more transparent where to put the -1.

PR #343, opened 2021-11-17
added numba version of get_vector_potential to ARZ module

sjoerd-bouma:
From profiling the ARZ signal generator, I noticed that ~95% of the time used to generate a trace is spent in the get_vector_potential_fast function. I added a numba version of this function to the ARZ module, which will be used by default if numba is available. The first call to this function will be very slow, as numba will compile the function, but subsequent calls should be a factor of ~3 quicker than the pure Python implementation.

Right now, the numba function is more or less a carbon copy of the Python function, with the following differences:
-  It does not access any internal variables of the ARZ class (e.g. self._Af_p etc.), so these need to be given explicitly. This is done in order to avoid rewriting the whole ARZ class in numba (though theoretically, I guess one could also rewrite the Python method in the same way if we want to keep this symmetric)
- Numba doesn't like objects which it can't type, so the logger seems to not be supported.

I realize it's probably not ideal to have all this code more or less duplicated, but I am not aware of a neater solution (e.g. a numba ```@jit``` decorator that can be turned on/off conditionally). I suppose one could drop the pure Python version completely, but I felt that might be a little bit premature : )

Output of test script which calls ARZ.get_time_trace() 100 times:
```
WARNING:SignalGen.ARZ:setting seed to 1234
WARNING:SignalGen.ARZ:loading shower library (/home/sb/Python/software/nu-radio/NuRadioMC/NuRadioMC/SignalGen/ARZ/shower_library/library_v1.2.pkl) into memory
Numba first run took 12.7 s (includes compilation)
Numba code took 7.9 s (79 ms / trace)
Python code took 25.7 s (257 ms / trace)
```

### DIFF
--- a/.github/workflows/run_tests.yaml
+++ b/.github/workflows/run_tests.yaml
@@ -69,6 +69,12 @@ jobs:
         export PYTHONPATH=$PWD:$PYTHONPATH
         export GSLDIR=$(gsl-config --prefix)
         NuRadioMC/test/SingleEvents/validate_ARZ.sh
+    - name: Test Numba version of ARZ
+      run: |
+        export PYTHONPATH=$PWD:$PYTHONPATH
+        export GSLDIR=$(gsl-config --prefix)
+        pip install numba
+        NuRadioMC/test/SingleEvents/validate_ARZ.sh
     - name: Signal generation test
       run : |
           export PYTHONPATH=$PWD:$PYTHONPATH
--- a/NuRadioMC/SignalGen/ARZ/ARZ.py
+++ b/NuRadioMC/SignalGen/ARZ/ARZ.py
@@ -34,252 +34,249 @@ xmu = 12.566370e-7 * units.newton / units.ampere ** 2
 c = 2.99792458e8 * units.m / units.s
 # e = 1.602177e-19 * units.coulomb
 
-if numba_available:
-    @jit(nopython=True)
-    def get_vector_potential_numba(shower_energy, theta, N, dt, profile_depth, profile_ce,
-                                Af, freq_pos, freq_neg, exp_pos, exp_neg, t0_pos, t0_neg,
-                                shower_type="HAD", n_index=1.78, distance=1 * units.m,
-                                interp_factor=1., interp_factor2=100., shift_for_xmax=False,
-                                em_factor=1.
-                                ):
+def get_vector_potential(
+    shower_energy, theta, N, dt, profile_depth, profile_ce,
+    Af, freq_pos, freq_neg, exp_pos, exp_neg, t0_pos, t0_neg,
+    shower_type="HAD", n_index=1.78, distance=1 * units.m,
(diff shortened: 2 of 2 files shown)


cg-laser:
awesome! A factor of 3 is great. 

I agree that code duplication (and no logging) is not ideal. I don't really get why the logger is not working, it seems like this would be a standard python thing to do. Let's discuss it in the simulation call if a better way of implementing exists. We can of course also make the default `get_vector_potential` function a non-member function. Would this work? 

Did you check that the output agrees perfectly? Can we add the numba version to our test suite? I think you can just copy the existing ARZ test and install numba only for this test.

sjoerd-bouma:
> I agree that code duplication (and no logging) is not ideal. I don't really get why the logger is not working, it seems like this would be a standard python thing to do. Let's discuss it in the simulation call if a better way of implementing exists. We can of course also make the default `get_vector_potential` function a non-member function. Would this work?
> 
> Did you check that the output agrees perfectly? Can we add the numba version to our test suite? I think you can just copy the existing ARZ test and install numba only for this test.

The logger is only turned off inside get_vector_potential_fast, not sure why it does not work, but I didn't think it was significant enough to worry about (errors are still raised correctly, logger messages in the rest of the ARZ module work fine). I will see if I can fix it - I guess if I can, and I can figure out how to conditionally decorate a function, than just moving get_vector_potential_fast out of the ARZ class would work.

I of course checked that the output agrees : ) both the tests (validate_ARZ & SignalGen/test_build) and np.allclose() on a couple of examples are happy.

sjoerd-bouma:
So with the loss of a grand total of 2 logger.info messages, the Python and Numba implementations are now the same, with the only difference being that Numba will compile the function and run much faster. Numba will now also try to 'cache' the compiled function, so in the use case of many different scripts that call ARZ only a couple of times, you still get the speed boost.

For now, I've left the .get_vector_potential and .get_vector_potential_fast methods in the ARZ class alone - the former was already never used. I guess I could remove them (there are a bunch of debugging plots in the code, but I suppose it is _unlikely_ that someone would want to use those in the future).

anelles:
Looks in principle good to me. Should you add a require numba to the project to actually be able to see whether the results are the same on the tests? Right now, it would never use numba due to the checking, no? 

I am a bit torn on leaving the old functions in there: 
- Pro: People like it, if the code is self-documenting. So clearly adding something in the doc string that says we keep the function here for being able to see what was done previously, but advising against using -- will probably be appreciated by many scientists.
- Con: I we all agree that the new is equally good but fast. It really is only cluttering of the code.

anelles (approved):
Looks good to me now. I trust that the tests pass.

cg-laser (approved):
looks good to me as well. 

Please leave the "slow" function in the code. It is a more straightforward (easier to understand) implementation of the method which is also more precise than the highly optimized `fast` version of the function.

PR #344, opened 2021-11-17
changed pyproject.toml to poetry, added installation script for devs

sjoerd-bouma:
At Christian's request, I've converted the pyproject.toml to now be buildable by poetry instead of flit. This means NuRadioMC/NuRadioReco can be made pip-installable with the simple command ```poetry build```. Additionally, I've added an interactive Python script for developers ```install_dev.py``` which:
- Installs all NuRadioMC dependencies (i.e. similar to ```pip install -r requirements.txt```)
- Adds ```NuRadioMC``` to the ```PYTHONPATH``` in ```.bashrc```, if it is not already found in ```PYTHONPATH```
- Adds a '[pre-commit hook](https://githooks.com/)' to ```.git/hooks``` that blocks any commits that include files over 5 MB
Finally, the version number for both NuRadioMC and NuRadioReco is stored only in one location - pyproject.toml. When calling `NuRadioReco.__version__`, Python first tries to obtain the version number from pyproject.toml in the parent directory. If this is not available (i.e. the user is using the pip-installed version), the version number of the pip installed package ```NuRadioMC``` is used instead.

Feedback is welcome!
Some notes, in no particular order:
- I go through the trouble of converting pyproject.toml to pip-style requirements in install_dev.py, because poetry **locks** requirements; i.e. if one installs dependencies using poetry, it will **always** pick a unique version number for each dependency, and then force the user to upgrade/downgrade. While this seems to be universally recommended for software developers, it is different from what I/we are used to. Changing this is straightforward if we should want to.
- Again Christian requested that I exclude proposal from the poetry requirements. I have not yet found a good way to include optional requirements (requirements that poetry recognises as optional will end up as forced requirements in the pip installable version), but I feel that we should at least have an overview of the optional requirements somewhere, particularly if we insist on specific versions (as with proposal). ~The easiest thing might be an additional optional_requirements.txt?~ Currently, optional dependencies are included as dev-dependencies - the `install_dev.py` script now asks which of these to install.
- On the to do list: ~adding a git workflow that automatically builds & pip releases NuRadioMC when a new release is made, as well as updating the installation instructions.~ Done - installation instructions can / will be found at https://nu-radio.github.io/NuRadioMC/Introduction/pages/installation.html.


anelles:
Looks like a nice addition. 
Just a couple of comments from my side: 
- 5 MB still seems pretty large. Do we have large files like this?
- It says (and had been saying Christian Glaser as author): does this require a natural person, otherwise it seems odd to just have one instead of all developers
I am happy to try this out, but I have no time at the moment. So I am hoping someone else goes first.

sjoerd-bouma:
> Looks like a nice addition. Just a couple of comments from my side:
> 
>     * 5 MB still seems pretty large. Do we have large files like this?
> 
>     * It says (and had been saying Christian Glaser as author): does this require a natural person, otherwise it seems odd to just have one instead of all developers
>       I am happy to try this out, but I have no time at the moment. So I am hoping someone else goes first.

* There are a couple  of files around / over 1 MB, mostly reference files for the tests (one is 6.9 MB), the ARZ shower library is also slightly over 1 MB. I can change the limit to be lower, of course - end users can always circumvent it by just deleting/renaming `.git/hooks/pre-commit` (I can make it print a line to make users aware of this).
* I've changed Christian Glaser -> Christian Glaser et al.; this seems to be what [numpy](https://pypi.org/project/numpy/) does. I can also add a list of maintainers, but it feels like that might be overkill

@cg-laser - I've added a workflow to publish to pypi (on pushes to **master**), and also to push to test-pypi on pushes to **develop** - the latter I can of course remove, but I thought it might be nice to see if it actually works. You will need to add API tokens for both of these - see instructions [here](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#saving-credentials-on-github).

Once the documentation has been merged & updated, I think this PR will be good to go.

sjoerd-bouma:
I think this is done now. We can of course argue a bit about core and optional dependencies still!
@cg-laser see previous comment for how to enable the automatic publishing to pypi & test-pypi.

install_dev.py
```diff
@@ -0,0 +1,255 @@
+import os
+import sys
+import shutil
+import subprocess
+import logging
+import tempfile
+import argparse
+logging.basicConfig()
+
```
cg-laser, line 9: can you add a doc string here explaining briefly what this script does?

.github/workflows/run_tests.yaml
```diff
@@ -29,8 +29,9 @@ jobs:
         python -m pip install --upgrade pip
         pip install flake8 pytest
         export GSLDIR=$(gsl-config --prefix)
-        if [ -f NuRadioMC/test/requirements.txt ]; then pip install -r NuRadioMC/test/requirements.txt; fi
-        export PYTHONPATH=$PWD:$PYTHONPATH
+        python install_dev.py --install --dev proposal --no-interactive
```
cg-laser, line 32: not ideal because it would be good if the "non python dependencies" like proposal would only be tested in the respective test. 
Can't you just call the `install_dev` script again with `--dev proposal` to install proposal only in the test that needs it? Then we can do the same for RadioPropa and other future optional C dependencies

.github/workflows/run_tests.yaml
```diff
@@ -29,8 +29,9 @@ jobs:
...
+        pip uninstall numba -y # easiest way to test ARZ without numba
```
cg-laser, line 33: I assume this is required because numba is a requirement in the install_dev.py script?

cg-laser (commented):
loooks good, awesome PR. Just a small number of comments

.github/workflows/run_tests.yaml
sjoerd-bouma: Yes - we should probably decide whether `numba` and `uproot` should be core dependencies or optional.
If optional, we keep NuRadioMC 'smaller', but a pip install will not install optional dependencies. A missing `numba` install will not result in errors, but just in a slower ARZ signal generation, which 'casual' users might never catch. Uproot on the other hand is only required for specific experiments (as far as I understand?) and will generate a simple `ImportError` if the user tries to use modules that depend on it. Hence why `uproot` is currently optional, and `numba` is core. But I am happy to change this.

.github/workflows/run_tests.yaml
sjoerd-bouma: Sure. My logic was that proposal is only used in modules that import it anyway, so it's unlikely to affect the other tests?

install_dev.py
sjoerd-bouma: Will add something. I do feel it's already 'documented': it's an interactive script (try running it with no options -  it will always ask before doing anything), optional command line arguments are also explained by `python install_dev.py --help` (as is standard), and comments in the code explain which each part does.

.github/workflows/run_tests.yaml
cg-laser: I'm worried that the proposal installation might break at some point. And it takes a while which slows down the initial tests.

sjoerd-bouma: If proposal breaks, it will break the tests anyway : ) And in terms of time it's not that slow - proposal 6.1.6 is cached, so it adds less than 20 seconds to the installation.

install_dev.py
cg-laser: then just write "Interactive script to install NuRadioMC/Reco and its dependencies, for more information call `python install_dev.py --help`"
 ;-)

.github/workflows/run_tests.yaml
cg-laser: to me that looks good, just wanted to make sure I understand it.

install_dev.py
sjoerd-bouma: Added a docstring, which can be displayed using
```
import install_dev
help(install_dev)
```

.github/workflows/run_tests.yaml
sjoerd-bouma: Both uproot and numba are now core dependencies again. Christian's comment below for completeness:
>I think numba is fine, uproot too. These are well maintained packages. And we always want numba because of the speed gain

.github/workflows/run_tests.yaml
sjoerd-bouma: Proposal installation now happens for the first test that actually uses proposal.

cg-laser (approved):

cg-laser (approved):

PR #345, opened 2021-11-19
Stationnumber bug detectordescription

iplaisier:
there were still three bugs in the detector description for the deployed stations of RNO-G. 
1. channel 5,6,7 of station 11 were labeled as station 22
2. the antenna positions of 6 and 7 were unchanged 
3. channel number 21 was duplicated and therefore not recognized
Hope everything is good now.


sjoerd-bouma:
Looks like you swapped channels 21 <-> 23 again on all stations - I thought we had decided that 21 is the HPol, and 23 is the VPol?

iplaisier:
Yes we have decided on that. I didn't do this on purpose, but this went wrong by copy-ing over the file :( 
It's good we only have one description now. I'll change it.

sjoerd-bouma (approved):
Looks good now! Thanks for fixing this.
On a side note, there seems to be a Matplotlib deprecation warning coming from the sphericalwavefitter when I run the reconstruction example - but that's probably something for another time.

PR #346, opened 2021-11-23
Feature/allow multiple reference stations

lpyras:
Change genericDetector in a way, that two or more default stations are possible by adding reference_station in the detector description. This is necessary to create the gen_2 detector layout.


lpyras:
So far, the station id format is not correct.

NuRadioReco/detector/generic_detector.py
```diff
@@ -196,20 +219,23 @@ def add_generic_station(self, station_dict):
             logger.warning('Station with ID {} already exists in buffer. Cannot add station with same ID'.format(
                 station_dict['station_id']))
             return
-
-        for key in self.__default_station.keys():
-            if key not in station_dict.keys():
-                station_dict[key] = self.__default_station[key]
+        if 'reference_station' in station_dict:
```
shallmann, line 222: the if "reference_station" should be removed... it might not be set...

cg-laser:
@lpyras is this ready to go from your side? Because it is still a draft

NuRadioReco/detector/generic_detector.py
```diff
@@ -28,8 +29,9 @@ class GenericDetector(NuRadioReco.detector.detector.Detector):
     This detector only accepts json detector descriptions or dictionary.
     """
 
-    def __init__(self, json_filename, default_station, default_channel=None, default_device = None, source='json', dictionary=None,
-                 assume_inf=True, antenna_by_depth=True):
+    def __init__(self, json_filename, default_station=None, default_channel=None, default_device=None, source='json',
+                 dictionary=None,
+                 assume_inf=True, antenna_by_depth=False):
```
cg-laser, line 34: I see that you changed the default of `antenna_by_depth`. Great. I was just about to suggest that ;-)

cg-laser (approved):
looks good to me. @christophwelling you know the code best. Does it look good to you?

NuRadioReco/modules/io/NuRadioRecoio.py
```diff
@@ -296,6 +296,11 @@ def get_detector(self):
                     # Detector is a generic detector, so we have to consider default
                     # station/channel and event-specific changes
                     self.__detectors[self._current_file_id] = NuRadioReco.detector.generic_detector.GenericDetector.__new__(NuRadioReco.detector.generic_detector.GenericDetector)
+                    # the use of default_station and default_channel is deprecated. Allow to set it for now, to ensure backward compatibility
+                    if 'default_station' not in detector_dict:
+                        detector_dict['default_station'] = None
+                    if 'default_channel' not in detector_dict:
+                        detector_dict['default_channel'] = None
                     self.__detectors[self._current_file_id].__init__(source='dictionary', json_filename='', dictionary=detector_dict, default_station=detector_dict['default_station'], default_channel=detector_dict['default_channel'])
```
christophwelling, line 304: I don't think you need these lines. You can just not pass any default channels and stations to the generic detector __init__ (and should, because these are no longer supposed to be passed).

NuRadioReco/modules/io/eventWriter.py
```diff
@@ -174,9 +174,9 @@ def __get_detector_dict(self, event, det):
             "channels": {},
             "stations": {}
         }
-        if is_generic_detector:
-            det_dict['default_station'] = det.get_default_station_id()
-            det_dict['default_channel'] = det.get_default_channel_id()
+        #if is_generic_detector:
```
christophwelling, line 177: Please remove these instead of commenting them out.

NuRadioReco/modules/io/event_parser_factory.py
```diff
@@ -93,6 +93,10 @@ def scan_files_2_2(self, iF, current_byte):
                     'stations': {}
                 }
             if is_generic_detector:
+                if 'default_station' not in detector_dict:
```
christophwelling, line 96: Why do you need to explicitly set this to none?

christophwelling (changes requested):
I don't think reading the detector description from a nur file is backward compatible this way. I am fine with that since it was a feature that was not used much, but in one comment you seem to imply that it should be backward compatible.

There is also a syntax error that makes the tests fail.

NuRadioReco/modules/io/NuRadioRecoio.py
christophwelling: Never mind, I get what you are doing now. Looks good to me.

NuRadioReco/modules/io/event_parser_factory.py
shallmann: @christophwelling maybe we can just remove the explicit setting and also ll 100/101 where they are put into the detector dicts? what do you think?

christophwelling: These entries need to exist for NuRadioRecoio to recognise that it is a genericDetector. As long as we want to support older files, this seems to be a decent solution.

christophwelling (approved):

NuRadioReco/modules/io/event_parser_factory.py
shallmann: ok added a comment.

NuRadioReco/modules/io/eventWriter.py
shallmann: done

NuRadioReco/detector/generic_detector.py
shallmann: Thanks go to Lilly

cg-laser (approved):

PR #353, opened 2021-12-08
Fix NuRadioReco being overwritten in tests

### LINKED ISSUE #351: Install dependencies in run_tests.yaml overwrites (?) code actually in commit
The automatic tests in the github workflow run_tests.yaml overwrite code in the NuRadioReco module with code from a zip archive:
```
wget https://github.com/nu-radio/NuRadioReco/archive/refactor-pa.zip -O /tmp/NuRadioReco.zip
unzip /tmp/NuRadioReco.zip
mv NuRadioReco-refactor-pa/NuRadioReco $PWD/NuRadioReco
```

This means that instead of testing the code in the pushed commit, it will actually run the tests on code in this archive (which will trivially complete).

I ran into this when testing the documentation. If there is some very good reason to test this archive, I suggest we make a separate test for it. Otherwise, this should just be removed from `run_tests.yaml` ASAP.

sjoerd-bouma:
The unit tests (run_test.yaml) workflow used to overwrite the NuRadioReco module with an older version from a zipped archive, and thus would potentially not test the committed code properly. I've removed this line (as well as some commented out lines) from run_tests.yaml.

Fixes #351

### DIFF
--- a/.github/workflows/run_tests.yaml
+++ b/.github/workflows/run_tests.yaml
@@ -30,20 +30,6 @@ jobs:
         pip install flake8 pytest
         export GSLDIR=$(gsl-config --prefix)
         if [ -f NuRadioMC/test/requirements.txt ]; then pip install -r NuRadioMC/test/requirements.txt; fi
-        #wget https://github.com/nu-radio/radiotools/archive/master.zip -O /tmp/radiotools.zip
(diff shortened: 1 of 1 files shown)


cg-laser (approved):
very good that you fixed it!

Issue #352, opened 2021-12-08
Potential issue in Hilbert envelope?


shallmann:
I was trying to use the get_hilbert_envelope function in the `BaseTrace` class, and didn't return what I expected (i expected `np.array(np.abs(h))`. Instead, only the first three items are returned (probably assumed that one has a 3D trace?), is that intended? Why do we want that?

https://github.com/nu-radio/NuRadioMC/blob/572d4f6382db4928b75581dda117c3f9e90b013d/NuRadioReco/framework/base_trace.py#L129

anelles:
This looks like it was made to return the hilbert envelope for three components of the electric field. So the main axis seems to be wrong for the type of array that you are using this for. This shouldn't happen.

anelles:
@cg-laser actually noticed this already in #304 , so this needs fixing.

shallmann:
Fixed

Issue #304, opened 2021-06-22
hilbert envelop helper broken for channels


cg-laser:
the helper function `get_hilbert_envelope` in `base_trace.py` assumed a 3-dim trace which results in wrong results for the channel object. Example
```
In [2]: channel.get_hilbert_envelope()                                                                                                                    
Out[2]: array([2.78170887e-09, 2.73316328e-09, 2.65892562e-09])
```
where it should be
```
from scipy.signal import hilbert
In [1]: np.abs(hilbert(channel.get_trace()))                                                                                                              
Out[1]: 
array([2.78170887e-09, 2.73316328e-09, 2.65892562e-09, ...,
       2.81679655e-09, 2.79517428e-09, 2.79550081e-09])
```

shallmann:
Fixed

PR #354, opened 2021-12-08
return the hilbert envelope correctly

shallmann:
Small change to correctly return Hilbert envelope as expected. Fixes issues #352 and #304 .

### DIFF
--- a/NuRadioReco/framework/base_trace.py
+++ b/NuRadioReco/framework/base_trace.py
@@ -125,10 +125,12 @@ class BaseTrace:
 
     def get_hilbert_envelope(self):
         from scipy import signal
+        # get hilbert envelope for either 1D (N) analytic trace or (3,N) E-field
         h = signal.hilbert(self.get_trace())
         return np.abs(h)
 
     def get_hilbert_envelope_mag(self):
+        # ensure taking axis 0 of a 2D trace (trace might be (N) for analytic trace or (3,N) for E-field
         return np.linalg.norm(np.atleast_2d(self.get_hilbert_envelope()), axis=0)
 
     def get_number_of_samples(self):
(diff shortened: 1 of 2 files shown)


sjoerd-bouma (dismissed):

cg-laser (changes requested):
the problem is that the trace is sometimes three dimensional (efields) and sometimes 1D (voltages). We need to implement both cases (see the `resample` method for how to do it. )

sjoerd-bouma:
> the problem is that the trace is sometimes three dimensional (efields) and sometimes 1D (voltages). We need to implement both cases (see the `resample` method for how to do it. )

Up to Steffen, I guess, but I don't really see the issue? An (N)-shaped 1D trace returns an (N)-shaped 1D Hilbert envelope, and a (3,N)-shaped 3D E-field trace returns a (3,N)-shaped E-field trace with Steffen's implementation. What might be worth fixing, however, is the `get_hilbert_envelope_mag` method which is meant to reduce the field from 3D to 1D - it should probably be changed to not take the norm over the 0th axis if if len(self.get_hilbert_envelope().shape) == 1

shallmann:
ah @sjoerd-bouma thanks for spotting it. I will double check. Maybe we have to use `axis=-1` there as well

NuRadioReco/framework/base_trace.py
```diff
@@ -129,7 +129,7 @@ def get_hilbert_envelope(self):
         return np.abs(h)
 
     def get_hilbert_envelope_mag(self):
-        return np.linalg.norm(self.get_hilbert_envelope(), axis=0)
+        return np.linalg.norm(self.get_hilbert_envelope(), axis=-1)
```
sjoerd-bouma, line 132: I still don't think this is right. For a 3D trace, this will now take the norm over the trace length and return a (3)-shaped array, for a 1D array it will return a scalar. I still think it shouldn't do anything in 1D, and take the norm over the 3 dimensions for a 3D trace...

shallmann: Ahh... should have paid more attention to what I am doing. You are completely right :)

cg-laser (commented):
very good! I agree the new implementation is better. 

Can you add a comment to the code at the relevant places reminding people that the functions need to work for shape (N) and (3, N), and that therefore the axis keyword is important.

cg-laser (approved):
thanks @shallmann

PR #357, opened 2021-12-11
removed all occurences of deprecated np.int

shallmann:
Issue being addressed:
- replaced all `np.int` by `int` 
- made int-check in `cross_section.py` accept all numpy integer types (not only int64)

Did you remember to: 
* edit the changelog.txt file describing your changes? -> not needed for this minor edit


NuRadioMC/utilities/cross_sections.py
```diff
@@ -92,7 +92,7 @@ def csms(energy, inttype, flavors):
     if type(inttype) == str:
         inttype = np.array([inttype] * energy.shape[0])
 
-    if (type(flavors) == int or type(flavors) == np.int64):
+    if isinstance(flavors, (int, np.integer)):
```
anelles, line 95: That actually is the wrong direction. 
I think it should be no.int_ if we really want the int64 here. 
https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.int_

NuRadioMC/utilities/cross_sections.py
```diff
@@ -231,7 +231,7 @@ def get_nu_cross_section(energy, flavors, inttype='total', cross_section_type='c
         if type(inttype) == str:
             if inttype == 'total':
 
-                if (type(flavors) == int or type(flavors) == np.int64):
+                if isinstance(flavors, (int, np.integer)):
```
anelles, line 234: As above

anelles (changes requested):
Yeah, for grep. Apart from the cross-sections file, I think it is ok.

NuRadioMC/utilities/cross_sections.py
anelles: And all other instances. Comments were strange for me ...

shallmann: @anelles sorry for not explaining better: ... (I think) if you would define the flavors as `np.int32` the previous would fail, so the `isinstance` is more general since it accepts all `np.integer` types

shallmann: now, the check is: if flavors is a standard python int or *any* `np.integer` type.

anelles: Sorry, I copied the check over incorrectly and got another depreciation warning. My fault.

anelles (approved):

lpyras (approved):
Thanks for doing this.

PR #358, opened 2021-12-16
Muon flux

lpyras:
I added this to access the muon flux. The data provided so far is from MCEq. Additional fluxes can be added.


NuRadioMC/EvtGen/generator.py
```diff
@@ -265,7 +265,7 @@ def get_GZK_1(energy):
     and a proton fraction of 10% at E = 10^19.6 eV
     """
     E, J = np.loadtxt(os.path.join(os.path.abspath(os.path.dirname(__file__)),
-                                     '../examples/Sensitivities/ReasonableNeutrinos1.txt'))
+                                     '../examples/Sensitivities/data/ReasonableNeutrinos1.txt'))
```
shallmann, line 268: I think ReasonableNeutrinos1.txt is a really bad name, but am aware that this has been sticking around for some time. Let's leave it for now and rename when we (you) tidy up neutrino flux predictions/models at some point.

NuRadioReco/utilities/muon_flux.py
```diff
@@ -0,0 +1,154 @@
...
+         lower edge of zenith bin over which flux is integrated (in rad)
+    zenith_bin_edge_high: float
+         higher edge of zenith bin over which flux is integrated (in rad)
+
+    Returns
+    -------
+    integrated flux of muon in NuRadio units.
+    '''
+
+    def flux(E, theta):
```
shallmann, line 119: the call to dblquad looks wrong. You are integrating over cos(zenith) but the function is defined as theta angle. Can you check if it is correct and rename variables to something more appropriate if it is indeed correct?

NuRadioReco/utilities/muon_flux.py
```diff
@@ -0,0 +1,154 @@
...
+        return get_flux_per_energy_and_zenith(E, theta)
+
+    int_flux = dblquad(flux, np.cos(zenith_bin_edge_high), np.cos(zenith_bin_edge_low), lambda x: energy_bin_edge_low, lambda x: energy_bin_edge_high)
```
shallmann, line 122: I am referring to this dblquad call in my comment above

NuRadioReco/utilities/muon_flux.py
```diff
@@ -0,0 +1,154 @@
...
+    plt.xscale('log')
+    plt.yscale('log')
+    plt.ylabel(r'Flux data',fontsize=14)
+    plt.xlabel('Energy [eV]',fontsize=14)
+    #plt.show()
+    #plt.close()
+
+    plt.style.use({'figure.facecolor':'white'})
+    plt.plot(1e18, get_flux_per_energy_and_zenith_bin(1e18, 1e19, 0*units.deg, 90*units.deg), marker='x', color='k', label='test 2d')
+    plt.plot(1e18, get_flux_per_energy_bin(1e18, 1e19), marker='+', color='k', label='test 1d')
```
shallmann, line 145: did not try to run the plotting. This is just one data point at 1e18, but you are integrating from 1e18 to 1e19... maybe better put it at 18.5

NuRadioReco/utilities/muon_flux.py
```diff
@@ -0,0 +1,154 @@
...
+
+    plt.legend()
+    plt.xscale('log')
+    plt.yscale('log')
+    plt.ylabel(r'Flux',fontsize=14)
+    plt.xlabel('Energy [eV]',fontsize=14)
+    plt.tight_layout()
+    plt.show()
+    plt.close()
```
shallmann, line 154: (Note the now-newline coding style error...)

shallmann (changes requested):
Thanks for adding this. 
You know about my preference of integrating over logE rather than E, but I assume you have checked that results are reasonable.

Can you just check the cos(zenith) integration behaves as expected and modify or rename variables appropriately please?

NuRadioReco/utilities/muon_flux.py
```diff
@@ -0,0 +1,154 @@
...
+    J_e3 = np.array(data_surface_mu_flux['mu_total']) * units.GeV**2 * units.cm**-2 * units.s**-1 *units.sr**-1
+    E_data = np.array(data_surface_mu_flux['e_grid']) * units.GeV
+
+    plt.plot(E_data, J_e3/E_data**3)
+    plt.xscale('log')
+    plt.yscale('log')
+    plt.ylabel(r'Flux data',fontsize=14)
+    plt.xlabel('Energy [eV]',fontsize=14)
+    #plt.show()
+    #plt.close()
```
sjoerd-bouma, line 141: Just cosmetic (looks like Steffen is taking care of the real review), but I think these lines are superfluous (duplicated from line 148)?

NuRadioReco/utilities/muon_flux.py
```diff
@@ -0,0 +1,154 @@
...
+import numpy as np
+import json
+import scipy.signal
+from NuRadioReco.utilities import units, io_utilities
+import astropy
+import scipy.interpolate as interpolate
+from scipy.integrate import quad
+from scipy.integrate import dblquad
+
+plot_flux = False
```
sjoerd-bouma, line 11: Maybe rather than having a `plot_flux` variable, you can protect the script with an `if __name__ == "__main__"` line. Then the plotting part will only run if you call `python muon_flux.py` as a script. Optionally, you could also add a docstring to the script to explain what it plots here.

NuRadioReco/utilities/muon_flux.py
```diff
@@ -0,0 +1,154 @@
...
+    '''
+
+    def flux(E, theta):
+        return get_flux_per_energy_and_zenith(E, theta)
+
+    int_flux = dblquad(flux, np.cos(zenith_bin_edge_high), np.cos(zenith_bin_edge_low), lambda x: energy_bin_edge_low, lambda x: energy_bin_edge_high)
+    return int_flux[0]
+
+
+if plot_flux == True:
```
sjoerd-bouma, line 126: i.e. replace this line with `if __name__ == "__main__"`

NuRadioReco/utilities/muon_flux.py
shallmann: looks better now thanks!

NuRadioReco/utilities/muon_flux.py
shallmann: looks good to me now. Thanks!

shallmann (dismissed):

shallmann:
Looks better to me now. Thanks for implementing the changes.

NuRadioReco/utilities/muon_flux.py
lpyras: Thanks, ya, the plot was not really thought thru. I changed it :)

NuRadioReco/utilities/muon_flux.py
lpyras: Thanks, nice idea. I implemented it :)

NuRadioReco/utilities/muon_flux.py
lpyras: Thank you, I redid the plot :)

NuRadioReco/utilities/muon_flux.py
lpyras: thanks, I cleaned this up.

NuRadioMC/EvtGen/generator.py
lpyras: I will do that soonish. :)

shallmann (approved):

NuRadioReco/utilities/muon_flux.py
```diff
@@ -0,0 +1,159 @@
...
+    energy: float
+        energy at which flux is calculated (in eV)
+    file_surface_mu_flux: string
+        file with model for muon flux
+
+    Returns
+    -------
+    flux of muon in NuRadio units. (GeV**-1 * m**-2 * ns**-1 sr**-1)
+    '''
+    abspath = os.path.dirname(os.path.abspath(__file__))
```
sjoerd-bouma, line 24: One more comment before you merge - it's good that you added this abspath, so the script works regardless of which directory it's called from. It does mean a user who has a different `file_surface_mu_flux` will have trouble using the `get_flux` function. This may not be an issue if you think it makes sense to always have the data files in this directory, or if you don't care about this functionality. 

But a solution would be to define `abspath` at the module level, and change the default value for `file_surface_mu_flux` instead (i.e. to `file_surface_mu_flux=os.path.join(abspath, 'data/muon_flux_E...').

cg-laser:
thanks for spotting this, I agree, this is the better solution

lpyras:
muon_flux

Here the optional plot :)

NuRadioReco/utilities/muon_flux.py
lpyras: Thanks, that's a good one. I will remember it for the next modules.

sjoerd-bouma (dismissed):
I haven't double checked the other changes, but the abspath change looks good to me : )

PR #361, opened 2021-12-23
cleanup noise generator functions

shallmann:
Cleans up the PA noise generator module to make it more transparent which of the implementations is faster...

### DIFF
--- a/NuRadioReco/utilities/noise.py
+++ b/NuRadioReco/utilities/noise.py
@@ -8,6 +8,9 @@ import datetime
 import copy
 import time
 
+import logging
+logger = logging.getLogger('noiseTriggerSimulation')
+
 
 def rolled_sum_roll(traces, rolling):
     """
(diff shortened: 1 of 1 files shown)


NuRadioReco/utilities/noise.py
```diff
@@ -548,13 +497,6 @@ def generate_noise2(self, debug=False):
 
                 traces[iCh] = perfect_floor_comparator(trace, self.adc_n_bits, self.adc_ref_voltage)
 
-            phased_traces = np.zeros((len(self.beam_time_delays), self.n_samples * self.upsampling))
```
cg-laser, line 551: was this really unused?

NuRadioReco/utilities/noise.py
```diff
@@ -474,27 +401,28 @@ def __triggering_strided(self):
...
+                        logger.info(f"{self._traces[iCh].std():.2f}")
                     ax[4].plot(self._phased_traces[iBeam])
                     fig.tight_layout()
                     plt.show()
                 return True
         return False
 
-    def generate_noise_optimised(self, debug=False):
+
+    def generate_noise(self, phasing_mode="slice", trigger_mode="binned_sum", debug=False):
```
cg-laser, line 418: can you document the two modes here? phasing_mode="slice", trigger_mode="binned_sum"

NuRadioReco/utilities/noise.py
shallmann: I think you had started implementing the (anyway much slower) generate_noise2 from the `generate_noise` and this was a leftover. In this function only  the 1D `phased_trace` is used iteratively in the nested loops to sum up the `shifted_traces`

NuRadioReco/utilities/noise.py
shallmann: ok done

shallmann:
I think this is fine to merge.

Example code

```
import numpy as np
import logging
logging.basicConfig(level=logging.INFO, format='%(message)s')

from NuRadioReco.utilities import noise
import NuRadioReco.modules.channelBandPassFilter
import time

channelBandPassFilter = NuRadioReco.modules.channelBandPassFilter.channelBandPassFilter()

Vrms = 10.

ngen = noise.thermalNoiseGeneratorPhasedArray(detector_filename="pa_100m.json", station_id=1,
                                              triggered_channels=range(2, 6),
                                              Vrms=Vrms,
                                              threshold=30.85 * np.power(Vrms, 2.0),
                                              ref_index=1.75)


t0 = time.time()
for i in range(10):
    traces, phased_traces = ngen.generate_noise(debug=False) # alternative options from older implementation: phasing_mode="roll", trigger_mode="stride"
print(f"generate_noise took {time.time() -  t0:.0f}seconds for 10 events")
```

cg-laser (approved):

PR #362, opened 2021-12-23
fix generic detector

cg-laser:
generic detector has still some bugs with the new reference stations/channels option

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -42,7 +42,7 @@ jobs:
         if [ "$master_version" = "$(poetry version)" ]
         then
           echo Version number is the same as existing release. Please update the version number!
-          exit 1
+          # exit 1
         fi
 
     - name: Build NuRadioMC # let's always build - maybe this will occasionally catch errors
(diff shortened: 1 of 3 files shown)


shallmann:
Thanks for spotting this. I should have tested this more thoroughly with a non-default-station detector file ...

cg-laser:
there is annother problem. This part of simulation.py. The generic detector is only used if a default station is provided. But this option is deprecated. Is is possible to determine automatically if a detector description is "generic" or not? 
```# read in detector positions
        logger.status("Detectorfile {}".format(os.path.abspath(self._detectorfile)))
        self._det = None
        if(default_detector_station):
            logger.status(f"Default detector station provided (station {default_detector_station}) -> Using generic detector")
            self._det = gdetector.GenericDetector(json_filename=self._detectorfile, default_station=default_detector_station,
                                                 default_channel=default_detector_channel, antenna_by_depth=False)
        else:
            self._det = detector.Detector(json_filename=self._detectorfile, antenna_by_depth=False)
        self._det.update(evt_time)

shallmann (approved):
ok with me

PR #363, opened 2022-01-05
fix PROPOSAL min energy cut in correct units

shallmann:
In the NuRadioMC PROPOSAL class (secondary shower generation) all the private functions are in PROPOSAL units (MeV instead of NuRadioMC's eV). One of them accidentially used NuRadioMC units.

As a consequence, the `min_energy_cut_nu` needed to be given in NuRadioMC units x 1e6. This is fixed with this PR.

### DIFF
--- a/NuRadioMC/EvtGen/NuRadioProposal.py
+++ b/NuRadioMC/EvtGen/NuRadioProposal.py
@@ -404,10 +404,10 @@ class ProposalFunctions(object):
         # Checking if the secondary type is greater than 1000000000, which means
         # that the secondary is an interaction particle.
         if (sec.type > 1000000000):
-            energy = (sec.parent_particle_energy - sec.energy) * units.MeV
+            energy = (sec.parent_particle_energy - sec.energy) * pp_MeV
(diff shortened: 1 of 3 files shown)


NuRadioMC/EvtGen/NuRadioProposal.py
```diff
@@ -534,7 +534,7 @@ def __filter_secondaries(self,
                 distance += ((sec.position.z - lepton_position[2]) * units.cm) ** 2
                 distance = np.sqrt(distance)
 
-                energy = self.__secondary_energy(sec)
+                energy = self.__secondary_energy(sec) * units.MeV
```
cg-laser, line 537: Is this conversion done so that the output shower list has the correct NuRadioMC units? 
Did you verify that this `energy` is not used anywhere else within the proposal code (where it would have the wrong units)?

shallmann: Yes, I checked both

cg-laser (approved):

cg-laser (approved):

PR #365, opened 2022-01-10
fix hdf5 on-the-fly passing

cg-laser:
the attributes of the hdf5 files were not correctly parsed when the input file was just passed in memory.


shallmann (approved):
fix looks good to me!

PR #372, opened 2022-01-17
save trigger time in phased array module

cg-laser:
the phased array module did not calculate the trigger time and was just setting the trigger time to the trace start time. This PR fixes that.


sjoerd-bouma:
Looks good, though I'll have another look in the morning. 
Would be nice to have the docstring for `triggerSimulator.phased_trigger` updated to reflect the additional `trigger_time` returned. And of course the trigger test needs a new reference file.

dansmithphysics (approved):
This commit seems fair to me. The only issue I might raise is that I am not sure how the hardware defines a trigger time, it is something I will have to look up.

sjoerd-bouma (approved):

PR #373, opened 2022-01-20
add calculation of trigger time and cutting trace to correct length

cg-laser:
an important feature was still missing in the calculation of the triggered noise from a phased array: The trace needs to be cut to have the trigger time at roughly the correct position (defined by the pre_trigger_time). 

Furthermore, the generated traces needed to be downsampled to the detector sampling range (because the PA code works with a 2x/4x upsampling). 

Also, the random generator interphase in the noise generation is changed to the recommended Philox. 

I only implemented it for the "fast" implementation. The other options are still TBD (or maybe remove them?)


ZackfromEarth (commented):
Is the Philox change a depreciation thing? Or is it just a better random method?

cg-laser:
a better random method. We should have changed all random generators to use Philox but we must have missed this one

ZackfromEarth (commented):
So the A01 test script hangs when I try to run it on this branch (works on develop). 

I've gone through line by line and there are two issues I'm seeing: 
1. It doesn't like the generic_detector, if I remove that we get things working up until...
2. lines 498 - 505

` triggered_bin = triggered_bin // 2  # the trace is cut in the downsampled version. Therefore, triggered bin is factor of two smaller. 
                i_low = triggered_bin - self.pre_trigger_bins
                i_high = i_low + self.n_samples_trigger
                if (i_low >= 0) and (i_high < self.n_samples):
                    # traces need to be downsampled
                    # resample and use axis -1 since trace might be either shape (N) for analytic trace or shape (3,N) for E-field
                    self._traces = scipy.signal.resample(self._traces, np.shape(self._traces)[-1] // self.upsampling, axis=-1)
                    return self._traces[:, i_low:i_high], self._phased_traces`

Things work again if I add an else statement to the end that returns self._traces, self._phased_traces. The hang up happens when nothing satisifes the if statement.... something with the i_low i_high and downsampling?

anelles (dismissed):
Looks good to me. I am not sure whether the default values make sense in pre-trigger window etc. But I am sure that you thought about these ;)

cg-laser:
thanks, I'm not merging that yet because the tests fail most likely because of a change in the random generator, but I still wanted to double check that.

colemanalan (dismissed):

colemanalan (approved):

PR #374, opened 2022-01-21
Update NuRadioRecoio.py

leshanz:
Fixing station_time got from station_header.

In the old version:
`data=NuRadioRecoio.NuRadioRecoio(file)
  h=data.get_header()
  sh=h[station_id]
  station_time=sh[stnp.station_time]`

station_time is the built-in python module <time>.

This is because in NuRadioRecoio.py line 169:

`station_time = time`

where time is the imported python module.

It is fixed by the change


cg-laser (approved):
fix looks good, but @christophwelling should also have a look.

christophwelling (approved):
Yes, that looks like a pretty big bug. Thanks for fixing

PR #375, opened 2022-01-26
Hotfix/n interaction parameter

cg-laser:
the n_interaction parameter was accidentally overridden. n_interaction counts the number of interactions of taus and muons generated in the corresponding CC interactions. The bug resulted in the following behavior: The n_interaction value of the primary neutrino interaction was overridden with the n_interaction value of the last shower of the previous event group. If the previous event group was a CC interaction, the n_interaction value of the primary interaction was set to a value larger than 1. If that happened, and if the primary shower did not trigger the detector, this shower is not added to the output hdf5 file. This has consequences for analyzing hdf5 files for 
 secondary interactions. With this bug, the initial interaction needs to be identified with np.abs(flavor) == 14 or 16.  
This bug does not affect any effective volume calculation.

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -1311,7 +1311,6 @@ class simulation():
     def _read_input_particle_properties(self, idx=None):
         if idx is None:
             idx = self._primary_index
-        self._fin['n_interaction'][self._shower_index] = self._fin['n_interaction'][idx]
         self._event_group_id = self._fin['event_group_ids'][idx]
 
         self.input_particle = NuRadioReco.framework.particle.Particle(0)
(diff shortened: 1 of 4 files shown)


NuRadioMC/simulation/simulation.py
```diff
@@ -1311,7 +1311,6 @@ def _create_station_output_structure(self, n_showers, n_antennas):
     def _read_input_particle_properties(self, idx=None):
         if idx is None:
             idx = self._primary_index
-        self._fin['n_interaction'][self._shower_index] = self._fin['n_interaction'][idx]
```
shallmann, line 1314: can you also change the if in l 1328 from self._fin['n_interaction'][self._shower_index] to self._fin['n_interaction'][idx] ...

shallmann (changes requested):

NuRadioMC/simulation/simulation.py
cg-laser: good point, done

shallmann (approved):
Good now. (If checks pass)

cg-laser:
for some reason the documentation test fails although nothing was changed that would impact the documentation. Likely cause is a newer sphinx version that became more picky? I will merge anyways, especially because the documentation is never build for the master.

Issue #348, opened 2021-12-01
detector browser cannot handle two reference stations


lpyras:
with the feature: allow_multiple_referemce_station a detector description can have more than one default station, called reference station. So far the detector browser does not seem compatible with that.

christophwelling:
Was fixed in PR #376

PR #376, opened 2022-01-27
Detector browser reference stations

christophwelling:
Fix for #348 

By default, the detector browser no longer asks for reference stations and channels, but a checkbox allows you to set them if you still need them.


lpyras (approved):
Thanks for taking care of this. I tried it with a file and it worked.

PR #384, opened 2022-02-14
the extra phase factor has been removed for emitter model

pgiri4:
Unnecessary phase factor and other terms has been erased from simulation file.


shallmann:
@pgiri4 can you briefly describe here why the phase was there / is to be removed in the emitter model? This makes the changes more traceable... Thanks!

cg-laser (approved):
all time delays are taken care of in the signal propagation part (the ray tracing). Adding an additonal phase factor here will result in double counting. Furthermore, it resulted in wrap-around effects as the trace length is typically much smaller than the total propagation time.

shallmann (approved):
Great, fine!

PR #389, opened 2022-03-15
fix issue 387 by including reference_station to dictionary

jakobhenrichs:
station in readCoREASShower.py was missing a reference station when it was added manually


shallmann:
Fine with me!

lpyras (approved):
Looks good for me, thanks for adding it.

PR #390, opened 2022-03-15
change refraction index to match the function definition

lpyras:
The channelGenericNoiseAdder uses the function. In the previous version the noise went from ice to air, now the refraction indices are swapped so that the signal goes from air to ice.


shallmann (approved):
Looks correct. Sorry for taking long.

PR #394, opened 2022-03-31
Update noise.py

leshanz:
Changed occurrences of:
`np.argwhere(triggered_bins is True)`
to
`np.argwhere(triggered_bins == True)`

I don't know if this is a python version issue, but at least with the python I am using (3.8.9), using `is` incorrectly returns an empty list and raises an error, while using `==` works


anelles (approved):
Thanks for catching this.

PR #396, opened 2022-04-13
Add inelasticity distributions from BRG18

cg-laser:
This PR adds the more precise dsigma/dy distributions from BRG18 to NuRadioMC. This makes the inelasticity distribution energy and interaction type dependent which significantly improved over the previous model. This implementation was used in https://arxiv.org/abs/2204.04237 

Todo: Also update total cross sections to use this model.


NuRadioMC/utilities/inelasticities.py
```diff
@@ -13,8 +20,12 @@
 cspeed = constants.c * units.m / units.s
 G_F = constants.physical_constants['Fermi coupling constant'][0] * units.GeV ** (-2)
 
+nu_energies_ref, yy_ref, flavors_ref, ncccs_ref, dsigma_dy_ref = pickle.load(lzma.open(os.path.join(os.path.dirname(__file__), "data", "BGR18_dsigma_dy.xz")))
+ncccs_ref = np.array(ncccs_ref)
 
-def get_neutrino_inelasticity(n_events, rnd=None):
+
+def get_neutrino_inelasticity(n_events, model="CTW", rnd=None,
+                              nu_energies=1 * units.EeV, flavors=12, ncccs="CC"):
```
shallmann, line 28: Not a huge fan of the nccc ccnc inttype interaction_type channel at various places in NuRadioMC (generator/inelasticieite/cross_sections) ... maybe that is because there is no official nomenclature... will it get worse when someone adds "res"/"glashow" in the future?

shallmann: But probably not the task of this PR... :)

shallmann (approved):

shallmann:
... it is expected that the Veff test fails now?

cg-laser:
thanks for having a look. The Veff test fails because I changed the default to the new cross sections. I'm never sure what to do, I think it is better to update the defaults to the best model, but then users might be confused to see their results changing when they update to a new version. What do you think?

NuRadioMC/utilities/inelasticities.py
cg-laser: I agree, do you have a better suggestion?

shallmann:
I think it is better to have a more accurate model. But thanks for verifying the tests fail intentionally!

NuRadioMC/utilities/inelasticities.py
shallmann: not really... I'd lean towards using `interaction_type` but just because I `ncccs` would make me feel I fell asleep pressing the c key... but I see the advantage that nccc tells everyone directly what this variable describes... You may leave it as is

shallmann:
One thing that I realised, do we then want to keep default 'ctw' cross section in other functions like `get_interaction_length` , `get_nu_cross_section`, `get_number_of_events_for_flux`,  `get_core_mantle_crust_weight`?


Occurences:
```
./NuRadioMC/utilities/cross_sections.py
./NuRadioMC/utilities/earth_attenuation.py
./NuRadioMC/utilities/fluxes.py
./NuRadioMC/examples/Sensitivities/E2_fluxes.py
./NuRadioMC/examples/Sensitivities/E2_Gen2_Long_whitepaper.py
./NuRadioMC/examples/Sensitivities/E2_fluxes2.py
./NuRadioMC/examples/Sensitivities/E2_fluxes3.py
./NuRadioMC/examples/Sensitivities/E1_fluxes.py                                            
```

cg-laser:
yes good point, we should change it consistently.

fschlueter:
@cg-laser, what is the status of this PR? It feels important that we move a way from a constant inelasticity value

NuRadioMC/utilities/inelasticities.py
```diff
@@ -32,25 +45,55 @@ def get_neutrino_inelasticity(n_events, rnd=None):
...
+                    yyy = 0.5 * (yyy[:-1] + yyy[1:])
+                    dsigma_dyy = 10 ** get_y(np.log10(yyy))
+
+                    yy[mask] = rnd.choice(yyy, size=size, p=dsigma_dyy / np.sum(dsigma_dyy))
+        return yy
+    else:
+        raise AttributeError(f"inelasticity model {model} is not implemented.")
 
 
 def get_ccnc(n_events, rnd=None):
```
fschlueter, line 83: Shouldn't that function also become energy dependent with the new model BGR18?

fschlueter:
Regarding the cross-section implementation. Is that not already in NuRadio?
```
    elif parameterization == 'hedis_bgr18':
        """
        Parameterization as above fitted to GENIE HEDIS module (with BGR18) cross sections
        as in arXiv:2004.04756v2 (prepared for JCAP)
        Precalculated xsec tables for 'nu_mu(_bar)_O16/tot_cc(nc)'/16 for isoscalar target
        obtained from GHE19_00a_00_000.root in https://github.com/pochoarus/GENIE-HEDIS/tree/nupropearth/genie_xsec
        Fitted in the energy range above 1 TeV; do not use below
        """
  ```
  Or am I confusing things?

**Comment by fschlueter** [comment]
What is here labeled with "BGR18" is from https://arxiv.org/pdf/1808.02034. The other models are already implemented in NuRadio. @cg-laser is "HEDIS-BGR" from https://arxiv.org/pdf/2004.04756v2 (the links in NuRadio and the paper are not working anymore). Is that the better param we should use?
image

**Comment by fschlueter** [comment]
This shows the cc fraction for the different models + the result for the new implementation of `get_ccnc` (thin dashed lines)
image

**Comment by fschlueter** [comment]
@cg-laser and @shallmann I updated defaults of (hopefully) all relevant functions to use the new cross-section model. I implemented the cc/nc fraction energy dependent. I will need to update the unit tests but feel free to comment (also see my question above regarding the precise parameterization for BGR18)

**Comment by shallmann** [comment]
> What is here labeled with "BGR18" is from https://arxiv.org/pdf/1808.02034 (which is what @cg-laser references in his paper). The other models are already implemented in NuRadio. @shallmann is "HEDIS-BGR" from https://arxiv.org/pdf/2004.04756v2 (the links in NuRadio and the paper are not working anymore). Is that the better param we should use, to what param does the inelasticity distribution belong)? image

Hi Felix,
yes he arxiv paper is the correct reference ... Not 100% sure, but with not working links you mean the links to the cross section tables in Alfonso's Github? Looks like these were removed... maybe the high energy tune is now integrated in GENIE directly somewhere (might need to check... not sure I find it)

If I am not mistaken, the parameterisation I put into NuRadio at the time was fit to the HEDIS tune of these GENIE xsection files

**Comment by shallmann** [comment]
Good you updated the defaults, looks good to me!

**Comment by sjoerd-bouma** [comment]
> This shows the cc fraction for the different models + the result for the new implementation of `get_ccnc` (thin dashed lines)

What's the reason for the scatter? I would have naively expected smooth lines...

**Comment by fschlueter** [comment]
> > This shows the cc fraction for the different models + the result for the new implementation of `get_ccnc` (thin dashed lines)
> 
> What's the reason for the scatter? I would have naively expected smooth lines...

random fluctuation, the function `get_ccnc` draws from a uniform random distribution.

**Comment by fschlueter** [comment]
This compares the inelasticity distribution. Old energy-independent param (CTW) and new energy dependent params for 3 energy bins.  
image

**Comment by fschlueter** [comment]
Here with mean values in the legend. Does anyone know why there are these bumps in regular intervals in the BGR param?
image

**Comment by fschlueter** [comment]
The two unit tests which are failing are both testing the the simulated Veff with fixed seeds. We would expect a change there when we change the inelasticity and cc/nc distribution. However, the simulations are at 1EeV so I would assume that the  inelasticity has a small effect while the tests are reporting a change of ~-10%. Is that solely from the cc/nc distribution?

**Comment by shallmann** [comment]
> Here with mean values in the legend. Does anyone know why there are these bumps in regular intervals in the BGR param? image

Yes, the bumps look weird. Maybe some binning effect? This is just what comes out of the function in `inelasticities.py`? I would suspect it is due to bumps in the `dsigma_dy_ref` that is evaluated ... `BGR18_dsigma_dy.xz`

**Comment by fschlueter** [comment]
okay, so against my intuition the entire change of ~ -10% in eff. volume comes from the change in inelasticity which for me is surprising giving that the energy independent distributions we had actually fits the energy-dependent one at 1EeV (for cc):
image

**Comment by fschlueter** [comment]
> > Here with mean values in the legend. Does anyone know why there are these bumps in regular intervals in the BGR param? image
> 
> Yes, the bumps look weird. Maybe some binning effect? This is just what comes out of the function in `inelasticities.py`? I would suspect it is due to bumps in the `dsigma_dy_ref` that is evaluated ... `BGR18_dsigma_dy.xz`

you can actually also see it in the CTW which I find strange.

**Comment by fschlueter** [comment]
so are people happy with just updating the reference values?

**Comment by shallmann** [comment]
> 
> you can actually also see it in the CTW which I find strange.

That is indeed strange, because I can see nothing that would produce this for CTW in the code... is the code that generates the plot in this PR also?

**Comment by sjoerd-bouma** [comment]
I'm not surprised by a 10% difference in Veff, this is along the lines of what @cg-laser mentioned in #600. I'm a bit confused as to why your BGR inelasticity lines up with CTW when I found a large discrepancy in https://github.com/nu-radio/NuRadioMC/issues/600#issuecomment-2003651811, but maybe the comparison is not apples to apples - will try to make time to look into this PR this week. I would hold off on updating the reference values until we're confident that we've figured out what's going on with the bumps above.

And also, thanks for finishing this PR, @fschlueter !

**Comment by fschlueter** [comment]
> > you can actually also see it in the CTW which I find strange.
> 
> That is indeed strange, because I can see nothing that would produce this for CTW in the code... is the code that generates the plot in this PR also?

No, I used the following code which uses the dependency `plotting.draw_residual(ax, ax2)` from the rnog analysis tools (newest develop version)
```

if __name__ == "__main__":
    # Test the functions
    n_events = 10000000

    from matplotlib import pyplot as plt
    from utilities import plotting

    fig, (ax, ax2) = plt.subplots(2, 1, sharex=True, height_ratios=[2.5, 1], gridspec_kw={"hspace": 0.05})

    inelasticities = get_neutrino_inelasticity(n_events)

    bins = np.linspace(0, 1, 100)

    ax.hist(inelasticities, bins=bins, histtype="step", weights=np.ones_like(inelasticities) / n_events, lw=1, label=f'CTW, mean: {np.mean(inelasticities):.3f}')

    for idx, nu_energy in enumerate([1 * units.EeV]):
        inelasticities_cc_bgr = get_neutrino_inelasticity(n_events, model="BGR18", ncccs="CC", nu_energies=nu_energy)
        ax.hist(inelasticities_cc_bgr, bins=bins, histtype="step", ls="--", color=f"C{idx+1}",
                weights=np.ones_like(inelasticities_cc_bgr) / n_events, lw=1, label=f'BGR18 CC {nu_energy / units.EeV:.2f}EeV, mean: {np.mean(inelasticities_cc_bgr):.3f}')

        inelasticities_nc_bgr = get_neutrino_inelasticity(n_events, model="BGR18", ncccs="NC", nu_energies=nu_energy)
        ax.hist(inelasticities_nc_bgr, bins=bins, histtype="step", ls=":", color=f"C{idx+1}",
                weights=np.ones_like(inelasticities_nc_bgr) / n_events, lw=1, label=f'BGR18 NC {nu_energy / units.EeV:.2f}EeV, mean: {np.mean(inelasticities_nc_bgr):.3f}')


    plotting.draw_residual(ax, ax2)
    ax.grid()
    ax2.set_xlabel('Inelasticity')

    ax.set_yscale("log")
    ax.legend()
    fig.tight_layout()
    plt.show()
    ```

fschlueter:
okay, the bumps look like a binning artefact to me. This is for CTW with 101 instead of 100 bins:
image

fschlueter:
and with 1000 bins: 
ima@sjoerd-bouma src=
@sjoerd-bouma does that mean I have your blessing to update the reference values?

fschlueter:
image

fschlueter:
> I'm not surprised by a 10% difference in Veff, this is along the lines of what @cg-laser mentioned in #600. I'm a bit confused as to why your BGR inelasticity lines up with CTW when I found a large discrepancy in [#600 (comment)](https://github.com/nu-radio/NuRadioMC/issues/600#issuecomment-2003651811), but maybe the comparison is not apples to apples - will try to make time to look into this PR this week. I would hold off on updating the reference values until we're confident that we've figured out what's going on with the bumps above.

I thought it should be less but only because the tests run at 1 EeV for which the inelasticity of the dominant cc channels is more or less unchanged...

fschlueter:
> > I'm not surprised by a 10% difference in Veff, this is along the lines of what @cg-laser mentioned in #600. I'm a bit confused as to why your BGR inelasticity lines up with CTW when I found a large discrepancy in [#600 (comment)](https://github.com/nu-radio/NuRadioMC/issues/600#issuecomment-2003651811), but maybe the comparison is not apples to apples - will try to make time to look into this PR this week. I would hold off on updating the reference values until we're confident that we've figured out what's going on with the bumps above.
> 
> I thought it should be less because the tests run at 1 EeV for which the inelasticity of the dominant cc channels is more or less unchanged...

NuRadioMC/utilities/inelasticities.py
```diff
@@ -23,58 +37,112 @@ def get_neutrino_inelasticity(n_events, rnd=None):
...
+                        size = np.sum(mask)
+
+                    nccc = nccc.upper()
+                    iF = np.argwhere(flavors_ref == flavor)[0][0]
+                    inccc = np.argwhere(ncccs_ref == nccc)[0][0]
+                    iE = np.argmin(np.abs(energy - nu_energies_ref))
+                    get_y = intp.interp1d(np.log10(yy_ref), np.log10(dsigma_dy_ref[iF, inccc, iE]),
+                                          fill_value="extrapolate", kind="cubic")
+
+                    yyy = np.linspace(0, 1, 1000)
```
shallmann, line 81: for BGR I initially thought this quantisation here might lead to the binning effect... but for CTW this is not used...

shallmann (commented):
From my side this looks fine. If you are confident that the >~10% deviation is expected or just a fluctuation, it is fine from my side to merge

shallmann (dismissed):

sjoerd-bouma:
> > I'm not surprised by a 10% difference in Veff, this is along the lines of what @cg-laser mentioned in #600. I'm a bit confused as to why your BGR inelasticity lines up with CTW when I found a large discrepancy in [#600 (comment)](https://github.com/nu-radio/NuRadioMC/issues/600#issuecomment-2003651811), but maybe the comparison is not apples to apples - will try to make time to look into this PR this week. I would hold off on updating the reference values until we're confident that we've figured out what's going on with the bumps above.
> 
> I thought it should be less but only because the tests run at 1 EeV for which the inelasticity of the dominant cc channels is more or less unchanged...

In the comment I linked the inelasticity is very much not unchanged, also at 1 EeV. It's possible I got it wrong, but I'd like to double-check that I understand why my result there is wrong before merging this.

fschlueter:
> > > I'm not surprised by a 10% difference in Veff, this is along the lines of what @cg-laser mentioned in #600. I'm a bit confused as to why your BGR inelasticity lines up with CTW when I found a large discrepancy in [#600 (comment)](https://github.com/nu-radio/NuRadioMC/issues/600#issuecomment-2003651811), but maybe the comparison is not apples to apples - will try to make time to look into this PR this week. I would hold off on updating the reference values until we're confident that we've figured out what's going on with the bumps above.
> > 
> > 
> > I thought it should be less but only because the tests run at 1 EeV for which the inelasticity of the dominant cc channels is more or less unchanged...
> 
> In the comment I linked the inelasticity is very much not unchanged, also at 1 EeV. It's possible I got it wrong, but I'd like to double-check that I understand why my result there is wrong before merging this.

@sjoerd-bouma on what time scale is that double check?

sjoerd-bouma:
> > > > I'm not surprised by a 10% difference in Veff, this is along the lines of what @cg-laser mentioned in #600. I'm a bit confused as to why your BGR inelasticity lines up with CTW when I found a large discrepancy in [#600 (comment)](https://github.com/nu-radio/NuRadioMC/issues/600#issuecomment-2003651811), but maybe the comparison is not apples to apples - will try to make time to look into this PR this week. I would hold off on updating the reference values until we're confident that we've figured out what's going on with the bumps above.
> > > 
> > > 
> > > I thought it should be less but only because the tests run at 1 EeV for which the inelasticity of the dominant cc channels is more or less unchanged...
> > 
> > 
> > In the comment I linked the inelasticity is very much not unchanged, also at 1 EeV. It's possible I got it wrong, but I'd like to double-check that I understand why my result there is wrong before merging this.
> 
> @sjoerd-bouma on what time scale is that double check?

Feel free to merge by the end of this week if I haven't gotten back to you by then.

fschlueter:
Unfortunately some of the above plots had wrong labels. The CTW distribution was actually the BGR18 cc at 1EeV. Here are corrected plots. With that it makes sense that the unit tests are failing by ~ -10%
image
image

NuRadioMC/utilities/inelasticities.py
```diff
@@ -23,58 +37,112 @@ def get_neutrino_inelasticity(n_events, rnd=None):
...
+        # based on CTW cross sections https://link.aps.org/doi/10.1103/PhysRevD.83.113009
+        ccnc = np.where(random_sequence <= 0.7064, 'cc', 'nc')
+    else:
+        if not isinstance(energy, (float, int)):
+            assert len(energy) == n_events, "Energy must be a scalar or an array of the same length as n_events"
+
+        # Flavor only relevant to determine if its a neutrino or antineutrino. For cross section ratio it
+        # hopefully doesn't matter
+        cc = cross_sections.get_nu_cross_section(energy, flavors=12, inttype="cc", cross_section_type=model.lower())
+        nc = cross_sections.get_nu_cross_section(energy, flavors=12, inttype="nc", cross_section_type=model.lower())
```
sjoerd-bouma, line 133: It's a smallish effect... but it does seem to matter? At least neutrino vs antineutrino. See the plot here: https://github.com/nu-radio/NuRadioMC/issues/600#issue-2026805640. I would add it as an optional parameter to the function.

fschlueter:
Okay, I updated the cross_section.py to return the integral over the differential cross section and updated the necessary reference values (-14% and +8%) + the energy range of an example (keep in mind that currently we can only simulated until 10EeV).

fschlueter:
@cg-laser and @shallmann the changes are implemented as discussed. One important note, I had to change the neutrino energy range in a few examples. We should remind us to change them back once we have an update param. Feel free to also merge this PR after approval as I might be unavailable.

anelles:
@shallmann are you in contact with Alfonso about a new parameterization? 
And if we need a reminder about the changed energy / new parameterization, I would suggest making an issue before merging this here.

shallmann:
> @shallmann are you in contact with Alfonso about a new parameterization? Or do you need someone to take this over? And if we need a reminder about the changed energy / new parameterization, I would suggest making an issue before merging this here.

I cced Felix and Sjoerd in the email. If Alfonso suggests we should generate tables ourselves it would be good if someone (eg. at DESY) could take care of this.

shallmann (approved):
With the caveat of tables going only up to 1e19 this is fine to merge. 
I propose to open an issue on this and extend the energy range in a separate PR.

Issue #400, opened 2022-04-25
voltage2efield not working without particle type


boboeyen:
After running a simulation with the emitter I wanted to use the `voltageToEfieldConverter` but it didn't properly work because no particle type was given to the event in the station. The particle type was set in the sim_station, but since a reconstruction model does not look for this (as data does not have this structure) the module was not aware of it and not working for me. Also I was wondering if some more differentiation is then needed to some how extract whether it is a neutrino, CR or pulser event?

ikrav:
I may be missing something, but why would voltageToEfieldConverter care about what is the cause of the electric field? Naively, it seems that only what the antenna characteristics are should be relevant.

Does this mean that something needs to be changed in the example for the emitter use that Pawan has configured in NuRadioMC?

Ilya

On Apr 25, 2022, at 10:06 AM, Bob Oeyen ***@***.******@***.***>> wrote:

Non-NU Email
________________________________



After running a simulation with the emitter I wanted to use the voltageToEfieldConverter but it didn't work because no particle type was given to the event in the station. The particle type was set in the sim_station, but since a reconstruction model does not look for this (as data does not have this structure) the module was not aware of it and not working for me. Also I was wondering if some more differentiation is then needed to some how extract whether it is a neutrino, CR or pulser event?

shallmann:
@boboeyen I suspect that the error was thrown elsewhere, since the voltageToEfieldConverter does not (and should not) use MC info like particle type... can you attach a minimum not-working example, or the full error message?

boboeyen:
Well the code simply does check whether it is a neutrino or cosmic ray to do other calculations for the refractive index. It apparently also uses a `ice` import from NuRadioReco.utilities to get the refractive index which has nothing sensible in it, just a few lines. It is really strange. Is this maybe an old module that was replaced by `voltageToEfieldConverterPerChannel` but was not removed?

https://github.com/nu-radio/NuRadioMC/blob/develop/NuRadioReco/modules/voltageToEfieldConverter.py#L32
https://github.com/nu-radio/NuRadioMC/blob/develop/NuRadioReco/utilities/ice.py

christophwelling:
This check for the particle type is in there because cosmic ray and neutrino signals have to be treated slightly differently.
If you reconstruct the signal direction for a cosmic ray from the time delay, you get the direction the signal had in the air, which is different from the one it has in the ice because of refraction at the air-ice surface. This does not happen for signals from neutrinos.

[...]

PR #401, opened 2022-04-27
Feature/particle type

cg-laser:
add more documentation of when to use the particle type, addresses #400

### DIFF
--- a/NuRadioReco/framework/base_station.py
+++ b/NuRadioReco/framework/base_station.py
@@ -186,8 +186,11 @@ class BaseStation():
 
     def is_cosmic_ray(self):
         if self._particle_type == '':
-            logger.error("Stations particle type has not been set")
-            raise ValueError("Stations particle type has not been set")
+            msg = "Stations particle type has not been set. Please call the module `eventTypeIdentifier.run(event, station, mode='forced', forced_event_type='neutrino'/'cosmic_ray')`."
+            msg += " This flag is used to differentiate between signals that originate from air vs. signals that originate from within the ice which is needed to e.g. determine if refraction "
+            msg += "into the ice needs to be considered."
+            logger.error(msg)
+            raise ValueError(msg)
 
         return self._particle_type == 'cr'
 
--- a/NuRadioReco/framework/base_station.py
+++ b/NuRadioReco/framework/base_station.py
@@ -179,8 +179,11 @@ class BaseStation():
 
     def is_neutrino(self):
         if self._particle_type == '':
-            logger.error("Stations particle type has not been set")
-            raise ValueError("Stations particle type has not been set")
+            msg = "Stations particle type has not been set. Please call the module `eventTypeIdentifier.run(event, station, mode='forced', forced_event_type='neutrino'/'cosmic_ray')`."
+            msg += " This flag is used to differentiate between signals that originate from air vs. signals that originate from within the ice which is needed to e.g. determine if refraction "
+            msg += "into the ice needs to be considered."
+            logger.error(msg)
+            raise ValueError(msg)
 
         return self._particle_type == 'nu'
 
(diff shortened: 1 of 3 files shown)


boboeyen:
Great, thanks for the quick addition. I think It can maybe be even more clearly stipulated that this module is only intended to use for cosmic rays and surface antennas. That would make it easier to understand its use.

boboeyen (approved):
Great, thanks for the quick addition. I think It can maybe be even more clearly stipulated that this module is only intended to use for cosmic rays and surface antennas. That would make it easier to understand its use.

cg-laser:
but this is not correct. This module can also be used for neutrino signals as well as to reconstruct pulser signals. We used it for example for the ARIANNA SPICE reconstruction. The requirement is that the efield is similar enough at the antennas that are used for reconstruction.

boboeyen:
Ah, then I didn't understand it well enough

NuRadioReco/framework/base_station.py
```diff
@@ -186,8 +186,11 @@ def is_neutrino(self):
 
     def is_cosmic_ray(self):
         if self._particle_type == '':
-            logger.error("Stations particle type has not been set")
-            raise ValueError("Stations particle type has not been set")
+            msg = "Stations particle type has not been set. Please call the module `eventTypeIdentifier.run(event, station, mode='forced', forced_event_type='neutrino'/'cosmic_ray')`."
+            msg += " This flag is used to differentiate between signals that originate from air vs. signals that originate from within the ice which is needed to e.g. determine if refraction "
+            msg += "into the ice needs to be considered."
+            logger.error(msg)
+            raise ValueError(msg)
```
sjoerd-bouma, line 193: If we add this to the `is_cosmic_ray` method, shouldn't we maybe add it to the `is_neutrino` method too?
It also is very detailed for an error message (maybe you can copy it over to the docstring, too?), but I guess there's no such thing as a too helpful error message : )

NuRadioReco/modules/voltageToEfieldConverter.py
```diff
@@ -109,6 +109,13 @@ def stacked_lstsq(L, b, rcond=1e-10):
 
 
 class voltageToEfieldConverter:
+    """
+    This module reconstructs the electric field by solving the system of equation that related the incident electric field via the antenna response functions to the measured voltages 
+    (see Eq. 4 of the NuRadioReco paper https://link.springer.com/article/10.1140/epjc/s10052-019-6971-5).
+    The module assumed that the electric field is identical at the antennas/channels that are used for the reconstruction. Furthermore, at least two antennas with
+    orthogonal polarization response are needed to reconstruct the 3dim electric field. 
+    Alternatively, the polarization of the resulting efield could be forced to a single polarization component. In that case, a single antenna is sufficient.
```
sjoerd-bouma, line 117: Again, very minor, so feel free to ignore. But numpydoc recommends
```
"""
short single line description

(optional) longer description
(...)
""""
```
for docstrings. For a class, I don't actually think it matters at all, but for methods and functions, having long blocks of text without the single line description occasionally renders a bit funny on sphinx.

sjoerd-bouma (approved):
Thanks Christian for adding some explanation / docstrings here... couple of very small nitpicky points which you are free to ignore. In general, I'm not sure what's good practice with regards to this stuff - it seems unlikely (and also not desirable) that NuRadioMC will ever conform to _all_ of python's / numpy's style guidelines. So let's say I'm happy to be ignored from time to time when it comes to this kind of comments.

cg-laser:
Hey Sjoerd, thanks a lot for your comments. It is very good that you keep educating me/us on how to do better documentation. As you said, I wouldn't be too strict in keeping all standards, it is always a compromise between better documentation and reducing the initial threshold for new people. As any documentation is better than no documentation, we should allow for non-standard doc strings and keep educating people to improve future work.

sjoerd-bouma (approved):
Thanks!

PR #402, opened 2022-04-29
Fix/attenuation length positive z

shallmann:
The attenuation length returned for positive z was counterintuitive/misleading/wrong

I did the following

- printout warning if attenuation length is requested for positive z
- set attenuation length to inf for positive z
- added debug plot if "__main__"
- edit the changelog.txt


cg-laser (approved):
good to catch this case.

christophwelling:
Yes, I think by default NuRadioMC will use the C++ attenuation. I can take care of that next week.

shallmann:
Thank you!

> Yes, I think by default NuRadioMC will use the C++ attenuation. I can take care of that next week.

Thanks!

christophwelling:
I think you only need to add a check of z > 0 in line 50 of attenuation.h
I don't really know what to return in that case though. As far as I know, C++ doesn't have infinity.

shallmann:
@christophwelling right, should really just be that line. Maybe we can find an attenuation length in air (which is large but not infinite)? Might require some searching, we probably don't usually use attenuation in air...

cg-laser:
@shallmann @christophwelling Is this PR good to be merged? Or does the GL2 model needs adjustment? We should probably merge this first before #424

shallmann:
Now returning `inf` in `attenuation.h`. If `inf` is not ok, another option is using 1e130 for infinity as in `NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp`

@christophwelling this is good to go now?

christophwelling:
Looks good so far. But have you checked if the infinity from the C++ code gets translated into Python properly?

anelles (dismissed):
@shallmann this needs conflict resolving and the last question by christoph is open whether the inf is propagated usefully. Otherwise, this seems good for me.

shallmann:
For consistency with the `analytic_raytracing.cpp` I set infinity to `1e130` and accounted for conflicts

anelles (approved):

PR #403, opened 2022-05-02
Update emitter

pgiri4:
The existing idl data has been updated and new data for hvsp1 waveform has been added ( as discussed in **inicemc** call  April-27, 2022). And emitter.py file has also been updated to introduce new model hvsp1.


cg-laser (commented):
looks good, my only comment is that there is an hvsp2 option that is not documented anywhere. Do you want to remove this file?

pgiri4:
Hi Christian, you are right, hvsp2 is no longer in use. A new model HVSPM is under progress at KU. Ilya is also planning to use HVSPM in summer for RNO-G . We thought, once HVSPM become available we will replace HVSP2 with HVSPM. So lets Keep hvsp2 in emitter model until we replace it . Let me know if you have different suggestion..

cg-laser:
if you want to merge the PR already now, I think it is better to delete the unused file now, and also remove the description. 

Can you also document the change in the `changelog.txt` file?

cg-laser (approved):

PR #409, opened 2022-06-02
Small fixes

anelles:
Simplifying life for issues with new versions of externals in compiling the CPP ray-tracer and fixing typos.
Changes are minor so no need to edit the changelog


cg-laser (approved):

sjoerd-bouma (approved):
Thanks for cleaning up some of my trailing spaces!
I have added two small fixes to the C++ raytracer `setup.py`:
- `from distutils.core import setup` will be deprecated (PEP 632) and raises a deprecation warning as of python 3.10.
- Not specifying `language_level` causes the raytracer to be compiled using python 2 on systems where this is available. That doesn't seem to break anything, but it does raise a couple of errors during compilation.
Both are now fixed.

PR #415, opened 2022-07-13
improved validation workflow

cg-laser:
all test are executed even if some show an error

### DIFF
--- a/.github/workflows/run_tests.yaml
+++ b/.github/workflows/run_tests.yaml
@@ -42,14 +42,16 @@ jobs:
     - name: Lint with flake8 (unimportant)
       run: |
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-        flake8 . --count --max-complexity=10 --max-line-length=127 --statistics
-      continue-on-error: true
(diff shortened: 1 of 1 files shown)


anelles (approved):
Looks like you just added an always() everywhere, so that is fine with me. Not sure how these workflows work exactly. 
But one comment while reading this: While do we uninstall numba to test ARZ? Shouldn't we test it with it? (line 33)

anelles:
The I don't understand how this file works. It looks that the installation dependencies are always the same.

cg-laser:
```
- name: Single event test (ARZ)
      if: always()
      run: |
        export PYTHONPATH=$(pwd):$PYTHONPATH
        NuRadioMC/test/SingleEvents/validate_ARZ.sh
    - name: Test Numba version of ARZ
      if: always()
      run: |
        export PYTHONPATH=$(pwd):$PYTHONPATH
        pip install numba
        NuRadioMC/test/SingleEvents/validate_ARZ.sh
```

anelles:
Ah, I fell for the double negative. First uninstalling and then installing again. Never mind. All clear.

PR #417, opened 2022-07-21
RNO-G single station detector

christophwelling:
Adds a reference_channel entry to the detector descriptions
Fixes positions of LPDAs


lpyras (approved):
Since this is only used as a GenericDetector the station time is not important. Looks good to me.

PR #418, opened 2022-07-25
Generic detector simulations

christophwelling:
To use the GenericDetector class for simulations it is no longer required to pass a reference station of channel by hand. Instead, NuRadioMC will check if a reference station/channel is defined in the detector description and select the correcdt detector class.

Trying to still pass a defautl station will now cause a deprecation warning.


cg-laser (changes requested):
Hi Christoph, great fix, thanks. 

I'm wondering if lines 233-244 that check if we have a generic detector or not should be refactored into a function. And maybe even added as a utility function to the generic detector module. 

Didn't we also allow to pass the detector description directly as a JSON object? I think the detector class allows for it but it is never used in simulation.py?

sjoerd-bouma:
Maybe this is also the right place for this discussion - when I started out on NuRadioMC I found it a bit confusing that the `Detector` and `GenericDetector` are two separate classes. I can see why the two need to be distinguished, but from a user perspective, wouldn't it be easier to at least be able to open any json detector description from a single `Detector` class? We could consider always printing a warning that tells the user whether or not they have opened a `GenericDetector`.

cg-laser:
I agree, that this would be the better interface. Could we add the check that Christoph implemented into the Detector class which would then automatically return a GenericDetector object in case the JSON file is a generic detector?

christophwelling:
That is definitely possible: https://stackoverflow.com/questions/42963968/python-class-depending-on-variable

Though it feels a bit weird to call the constructor on a class, and then get an object from a different class.

christophwelling:
Done

NuRadioReco/detector/detector_base.py
```diff
@@ -0,0 +1,995 @@
...
+        elif source == 'dictionary':
+            self._db = TinyDB(storage=MemoryStorage)
+            self._db.truncate()
+            stations_table = self._db.table('stations', cache_size=1000)
+            for station in dictionary['stations'].values():
+                stations_table.insert(station)
+            channels_table = self._db.table('channels', cache_size=1000)
+            for channel in dictionary['channels'].values():
+                channels_table.insert(channel)
+        else:
```
sjoerd-bouma, line 179: Another mistake I've definitely made before is to forget about the `source` argument and called `Detector(path_to_detector_json)` - this (unexpectly) loads the Arianna detector. This could be fixed by changing this to another `elif source == 'json':` + perhaps an error emitted if `source` is not one of `json, dictionary` or `sql`.

(Thanks a lot for these changes by the way Christoph! I'm not sure if github etiquette is to point out these suggestions or just make commits to your PR? I can also do the latter)

christophwelling: This check is already done in the "new" Detector class.

Unfortunately the ARIANNA detector is set as the default argument for the JSON filename, that's why it is loaded. You can see that I wanted to remove this default, but that would require changing the order of parameters, since `source` has a default as well. And may break a lot of things.

NuRadioReco/detector/detector.py
```diff
@@ -1,995 +1,80 @@
...
-            Can be used to force the creation of a new detector object. By default, the __init__ will anly create a new
-            object of none already exists.
         """
-        self._serialization = SerializationMiddleware()
-        self._serialization.register_serializer(DateTimeSerializer(), 'TinyDate')
-        if source == 'sql':
-            self._db = buffer_db(in_memory=True)
+        if source == 'json':
+            f = open(json_filename, 'r')
+            station_dict = json.load(f)
```
sjoerd-bouma, line 45: This gets rid of the (quite convenient feature) of being able to call detector paths relative to the `detector` module path, e.g. `RNO_G/RNO_season_2021.json`. This never worked for the GenericDetector anyway, but maybe you can copy the implementation from the DetectorBase class (old Detector class)? (lines 180-188)

cg-laser (changes requested):
thanks Christoph. I think the changes are good but I'm confused about the naming of the file names. I would have thought that `Detector.py` is the most general class that users would always call. This class would redirect to the specific detector the user wants. 
If I see it correctly, the only thing you would need to change is to add to "Detector.py" to allow calling it with the SQL option?

christophwelling:
That is exactly how it works. If you call the __init__ for the Detector class, it will check if reference stations or channels are specified and then initialize itself as a GenericDetector or as DetectorBase (what used to be the default detector class, I figured the new name made sense since the GenericDetector is just an extension and inherits from it.)
I will put in the SQL option. Do we have an easy way to test that?

christophwelling:
Done

cg-laser (changes requested):
it is still missing in the docstring :-(

and you initialize a new GenericDetector. Shouldn't it be a base_detector?

cg-laser (approved):

PR #421, opened 2022-08-01
Singleton metaclass no longer alters class signature

sjoerd-bouma:
The [Singleton](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.utilities.metaclasses.html?highlight=singleton#NuRadioReco.utilities.metaclasses.Singleton) metaclass used to alter the signature of the classes it was applied to. In particular, calling a Singleton class with positional arguments resulted in the first positional argument being interpreted as the `create_new` boolean argument - this behaviour was (to me) unexpected, and is not reflected in the docstrings / documentation of the existing Singleton classes.

This pull request changes the behaviour of Singleton classes - now, `create_new` **must** be passed as a keyword argument, but the signature of the class stays the same. I think this is more intuitive and less error-prone when creating new Singleton classes than the old behaviour. Alternatively, one could consider (at minimum) to change all docstrings to accurately reflect that the first argument is `create_new`.

### DIFF
--- a/NuRadioReco/utilities/metaclasses.py
+++ b/NuRadioReco/utilities/metaclasses.py
@@ -1,21 +1,31 @@
 class Singleton(type):
     """
     Can be assigned to classes as a metaclass.
-    By default, only one instance of a Singleton can exist at a time, as the __call__ method is overwritten to
+
+    By default, only one instance of a Singleton can exist at a time,
+    as the ``__call__`` method is overwritten to
     return the existing instance if one exists.
     """
     _instances = {}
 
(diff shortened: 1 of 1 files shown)


cg-laser (approved):
very good. This was indeed quite error-prone before. Maybe @christophwelling should also have a look before we merge it.

anelles:
I think this is waiting for an ok from @christophwelling -- but I am in favor also.

christophwelling:
Sorry, I must have missed this one. Looks good to me

anelles (approved):

PR #423, opened 2022-08-18
Station 11 helper string

iplaisier:


NuRadioReco/detector/RNO_G/RNO_season_2021.json
```diff
@@ -1300,7 +1300,7 @@
             "ant_orientation_theta": 0.0,
             "ant_position_x": -17.2,
             "ant_position_y": 29.097,
-            "ant_position_z": -93.72,
+            "ant_position_z": -92.72,
```
shallmann, line 1303: This value does not seem correct, since the HPol is at ~ the same height on that string.

shallmann (changes requested):
Hi Ilse, thanks for updating these values. Please check the Cal Vpol depth of station 21 / B. It should be a metre higher (or the antenna positions on that string lower) in order to not overlap with the HPol.

I asked that someone a while back, but do not remember if I got a satisfying answer: The Helper strings are all built similarly? I.e. The Cal Vpol is highest all stations in both the B and C strings? Can you confirm?

NuRadioReco/detector/RNO_G/RNO_season_2021.json
iplaisier: good call. 

The calpulsers are indeed always deployed as the highest antenna of the helper strings.

shallmann (approved):
Good to merge. Thanks Ilse!

PR #424, opened 2022-08-18
Greenland attenuation model

christophwelling:
Adds the updated attenuation model from our 2021 paper https://arxiv.org/pdf/2201.07846.pdf
The paper originally used the Bogorodsky model for its depth dependence, which was implemented as "GL3" into NuRadioMC. The reviewer then pointed out  that this model is outdated and we should use the MacGregor model instead. This revised model is now implemented as "GL3" with this pull request

This plot shows the values the model returns for different frequencies between 100 and 500 MHz. The dotted black line is the C++ implementation at 300MHz. 

GL3_implementation
.


cg-laser (changes requested):
Thanks for adding the new attenuation length measurement. 

The depth dependence is quite spikey. Have you checked that the line integral along the path to get the attenuation factor still converges well? 

If I see it correctly, in the python implementation, the file is loaded and interpolated verytime you call the function. Better move this to the beginning of the file so that it is done only once. Or even better, implement a buffer so that the data is loaded and interpolated when the GL3 model is called for the first time.

christophwelling:
Good point. I thought it wouldn't be too bad since the attenuation is almost always run on entire numpy arrays at once, but it slows stuff down quite a bit.
I fixed it, so the CSV data is read in whenever the module is loaded now.

NuRadioMC/utilities/attenuation.py
```diff
@@ -123,6 +133,23 @@ def get_attenuation_length(z, frequency, model):
             att_length_f[att_length_f < min_length] = min_length
         return att_length_f
 
+    if model == 'GL3':
+        slope_interpolation = scipy.interpolate.interp1d(
```
cg-laser, line 137: don't you want to move the interpolation ll 137-148 also to the start of the module?

cg-laser:
> 

good that you checked it. I was a bit worried that the numerical integration routine gets into problems if the function is not smooth enough.

NuRadioMC/utilities/attenuation.py
christophwelling: It doesn't slow down the code too much since it is just a simple linear interpolation, but I guess it can't hurt. Done

cg-laser (approved):

PR #425, opened 2022-08-23
Index of refraction with numpy arrays

christophwelling:
Makes it possible to pass an array of points to the get_index_of_refraction method of the IceModelSimple class instead of only one point at a time.

### DIFF
--- a/NuRadioMC/utilities/medium_base.py
+++ b/NuRadioMC/utilities/medium_base.py
@@ -185,19 +185,25 @@ class IceModelSimple(IceModel):
 
         Parameters
         ----------
-        position:  3dim np.array
-                    point
+        position:  1D or 2D numpy array
+                    Either one position or an array
+                    of positions for which the indices
+                    of refraction are returned
 
         Returns
         -------
         n:  float
             index of refraction
         """
-        if (position[2] - self.z_air_boundary) <=0:
-            return self.n_ice - self.delta_n * np.exp((position[2] - self.z_shift) / self.z_0)
+        if position.ndim == 1:
+            if (position[2] - self.z_air_boundary) <=0:
+                return self.n_ice - self.delta_n * np.exp((position[2] - self.z_shift) / self.z_0)
+            else:
+                return 1
         else:
-            return 1
-
+            ior = self.n_ice - self.delta_n * np.exp((position[:, 2] - self.z_shift) / self.z_0)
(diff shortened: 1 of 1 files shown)


anelles:
I think you are breaking backwards compatibility now with people entering lists (like the tests do) and not arrays.

christophwelling:
Docstring said `position` should be a numpy array. But I can make it work for lists as well.

cg-laser (commented):
looks good but you could improve readability by putting it into one line:

`if isinstance(position, list) or position.ndim == 1:`

boboeyen:
Just a thought, but shouldn't we do this for all the ice models then? I think I know an elegant way of doing this:

introduce a specific helper function `get_index_of_refraction_point` which is different for each ice model and only takes one 3D point (basically the current `get_index_of_refraction` function in the `IceModel` class that has to be overwritten with each new ice model inheriting from it). Then define a new `get_index_of_refraction` function which calls the previous function but can handle point, list of points and arrays of points. By implementing the last function in the parent class `IceModel`, every model that will be made in the future will have this kind of capability and people inventing new models only have to worry about implementing it for a point.

p.s.: maybe this should also be done then for the average and the gradient so all the function have the same functionality?

cg-laser:
Interesting idea. I think the downside is that it will slow things down as you would need to loop over all points. Accepting an array of points makes only sense if the function is vectorized but I'm not sure if this can be done for all ice models.

boboeyen:
Okay, hadn't think of the fact that a vectorised function is used and not a loop
If a function can be vectorised one could overwrite the `get_index_of_refraction` function again but then all the ice models have the same functionality. Or doesn't that make sense either?

christophwelling:
I don't understand what the `get_index_of_refraction_point` would do that you would not already get from the `get_index_of_refraction` function if you just pass a single point.

I thought about also implementing support for a list of points but then decided against it. Applying a function to an array of inputs is really only done in the context of numpy and not something you normally see for lists. If you need it, there is nothing stopping you from converting your list into a numpy array.

boboeyen:
Well the things with this implementation is that `get_index_of_refraction` is inherited and automatically wraps the new function into a function that can handle both points, lists of points and arrays. So when implementing a new ice model you don't need to worry about righting your function for multiple variable. You just adjust the point function and inheritance does the rest for you

anelles (approved):

PR #430, opened 2022-09-06
Update CI and gitignore.

fschlueter:
In this MR the validation tests are updated (their shebangs) and the housekeeping of the produced output files in improved (removing them if tests where successful, adding them to the gitignore).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,13 @@
 # .gitignore files to exclude specific files.
 #############################################################
 
+# Ignore test output - this will also make sure that you do not accidentally add large files to your commit
+*output*.nur
+*output*.hdf5
+# Exclude reference files from being ignored here
+!*reference*.nur
+!*reference*.hdf5
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
@@ -130,9 +137,3 @@ venv.bak/
 # ignore .nur files in NuRadioReco/examples/
 # this should not apply for NuRadioReco/examples/example_data/
 NuRadioReco/examples/*.nur
-
-# ignore test output (not include input or reference files here!)
-*output*.nur
-*output*.hdf5
-!*reference*.nur
-!*reference*.hdf5
--- a/.gitignore
+++ b/.gitignore
@@ -91,7 +91,6 @@ celerybeat-schedule
 # Environments
 .env
 .venv
-env/
 venv/
 ENV/
 env.bak/
@@ -132,9 +131,8 @@ venv.bak/
 # this should not apply for NuRadioReco/examples/example_data/
 NuRadioReco/examples/*.nur
 
-# ignore test output (not include input files here!)
-NuRadioMC/test/SingleEvents/1e18_output.nur
-NuRadioMC/test/SingleEvents/1e18_output_ARZ.hdf5
-NuRadioMC/test/SingleEvents/1e18_output_noise.hdf5
-NuRadioMC/test/SingleEvents/MB_1e18_output.hdf5
-NuRadioMC/test/Veff/1e18eV/output.nur
+# ignore test output (not include input or reference files here!)
+*output*.nur
+*output*.hdf5
+!*reference*.nur
+!*reference*.hdf5
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
+#############################################################
+# Keep in might that sub-directories might contain additional
+# .gitignore files to exclude specific files.
+#############################################################
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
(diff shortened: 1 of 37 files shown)


cg-laser:
I support this but in addition all tests should clean up after themselves, i.e., in `run_all_tests.sh` should be `rm output_file`

fschlueter:
Can the github ci store artifacts? Are (some of) those tests also validating their output? If so it might make sense to keep the output (if the validation fails and one want to compare the output/result)

fschlueter:
I implemented the clean up for the few files I added to the gitignore. However a few remarks:

- I found gitignores in several sub directories which already contain some of the output files from the automatic testing. Should we rather go with this idea or remove them and implement all in one top level file?
- the file  `output.nur` in `NuRadioMC/test/Veff/1e18eV/` is not validated as it seems (but  `output.hdf5` is)
- We probably want to extend that to all tests but I would first like to ask if that is the correct approach.

cg-laser:
good point about the global/local gitignore files. I think it is often more practical to create a local gitignore in the test folder. That seems to be more structured for me but I don't have a strong opinion on that.

fschlueter:
Okay, I can shift the entries to the local gitignore files. Just one thought, wouldn't we make our lives easier if we add all `*.hdf5` and `*.nur` files but exclude from this all `*reference.hdf5` and `*reference.nur` files or am I overseeing something here?

cg-laser:
yes, hdf5 files are input to some of the tests.

fschlueter:
@cg-laser some of the tests when executed with `run_all_tests.sh` would fail for me due to an import error when executed without specifying an interpreter. I solved it when changing the shebang in the first line from `#!/usr/bin/env python` to `#!/usr/bin/env python3`. Can I do this for all files (do we support python2 still?) or is it rather a problem of my environment?

fschlueter:
or should we specify the interpreter when executing i.e., write `python3 NuRadioMC/.../...py ... ` instead of  `NuRadioMC/.../...py ... `

cg-laser:
`#!/usr/bin/env python3` is the better option. On newer systems, a separate `python3` executable might not exist anymore. We don't support p2 anymore. This change can be done for all tests. Thanks

fschlueter:
alright, If the checks pass i consider this PR ready. All files which are produced should now be removed if the tests is successful. If not they should be ignored by git (there might be still a few files which are not but finding them now would be very cumber sum and adding them later should not be a big issue).

anelles (approved):
I think you have caught most. Following the discussion above this seems good to me.

.gitignore
```diff
@@ -123,3 +130,9 @@ venv.bak/
 # ignore .nur files in NuRadioReco/examples/
 # this should not apply for NuRadioReco/examples/example_data/
 NuRadioReco/examples/*.nur
+
+# ignore test output (not include input or reference files here!)
+*output*.nur
+*output*.hdf5
```
cg-laser, line 136: I don't like this wildcard. We might want to add files matching this string later. It will be difficult to find out for developers why they don't show up in git. I think we need to go though the extra effort of adding the exact file names. 
Or what do other people think? This of course is an additional safeguard against adding larger files by mistake.... but sometimes we want to add new nur files which would then always require to update the global gitignore file first.

cg-laser (changes requested):

.gitignore
fschlueter: just to mention it. I explicitly exclude from being ignored those:
```
!*reference*.nur
!*reference*.hdf5
```
My intention was: Its annoying to maintain gitignore files. If we come up with a naming scheme for our input, output, reference files we could use wildcards and also remove quite some local gitignores. 

That being said I can add explicit entries again.

cg-laser: Yes I saw that. But it is a very hidden implicit convention for a naming scheme. I think there a pro and cons to both options. Let's discuss it in one of the next inicemc calls.

fschlueter:
Can this be merged now?

sjoerd-bouma (approved):
Looks good to me, but I think @cg-laser needs to approve before you can merge.

.gitignore
```diff
@@ -1,3 +1,15 @@
+#############################################################
+# Keep in might that sub-directories might contain additional
+# .gitignore files to exclude specific files.
+#############################################################
+
+# Ignore test output - this will also make sure that you do not accidentally add large files to your commit
+*output*.nur
```
cg-laser, line 7: I think we discussed to make it even more agressive and exclude all *.nur and *.hdf5 so that people don't accidentally commit large files. Other than that it looks good.

fschlueter:
ping

cg-laser (approved):

PR #431, opened 2022-09-06
Remove deprecated interface from visualize_detector.py; Update RNO_single_station.json 

fschlueter:
Not sure if this tool is very useful and if the fix really makes sense. Also, if the tool is useful should we added to the CI?

### DIFF
--- a/NuRadioReco/detector/RNO_G/RNO_single_station.json
+++ b/NuRadioReco/detector/RNO_G/RNO_single_station.json
@@ -344,7 +344,7 @@
       "station_id": 11
     },
     "22": {
-      "ant_comment": "Helper string B",
+      "ant_comment": "Helper string A",
       "amp_type": "iglu",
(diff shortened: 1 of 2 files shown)


anelles:
I would like @iplaisier to comment on this. I think she mostly handled the RNO-G json. Is this a useful tool?
My feeling is that I wouldn't add it to CI -- it may be useful but not that important.

cg-laser (approved):
I think it is a useful feature to generalize the script. Users can still adapt it to take a different station.

cg-laser (commented):
But I'm not sure about the renaming of the RNO-G detector description. Shouldn't this be better done in a separate PR?

fschlueter:
@cg-laser I updated the MR title to reflect the fact that I fix the json file. I dont not think such a uncritical change needs to be an own PR.

PR #432, opened 2022-09-06
minimal refactor of 06_webinar/W02RunSimulation.py. Update manual acc…

fschlueter:
Again fixing small things I stumble upon. The manual is very nice, now it should be correct again. An expert might want to add more/updated doc strings.

### DIFF
--- a/NuRadioMC/examples/06_webinar/W02RunSimulation.py
+++ b/NuRadioMC/examples/06_webinar/W02RunSimulation.py
@@ -1,6 +1,5 @@
 import argparse
 # import detector simulation modules
-import NuRadioReco.modules.efieldToVoltageConverter
 import NuRadioReco.modules.trigger.simpleThreshold
 import NuRadioReco.modules.trigger.highLowThreshold
 import NuRadioReco.modules.channelResampler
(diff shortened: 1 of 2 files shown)


cg-laser (changes requested):
thanks Felix.

We recently changed the interface to the generic detector. It is no automatically identified if a generic detector was used. This means, that the arguments `default_detector_station` and `default_detector_channel` can be removed. For now we kept them still in for backward compatibility. Could you change (and test) that?

fschlueter:
@cg-laser, I removed the both deprecated arguments. The code is running and I do not see any difference in the description of channels 1-3 w.r.t. channel 0 (which was the default channel. However when I run the example before and after the results changed (judged from the terminal output). However I suspect this is because the random seed to generate the event list is not fixed. Should we fix the seed? Do we have to validate that the results are exactly the same or is this superficial check okay?

fschlueter:
just to learn, can I signify that I have addressed the requested changes or does the person who requested have to resolve them?

cg-laser (approved):
This is just an example that should run and be up to date but we don't use it to validate the code. Fixing the seed is in general a bad idea so we shouldn't do it in the example. 

I checked the example and the "default_channel" and "default_station" feature is not used in the detector description. (But with the old interface, you still needed to provide a default channel or station to tell NuRadioMC that you have a `GenericDetector`. This is no longer needed with the recent fix. )

Issue #325, opened 2021-10-18
Pulser example throws confusing warning


anelles:
When executing example, 05_pulser_calibration_measurement

the warning
WARNING:cross sections:CTW neutrino nucleon cross sections not valid for energies below 1e4 GeV

fills up the terminal. Obviously for a pulser no cross-section is needed, but we should think about a clever way to suppress this.

PR #434, opened 2022-09-08
Fix example 05_pulser_calibration_measurement. Uncomment dummy proper…

fschlueter:
…ties for neutrino event
with this PR the example `05_pulser_calibration_measurement` is "fixed". Which means it runs through. However it splits out a lot of warnings which were/are already addressed with the issue #325. By the way this example is not in the automatic testing (CI).

Can we maybe avoid the warning by increasing the dummy energy? Is is dangerous to put a "reasonable" energy there?

### DIFF
--- a/NuRadioMC/examples/05_pulser_calibration_measurement/A01generate_pulser_events.py
+++ b/NuRadioMC/examples/05_pulser_calibration_measurement/A01generate_pulser_events.py
@@ -63,11 +63,11 @@ def generate_my_events(filename, n_events):
     data_sets["interaction_type"] = np.array(['had'] * n_events)
 
     # again these parameters are irrelevant for our simulation but still need to be set
-    # data_sets["flavors"] = np.array([12 for i in range(n_events)])
-    # data_sets["energies"] = np.ones(n_events) * 1 * units.eV
-    # data_sets["inelasticity"] = np.ones(n_events) * 0.5
-    # data_sets["shower_energies"] = data_sets["inelasticity"] * data_sets["energies"]
-    # data_sets["shower_type"] = np.array(['had'] * n_events)
+    data_sets["flavors"] = np.array([12 for i in range(n_events)])
+    data_sets["energies"] = np.ones(n_events) * 1 * units.eV
+    data_sets["inelasticity"] = np.ones(n_events) * 0.5
+    data_sets["shower_energies"] = data_sets["inelasticity"] * data_sets["energies"]
+    data_sets["shower_type"] = np.array(['had'] * n_events)
 
     # write events to file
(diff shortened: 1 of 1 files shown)


pgiri4:
This is not in test example section(automatic test), the effect of commenting those lines were not seen easily. Thanks Felix for running the example.  Since, we do have fully developed radio emitting Pulser example available, I think we rarely use this example. May be we can consider to transfer all the essence of this example to radio emitter example and keep only one pulser example inside NuRadioMC. This is just my opinion. May be @cg-laser , @ikrav have different suggestions..

anelles:
I think @pgiri4 we have been waiting for your verdict on whether the example is ready and should be added to the automatic testing. I think it would be good to have a minimal working example, really only the basics.

cg-laser:
We removed this example because it is outdated. It should be fully replaced by pawans new example. However, this old example has additional nice features as it also provides an example reconstruction of pulser events. I see tow options. 1) merge the two examples, i.e., add an example reconstruction also to Pawans new example or 2) properly fix the current example (I think this means changing the event generation part to specify that the generated events are an emitter. )

pgiri4:
I think doing (1) is a better idea, so that we will have only one example for pulser calibration. Okay, I will try to convert the reconstruction script for radio emitter and will ask for help if needed.

fschlueter:
I think the consent from the colab meeting was: Merging that fix now but than merging this example with an other one.

anelles:
So let's merge this fix now and then create an issue to merge the examples, @fschlueter

anelles (approved):

fschlueter:
@anelles should I add an issue to merge this example? With which should it be merged?

anelles:
You wrote "merging this example with an other one." ;)
I think the idea was that pawan writes a new example and once that is there, we see if there is anything in here that is not yet covered in the newer one and merge.

anelles:
By the way you just violated our general rules "Wait 24 hours after approval before merging." 
I think it is uncritical here, but good practice and it has caught mistakes in the past.

pgiri4:
Hi @anelles , @fschlueter , @cg-laser . Sorry for the delay from my side.  As far I remember, our goal here is to include the Efield reconstruction idea to new pulser example. I have come up with a simple Efield-reconstruction script from existing file.  The link for the file is here: https://github.com/nu-radio/NuRadioMC/blob/to_update_pulser_example/NuRadioMC/examples/05_pulser_calibration_measurement/radioEmitting_pulser_calibration/A03reco.py  . I need your review to make sure that if this file includes the essence of Efield reconstruction idea or not or if there is more to include. Once we finalize this  script then I can step to create a pull request.

PR #436, opened 2022-09-09
add trigger times to HDF5 output

sjoerd-bouma:
This PR adds trigger times to the HDF5 files. Although the ray tracing times are already stored, the hdf5 file does not know which ray triggered. This means it was necessary to use the `.nur` files to determine the trigger times previously in order to determine e.g. ideal trigger delays in a detector with cable delays. This PR adds the keys
```
f['trigger_times']
f[station_id]['trigger_times_per_event']
```
to enable such a study with HDF5 output only.


cg-laser (commented):
this is a cool feature. 

Does it make sense to update the hdf5 files of the tests to also test for the new trigger times field?

sjoerd-bouma:
The `trigger_times` entries are now also included in the tests, and the reference hdf5 files updated correspondingly. 

I will add the appropriate entry in the documentation once #477 has been merged, unless Felix doesn't care about the merge conflict.

cg-laser (approved):

fschlueter:
Is it necessary to update the HDF5 version (specified in generator.py)?

sjoerd-bouma:
Sorry @cg-laser - I went over this once more and realized that the file-level `trigger_times` were not meaningful for multiple stations. Now they contain the first trigger time per trigger over all stations.

@fschlueter I have no strong opinion on this - this doesn't modify the event generation, so the output from `generator.py` doesn't actually change. But I'm happy to bump the minor version anyway if that's the consensus.

cg-laser (approved):
looks good, thanks Sjoerd. 

I don't think we need to update the hdf5 version here. The version is set in the generator and you didn't touch it.

PR #439, opened 2022-09-12
Store non-triggered events if config save_all is true

### LINKED ISSUE #438: The config "save_all" does not seem to work
From the description if set to `True` also non-triggered events should be stored. However this is not the case. In fact I do not see `save_all` to be anywhere used.

fschlueter:
Fixes #438 

This change seems to fix the issue. However, I have just re-run a simulation and saw that more events are written out! Also, event which skipped to speedup the simulation are still not stored. Is that desired? Is that information in the hdf5 files? I am not experience yet to judge whether the use case is strong enough, but  I would like to have possibilities to have all events stored (even if their signals are not properly simulated to speed up the simulation), this could be useful for efficiency calculation, right?

### DIFF
--- a/NuRadioMC/simulation/config_default.yaml
+++ b/NuRadioMC/simulation/config_default.yaml
@@ -9,6 +9,8 @@ split_event_time_diff: 1e6  # the minimal time difference (in ns) between two vo
 
 seed: 1235
 
+save_all: False  # If True store all (also non-triggered) events
+
 speedup:
   minimum_weight_cut: 1.e-5
   delta_C_cut: 0.698  # 40 degree
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -203,6 +203,8 @@ class simulation():
 
         self._antenna_pattern_provider = antennapattern.AntennaPatternProvider()
 
+        self.__number_of_skipped_events = 0
+
         # initialize propagation module
         self._prop = propagation.get_propagation_module(self._cfg['propagation']['module'])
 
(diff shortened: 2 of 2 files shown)


fschlueter:
shall I do some refactoring in the MR or better in a separated one?

fschlueter:
Oh apparently is `save_all` not in the default_config. Was this option removed because it is undesired or was this feature never really implemented and should be removed from  the manual?

anelles:
I remember that we have discussed this previously. I am not sure whether we have come to a conclusion what the intention of "all" actually was -- like which events are still skipped etc. 
Maybe something to discuss more widely.

cg-laser:
A few general comments. 

- All events are present in the input hdf5 file, so you can check that file if you need it. 
- the total number of generated neutrino interactions are still present in the output hdf5 file, so you can still calculate efficiencies, effective areas etc. just fine. 
- to get effective areas as a function of zenith angle for example, you can run simulations for separate zenith angle bins, or you just bin your results aftewards. As one should always start with isotropic incoming directions, this can be achieved easily and Steffen already implemented this rebinning in the Veff tool. 
- speedup cuts can be disabled by adapting the settings in the config file

There was a reason why we removed this feature but I don't recall why exactly. I think the problem was with secondary interactions where one neutrino interaction generates many showers where we don't have a 1:1 mapping to events.

cg-laser:
I remember the problem now. For simulations with secondaries, there will be many many showers along the muon/tau track. If all these showers are considered, it will result in a single event with a very long trace. Remember that events are created dynamically depending on the time separation between the pulses arriving at the DAQ. If all showers are considered, there will likely not be a sufficiently large time delay to create separate events.

PR #440, opened 2022-09-12
add useful function to event: has_triggered

fschlueter:
Is that useful?

### DIFF
--- a/NuRadioReco/framework/event.py
+++ b/NuRadioReco/framework/event.py
@@ -119,8 +119,6 @@ class Event:
             raise ValueError("generator information key needs to be of type NuRadioReco.framework.parameters.generatorAttributes")
         return key in self._generator_info
 
-
-
     def get_id(self):
         return self._id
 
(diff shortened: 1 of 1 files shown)


anelles:
This is intended as if any of the triggers has fired, right?
I also think this probably goes on the pile: doesn't harm. However, in the current state it also may be risky, if people don't think about the trigger type anymore.

sjoerd-bouma:
The functionality to return whether 'any' trigger has triggered is not very useful currently, as I don't think there is a way to record events that do not trigger in simulation.py. But I can see how checking for specific triggers would be a useful shortcut from time to time.

shallmann:
@sjoerd-bouma I agree, checking for specific triggers can be handy. Also, I think if the `write_all` will work as expected (another discussion started again by Felix) that might become relevant at some point as well

fschlueter:
@sjoerd-bouma with PR #439 also non-triggered events can be stored if configured.

cg-laser (approved):
useful function, thanks for adding. 
BTW: You can just pass a trigger name to the function to check for individual triggers. 

A possible improvement could be to allow passing a list of trigger names and then the function calculates the OR of all specified triggers. But this can also be implemented later if anyone needs it.

PR #441, opened 2022-09-12
add utility to convert PDG ids to names

fschlueter:
useful or superfluous?


anelles:
Doesn't harm I would think. This is related to what you discussed with @shallmann right:
`Maybe we should just search and replace all particle_id in Particle and event.py by something less ambiguous`

shallmann:
I think it is useful, and the place is better than something similar already in
 `NuRadioMC/NuRadioMC/EvtGen/NuRadioProposal.py`
 
 `NuRadioProposal` should import from the new utilities. Although I don't think they are, I didn't check if the clear names are used somewhere and re-naming could cause trouble.

shallmann:
@anelles, the discussion on `particle_id` was something different - the name `particle_id` there referred to an internal counter on MC particles (and is hence ambiguous)

 I think the new dict is useful and should replace the one in the Proposal wrapper. Can you adapt the NuRadioProposal to import particle codes @fschlueter ?

anelles:
Thanks for clearing this up. We should then probably add the docstring, since I was also confused.

fschlueter:
@shallmann done. I had to add two pseudo particles  and adapted the Manual as well. Btw, when I google for the PDG code scheme I find that only 81-100 are reserved for MC specific codes (not including 80). It might be just picky of me... but do you know if that code 80 "particle" is actually used?

fschlueter:
> Thanks for clearing this up. We should then probably add the docstring, since I was also confused.

Okay but better in a separated PR

cg-laser (commented):
I think that is a useful refactoring. A few comments

Is there an official naming scheme for the particles? Currently, it is mixed. E.g. "Tau (negative)" and "Antitau (positive)" but just "Pion (positive)" and "Pion (negative)"

Shall we make it a map? Or add a function that gives the particle code based on the name?

fschlueter:
@cg-laser, I will check if there is a official naming scheme. I could add a function  returning the ID base on the name using this dict but I do not know how efficient such a function would be:
```

for id, name in dict.items():
    if name == test_name:
        return name
```
Should we added once someone needs it?

anelles:
Yeah, I concur. Let's add a dict once someone needs it. Right now, I don't see the use case of having all names available.

fschlueter:
I did not found a official naming scheme. I found this list: https://github.com/scikit-hep/particle/blob/master/src/particle/data/pdgid_to_evtgenname.csv (Initially I also thought about using this repo but than I though having another dependency for this easy task in not necessary).

I would suggest to fix the inconsistencies you pointed out but keep most of what we have so far.

shallmann:
> @shallmann done. I had to add two pseudo particles and adapted the Manual as well. Btw, when I google for the PDG code scheme I find that only 81-100 are reserved for MC specific codes (not including 80). It might be just picky of me... but do you know if that code 80 "particle" is actually used?

I am not sure what your question is, but. "particle" exists as one of the possible outputs in proposal.
Wether 80 should be used or not I am not sure. In deed the 81 is generator specific. I don't see that 80 is used at all:
https://github.com/scikit-hep/particle/blob/v0.20.1/src/particle/data/conversions.csv

Concerning naming conventions @cg-laser, One could use the same as in scikit-hep's `particle` module, but not sure this is any better...
Screenshot 2022-09-15 at 11 45 42

shallmann:
:D didn't realise you found the same package in the meantime

cg-laser:
I think it would be good to keep standards as much as possible. Importing the scikit-hep library just for the particle names is too much, but we can just use the same strings. 

I would also opt for adding the inverse function right now. Then we don't need to touch it again. @fschlueter instead of calculating the inverse dict on the fly all the time, you can just more this to the beginning of the module where the other dict is initiated. Then the inverse map is also in memory.

fschlueter:
I rebased and force pushed this branch to recreate the original particle names which should now match the names Steffen posted

shallmann (approved):
Looks good to me

PR #443, opened 2022-09-14
Hotfix/rnog uproot readout

sjoerd-bouma:
Two small fixes for reading out RNO-G .root files with the `rnogDataReader`:
- Read out some header information event by event, speeding up the readout of single events
- Replaced the trigger time with `readout_time` from the .root files, as the `trigger_time` is incorrectly stored in some of the .root files.


shallmann (approved):
thanks Sjoerd for the quick fix. Ideally we would use trigger_time, however that parameter in the root files (not in the DAQ) produces sometimes infinities.

Fine to merge

PR #444, opened 2022-09-15
Add docstring and comment to explain particle.get_id(). Also remove default value for get_id().

fschlueter:
The change in the init function probably breaks code. But I think this is safer. Do you agree?
Cheers,
Felix


cg-laser (approved):
Thanks Felix. I agree, it is saver to remove the default value.

PR #445, opened 2022-09-19
Radiopropa/default ice

boboeyen:
Make it possible to adjust the RadioPropa ice model imbedded in a medium and set it to this new object without losing the default RadioPropa ice model configuration in the medium

### DIFF
--- a/NuRadioMC/utilities/medium.py
+++ b/NuRadioMC/utilities/medium.py
@@ -247,22 +247,22 @@ class greenland_firn(medium_base.IceModel):
         return self._scalarfield.getGradient(pos) * (1 / (units.meter/RP.meter))
 
     
-    def get_ice_model_radiopropa(self):
+    def compute_default_ice_model_radiopropa(self):
         """
-        Returns an object holding the radiopropa scalarfield and necessary radiopropa moduldes 
-        that define the medium in radiopropa. It uses the parameters of the medium object to 
-        contruct some modules, like a discontinuity object for the air boundary. Additional modules
-        can be added in this function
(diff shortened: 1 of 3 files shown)


cg-laser (commented):
Hej Bob, seems like a useful addition but I'm wondering if it is always clear to the user what he will get?

boboeyen:
> Hej Bob, seems like a useful addition but I'm wondering if it is always clear to the user what he will get? Maybe you can describe better the use case when the user needs to override the default ice model. If I see it correctly, only the python version is changed but RadioPropa will still use a different ice model? Isn't this error prone?

So the `compute_default_ice_model_radiopropa()` should always be overridden by a newly implemented ice model. This will be the function that holds what you want as a default and is a helper function for the getter. Accessing the ice model is always done through the `get_ice_model_radiopropa`, which returns to you the latest model it knows. When nothing specific is set for the object you created, the function uses the default model stored in the helper function. This way you never lose the default but you can set and get more adjusted models if you want (adding an extra boundary for example).

What do you mean with your last question. The python version is only a way to access the underlaying radiopropa ice model.

boboeyen:
I hope the added comments make it more clear.

Don't know why the documentation test is failing though :/

boboeyen:
Is this alright with you guys? Can this be merged? I need a approving review for this

anelles:
Looks good to me, but @cg-laser had an opinion, I will leave it up to him to approve.

anelles (approved):
I think they are fine now. I checked the changes and Christian's comments.

PR #446, opened 2022-09-20
improve log output of C++ ray tracer

cg-laser:
If a new attenuation model is added, the C++ ray tracer needs to be recompiled. Otherwise, one gets a cryptic error message that a certain attenuation model is not present. I added this hint to the error message.


shallmann (approved):
That log message can be very helpful!

PR #447, opened 2022-09-21
Fix wrong function call in radiopropa tracer

boboeyen:
Fix wrong function call in radiopropa tracer. Changelog was not edited since this bug was introduced by a previous new feature in the minimizer pull request


shallmann (approved):
From the config field this looks indeed more correct now!

PR #448, opened 2022-09-26
Updated sensitivity example

cg-laser:
This PR adds PUEO and beacon expected sensitivies to the plotting library and updates the icecube numu flux measurement. 
Annotations are also adjusted.


shallmann (approved):
Thanks for adding/updating those. That is certainly a good enhancement to have. You want to keep the 3 in the filename?

PR #450, opened 2022-09-30
sort ray tracing solutions by C0

sjoerd-bouma:
Currently, ray tracing solutions from the analytic ray tracing module are sorted by ray type. This is ambiguous in the case of two refracted solutions, in which the solutions will be returned in the (random?) order in which they were found by the solver. Thanks to @jannavischer for pointing this out!

This pull request orders the solutions by the `C0` parameter instead. From equation C.33 in the [NuRadioMC paper](https://arxiv.org/pdf/1906.01670.pdf#appendix.C), it can be seen that this is equivalent to sorting by the z coordinate of the turning point, i.e. the 'lower' ray will come before the 'higher' ray. This is consistent with the previous ordering in the case of distinct ray types but extends naturally to the case of two refracted rays (in particular, angles, path lengths, travel times of the i'th solution now evolve continuously as a function of vertex position).


sjoerd-bouma:
I've also replaced the warning/error due to disagreeing ray types (#470), as I think the correct 'refracted' solution will now be associated to the 'direct' solution in that case. I've replaced it with a warning if the number of reflections (or the reflector) is different, because I guess in that case the Fresnel coefficients complicate matters. Although this probably(?) never happens.

The disagreements in the reference files were only in the cases of refracted (or, for Moore's Bay, reflected) solutions, which is expected, as these would have been sorted 'randomly' previously. This PR should also make it possible to merge #318

cg-laser (approved):
thanks Sjoerd

boboeyen:
Should we implement something similar in the radiopropa raytracer?

sjoerd-bouma:
@boboeyen Good question, though I'm not quite sure how to generalize the above to more complex ice models. Maybe you could use the launch angle?
In any case, from what I can tell, there is no sorting done in the python interface to radiopropa right now, so if you'd be willing to implement this, I think that would be nice!

PR #452, opened 2022-10-11
minor fix on get_channel in envelope trigger

lpyras:
I accidentally changed that, it should work again with calling only the channel.


sjoerd-bouma:
Yeah, I noticed this recently, too, thanks for fixing. I would suggest moving this to after line 98, as well - the performance difference will be small but there's no point in bandpass filtering channels that are not in the trigger.

shallmann:
Looks good, moving down will not gain a lot, but it doesn't hurt. Good catch @sjoerd-bouma .

shallmann (approved):
Thanks!

sjoerd-bouma (approved):

Issue #391, opened 2022-03-29
Wrong showers written to `station_xxx` substructure in .hdf5 output


shallmann:
**Describe the issue**
In the hdf5 output the wrong `shower_id`s are written to the nested tables for each station. This leads to max receive vectors being all nan for some showers

**Expected behavior**
In the `station_xxx` sub-tables the shower ids should match the ones triggered in the main table

**Additional context**
I am attaching a notebook that demonstrates the issue

**To-dos**

- [ ] Find why wrong showers are written and fix

- [ ] Add test checking not all receive vectors are nan for each showers, i.e something like
```
receive_vectors = np.array(f['station_1001']['receive_vectors'])
# for each shower, see if ANY receive vector (in any antenna) is not NaN... this should never happen
[np.isnan(receive_vectors[i,:,:]).all() for i in range(np.shape(receive_vectors)[0])]
```

shallmann:
[hdf5_shower_numbering.tar.gz](https://github.com/nu-radio/NuRadioMC/files/8371428/hdf5_shower_numbering.tar.gz)

shallmann:
@lpyras maybe you can check if this is an issue and tackle this issue when simulating atm muons?

cg-laser:
Hej Steffen, I think Nans can occur under certain conditions. The fixed dimensionality of the hdf5 output arrays forces it to contain always all antennas and ray tracing solutions. It might be that certain antennas don't have any RT solutions. Then their values are set to Nan. However that should not happen is that all receive vectors of a shower are Nan. 
But you also say that the shower ids are wrong? The shower ids in the station_xxx tables should be a subset of the shower_ids on global detector level.

fschlueter:
This can be close right?

PR #437, opened 2022-09-12
Refactoring some code to obey coding style and increase readability

fschlueter:
Hi folks,
while reading through the modules in order to understand them I did some refactoring: mostly adding whitespaces which help me reading and understanding the code. Should I continue this and should this be merged or is that undesired? 
Cheers

PR #453, opened 2022-10-14
Fix indices in simulation.py

lpyras:
This PR adresses issue  #391 Wrong showers written to station_xxx substructure in .hdf5 output.
Reason for this was a wrongly defined function find_indices. This function is now replaced in line 1025.
The variables local_shower_index and global_shower_index are replaced by event_indices_of_triggered_shower_in_eventgroup and shower_indices_of_triggered_showers. They are too long, please make suggestions. 
I also added to entries to each station: 'event_id_per_shower' and 'event_group_id_per_shower' which should make it easier to match information of station level to a single particle event. 

A test to check if launch_vectors are NaN is not implemented yet.


lpyras:
I made three PR out of this on: fix_indices_sim (#454), style improvement (#437) and channel_id (#455).

PR #454, opened 2022-10-17
Fix indices in simulation.py

lpyras:
This PR addresses issue https://github.com/nu-radio/NuRadioMC/issues/391 Wrong showers written to station_xxx substructure in .hdf5 output.
Reason for this was a wrongly defined function find_indices. This function is now replaced in line 1025.
I suggest replacing the variables local_shower_index and global_shower_index are replaced by event_indices_of_triggered_shower_in_eventgroup and shower_indices_of_triggered_showers. They are too long, please make suggestions.
I also added to entries to each station: 'event_id_per_shower' and 'event_group_id_per_shower' which should make it easier to match information of station level to a single particle event.

A test to check if launch_vectors are NaN is not implemented yet.

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -974,36 +974,9 @@ class simulation():
                     triggered_showers[self._station_id].extend(self._get_shower_index(self._shower_ids_of_sub_event))
                     self._calculate_signal_properties()
 
-                    def find_indices(x, y):
-                        """
-                        finds the indices for the values `x` in array `y`
-
-                        modified from https://stackoverflow.com/questions/8251541/numpy-for-every-element-in-one-array-find-the-index-in-another-array
-                        the original solution returned a masked array which also indicated the elements in y that were
-                        not available in x. We don't need that. x will be always a subset of y, and we want only the
-                        indices in y for the subset x.
-
-                        Parameters
-                        ----------
-                        x: array
-                            the values for which the indices should be found
-                        y: array
-                            the larger array with many values
-
-                        Returns: array of integers
-                        """
-
-                        index = np.argsort(x)
-                        sorted_x = x[index]
-                        sorted_index = np.searchsorted(sorted_x, y)
-
-                        yindex = np.take(index, sorted_index, mode="clip")
-                        mask = x[yindex] != y
-                        result2 = yindex[~mask]
-                        return result2
-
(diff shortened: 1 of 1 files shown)


shallmann:
I think this simple example shows that the function that was in there did not do the right thing:
[
Screenshot 2022-11-01 at 13 31 06
](url)

cg-laser (commented):
thanks for fixing this! It seems to do the right thing. I didn't approve it yet because I think I went to this more complicated approach because the solution you implemented is slow. The arrays can be huge and you have a loop in there. This potentially slows down the simulation. Did you do any tests about the speed?

shallmann:
@cg-laser no we didn't benchmark it. @lpyras can you check?

(I don't recall if we need to keep ordering / duplicates from the `for` loop? Otherwise `np.argwhere(np.isin(event_indices, global_shower_indices)))` potentially with `assume_unique=True` may be slightly faster)

cg-laser (dismissed):

cg-laser:
good point. The shower indices are unique. If you could run a speed comparison between the different methods, that would be great. The arrays should be at least 10k long.

lpyras:
> good point. The shower indices are unique. If you could run a speed comparison between the different methods, that would be great. The arrays should be at least 10k long.

I checked, typical times for 10k showers are
time for loop  0.0031s
time for array  0.0018s
The array is a ~1.5x faster, therefore I implemented Steffens suggestion. The output of both solutions is equal.

cg-laser (approved):
thanks! Timing looks ok now. Thank for checking.

fschlueter:
Hi, I want to check for the TDR simulations how many radio events were triggered not by the first interaction (similar to what you did in https://arxiv.org/abs/2003.13442 Fig. 9 & 11) . Can I use shower_id  to do this? As far as I understand is the bug preventing this? If not, is there any other way of doing it and was this bug present for the paper above?

fschlueter:
event_group_id_per_shower would have been handy now :)]

fschlueter:
> Hi, I want to check for the TDR simulations how many radio events were triggered not by the first interaction (similar to what you did in https://arxiv.org/abs/2003.13442 Fig. 9 & 11) . Can I use shower_id to do this? As far as I understand is the bug preventing this? If not, is there any other way of doing it and was this bug present for the paper above?

Ah... I realised that I do not need the information on the station level.... sorry for the noise

PR #455, opened 2022-10-17
Separate channel id and channel indices

lpyras:
I started to disentangle channel id and channel indices


fschlueter:
Is this PR obsolet with the refactoring by @christophwelling, lets close it then.

PR #456, opened 2022-10-23
Refactoring generator.py

fschlueter:
See commit msg. Bug fix in new line 1266


fschlueter:
uhh, can it be that I checkout my hotfix branch from the wrong branch (master instead of develop)?! Is that a problem?

fschlueter:
Ah what I actually forgot to fix: Also the interaction type has to change for the electron/positron. What should it be? I found the bug because I filtered for interaction types to find the primary neutrinos ..

fschlueter:
And while we talking about it. Is it not weird that the "energy" of the produced positron is the same as the neutrino energy? In my opinion it should be equal to its "shower energy". But maybe I am misinterpreting the definition of the different quantities

shallmann:
yes the `energy` is always the primary energy (i.e. the energy before interaction) up to now. For the electron energy the shower energy is used in `simulation.py` (at least I hope) thereafter... 

I think we could/should define this better.
While we are at it, wouldn't it make sense to also redefine the energy in the proposal showers, such that "energy" is the "energy before interaction"?

fschlueter:
yes sound reasonable. I am not sure that I exactly know where to make the changes you suggested. Would you jump in?  Should we discuss this in a call?

fschlueter:
> sry that it took me a while to jump in here. This is not a bug but a convention we had so far. The confusion is probably because we don't clearly differentiate between the interacting particle and the produced showers (we don't save other interaction products). Therefore, the flavor of an EM shower is still "nu_e" and not "electron". Otherwise it will also be inconsistent with the hadronic showers where the flavor is still "neutrino" and not "hadron bundle".
This is actually not true. For taus and muons (with proposal) the flavour changes to hadronic decay bundle, .... .
Only for electrons the flavour is kept the neutrino flavour. I think this is another subject we should discuss.

fschlueter:
Hi,
I reverted change to the flavour and also changed the same of the PR to indicated that those changes now are only cosmetics.

anelles (commented):
Just one minor comment. I agree that it is mostly cosmetics at this point.

NuRadioMC/EvtGen/generator.py
```diff
@@ -1285,21 +1290,18 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
 
             E_all_leptons = (1 - data_sets["inelasticity"]) * data_sets["energies"]
             lepton_codes = copy.copy(data_sets["flavors"])
-            lepton_codes[lepton_codes == 14] = 13
-            lepton_codes[lepton_codes == -14] = -13
-            lepton_codes[lepton_codes == 16] = 15
-            lepton_codes[lepton_codes == -16] = -15
+            # convert neutrino flavors (makes only sense for cc interaction which is ensured with "mask_leptons")
+            lepton_codes = lepton_codes - 1 * np.sign(lepton_codes)
```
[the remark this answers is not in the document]
fschlueter: You mean when there is a lepton code which is not referring to a neutrino? The array is later only used with the `lepton_mask`. If that is a concern we should rather add an `assert` to verify that the flavors (which are an input) are legit.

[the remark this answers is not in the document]
fschlueter: I added a sanity check that only neutrino flavors are used within the function. I think we are good to go here.

anelles (dismissed):
looks good to me now.

NuRadioMC/EvtGen/generator.py
```diff
@@ -1279,60 +1296,56 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
             # we need to be careful to not double cound events. electron CC interactions apear twice in the event list
             # because of the two distinct showers that get created. Because second interactions are only calculated
             # for mu and tau cc interactions, this is not a problem.
-            mask_tau_cc = (data_sets["interaction_type"] == 'cc') & (np.abs(data_sets["flavors"]) == 16)
-            mask_mu_cc = (data_sets["interaction_type"] == 'cc') & (np.abs(data_sets["flavors"]) == 14)
-            mask_leptons = mask_tau_cc | mask_mu_cc
+            mask_tracks = data_sets["interaction_type"] == 'cc' & np.abs(data_sets["flavors"]) != 12
```
cg-laser, line 1299: I'm against this change. You might save a line of code but it makes it more error prone in case we might have more initial particles in the future. It is always better to be explicit than implicit.

NuRadioMC/EvtGen/generator.py
```diff
@@ -1279,60 +1296,56 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
...
 
             E_all_leptons = (1 - data_sets["inelasticity"]) * data_sets["energies"]
             lepton_codes = copy.copy(data_sets["flavors"])
-            lepton_codes[lepton_codes == 14] = 13
-            lepton_codes[lepton_codes == -14] = -13
-            lepton_codes[lepton_codes == 16] = 15
-            lepton_codes[lepton_codes == -16] = -15
+            
+            # convert neutrino flavors (makes only sense for cc interaction which is ensured with "mask_leptons")
+            lepton_codes = lepton_codes - 1 * np.sign(lepton_codes)
```
cg-laser, line 1305: same here, better to it explicit as it was. The new code is harder to understand. 
I know that you added a check at the beginning but a likely future scenario is that someone might add another intial particle type and forgets about the implicit assumptions that are done further below in the cod.e

cg-laser (changes requested):
Hey Felix, sorry for chiming in late. 

I have two comments: When refactoring code, we need to be really careful to test that it still does the same as the changes are not covered by any test. I find it difficult to check the new indexing variables just by looking at the code. Did you run a test with the old and new version? 

I'm also against code reduction at the expense of being explicit. You saved a few lines of code but are now making implicit assumptions. This is error prone for future changes. I added two in-line comments to point you to the places.

NuRadioMC/EvtGen/generator.py
fschlueter: I do not really see your point here. It should be quite easy to understand, is it really that confusing? And I have a hard time to see that one would add another initial particle type which is no neutrino to this function (without heavy rewriting it), it regardless of the above line specific to neutrinos. And even if one would accommodate a different particle type, is that desired if it makes the code harder to understand (because more  generic). We have already a different function to generate muon events

NuRadioMC/EvtGen/generator.py
fschlueter: Also here I have a hard time to see that we would want to use this function for other particles than neutrinos. But this is less explicit and I can change it again

fschlueter:
I did not check myself if the results change. I have hoped that this is covered by CI. I can do so if that is really needed

NuRadioMC/EvtGen/generator.py
fschlueter: @cg-laser do you agree with that or do you want that to be changed?

fschlueter:
The pipeline fails with an import error for proposal. Why is that?

NuRadioMC/EvtGen/generator.py
fschlueter: I reverted the changes here

fschlueter:
Okay, I resolved any merge conflict and addressed the comments (reverted a few changes). Are you okay with the current state? @cg-laser @shallmann

cg-laser:
fine with me, but you still need to resolve a conflict ;-)

fschlueter:
> fine with me, but you still need to resolve a conflict ;-)

done

cg-laser (approved):

PR #458, opened 2022-11-01
Upgrade PROPOSAL interface to PROPOSAL version 7

Jean1995:
In this PR I update the PROPOSAL interface (which is currently written for version 6) to the most-recent version of PROPOSAL (version 7). A list of all changes with the transition from PROPOSAL 6 to PROPOSAL 7 can be found in the [release notes](https://github.com/tudo-astroparticlephysics/PROPOSAL/releases). For NuRadioMC, the most important technical changes are the new configuration file format as well as the restructured output of the Propagator class. 

There are some questions that I would like to discuss, both concerning the structure of the new interface, but also the logic of old code.

- Why are the configuration files saved as `x.json.sample`, and need to be copied by the user to `x.json`? If this has to to with the interpolation tables paths that need to be adjusted - These tables path are now set in the python script, not in the configuration file anymore
- In the `get_decays` method, we return the energy sum of all decay particles, including invisible ones like neutrinos. Is this intentional? 
- In the `get_decays` method, if a lepton did not decay before it reached the lower energy limit of the simulation, the propagation is just repeated until a lepton did decay. This means that physical information ("In how many cases does my particles decay before energy X?") is lost. Is this intentional? 
- The upper energy limit of the tables that are created by PROPOSAL has been set to 1e14 MeV in the old configuration files. In the new interface, this upper limit is now set inside the script (`pp.InterpolationSettings.upper_energy_lim`). This is now a parameter that can be set when calling `ProposalFunctions`. It will also raise an error if a primary above this energy is injected. However, is the default value of 1e14 MeV sufficient? 
- At the moment, we group all decay products which are listed in `particle_names.primary_names` to a "Hadronic Decay bundle". However, we group both electromagnetic and hadronic secondaries together. For example, if a tau decays to an electron, this electron would be counted as a "Hadronic Decay Bundle". It this intentional?
- When a tau decays to a muon, we can propagate these muons and also include their secondaries in the output. Are we possibly interested in the decay products of this muon (i.e. the electron)? At the moment, this electron is neglected in the code, but it would only be two more lines to consider the electron as well.
- In [`particle_names.py`](https://github.com/nu-radio/NuRadioMC/blob/3550584f62769a91afe62d645ca71718d47b1e4e/NuRadioReco/utilities/particle_names.py#L48), "weak interaction" is listed as an interaction creating electromagnetic secondaries. However, a weak interaction should rather create hadronic primaries. Since "weak interaction" is per-default disabled as a process in PROPOSAL, this is not an urgent issue.
- Multiple Scattering has been disabled so far, and is still disabled by default. While multiple scattering should have no significant effect for energies that high, is there any other (technical) reason why it is disabled?
- In [ `particle_names.py`](https://github.com/nu-radio/NuRadioMC/blob/3550584f62769a91afe62d645ca71718d47b1e4e/NuRadioReco/utilities/particle_names.py#L8), the PDG code of photons is saved as `0` instead of `22`. Since the interface to PROPOSAL is not working with Photons as the moment, there is no problem here, but I was still wondering why this notation has been chosen. If photons would be included in the interface to PROPOSAL at some day, one would have to keep this in mind, since `0` is already defined within PROPOSAL as an "undefined particle"

I am happy about comments, code review, and discussion 🙂


shallmann:
Hi, wow, that gets picked up quickly...
will have a more detailed look later, here are a few comments:
- `...json.sample` indeed that was because of the tables paths only. So in proposal 7 it seems we can drop this, that simplifies things.
- the `get_decays` probably should be adjusted then. In normal simulations we don't typically use it. If the propagation energy lower bound is low enough it shouldn/t be such a strong bias (I guess)
- how far up can proposal7 go? Is it safe above 1e20?
- particle code we can maybe adjust

shallmann:
@Jean1995 updated my comments above.

Jean1995:
> Upper energy limit: How far up can proposal7 be trusted? Is it safe to trust cross sections above 1e20 eV? In general, 1e21 eV or even 1e22 eV would be better if possible.

From a technical point of view, there should be no limit (as long as the upper limits of the tables are set correctly). From a physical point of view, it depends on the question how far the underlying parametrizations can be extrapolated to higher energies. AFAIK, NuRadioMC has already used PROPOSAL for energies as high as 1e30 eV without a problem, so using a default such as 1e22 eV should be no problem.

> "Muon decay": could be done yes, but aren't decays with above PeV (100 TeV to be on the safe side) super rare?

I agree, they should be very rare at these energies.

> "Multiple scattering": I see no reason for not having it. One should probably doublecheck if the interaction vertices are used or recalculated from the interaction_time * c. Is there any downside, i.e. is it computationally expensive to enable?

Computationally, multiple scattering calculations should be negligible. I have just tested some NuRadioMC simulations, and the difference in the simulation time was not significant. 
Enabling multiple scattering is anyways just changing one parameter in the configuration file.

shallmann:
Unpacking the iterator is much more elegant! Good catch @Jean1995

shallmann:
@Jean1995 I think you can simply rename the `.json.sample` to `.json`. 

One additional thing I noticed when running array simulations on a cluster with proposal 6.1.6 for tau and mu at the same time required two independent table paths (so I passed a dedicated json for tau and mu to prevent failure). In PROPOSAL 7 would I still need to pass two separate table paths?

Jean1995:
> @Jean1995 I think you can simply rename the `.json.sample` to `.json`.

Done!

> One additional thing I noticed when running array simulations on a cluster with proposal 6.1.6 for tau and mu at the same time required two independent table paths (so I passed a dedicated json for tau and mu to prevent failure). In PROPOSAL 7 would I still need to pass two separate table paths?

I don't think that this should still be a problem in PROPOSAL 7, but this is something that should be tested.

fschlueter:
At the time I introduced `particle_names` I used the dicts already defined in `NuRadioProposal` not actually thinking what makes sense and not. So I guess we should fix those things. W.r.t weak interaction you refer to the fact that ~ 2/3 of the time the tau decays via the weak interaction into hadron and only ~ 1/3 into leptons?

Jean1995:
> W.r.t weak interaction you refer to the fact that ~ 2/3 of the time the tau decays via the weak interaction into hadron and only ~ 1/3 into leptons?

The `weak_interaction` is not referring to the tau decay, but to the (very rare) conversion of a charged lap lepton to a tau neutrino, where parts of tau energy is transferred to an nucleus. Therefore, I would have interpreted the (visible) part of this interaction as an hadronic cascade.

shallmann:
Yes, I think that makes sense.

documentation/source/Introduction/pages/installation.rst
```diff
@@ -189,7 +189,7 @@ These packages are recommended to be able to use all of NuRadioMC/NuRadioReco's
 
   .. code-block:: bash
 
-    pip install proposal==6.1.6
+    pip install proposal
```
cg-laser, line 192: should we add a `==7.*` here?

cg-laser (commented):
Hey everyone, I went through the changes, and it looks good to me (as far as I can tell). I didn't approve it yet, because we still need a larger production run using this branch. Janna and Steffen are running larger tests now. Can you report on the outcome here (including relevant plots)? E.g. a scatter plot of energy deposit vs. propagation distance. 
After that, I think we can merge it.

shallmann:
Hi Jean, what is the proposal 7 equivalent for `pp.particle.DynamicData(particle_def.particle_type)` that one seems to throw errors when I am trying to test it your PR version

documentation/source/Introduction/pages/installation.rst
Jean1995: Good point. Although I hope that we would notify you in advance if there is a major update of PROPOSAL planned :D

pyproject.toml
```diff
@@ -46,7 +46,7 @@ numba = "*"
 Sphinx = "*"
 sphinx-rtd-theme = "*"
 numpydoc = "1.1.0"
-proposal = "6.1.6"
+proposal = ">=7.0.0"
```
fschlueter, line 49: To keep track which tag/release of NuRadioMC used which proposal version I think we should fix that?

fschlueter (commented):
Is the description of the [returns](https://github.com/nu-radio/NuRadioMC/pull/458/files#diff-efe87c45ac930759770f94581e77f636d2d064338dac2c9901e8f9a7ac83fe27R552-R557) actually correct? I checked that with the function of proposal6 and it returned a list: `[[...]]` but not `[[...], [...]]` as I would understand the doc string

pyproject.toml
Jean1995: I'd say that we should either fix the major version (`proposal = "^7"`) so that we get all but major updates.

Or we also fix the minor version (`proposal = "~7.4"`) so that we don't automatically get all minor updates, but only fix releases.

But to be completely reproducible, the best idea is probably to always have a fixed proposal version for a fixed NuRadioMC tag/release.

shallmann: I would fix it for reproducibility (just to be on the safe side) to `proposal = "7.4.2"`

Other than that I think we can merge and then add additional changes on the NuRadio interface side in a separate PR

shallmann:
> Is the description of the [returns](https://github.com/nu-radio/NuRadioMC/pull/458/files#diff-efe87c45ac930759770f94581e77f636d2d064338dac2c9901e8f9a7ac83fe27R552-R557) actually correct? I checked that with the function of proposal6 and it returned a list: `[[...]]` but not `[[...], [...]]` as I would understand the doc string

Unless I am mistaken I think it is correct as is... I would defer future improvements to a new PR

NuRadioMC/EvtGen/NuRadioProposal.py
```diff
@@ -88,69 +85,77 @@ def __str__(self):
...
-                               1000000006: 87,
-                               1000000007: 84,
-                               1000000008: 88,
-                               1000000009: 89,
-                               1000000010: 90,
-                               1000000011: 91 }
-
-
-def particle_code(particle):
+proposal_interaction_names = { pp.particle.Interaction_Type.particle.value: 'particle',
```
Jean1995, line 95: I experienced a weird bug with these lines.
I tried to run NuRadioMC, and got the error
`AttributeError: 'proposal.particle.Interaction_Type' object has no attribute 'value'`

Took some time to figure this out: 
I have installed PROPOSAL manually (i.e. without conan) on my Ubuntu 18.04 machine.
However, Ubuntu ships with an old pybind11 version (2.4.2, from October 2019), to create the python bindings of PROPOSAL. This version does not include the `value` option for enums, since this has only been included with [pybind11/2.6.2](https://github.com/pybind/pybind11/releases/tag/v2.6.2).

This means that these lines will not work with people who built PROPOSAL with a pybind11 version older than 2.6.2 (this will not be the case for people using pip or conan, but only for users who manually built and installed PROPOSAL with cmake, AND who have an outdated pybind11 version).

What we can do:

1. Keep this with `.value`, which might cause problems in the very-specific case I described above
2. Just go back to how it was before this PR, so explicitly writing the numbers in the code
3. Write `int(pp.particle.Interaction_type.brems)` instead, which appears to be working in both versions
4. ... other ideas?

Jean1995:
> Is the description of the [returns](https://github.com/nu-radio/NuRadioMC/pull/458/files#diff-efe87c45ac930759770f94581e77f636d2d064338dac2c9901e8f9a7ac83fe27R552-R557) actually correct? I checked that with the function of proposal6 and it returned a list: `[[...]]` but not `[[...], [...]]` as I would understand the doc string

Nothing has changed about the return type of this function, and my runs with proposal6 also returned a "list of lists" with `NuRadioProposal.SecondaryProperties` objects, so the type that I would expect from the docstring.

Jean1995:
I realized there are some other scripts and files that would need to be updated. For example, in `EvtGen\generator.py` the docstring still says that one can change the path for the proposal tables in the config file, however, the path is now set directly in `ProposalFunctions`. This means that the scripts there should also have an argument with the proposal_tables_path that will be passed to PROPOSAL. 

This would also need to be updated in the tests (of which some currently fail because the config files have not been updated). 

Should this also be included in this PR?

fschlueter:
> > Is the description of the [returns](https://github.com/nu-radio/NuRadioMC/pull/458/files#diff-efe87c45ac930759770f94581e77f636d2d064338dac2c9901e8f9a7ac83fe27R552-R557) actually correct? I checked that with the function of proposal6 and it returned a list: `[[...]]` but not `[[...], [...]]` as I would understand the doc string
> 
> Nothing has changed about the return type of this function, and my runs with proposal6 also returned a "list of lists" with `NuRadioProposal.SecondaryProperties` objects, so the type that I would expect from the docstring.

The comment was meant for proposal 6 & 7, I left it here because this MR is so active. My issue is with

>  **The first dimension indicates the primary lepton** and the second dimension
            navigates through the secondaries produced by that primary.
        
For me it seems that no information of the primary lepton are returned, correct?. (Sorry for being so anal)

Jean1995:
> For me it seems that no information of the primary lepton are returned, correct?. (Sorry for being so anal)

No, it does not. If `return` is the returned array, then `return[i][:]` just means that we are looking at the secondaries that belong to the primary with the properties `energy_leptons_nu[i]`, `lepton_codes[i]`, etc.

If this is not clear from the current docstring, I am open for an improved description 😄

fschlueter:
> > For me it seems that no information of the primary lepton are returned, correct?. (Sorry for being so anal)
> 
> No, it does not. If `return` is the returned array, then `return[i][:]` just means that we are looking at the secondaries that belong to the primary with the properties `energy_leptons_nu[i]`, `lepton_codes[i]`, etc.
> 
> If this is not clear from the current docstring, I am open for an improved description 😄

That makes sense. And yes I would say the doctoring is not clear! Btw, is there a reason why we use the for loop to propagate all leptons and not in fact give the proposal functions arrays of leptons (apparently they can handle them). Is the performance improvements not significant?

Jean1995:
> And yes I would say the doctoring is not clear! 

Maybe you could make a suggestion to improve the clarity of the doctring? I feel a bit biased since it was clear to me, however, I can definitely understand that it is not for first-readers of the docstring 😅  

> Btw, is there a reason why we use the for loop to propagate all leptons and not in fact give the proposal functions arrays of leptons (apparently they can handle them). Is the performance improvements not significant?

Do you mean [this for-loop](https://github.com/nu-radio/NuRadioMC/pull/458/files#diff-efe87c45ac930759770f94581e77f636d2d064338dac2c9901e8f9a7ac83fe27L570-L571)? `proposal.Propagator.propagate()` can (at the moment) only propagate one particle at the same time. You can't pass an array of particles to PROPOSAL, only a single one.

fschlueter:
> Maybe you could make a suggestion to improve the clarity of the doctring? I feel a bit biased since it was clear to me, however, I can definitely understand that it is not for first-readers of the docstring 

Can try.

> Do you mean [this for-loop](https://github.com/nu-radio/NuRadioMC/pull/458/files#diff-efe87c45ac930759770f94581e77f636d2d064338dac2c9901e8f9a7ac83fe27L570-L571)? proposal.Propagator.propagate() can (at the moment) only propagate one particle at the same time. You can't pass an array of particles to PROPOSAL, only a single one.

Ah sorry did not check carefully enough. What I mean `get_secondaries_array` is "designed" to take arrays for several leptons at a time (you can actually also just give floats and it will extend them. But then internally it will loop over them. We somehow do quite one back and forth conversions but never mind

shallmann:
> I realized there are some other scripts and files that would need to be updated. For example, in `EvtGen\generator.py` the docstring still says that one can change the path for the proposal tables in the config file, however, the path is now set directly in `ProposalFunctions`. This means that the scripts there should also have an argument with the proposal_tables_path that will be passed to PROPOSAL.
> 
> This would also need to be updated in the tests (of which some currently fail because the config files have not been updated).
> 
> Should this also be included in this PR?

Done!

NuRadioMC/EvtGen/NuRadioProposal.py
shallmann: Is this still an issue?

shallmann:
@cg-laser @fschlueter ... after fixing the tests and docstring/tables paths in the generator, all tests are running. I went through the comments above and my notes and tried to see if we are missing something before merging/upgrading to proposal7:

- [ ] https://github.com/nu-radio/NuRadioMC/pull/458#discussion_r1082356985 this might still be a last open issue in the particle class? Not sure? Maybe @fschlueter you can check?

- [ ] We wanted to add installation instructions for proposal7 on systems where `pip install` currently does not work, I tentatively included this in the `documentation/source/Introduction/pages/install.rst`. @Jean1995  is the link to the step-by-step guide persistent or are you planning to include this in the proposal install instructions? Alternatively we can copy paste the linked contents to the `install.rst` ...

- [ ] Originally (I thought) our plan was to create a tag with the final proposal6.1.8, then upgrade the interface to proposal7 (i.e. this PR) and only then enhance/adapt the output information from the interface. @fschlueter in the meanwhile has opened [this]( https://github.com/nu-radio/NuRadioMC/pull/478) PR on proposal6 to enhance/adapt the output already there. If we proceed with that, we should create the final proposal6 tag only after [PR478]( https://github.com/nu-radio/NuRadioMC/pull/478) is merged and hold of this PR until we make sure (@fschlueter) that the changes are also correctly implemented here.

Otherwise I think this PR would be done.

NuRadioMC/EvtGen/NuRadioProposal.py
fschlueter: I would go for option 3 for the time being and add an appropriate comment

fschlueter:
> Originally (I thought) our plan was to create a tag with the final proposal6.1.8, then upgrade the interface to proposal7 (i.e. this PR) and only then enhance/adapt the output information from the interface. @fschlueter in the meanwhile has opened https://github.com/nu-radio/NuRadioMC/pull/478 PR on proposal6 to enhance/adapt the output already there. If we proceed with that, we should create the final proposal6 tag only after https://github.com/nu-radio/NuRadioMC/pull/478 is merged and hold of this PR until we make sure (@fschlueter) that the changes are also correctly implemented here.

The implementation in #478 is "ready". However, there is still the issue with the interaction_type item and I have not extensively tested the output yet. 
Alternatively we could push this PR forward, implement the changes of #478 in a 6.1.8 branch and tag later?

fschlueter:
Should we address `this:`

        # TODO: At the moment, all decay_products in primary_names are grouped and identified
        #       as a 'Hadronic Decay bundle', even electromagnetic decay products such as
        #       electrons. Is that ok? Should they be negleceted? Or grouped separately?

NuRadioMC/EvtGen/NuRadioProposal.py
Jean1995: I think it is an issue, although a minor one. I would agree with Felix.

Jean1995:
> * [ ]  We wanted to add installation instructions for proposal7 on systems where `pip install` currently does not work, I tentatively included this in the `documentation/source/Introduction/pages/install.rst`. @Jean1995  is the link to the step-by-step guide persistent or are you planning to include this in the proposal install instructions? Alternatively we can copy paste the linked contents to the `install.rst` ...

I have added these instructions page to a wiki page on the PROPOSAL GitHub under this link:

https://github.com/tudo-astroparticlephysics/PROPOSAL/wiki/Installing-PROPOSAL-on-a-Linux-kernel---4.11

Feel free to reference this wiki page, or to just copy the instructions directly to your `install.rst`. I won't claim any copyrights for it 😄

NuRadioMC/EvtGen/NuRadioProposal.py
fschlueter: Agreed? @Jean1995 @shallmann

shallmann: yes, agreed @fschlueter ... are you going to change it or should I? We need to change all the left hand sides in the dict to something like `int(pp.particle.Interaction_type.brems)`, right?

fschlueter: can do. Yes, I think this is it.

fschlueter:
Hey @shallmann, I used a couple of commits. Let me know what you think.

[f22ef63](https://github.com/nu-radio/NuRadioMC/pull/458/commits/f22ef63ab3732615afe92e5ae77de1ef8b8b05b0): make code much nicer. I did not check if it is slower (would that be a concern?)

[1f923c](https://github.com/nu-radio/NuRadioMC/pull/458/commits/1f923c01d3caa3cb51f48d2dacaa8a71c02ba840): Changes where the tables are stored. This might be a bit invasive but I do not think that tmp is a good location. In particular for notebooks which once in a while are shut down.

NuRadioMC/EvtGen/NuRadioProposal.py
```diff
@@ -40,6 +40,10 @@
 pp_km = 1.e5
 
 
+def calculate_distance(pp_position, pos_arr):
```
shallmann, line 43: indeed makes code much nicer, but unfortunately also WRONG if I am not mistaken! ;) pp_position is in cm!

shallmann (changes requested):
please @fschlueter check if units are correct. 
Maybe let's defer further beautifications to a separate PR to keep changes traceable...

shallmann:
> Hey @shallmann, I used a couple of commits. Let me know what you think.
> 
> [f22ef63](https://github.com/nu-radio/NuRadioMC/pull/458/commits/f22ef63ab3732615afe92e5ae77de1ef8b8b05b0): make code much nicer. I did not check if it is slower (would that be a concern?)

`np` functions should not be very slow, performance is fine.

> 
> [1f923c](https://github.com/nu-radio/NuRadioMC/pull/458/commits/1f923c01d3caa3cb51f48d2dacaa8a71c02ba840): Changes where the tables are stored. This might be a bit invasive but I do not think that tmp is a good location. In particular for notebooks which once in a while are shut down.

The default to `None` change I like very much. The default location is a matter of taste. Originally we chose a temporary directory so that no cleanup is necessary; and the user still has the possibility to pass a persistent directory in the arguments if required. But I agree that table creation takes long and redoing it should be avoided where possible. IMO this is a good default, additional ~20MB of tables in the installation dir shouldn't be a problem. Maybe we could add a `.gitignore` with `proposal_tables/*.dat` just to avoid accidential commits?

Jean1995:
We have just released version 7.5.0 of PROPOSAL (see [here](https://github.com/tudo-astroparticlephysics/PROPOSAL/releases/tag/7.5.0)).

Apart from some bugfixes (e.g. [out-of-the-box compatability with Python3.11](https://github.com/tudo-astroparticlephysics/PROPOSAL/pull/325) when using `pip install proposal`), there are [changes for the LPM effect parametrization](https://github.com/tudo-astroparticlephysics/PROPOSAL/pull/336), which may have a small effect for simulations at very-high-energies. 

We might directly want to include 7.5.0 in this PR, or have a separate PR for that (although I don't see a reason that speaks against the first approach)

shallmann:
> We have just released version 7.5.0 of PROPOSAL (see [here](https://github.com/tudo-astroparticlephysics/PROPOSAL/releases/tag/7.5.0)).
> 
> Apart from some bugfixes (e.g. [out-of-the-box compatability with Python3.11](https://github.com/tudo-astroparticlephysics/PROPOSAL/pull/325) when using `pip install proposal`), there are [changes for the LPM effect parametrization](https://github.com/tudo-astroparticlephysics/PROPOSAL/pull/336), which may have a small effect on simulations at very-high-energies.
> 
> We might directly want to include 7.5.0 in this PR, or have a separate PR for that (although I don't see a reason that speaks against the first approach)

Yes thanks, I saw that already yesterday and successfully pip-installed it on my Mac. The LPM comparison plots are also helpful. Just to be sure: the LPM is enabled by default in our interface, right?

Jean1995:
>  Just to be sure: the LPM is enabled by default in our interface, right?

It needs to explicitly enabled in the configuration files, which it is, see for example here:

https://github.com/nu-radio/NuRadioMC/pull/458/files#diff-fab1a995316b54d60dfec6e2bbc12ff7d3ac8fdc4a08eacee0f95202a645275fR14

NuRadioMC/EvtGen/NuRadioProposal.py
fschlueter: You are right, `pp_position` is in proposal units. But I thought `pos_arr`, which (after checking) seems not to be the case... 
I just tried to replace the old code which is the one below which is identical (wrong?) to what I wrote, right?
```
distance = ((sec.position.x - lepton_position[0]) * units.cm) ** 2
distance += ((sec.position.y - lepton_position[1]) * units.cm) ** 2
distance += ((sec.position.z - lepton_position[2]) * units.cm) ** 2
distance = np.sqrt(distance)
```

fschlueter:
> np functions should not be very slow, performance is fine.

What I noticed is, that when you do `np` on single floats (for which they are of course not meant), they are much slower than the build-in functions or from `math`.... But I guess this is okay

fschlueter:
> Maybe we could add a .gitignore with proposal_tables/*.dat just to avoid accidential commits?

Will do

NuRadioMC/EvtGen/NuRadioProposal.py
fschlueter: Apropos units. I do not quite understand where we should use proposal units and where NuRadioMC units. Things like `SecondaryProperties` are in NuRadioMC units. So the distances should be converted to them (i.e. multiplied with units.cm). I think the problem is that the `lepton_position` has not been converted to proposal units?

fschlueter: And how could that not be catched by the CI or did it?

shallmann:
> > np functions should not be very slow, performance is fine.
> 
> What I noticed is, that when you do `np` on single floats (for which they are of course not meant), they are much slower than the build-in functions or from `math`.... But I guess this is okay

True, but this is an array calculation... anyway `math` vs `np` should almost never be the bottleneck even if you are sure it is floats (I guess)... I don't really see the advantage of using `math`, since we anyway install `np`

shallmann:
> > Just to be sure: the LPM is enabled by default in our interface, right?
> 
> It needs to explicitly enabled in the configuration files, which it is, see for example here:
> 
> https://github.com/nu-radio/NuRadioMC/pull/458/files#diff-fab1a995316b54d60dfec6e2bbc12ff7d3ac8fdc4a08eacee0f95202a645275fR14

Right, great! Thanks for the reminder where this was set ;)

NuRadioMC/EvtGen/NuRadioProposal.py
shallmann: oh sorry, my bad. yes you were indeed right! Which underlines that your code is much nicer to read now. Thanks!

shallmann (dismissed):

NuRadioMC/EvtGen/NuRadioProposal.py
shallmann: concerning units, it depends if it is an internal function starting with `__` or not:
```
class ProposalFunctions(object):
    """
    This class serves as a container for PROPOSAL functions. The functions that
    start with double underscore take PROPOSAL units as an argument and should
    not be used from the outside to avoid mismatching units.
    """
```

fschlueter: But now I am confused: `lepton_positions_nu` is not converted into proposal units (i.e., cm) at the beginning of function `get_secondaries_array` but it should?

fschlueter: And should we rename `calculate_distance` to `__calculate_distance` because it takes proposal units? But then it returns NuRadioMC units which make sense when the distances it calculates it returns but its awkward.

fschlueter:
Shall we merge that right now or do we wait another 24h?

shallmann (approved):
We decided it is good to go.

shallmann:
> @cg-laser @fschlueter ... after fixing the tests and docstring/tables paths in the generator, all tests pass. I went through the comments above and my notes and tried to see if we are missing something before merging/upgrading to proposal7:
> 
> * [x]  [Upgrade PROPOSAL interface to PROPOSAL version 7 #458 (comment)](https://github.com/nu-radio/NuRadioMC/pull/458#discussion_r1082356985) this might still be a last open issue in the particle class? Not sure? Maybe @fschlueter you can check?
> * [x]  We wanted to add installation instructions for proposal7 on systems where `pip install` currently does not work, I tentatively included this in the `documentation/source/Introduction/pages/install.rst`. @Jean1995  is the link to the step-by-step guide persistent or are you planning to include this in the proposal install instructions? Alternatively we can copy paste the linked contents to the `install.rst` ...
> * [x]  Originally (I thought) our plan was to create a tag with the final proposal6.1.8, then upgrade the interface to proposal7 (i.e. this PR) and only then enhance/adapt the output information from the interface. @fschlueter in the meanwhile has opened [this](https://github.com/nu-radio/NuRadioMC/pull/478) PR on proposal6 to enhance/adapt the output already there. If we proceed with that, we should create the final proposal6 tag only after [PR478](https://github.com/nu-radio/NuRadioMC/pull/478) is merged and hold of this PR until we make sure (@fschlueter) that the changes are also correctly implemented here.
> 
> Otherwise I think this PR would be done.

The open questions are done. Felix will implement point 3 directly on proposal 7

fschlueter (approved):

PR #463, opened 2022-11-09
Simulation.py cleanup

christophwelling:
Some code cleanup.
Removes the deprecated parameter write_mode from the simulation class __init__
Others are just styling changes that do not affect behavior.


sjoerd-bouma (approved):
Thanks for helping crush the bracket inflation.

PR #465, opened 2022-11-15
Removes unneeded files from simulation folder

christophwelling:
Removes some detector descriptions and example scripts that did not belong there from the NuRadioMC/simulation folder.


cg-laser (commented):
we still need the T04-T07 scripts. Do we have them somewhere else? 
I agree, the rest can be deleted.

christophwelling:
I made a new PR (#491 ) where the scripts are moved into the examples folder.

PR #467, opened 2022-11-23
Docu hdf5 output

lpyras:
I added a sketch how events are generated and explained event_group_id, shower_id and event_id. I also added an explanation for event_group_id_per_shower and event_id_per shower to the table.


documentation/source/NuRadioMC/pages/HDF5_structure.rst
```diff
@@ -32,7 +47,8 @@ The HDF5 files can be thought of as a structured dictionary:
 - Finally, the ``station_<station_id>`` key contains slightly more detailed information (triggers, propagation times, amplitudes...) at the level of individual channels :ref:`for each station <NuRadioMC/pages/HDF5_structure:Station data>`.
 
 HDF5 file attributes
-____________________
+----------------------
+
```
sjoerd-bouma, line 51: The line style controls the level of the heading - I would have left `f.attrs` as a subheading of the HDF5 structure heading, at the same level as `f[key]` and `f[station]`, but I'm not going to argue if you disagree.

sjoerd-bouma (dismissed):
Thanks for the update! Not sure how I missed the `event_id_per_shower` keys, but I think it's definitely clearer now.

You'll need to merge in the latest `develop` if you want the tests to pass, but it seems you can merge even if they don't.

documentation/source/NuRadioMC/pages/HDF5_structure.rst
lpyras: I am happy to change that back. I think I just wanted to not throw an error.

sjoerd-bouma (approved):

PR #468, opened 2022-11-23
run actions on pull requests also + move tests to ubuntu 20.04

sjoerd-bouma:
see title

`ubuntu-latest` has been updated to 22.04, meaning our unit tests fail until #458 has been merged. This PR reverts the unit tests to 20.04 until that happens. We should remember to go back to `ubuntu-latest` once this is the case.

The second issue I also noticed based on #458  - our actions only run on our own branches. It's _probably_ useful to run them on PRs from forks, too? (Note that, as far as I understand, actions still will not run *by default* on PRs from forks due to the potential security risk, but anyone with write permission should be able to click a button to allow them to run in that case now).

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -8,18 +8,18 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2.4.0
+    - uses: actions/checkout@v3
     - name: Get the master branch (for checks)
-      uses: actions/checkout@v2.4.0
+      uses: actions/checkout@v3
       with:
         ref: 'refs/heads/master'
 
     - name: Set up Python 3.7
-      uses: actions/setup-python@v2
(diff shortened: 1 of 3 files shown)


shallmann:
Thanks for the quick fix Sjoerd! Looks good to merge to me

shallmann (dismissed):

sjoerd-bouma:
Sorry Steffen - it seems in order to avoid running tests twice (they trigger both on 'pull request' and on 'push', a more complicated fix was needed. I took the opportunity to also update our 'marketplace actions' (like /checkout, /cache...) as we were getting deprecation warnings on the old ones.

I realize this has made this PR a little bit more cumbersome. If you prefer, I can split out the temporary `ubuntu-20.04` fix and turn this into a separate PR so we can discuss an appropriate solution for the second issue (running tests on PRs from forked repositories).

shallmann (approved):

Issue #420, opened 2022-07-28
Missing antenna "KU_dipole"


christophwelling:
The detector description `NuRadioReco/detector/ARIANNA/arianna_detector_db.json` contains the antenna model "KU_dipole". This model seems to not exist and causes an error.

shallmann:
Maybe @cg-laser has any idea if the KU_dipole model was renamed / who still has it? The updated list of available antenna models is here https://nu-radio.github.io/NuRadioMC/NuRadioReco/pages/detector/antennamodels.html

PR #473, opened 2022-12-08
Docu/antenna model

shallmann:
Updates antenna model names to match the ones available on the cluster.

### DIFF
--- a/documentation/source/NuRadioReco/pages/detector/antennamodels.rst
+++ b/documentation/source/NuRadioReco/pages/detector/antennamodels.rst
@@ -17,7 +17,7 @@ the antenna model so that he can put it on our central server.
 Inf means usually an infinite medium, firn has a refractive index n = 1.3-1.4. Since this is a bit imprecise
 we changed the naming to the actual refractive index.
 
-bicone_v8_infAir
+bicone_v8_InfAir
 -----------------
 WIPL-D simulation of ARA Bicone antenna.
 This antenna has been used by ARIANNA at the South Pole.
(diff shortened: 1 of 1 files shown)


shallmann:
I want the models listed here to reflect what is on the server - We were just asked the ~5th time why a certain antenna model can not be loaded - the reason always being missing `n`'s or slightly different filenames.
 
Not sure who is best to review these changes - the changes I made should change adjust filename changes to reflect what is on the cluster;
I am not the right person to revisit existing documentation for existing models, and add new documentation to the additional models.

shallmann:
@anelles do you have a suggestion who might fill in information on the models without documentation? If we don't I would tend to just merge it - non-detailed documentation is certainly less confusing than wrong documentation

anelles:
If I am not mistaken, all the bicone models are from Chicago, likely @dansmithphysics ? The LPDA models should be from UCI, but probably only @cg-laser and @ggaswint know details. 
Yes, I think we maybe a bit late in this game for thorough documentation. SO unless there is suddenly a flurry of activity,  I agree, it is more useful to merge and  create an issue.

dansmithphysics:
Hi all, 

The RNO-G dipole antennas are saved in a repo here: https://github.com/dansmithphysics/rnog_vpol_fat_dipole. There is some amount of documentation and simulation output on them here: https://radio.uchicago.edu/wiki/index.php/Science_Antennas#Downhole_V-Pol_Antenna 

I have a lot less information on the ARA bicones, but I did import the model into xFDTD (but was never able to simulate it since they were very slow simulations). Those files are saved here: https://github.com/dansmithphysics/ara_bicone

Hope that helps!

christophwelling:
While we are talking about missing antenna models, there is also this issue: #420  about a missing antenna model in one of the detector descriptions.

shallmann:
Ok, from @dansmithphysics linked repos I don't see additional information that could be put into the documentation now. The KU_dipole is a separate issue, but worth following up. Will merge the updated antenna documentation as is.

PR #475, opened 2023-01-09
Fix propagation of secondary muons from tau decay in NuRadioProposal.py

Jean1995:
When updating the NuRadioMC interface to PROPOSAL 7, we figured out that there was a disagreement in the energy loss spectra of taus.

After some debugging, we found out that the spectra do agree if we disable the propagation of secondary muons in the new interface, and with that, we learned that in the old version of the interface, secondary muons are not propagated at all!

This PR fixes the bug in the interface with proposal 6. The bug will also be fixed in PR #458.

### Validation plots

As a validation, these are the energy losses for the propagation of 1e5 tau leptons with an initial energy of 1e11 MeV, using the greenland configuration file.

**Current development branch:**

image

**This PR:**

image

**PR #458 (i.e. PROPOSAL version 7), with secondary muon simulation enabled**:

image

**PR #458 (i.e. PROPOSAL version 7), with secondary muon simulation disabled**:

image

### DIFF
--- a/NuRadioMC/EvtGen/NuRadioProposal.py
+++ b/NuRadioMC/EvtGen/NuRadioProposal.py
@@ -578,19 +578,13 @@ class ProposalFunctions(object):
 
             # Checking if there is a muon in the products
             if propagate_decay_muons:
-
                 decay_muons_array.append([None, None, None, None])
 
                 for sec in secondaries:
-
-                    if (sec.type not in proposal_interaction_codes) or (sec.type not in particle_names.particle_names):
-                        continue
-
-                    if (sec.particle_def == pp.particle.MuMinusDef()) or (sec.particle_def == pp.particle.MuPlusDef()):
-
-                        if sec.particle_def == pp.particle.MuMinusDef():
+                    if (sec.type == 13) or (sec.type == -13):
+                        if sec.type == 13:
                             mu_code = 13
-                        elif sec.particle_def == pp.particle.MuPlusDef():
+                        elif sec.type == -13:
                             mu_code = -13
 
                         mu_energy = sec.energy
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,7 +11,7 @@ new features:
 - add uniform ice model with 1.78 refractive index
 - update ARA detector description and add a file to Print ARA detector json file
 bugfixes:
-- fix bug in NuRadiProposal which pervented muons from tau decays to be propagated
+- fix bug in NuRadioProposal which pervented muons from tau decays to be propagated
 
 
 version 2.1.6
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,6 +11,7 @@ new features:
 - add uniform ice model with 1.78 refractive index
 - update ARA detector description and add a file to Print ARA detector json file
 bugfixes:
+- fix bug in NuRadiProposal which pervented muons from tau decays to be propagated
 
 
 version 2.1.6


shallmann (approved):
@Jean1995 this looks good to merge to me. For completeness (do you still remember?): before the muons were never propagated and skipped.

Was that due to this
` if (sec.type not in proposal_interaction_codes) or (sec.type not in particle_names.particle_names)` line or
the statements like those: `(sec.particle_def == pp.particle.MuMinusDef()`

shallmann:
also, @Jean1995 can you still edit the changelog.txt and add
a `bugfixes:` line plus description that decay muons were not propagated before under the next future tag (`2.2`)

Jean1995:
> @Jean1995 this looks good to merge to me. For completeness (do you still remember?): before the muons were never propagated and skipped.
> 
> Was that due to this ` if (sec.type not in proposal_interaction_codes) or (sec.type not in particle_names.particle_names)` line or the statements like those: `(sec.particle_def == pp.particle.MuMinusDef()`

It was due to the line 

```
    if (sec.type not in proposal_interaction_codes) or (sec.type not in particle_names.particle_names):
```

because the if condition was always fullfilled: Every secondary of PROPOSAL is either an energy loss (so the first part is true) or a decay particle (so the second part is true). This means that the following code, which repropagates the muons, has always been skipped.

The code afterwards has just been changed by me because it doesn't work: `sec` does not have a member `particle_def`. It just wasn't noticed because the code has never been reached.

shallmann (approved):
Very good! Thanks for clarifying and fixing!

cg-laser:
very good that you tracked it down. I agree this can/should be merged but I was wondering if we shouldn't add a test that tests for a correct propagation? Or is this not feasible because the random sequence will be different between different proposal version?

shallmann:
I guess the random sequence could be different for different proposal versions and complicate comparison against a pre-computed particle list, but @Jean1995 maybe has more insight.

If we only want to test if the tau decay works, we can check wether the number of secondary particles is longer if muon propagation is enabled (depending on decay mode equal or strictly longer), something like

```
import numpy as np
from NuRadioMC.EvtGen.NuRadioProposal import ProposalFunctions
import proposal

propf = ProposalFunctions(config_file='config_PROPOSAL_infice.json')
#'InfIce')

# generate one test event per cycle
n_test_events = 1
proposal_args = {
        "energy_leptons_nu": np.full(n_test_events, 1e18*units.eV),
        "lepton_codes": np.random.choice((-15, 15),n_test_events),
        "lepton_positions_nu": None,
        "lepton_directions": None,
        "low_nu": 0.5 * units.PeV,
        "propagation_length_nu": 1000 * units.km,
        "min_energy_loss_nu": 0.5 * units.PeV,
}

diffs = []
for i in range(100):
    # note: seed should not be 0, to avoid random behaviour
    
    proposal.RandomGenerator.get().set_seed(i+1234)
    secondaries_propagated = propf.get_secondaries_array(
          propagate_decay_muons=True, **proposal_args)
    
    proposal.RandomGenerator.get().set_seed(i+1234)
    secondaries_not_propagated = propf.get_secondaries_array(
          propagate_decay_muons=False, **proposal_args)

    diffs.append(len(secondaries_propagated[0])-len(secondaries_not_propagated[0]))

diffs = np.array(diffs)
np.sum(diffs>0), np.sum(diffs==0), np.sum(diffs<0)
```

For proposal 6 yielding here `(21, 79, 0)`, i.e. test for np.sum(diffs>0) > 0 and np.sum(diffs<0) == 0

I would defer this to the proposal7 PR

Jean1995:
> I guess the random sequence could be different for different proposal versions and complicate comparison against a pre-computed particle list, but @Jean1995 maybe has more insight.

This is true. While results of individual calculations should stay consistent between versions, we can not guarantee that the result of entire propagation processes will stay the same (because a single bugfix might change the order that random numbers are called).

I like the idea of checking the length of the propagation output.

shallmann:
Test is to be included in the proposal7 upgrade PR.

PR #476, opened 2023-01-10
return useful debug message if .nur file is empty

lpyras:
I added a file size check in NuRadioReco/modules/io/NuRadioRecoio.py and added an elif for version 0.0 in  NuRadioReco/modules/io/event_parser_factory.py. Please contribute if you have more ideas.

### DIFF
--- a/NuRadioMC/EvtGen/NuRadioProposal.py
+++ b/NuRadioMC/EvtGen/NuRadioProposal.py
@@ -578,19 +578,13 @@ class ProposalFunctions(object):
 
             # Checking if there is a muon in the products
             if propagate_decay_muons:
-
                 decay_muons_array.append([None, None, None, None])
 
                 for sec in secondaries:
-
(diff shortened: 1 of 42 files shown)


christophwelling (changes requested):
Could you make this more robust and also check if the file version is a reasonable number?
The reason is that if NuRadioReco tries to read in a corrupted file, it often interprets some random byte as the version, which often results in very weird version numbers.
So could you check if `abs(self.__file_version) <= VERSION` and give out a warning that the file is probably corrupted otherwise?

cg-laser:
@lpyras did you unintentionally commit many other changes? Could you clean this up so that we can properly review it?

lpyras:
> @lpyras did you unintentionally commit many other changes? Could you clean this up so that we can properly review it?

I think I rebased it. Sorry, this changes are already merged into develop.

PR #478, opened 2023-01-16
Modify hdf5 simulations files. 

fschlueter:
The flavor contains now the type of th…e parent particle. The interaction type contains now, in case of a charge lepton, the type of energy loss.

Implement the recently contained changes to the hdf5 files: 

- [x] `flavors` contains now the PDG particle code of the parent (neutrinos or muon, tau leptons)
- [x] `interaction_type` not anymore contains the neutrino interaction type ("cc" or "nc") but can also contain the PDG code for the shower creating secondary (when a charge lepton produces a shower)
- [x] Store the parent energy (directly before energy loss) in `energies`

A few thoughts:
 - Do we want/need to bump the version number 
 - Is it a problem to store strings and int both in `interaction_type`

### DIFF
--- a/NuRadioMC/EvtGen/generator.py
+++ b/NuRadioMC/EvtGen/generator.py
@@ -1370,8 +1370,8 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
                                 n_interaction += 1
                                 data_sets_fiducial['shower_energies'][-1] = product.energy
                                 data_sets_fiducial['inelasticity'][-1] = np.nan
-                                # interaction_type is either 'had' or 'em' for proposal products
-                                data_sets_fiducial['interaction_type'][-1] = product.shower_type
+                                # interaction_type is particle codes taken from NuRadioProposal.py
+                                data_sets_fiducial['interaction_type'][-1] = product.code
                                 data_sets_fiducial['shower_type'][-1] = product.shower_type
 
                                 data_sets_fiducial['xx'][-1] = x
@@ -1381,8 +1381,8 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
                                 # Calculating vertex interaction time with respect to the primary neutrino
                                 data_sets_fiducial['vertex_times'][-1] = vertex_time
 
-                                # Flavors are particle codes taken from NuRadioProposal.py
-                                data_sets_fiducial['flavors'][-1] = product.code
+                                # Flavors from parent lepton
+                                data_sets_fiducial['flavors'][-1] = lepton_codes[iE]
             time_proposal = time.time() - init_time
         else:
             if(n_batches == 1):
--- a/NuRadioMC/EvtGen/generator.py
+++ b/NuRadioMC/EvtGen/generator.py
@@ -1340,33 +1340,13 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
                                                           [data_sets['xx'][iE], data_sets['yy'][iE], data_sets['zz'][iE]],
                                                           lepton_directions[iE])
                     if geometry_selection:
-                        products_array, leptons_energies = proposal_functions.get_secondaries_array(
+                        products_array = proposal_functions.get_secondaries_array(
                             np.array([E_all_leptons[iE]]), np.array([lepton_codes[iE]]),
                             np.array([lepton_positions[iE]]), np.array([lepton_directions[iE]]),
-                            return_lepton_energy=True,
                             propagate_decay_muons=False,
                             **proposal_kwargs)
-                        
+
                         products = products_array[0]
-                        print(len(products), len(leptons_energies))
-                        
-                        if 0:
-                            from matplotlib import pyplot as plt
-                            fig, ax = plt.subplots()
-                            
-                            energy_proxy = E_all_leptons[iE] - np.cumsum([p.energy for p in products])
-                            ax.plot(np.array(leptons_energies) * 1e6, label="parent energy")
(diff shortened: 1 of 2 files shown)


NuRadioMC/EvtGen/generator.py
```diff
@@ -1370,8 +1370,8 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
                                 n_interaction += 1
                                 data_sets_fiducial['shower_energies'][-1] = product.energy
                                 data_sets_fiducial['inelasticity'][-1] = np.nan
-                                # interaction_type is either 'had' or 'em' for proposal products
-                                data_sets_fiducial['interaction_type'][-1] = product.shower_type
+                                # interaction_type is particle codes taken from NuRadioProposal.py
+                                data_sets_fiducial['interaction_type'][-1] = product.code
```
shallmann, line 1374: I don't think this line does what we want.
The `interaction_type` being `cc` or `nc` might play a role in what the Askaryan model makes out of it - but I haven't checked.


 Was unaware that you are taking over implementing these changes. @jannavischer in that case please coordinate and make sure you are not doing things twice.
 
  I think you want to get the flavor of the `parent` not the product. Might make sence to wait until the proposal 7 branch is merged - since the interface changes

shallmann (changes requested):

NuRadioMC/EvtGen/generator.py
fschlueter: For the interaction type I thought we want the particle code of the produce (to know what initiated the shower). The flavour/code of  the parent is stored in the item "flavor"

@jannavischer, sorry I did not know that you were working on this. As part of writing a tool to convert nur to hdf5 files I thought starting with this actually. I am currently trying to understand how to get the energy of the muon and tau from proposal exactly before the energy loss

> The interaction_type being cc or nc might play a role in what the Askaryan model makes out of it - but I haven't checked.

Are you sure? Or do you confuse it with shower_type (which has to be em or had).

fschlueter: > Might make sence to wait until the proposal 7 branch is merged - since the interface changes

hm very good point.

fschlueter:
Hello @shallmann and @jannavischer, I was a little bit too impatient and started with getting the lepton energy. I get something which seems somewhat reasonable (see attached plot) but not jet correct (other plots do not look so good). There are some subtleties when shower secondaries are merged... maybe we can discuss this on Monday?
lepton_energy_379

fschlueter:
> Hello @shallmann and @jannavischer, I was a little bit too impatient and started with getting the lepton energy. I get something which seems somewhat reasonable (see attached plot) but not jet correct (other plots do not look so good). There are some subtleties when shower secondaries are merged... maybe we can discuss this on Monday?
> lepton_energy_379

For explanation, the blue curve shows what I am not getting from proposal as the energy of the lepton, the orange curve shows what one can calculate when one subtract the shower energies from the initial lepton energy (this ignores continues energy losses and/or energy losses below 1 PeV, hence it is expected that the two curves do not agree)

fschlueter:
@cg-laser, with these changes we store str ("nc", "cc") and integer (pdg codes of shower primaries (e.g. 83 or 85) in the same item `interaction_type`. Do you think that is okay? Would you rather have that avoided?

shallmann:
I would avoid having different types in one variable. Aren't we mixing `interaction-type identifiers` with `product codes` then? (Sorry I am slow here)

According to the existing documentation:
```
``interaction_type``, the interaction type. `'cc'` for charged current, and `'nc'` for neutral current. `'tau_had', 'tau_em', 'tau_mu'` indicate the tau decays into the hadronic, electromagnetic and muonic channels respectively, calculated using the CSDA for taus. If the vertex has been created by a secondary particle propagated by PROPOSAL, `had` indicates a hadronic shower, and `em` an electromagnetic shower.
```
it was always a string before; at least for the `tau_...` the only function that relies on that naming convention is `inelasticity_tau_decay(tau_energy, branch, rnd=None)` and that one is not used anywhere in the processing chain.

Not sure what is best to do.

fschlueter:
> I would avoid having different types in one variable. Aren't we mixing `interaction-type identifiers` with `product codes` then? (Sorry I am slow here)
> 
> According to the existing documentation:
> 
> ```
> ``interaction_type``, the interaction type. `'cc'` for charged current, and `'nc'` for neutral current. `'tau_had', 'tau_em', 'tau_mu'` indicate the tau decays into the hadronic, electromagnetic and muonic channels respectively, calculated using the CSDA for taus. If the vertex has been created by a secondary particle propagated by PROPOSAL, `had` indicates a hadronic shower, and `em` an electromagnetic shower.
> ```
> 
> it was always a string before; at least for the `tau_...` the only function that relies on that naming convention is `inelasticity_tau_decay(tau_energy, branch, rnd=None)` and that one is not used anywhere in the processing chain.
> 
> Not sure what is best to do.

Okay it's exactly the last sentence in documentation we want to change. 
Last call, we agreed storing the particle code in the `interaction_type` which was previously stored (for secondary showers) in `flavor`. However, to be more consistent it makes sense to store the code of the parent particle (in this case the mu/tau lepton) in `flavor`. Hence, to keep all information we decided to store the code of the shower-initiating particle in `interaction_type`. Keep in mind that all other information are stored already. 

To "fix" the problem with the data type we could choose to not store the particle code (integer) but the particle names:
`particle_names.particle_name(product.code)`

fschlueter:
This PR will be closed in the favour of #486 which does the same implementation directly for proposal7. 
This implementation is almost complete, just convert the `product.code` to a particle name to store it in `interaction_type`

PR #479, opened 2023-01-16
Update proposal dependency to 6.1.8 to avoid problems with pypi

Jean1995:
Installing proposal 6.1.6 on ubuntu22 (and other distributions) has been problematic due to a broken submodule. See [this issue](https://github.com/tudo-astroparticlephysics/PROPOSAL/issues/331) for details, including the current workaround.

I've created a "legacy release" 6.1.8, where this issue has been fixed, so a simple `pip install proposal==6.1.8` should be working now.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -12,6 +12,7 @@ new features:
 - update ARA detector description and add a file to Print ARA detector json file
 bugfixes:
 - fix bug in NuRadioProposal which pervented muons from tau decays to be propagated
+- update proposal version to 6.1.8 to avoid problems with pypi
 
 
 version 2.1.6
--- a/documentation/source/Introduction/pages/installation.rst
+++ b/documentation/source/Introduction/pages/installation.rst
@@ -189,7 +189,7 @@ These packages are recommended to be able to use all of NuRadioMC/NuRadioReco's 
 
   .. code-block:: bash
 
-    pip install proposal==6.1.6
+    pip install proposal==6.1.8
 
 - To use the channelGalacticNoiseAdder, you need the `PyGDSM <https://github.com/telegraphic/pygdsm>`_ package.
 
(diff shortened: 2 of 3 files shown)


anelles:
Thanks a lot for the custom support here!
I will leave @shallmann to scope out the details (FYI, he is on vacation this week).

fschlueter:
We should merge that into develop? Or is the plan to soon merge the proposal7 branch?

Jean1995:
> We should merge that into develop? Or is the plan to soon merge the proposal7 branch?

If I remember correctly, we wanted to tag a "final version" of NuRadioMC which uses proposal6 before merging the proposal7 branch.
And this final tag should use proposal version 6.1.8 (i.e. this PR), to that users using the tagged NuRadioMC version won't have the problem with pypi and proposal 6.1.6. Note that there are no other differences between 6.1.6 and 6.1.8.

fschlueter:
Out of curiosity, did the update to proposal7 cause the unit tests to fail? I.e., are we sensitive to minor changes in proposal?

anelles:
@fschlueter, if you have tested it and checked the code, then review it and then @Jean1995  can merge it tomorrow ;)

fschlueter:
I have checked the branch out and installed it (not more). Since the unit tests are running I consider everything fine.

fschlueter (approved):

Jean1995:
> Great, all tests are running through. So I think there is no reason to not merge this?

Not from my side, this should be good to merge.



> Out of curiosity, did the update to proposal7 cause the unit tests to fail? I.e., are we sensitive to minor changes in proposal?

I guess this is rather related to #458 there it looks like some tests are failing, but to me it looks like some paths are still incorrect. @shallmann told me that he wanted to have at look at this.

shallmann (approved):
That is good to merge!

@cg-laser - can you generate a tag after this ... then we can proceed and also merge the proposal7 PR

PR #482, opened 2023-01-21
implement __str__ for particle class

fschlueter:
I thought this is fun to do. I included the `hex(id(self))` to be able to identify the object (as with the default __str__). Any other suggestions? particle id, parent id, ...


shallmann (approved):
Good idea, I like it. I think the info is fine, unless you find anything in addition useful, eg. the `particle id` or the `vertex` ... maybe not much more than that to keep it short

PR #483, opened 2023-01-21
Fix doc strings which cas deployment of docu to fail

fschlueter:
Sometimes the deployment of the doc strings is skipped sometimes not (and then it fails). Not sure why but maybe we just fix it.

### DIFF
--- a/NuRadioMC/EvtGen/generate_unforced.py
+++ b/NuRadioMC/EvtGen/generate_unforced.py
@@ -161,7 +161,7 @@ def generate_eventlist_cylinder(filename, n_events, Emin, Emax,
         calculate distance to center of Earth as a function of travel distance
         
         Parameters
-        -----------
+        ----------
         t: 3dim array
             travel distance
         v: 3dim array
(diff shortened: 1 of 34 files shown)


shallmann:
Hmm.. don't know about the documentation, but the `Deploy Documentation / build (pull_request) ` seems to be skipped and does not run through when kicked manually if I am not mistaken?

shallmann (approved):
These changes are all ok to me

fschlueter:
Did you init it manually now?

sjoerd-bouma (approved):
Thanks Felix!
@shallmann - the documentation test (build) is run on either `push` or `pull request`, in this case because it runs on `push` it skips the one on pull request.

FYI, having underlines too long (rather than too short) raises an error only with numpydoc>=1.2, I was too lazy to fix all of them previously, which is the reason we stuck with numpydoc 1.1. Maybe we can upgrade now, if they're all fixed?

fschlueter:
> FYI, having underlines too long (rather than too short) raises an error only with numpydoc>=1.2, I was too lazy to fix all of them previously, which is the reason we stuck with numpydoc 1.1. Maybe we can upgrade now, if they're all fixed?

Why was I getting an error than?

PR #484, opened 2023-01-24
Fix bug which caused that the first interaction of a triggered event …

fschlueter:
…was not stored when the interaction ifself did not trigger

This little typo caused the issue that the first interaction was not stored even if the event triggered. 

I think such code a 'la `if "key" in f:` is dangerous and we should avoid it. Can we add this to the validation (i.e., do we validate a simulation with propagated leptons?)

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -1370,7 +1370,7 @@ class simulation:
             # a reference! saved indicates the interactions to be saved, while
             # triggered should indicate if an interaction has produced a trigger
             saved = np.copy(self._mout['triggered'])
-            if 'n_interactions' in self._fin:  # if n_interactions is not specified, there are not parents
+            if 'n_interaction' in self._fin:  # if n_interactions is not specified, there are not parents
                 parent_mask = self._fin['n_interaction'] == 1
                 for event_id in np.unique(self._fin['event_group_ids']):
                     event_mask = self._fin['event_group_ids'] == event_id


fschlueter:
That I do not understand yet: when inspecting hdf5 files from the TDR library I find showers which are not from a first interaction (they have a `vertex_time > 0`) but have a `n_interaction == 1`
Tested by running `np.any(n_interaction[vertex_time > 0] == 1)` of a random file. However, in the input hdf5 files I do not find those...

cg-laser (approved):
very good that you found it. This is indeed a difficult to find bug... I agree that such statements are potentially dangerous. Do you have a suggestion to avoid it?

fschlueter:
> very good that you found it. This is indeed a difficult to find bug... I agree that such statements are potentially dangerous. Do you have a suggestion to avoid it?

In this particular case: Do we even need the if condition? Do we want to process files which do not have the item?

fschlueter:
> That I do not understand yet: when inspecting hdf5 files from the TDR library I find showers which are not from a first interaction (they have a `vertex_time > 0`) but have a `n_interaction == 1`
> Tested by running `np.any(n_interaction[vertex_time > 0] == 1)` of a random file. However, in the input hdf5 files I do not find those...

In the file `tau_18.00eV_0.0_0.1.hdf5` from the TDR library, I find `n_interaction == 1` in 2270 out of 7419 showers with `vertex_time > 0`. Now, I simulated 1e4 neutrino events which results in 202 triggered showers with `vertex_time > 0`. Non of those had  `n_interaction == 1`. Hence, I consider this issue also resolved.

PR #486, opened 2023-01-26
Changes to the hdf5 generator output files for proposal7

fschlueter:
This PR implements the same as #478 directly for proposal 7.

I also refactored the imports.

This PR is based on PR #458 and can be merged once this in merged.

### DIFF
--- a/NuRadioMC/EvtGen/NuRadioProposal.py
+++ b/NuRadioMC/EvtGen/NuRadioProposal.py
@@ -42,11 +42,6 @@ pp_m = 1.e2
 pp_km = 1.e5
 
 
-def convert_energy_units(pp_energy):
-    """ Convert energy from proposal unit to NuRadioMC unit. """
-    return pp_energy / pp_MeV * units.MeV
-
-
 class SecondaryProperties:
     """
     This class stores the properties from secondary particles that are
@@ -513,13 +508,13 @@ class ProposalFunctions(object):
 
                 distance = ProposalFunctions.__calculate_distance(sec.position, lepton_position)
 
-                energy = convert_energy_units(sec.energy)
+                energy = sec.energy / pp_MeV * units.MeV
 
                 shower_type, code, name = self.__shower_properties(sec.type)
                 
                 shower_inducing_prods.append(
                     SecondaryProperties(distance, energy, shower_type, code, name, 
-                                        convert_energy_units(sec.parent_particle_energy)))
+                                        sec.parent_particle_energy / pp_MeV * units.MeV))
 
         return shower_inducing_prods
 
@@ -564,7 +559,7 @@ class ProposalFunctions(object):
             # all decay_particles have the same position, so we can just look at the first in list
             distance = ProposalFunctions.__calculate_distance(decay_products[0].position, lepton_position)
 
-            return SecondaryProperties(distance, convert_energy_units(sum_decay_particle_energy), 
+            return SecondaryProperties(distance, sum_decay_particle_energy / pp_MeV * units.MeV,
                                        'had', 86, particle_names.particle_name(86), 
                                        decay_energy)
         return None
@@ -667,7 +662,7 @@ class ProposalFunctions(object):
                         # We have already handled the muon, remove it to avoid double counting.
                         decay_products.remove(decay_particle)
 
-            decay_energy = convert_energy_units(secondaries.final_state().energy)  # energy of the lepton before decay
+            decay_energy = secondaries.final_state().energy / pp_MeV * units.MeV  # energy of the lepton before decay
             grouped_decay_products = self.__group_decay_products(decay_products, min_energy_loss, lepton_position, decay_energy)
 
             if grouped_decay_products is not None:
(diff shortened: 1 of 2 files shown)


fschlueter:
@Jean1995 you might want to check too.

fschlueter:
@shallmann

NuRadioMC/EvtGen/NuRadioProposal.py
```diff
@@ -508,15 +511,18 @@ def __filter_secondaries(self,
                 energy = sec.energy * units.MeV
 
                 shower_type, code, name = self.__shower_properties(sec.type)
-
-                shower_inducing_prods.append(SecondaryProperties(distance, energy, shower_type, code, name))
+                
+                shower_inducing_prods.append(
+                    SecondaryProperties(distance, energy, shower_type, code, name, 
+                                        sec.parent_particle_energy * units.MeV))
```
cg-laser, line 517: is this unit conversion save? This only works if the base unit it MeV. It is cleaner if you also divide by `pp_MeV` defined in this file.

cg-laser (commented):
looks good but a quick comment about units.

NuRadioMC/EvtGen/NuRadioProposal.py
fschlueter: A few lines above the same is done with `sec.energy`. So if we want to do it we should be consistent and do it everywhere

NuRadioMC/EvtGen/NuRadioProposal.py
```diff
@@ -42,35 +42,44 @@
...
     def __str__(self):
-        s = "Particle and code: {:} ({:})\n".format(self.name, self.code)
-        s += "Energy in PeV: {:}\n".format(self.energy / units.PeV)
-        s += "Distance from vertex in km: {:}\n".format(self.distance / units.km)
-        s += "Shower type: {:}\n".format(self.shower_type)
+        s =  "Particle and code    : {:} ({:})\n".format(self.name, self.code)
+        s += "Energy               : {:} PeV\n".format(self.energy / units.PeV)
+        s += "Distance from vertex : {:} km\n".format(self.distance / units.km)
+        s += "Shower type          : {:}\n".format(self.shower_type)
+        s += "Parent energy        : {:} PeV".format(self.parent_energy / units.PeV)
```
shallmann, line 82: distances/energies maybe easier to read with `{:.2e}` if crossing many energy scales?

NuRadioMC/EvtGen/NuRadioProposal.py
```diff
@@ -42,35 +42,44 @@
 pp_km = 1.e5
 
 
+def convert_energy_units(pp_energy):
```
shallmann, line 45: naming is not clear in which direction that goes... don't think this function actually helps

NuRadioMC/EvtGen/NuRadioProposal.py
```diff
@@ -654,7 +667,8 @@ def get_secondaries_array(self,
                         # We have already handled the muon, remove it to avoid double counting.
                         decay_products.remove(decay_particle)
 
-            grouped_decay_products = self.__group_decay_products(decay_products, min_energy_loss, lepton_position)
+            decay_energy = convert_energy_units(secondaries.final_state().energy)  # energy of the lepton before decay
```
shallmann, line 670: I did not check... but "final state" sounds suspicious? I hope you verified that this is the correct quantity?

NuRadioMC/EvtGen/NuRadioProposal.py
fschlueter: Its already in units of PeV and km, I think it is okay like this (can be improved later)

NuRadioMC/EvtGen/NuRadioProposal.py
fschlueter: yeah I agree...

NuRadioMC/EvtGen/NuRadioProposal.py
fschlueter: I have to admit I relied on what @Jean1995 told me.

fschlueter: okay, I x-checked `decay_energy` with 
```
 decay_energies = np.array([p.energy for p in secondaries.decay_products()])
 decay_energy2 = np.sum(decay_energies) / pp_MeV * units.MeV
```
if the particle decays they agree perfectly, otherwise not but otherwise the variable is also not used!

fschlueter: @shallmann fine?

cg-laser (approved):
looks good from my side, thanks

NuRadioMC/EvtGen/NuRadioProposal.py
Jean1995: Yes, final_state() returns the last particle state of the propagated particle, so the energy immediately before the particle decayed (or before the propagation stopped for any other reason).

Jean1995:
LGTM

PR #487, opened 2023-01-30
Feature/proposal table downloader

shallmann:
Table creation for proposal 7 is super slow and takes ~1h extra for the "Greenland" config in the tests. Here I provide the option to download pre-calculated tables from the DESY storage.

### DIFF
--- a/NuRadioMC/EvtGen/NuRadioProposal.py
+++ b/NuRadioMC/EvtGen/NuRadioProposal.py
@@ -1,6 +1,7 @@
 import proposal as pp
 import numpy as np
 from NuRadioReco.utilities import units, particle_names
+from NuRadioReco.utilities.metaclasses import Singleton
 import os
 import six
 import json
@@ -38,17 +39,6 @@ pp_ZeV = 1.e15
 
 pp_m = 1.e2
 pp_km = 1.e5
-
-
-
-
-class Singleton(type):
-    _instances = {}
-
-    def __call__(cls, *args, **kwargs):
-        if Singleton._instances.get(cls, None) is None:
-            Singleton._instances[cls] = super(Singleton, cls).__call__(*args, **kwargs)
-        return Singleton._instances[cls]
 
 
 class SecondaryProperties:
--- a/NuRadioMC/EvtGen/NuRadioProposal.py
+++ b/NuRadioMC/EvtGen/NuRadioProposal.py
@@ -260,6 +260,7 @@ class ProposalFunctions(object):
         self.__propagators = {}
         self.__config_file = config_file
         self.__tables_path = tables_path
+        self.__tables_downloaded = False
         self.__upper_energy_limit = upper_energy_limit * pp_eV # convert to PROPOSAL units
 
 
@@ -319,6 +320,12 @@ class ProposalFunctions(object):
                     "path {} exists.".format(config_file_full_path))
 
             pp.InterpolationSettings.tables_path = self.__tables_path
+            # download pre-calculated tables for default configs, but not more than once
+            if self.__config_file in config_files and not self.__tables_downloaded:
+                from proposal_tables_manager import download_proposal_tables
+
+                download_proposal_tables(self.__config_file, tables_path=self.__tables_path)
+                self.__tables_downloaded = True
 
             # upper energy lim for proposal tables, in PROPOSAL units (MeV)
             pp.InterpolationSettings.upper_energy_lim = self.__upper_energy_limit
--- a/NuRadioMC/EvtGen/NuRadioProposal.py
+++ b/NuRadioMC/EvtGen/NuRadioProposal.py
@@ -1,7 +1,7 @@
 import proposal as pp
 import numpy as np
 from NuRadioReco.utilities import units, particle_names
-from NuRadioReco.utilities.metaclasses import Singleton
+import NuRadioReco.utilities.metaclasses
 import os
 import six
 import json
@@ -203,7 +203,7 @@ def is_shower_primary(pp_type):
         return False
 
 
-@six.add_metaclass(Singleton)
+@six.add_metaclass(NuRadioReco.utilities.metaclasses.Singleton)
 class ProposalFunctions(object):
     """
     This class serves as a container for PROPOSAL functions. The functions that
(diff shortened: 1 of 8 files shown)


shallmann:
A description on how to produce tables I also put on the DESY cluster: 

# Pre-calculated PROPOSAL tables for default NuRadioMC detector configs
## Directory contents
Pre-calculated tables exist for each version of imported proposal (eg. 7.5.0) and
should be re-generated if a newer version of proposal is used

The tables unique sha numbers depend on the compiler used, hence separate calculation
for both (eg. on a Mac and an Ubuntu machine) is needed as of v.7.5.0.
An explanation for this behaviour is summarized here:  https://github.com/tudo-astroparticlephysics/PROPOSAL/issues/174

## HowTo: Add new set of tables to NuRadioMC data storage
Execute
```
python NuRadioMC/EvtGen/proposal_table_manager.py create "all" --tables_path <your_preferred_path>
```
and put the tarballs of the resulting tables in `<your_preferred_path>` to the download directory on the DESY cluster.
If you do not want to produce `"all"` tables, this may be replaced by a specific default configuration, eg. `"Greenland"`.

The functionality to download these tables is implemented in `proposal_table_manager.py`. 
This is also used in `NuRadioProposal.py` for the default configurations.

shallmann:
I put tables for both clang and gcc on the DESY cluster:
```
[wgs16] /pnfs/ifh.de/acs/radio/diskonly/NuRadioMC/proposal_tables % tree
.
|-- README.md
`-- v7.5.0
    |-- clang
    |   |-- Greenland.tar.gz
    |   |-- InfIce.tar.gz
    |   |-- MooresBay.tar.gz
    |   `-- SouthPole.tar.gz
    `-- gcc
        |-- Greenland.tar.gz
        |-- InfIce.tar.gz
        |-- MooresBay.tar.gz
        `-- SouthPole.tar.gz

3 directories, 9 files
```

shallmann:
Time comparison:

- webinar examples test:
  -  NOW (w/ download): `1:39 min`
  - Proposal7 (no download): `49 min`
  - Proposal6 (no download): `3 min`

- atm. Aeff example test:
  -  NOW (w/ download): `1:57 min`
  - Proposal7 (no download): `23 min`
  - Proposal6 (no download): `3 min`

So tests are fast again, and the automatic download in NuRadioProposal removes the 15-45 min waiting time to build proposal tables when first running on of the 4 default configs in a proposal tables path.

From my side this is ready.

NuRadioMC/EvtGen/NuRadioProposal.py
```diff
@@ -325,6 +320,12 @@ def __get_propagator(self, particle_code=13):
                     "path {} exists.".format(config_file_full_path))
 
             pp.InterpolationSettings.tables_path = self.__tables_path
+            # download pre-calculated tables for default configs, but not more than once
+            if self.__config_file in config_files and not self.__tables_downloaded:
+                from NuRadioMC.EvtGen.proposal_table_manager import download_proposal_tables
+
+                download_proposal_tables(self.__config_file, tables_path=self.__tables_path)
+                self.__tables_downloaded = True
```
sjoerd-bouma, line 328: If I understand correctly, this means e.g. every time a new Python script is launched, the tables will be redownloaded? Is that intentional?

NuRadioMC/EvtGen/proposal_table_manager.py
```diff
@@ -0,0 +1,138 @@
...
+    tables_path: string
+        output path for the generated tables
+    """
+
+    if tables_path is None:
+        proposal_func = ProposalFunctions(config_file=config_file, create_new=True)
+        tables_path = proposal_func._ProposalFunctions__tables_path
+
+    download_file = True
+    if download_file:
```
sjoerd-bouma, line 92: consider removing

NuRadioMC/EvtGen/NuRadioProposal.py
```diff
@@ -248,7 +238,11 @@ def __init__(self, config_file='SouthPole', log_level=logging.INFO, tables_path=
             upper_energy_limit of tables that will be created by PROPOSAL, in NuRadioMC units (eV).
             There will be an error if primaries with energies above this energy will be injected.
             Note that PROPOSAL will have to regenerate tables for a new values of upper_energy_limit
-
+        create_new: bool (default:False)
+            Can be used to force the creation of a new ProposalFunctions object.
+            By default, the __init__ will only create a new object if none already exists.
+            For more details, check the documentation for the
+            `Singleton metaclass <NuRadioReco.utilities.html#NuRadioReco.utilities.metaclasses.Singleton>`_.
```
sjoerd-bouma, line 245: This doesn't work. To cross-reference a class, you need to use the `:class:` reference (see [the documentation](https://nu-radio.github.io/NuRadioMC/Introduction/pages/contributing.html#links-and-cross-references)), which also doesn't have a trailing underscore. (rst is a pain, don't blame me!)
Alternatively, to have  a 'normal' hyperlink, you need to link to the actual webpage (which starts with https://...), at the risk that this will break at some point.

sjoerd-bouma (commented):
Thanks Steffen! Two very minor comments/questions, then I think this can be merged.

NuRadioMC/EvtGen/NuRadioProposal.py
shallmann: hmm... yes maybe not too wise... especially when launching thousands of jobs.

I wanted to avoid that a downloader script must be run manually. 
Do you have a good suggestion to check if latest set of tables exists already (should only change if we swap to different proposal version at some point)?

 ... without thinking in detail, one could eg. create a subdirectory structure `<proposal_tables>/Greenland` (or a more detailed one not omitting the compiler and proposal version used) or dumping the already downloaded paths to a `<proposal_tables>`/downloaded_tables.csv after successful download and checking that?

sjoerd-bouma: My first idea was also along the lines of your first suggestion (although I would include the proposal version, at least, to avoid unexpected behaviour in the future if something should change), though I guess keeping a record of downloaded proposal tables (probably with hashes?) also works. 
The former means things get a bit trickier if the user wants to modify the default download path, the latter would lead to errors if the user manually (re)moves tables. I think I would be fine with either.

shallmann: Nothing would lead to errors. Worst case tables are re-computed if not found / if table hashes changed.

The list with hashes you proposed to check for equivalence is a bit difficult because we are downloading and extracting a tarball. But then I'll go with the first and put default tables to a sub-directory structure `<tables_path>/<config_file>`

NuRadioMC/EvtGen/NuRadioProposal.py
shallmann: should be fixed

NuRadioMC/EvtGen/NuRadioProposal.py
shallmann: With the implemented sub-directory structure the `<config_file>` part in the `self.__tables_path` now set in the init can easily be adapted should version/etc for the download ever become an issue in the future.

PR #488, opened 2023-02-07
Fix set_trigger_time

jannavischer:
Change of `set_trigger_time` such that it is always in relation to the time of interaction and not (as previously in some cases and in the documentation) with respect to the beginning of the trace.

Additionally I fixed two typos in the **NuRadioProposal as a standalone module** example, I came across.

### DIFF
--- a/NuRadioReco/modules/phasedarray/triggerSimulator.py
+++ b/NuRadioReco/modules/phasedarray/triggerSimulator.py
@@ -569,7 +569,7 @@ class triggerSimulator:
             #trigger_time(s)= time(s) from start of trace + start time of trace with respect to moment of first interaction = trigger time from moment of first interaction; time offset to interaction time (channel_trace_start_time) already recognized in self.phased_trigger
             trigger.set_trigger_time(trigger_time)# 
             trigger.set_trigger_times(trigger_times)
-	else:
+        else:
             trigger.set_trigger_time(None)
 
         station.set_trigger(trigger)
--- a/NuRadioReco/modules/phasedarray/triggerSimulator.py
+++ b/NuRadioReco/modules/phasedarray/triggerSimulator.py
@@ -333,9 +333,9 @@ class triggerSimulator:
             the delays for the primary channels that have caused a trigger.
             If there is no trigger, it's an empty dictionary
         trigger_time: float
-            the earliest trigger time with respect to interaction time.
+            the earliest trigger time with respect to first interaction time.
         trigger_times: dictionary
-            all time bins that fulfil the trigger condition per beam. The key is the beam number. Time with respect to interaction time.
+            all time bins that fulfil the trigger condition per beam. The key is the beam number. Time with respect to first interaction time.
         """
 
         if(triggered_channels is None):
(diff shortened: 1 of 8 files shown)


shallmann (changes requested):
Thanks for fixing the documentation here!
Most importantly, revert the duplicated added trace start time in one of the functions, rest is just cosmetics

jannavischer:
@shallmann I implemented your comments and tested trigger.py again. It should work now how it should.

fschlueter:
Sorry for being so anal...
is "time w.r.t. interaction time" the best description? I would immediately ask first or secondary interaction. I guess it should always be the first/primary interaction?!

shallmann (changes requested):

cg-laser:
> Sorry for being so anal... is "time w.r.t. interaction time" the best description? I would immediately ask first or secondary interaction. I guess it should always be the first/primary interaction?!

I agree "w.r.t. first interaction" is the better descirption. @jannavischer if you implement this quick change, the PR should be good to be merged.

sjoerd-bouma:
I guess the tests failing is expected, if the trigger times are now defined differently? If this is the case, the reference file should also be updated.
(The `Trigger tests` currently seem to fail)
```
Property trigger_time of trigger simple_phased_threshold differs from reference

Not equal to tolerance rtol=1e-07, atol=0

x and y nan location mismatch:
 x: array([nan, nan, nan, nan, nan, nan, nan, nan, nan])
 y: array([ 6939.478555,  6886.510438, 15029.516978,  8868.88317 ,
       13204.640001,  6756.579553, 10906.32335 ,  8502.376915,
        2521.756223])
```

cg-laser:
having `nan` as the trigger time seems more like a bug doesn't it?

shallmann (commented):
Let me see why tests are failing.


Trigger modules have two internal variables "_trigger_time" and "_trigger_times", during init both are set to nan. The latter is only set in some modules (eg. where the whole trace is tested for multiple positions where the signal is above threshold)... If trigger times are not set, we think the function `get_trigger_times` should return the single trigger time rather than nan.

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -564,11 +564,11 @@ def run(self, evt, station, det,
...
 
         trigger.set_triggered(is_triggered)
-
+        
         if is_triggered:
-            trigger.set_trigger_time(trigger_time)
+            #trigger_time(s)= time(s) from start of trace + start time of trace with respect to moment of first interaction = trigger time from moment of first interaction; time offset to interaction time (channel_trace_start_time) already recognized in self.phased_trigger
+            trigger.set_trigger_time(trigger_time)# 
             trigger.set_trigger_times(trigger_times)
-        else:
```
shallmann, line 571: I overlooked that before, but you accidentially deleted this line. I.e. all trigger times are now set to Nan

shallmann (changes requested):

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -564,11 +564,12 @@ def run(self, evt, station, det,
...
+	else:
```
sjoerd-bouma, line 572: Hi @jannavischer, I hope you can see the test output? It seems you used spaces instead of tabs (or the other way round) here:
```
./NuRadioReco/modules/phasedarray/triggerSimulator.py:572:6: E999 TabError: inconsistent use of tabs and spaces in indentation
	else:
	    ^
1     E999 TabError: inconsistent use of tabs and spaces in indentation
1
Error: Process completed with exit code 1.```

which means this code doesn't run.

shallmann (approved):

PR #489, opened 2023-02-07
RNO-G detector description for 2022

christophwelling:
This adds a description of the RNO-G detector as of 2022. The scripts to build the detector description are here: https://github.com/RNO-G/analysis-tools/tree/main/detector_json
Antenna positions are taken from the 2022 and 2021 GPS surveys. Unfortunately, there were issues with the GPS both years, so some positions won't be too accurate, but it's the best we have for now.
Depths are taken from tape measure readings during deployment and should be pretty accurate.
Fiber delays are from Kaeli's measurements, should also be very accurate. Cable delays for surface antennas assume same delay for all cables, then adds the extra delays for jumper cables.


shallmann:
I think this is useful to merge, to at least have a first estimate of the antenna positions (although indeed for the LPDAs it seems they are not the final ones) . 
Just to be sure:

- [ ] you took care also of the antenna orientations for the LPDAs right?
- [ ] the "old" stations' positions are not the same as in the 2021, right? Which one do you trust more? Maybe it is worth linking to a comparison (if possible... I guess someone did compare?)?
- [ ] we don't want to include rough positions of the other "devices" (DAQ-Box, Solar panels) here - that is ok for me, as we anyway should prioritize changing to the DB readout detector

shallmann (dismissed):

christophwelling:
Hi,
this is our best guess for now. We probably want to switch to using the DB, but that's gonna take a bit of time, and is nice to have something in the meantime. I will need to do an update though. I didn't know we also had measurements for the  surface cables and just put in generic values. But I have found the measurements and will add them.

- For the 2021 stations, we measured the orientation during deployment, and I used those notes. For the new stations, I assumed they are oriented perpendicular to the cables. I think the reason some of them (and for station 24 this is definitely the reason) look odd is because the GPS positions are off.
- The reason the station positions are different is just a different coordinate system. The old detector description was relative to the GPS basestation, while this one is relative to the DISK hole. But the positions are the exact same.
- I didn't add them yet, because I'm not sure if this is really the best place to put them, and I didn't have these positions for  the old stations.

christophwelling:
Sorry for the delay. I fixed the mislabeling of the surface LPDAs, and the cable delays for the surface channels are now based on the pre-deployment measurements.

PR #492, opened 2023-02-20
Update RNO-G amp responses

christophwelling:
NuRadioMC is still using the first version of the RNO-G amplifiers, which is different from the ones installed in the field.
This PR uses one calibration measurement to represent the response of all channels, after applying a Savitsky-Golay filter for smoothing to get rid of noise effects.
amp_responses_iglu
amp_responses_surface


fschlueter (approved):
Hey, I am not sure if I am already in the position to approve those changes already (judge for your self:-)) but on the other hand those changes seem pretty straight forward.

If you want you could use a list as argument for `usecols` in `np.loadtxt` to have this function called just ones but I guess efficiencies is not a main concern here.

sjoerd-bouma (approved):
Thanks, I think this will be very helpful!

fschlueter:
A general question. It seems that no test failed after changing the response. Is the hardwareResponseIncoperator validated in some test for a different response? Does it make sense to include it in the testing?

PR #493, opened 2023-02-22
Update tests and install_dev script

### LINKED ISSUE #471: Update tests to latest Ubuntu
See #468 

Tests are currently run on Ubuntu 20.04 (instead of `ubuntu-latest`) because they fail with proposal 6.1.6 otherwise. Once #458 has been merged, we should switch back.

Also - currently, tests can run on either pull requests *or* pushes to any branch. I was unable to come up with a simpler setup to avoid running tests multiple times, so the number of individual actions per PR is now kind of high. We may want to switch to running tests on *pull requests* or *on demand* only, with the downside that branches not in a (draft) PR are not automatically checked. I have no strong opinion on which is preferable.

sjoerd-bouma:
Minor changes to the tests:
- Update to latest Ubuntu again, now that proposal can be pip installed on it... (although we seem to have a new issue now), closes #471 
- Include `workflow_dispatch` hook in tests, so they can all be manually launched from the Actions tab
- Update `install_dev.py` such that it raises an error if an error occurs during dependency installation. Previously, these errors would be ignored, making troubleshooting the tests harder. (Sorry @fschlueter - I didn't see you had also fixed this, feel free to change to your implementation if you prefer).

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -1,6 +1,6 @@
 name: Build and publish
 
-on: [push]
+on: [push, workflow_dispatch]
 
 jobs:
   build:
--- a/.github/workflows/deploy_documentation.yaml
+++ b/.github/workflows/deploy_documentation.yaml
@@ -1,6 +1,6 @@
 name: Deploy Documentation
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 jobs:
   prepare:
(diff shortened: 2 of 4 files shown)


fschlueter (approved):

fschlueter:
Why the hack is proposal installing of #492?

fschlueter:
With the temporal fix of #502 this PR should be merged?!

PR #497, opened 2023-02-23
Fix issue with unkown numpy data type on my mac

fschlueter:
On my Mac with python3.10 the datatype numpy.float128 does not exist, on a linux server it does... 

However, the new code should be anyway better/more general.


sjoerd-bouma (approved):

cg-laser (approved):
looks good, this is the better way of checking for variable types

PR #498, opened 2023-02-23
Feature/multiband noise gen

colemanalan:
Issue being addressed:
In an experiment, we might trigger on a more narrow band than what we eventually record. The biased high-low noise generator only would generate and return the noise on this full band. This PR adds a flag to allow the user to save the full-band noise waveform instead.


colemanalan:
The checks are failing, in the build, but they are for all branches at the moment.

cg-laser (approved):
changes look good to me. We can also merge it now as this part of the code is anyway not tested so that we can ignore the failing tests (due to the proposal dependency).

PR #501, opened 2023-02-27
modify event.get_station() to return the first and only station of no…

fschlueter:
… station id is given

The idea is to avoid code like:

```
for station in evt.get_stations():
   ....
```
or worse 
```
station = evt.get_station(evt.get_station_ids()[0])
```
if it is clear that the event has only one station (which is typically the case ?!)

What do you think about that? Unnecessary?

### DIFF
--- a/NuRadioReco/framework/event.py
+++ b/NuRadioReco/framework/event.py
@@ -132,15 +132,15 @@ class Event:
         """
         Returns the station for a given station id.
 
-        Paramters
-        ---------
+        Parameters
+        ----------
 
         station_id: int
             Id of the station you want to get. If None and event has only one station
             return it, otherwise raise error. (Default: None)
 
-        Return
-        ------
+        Returns
+        -------
 
         station: NuRadioReco.framework.station
         """
--- a/NuRadioReco/framework/event.py
+++ b/NuRadioReco/framework/event.py
@@ -128,7 +128,30 @@ class Event:
     def get_run_number(self):
         return self.__run_number
 
-    def get_station(self, station_id):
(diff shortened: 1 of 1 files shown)


NuRadioReco/framework/event.py
```diff
@@ -128,7 +128,30 @@ def set_id(self, evt_id):
     def get_run_number(self):
         return self.__run_number
 
-    def get_station(self, station_id):
+    def get_station(self, station_id=None):
+        """
+        Returns the station for a given station id.
+
+        Paramters
```
sjoerd-bouma, line 135: Parameters

NuRadioReco/framework/event.py
```diff
@@ -128,7 +128,30 @@ def set_id(self, evt_id):
...
+        ---------
+
+        station_id: int
+            Id of the station you want to get. If None and event has only one station
+            return it, otherwise raise error. (Default: None)
+
+        Return
```
sjoerd-bouma, line 142: Returns

sjoerd-bouma (changes requested):
I'm all for quality of life improvements, and I guess this shouldn´t break any backwards compatibility. Please do correct the docstring headings, as they will cause errors in compiling the documentation otherwise (which I guess is hard to see at the moment because the tests fail in any case...)

NuRadioReco/framework/event.py
fschlueter: Even if we return a single station :o ;)

sjoerd-bouma (approved):
I've double checked that the documentation compiles locally, seeing as the tests don't work currently

Issue #494, opened 2023-02-22
`proposal` cannot be pip installed


sjoerd-bouma:
This is probably due to one of its dependencies (`conan`) having recently been updated. I think we'll have to wait for the proposal devs to update proposal to be installable with the new conan, or to fix their conan version to 1.59.0. See [this issue](https://github.com/tudo-astroparticlephysics/PROPOSAL/issues/354). This also means our tests involving `proposal` fail at the moment.

In the meantime, I've created a fork with conan fixed to <2.0, so proposal 7.5.0 can be installed using `pip install git+https://github.com/sjoerd-bouma/PROPOSAL`

fschlueter:
What should we do @cg-laser, @shallmann? Should we wait until proposal fixes their problem or should we temporarily switch to sjoerd's fork? And in the meant time should we continue merging stuff a few of the PR have successful pipelines.

shallmann:
Would it be possible to pip install a fixed Conan version prior to proposal?

shallmann:
Or would that be upgraded when pip installing proposal afterwards?

sjoerd-bouma:
@fschlueter I think we can wait, it's already been fixed and approved by the proposal devs - as soon as it gets pushed to pip we can just upgrade proposal.
I reckon in the meantime anything that doesn't mess with proposal and passes all the other tests can safely be merged.

@shallmann unfortunately, that doesn't work. But my fork or now the `pin_conan` branch on proposal can also be pip installed, so I think that's straightforward enough as  a workaround for the next couple of days.

sjoerd-bouma:
i.e. `pip install git+https://github.com/tudo-astroparticlephysics/PROPOSAL` should work as well, now.

shallmann:
Thanks Sjoerd for explaining. So it seems we can just wait 1 or 2 days and use the 7.5.1 or whatever version

PR #502, opened 2023-02-27
Fix Actions / temporarily switch to git version of proposal

sjoerd-bouma:
See #494
 
Seeing as so many people are actively working on stuff, this was a simple one-line change, and I don't want to end up having to fix a load of docstrings next week - this PR changes the proposal version to the [current version of the master branch](https://github.com/tudo-astroparticlephysics/PROPOSAL/commit/b95628a57befe9650ab6bb3c8ce21442880ea382) which enables it to be pip installed again, such that the tests can run again.

Once the proposal pypi release has been updated, we can switch back to that.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ numba = "*"
 Sphinx = "*"
 sphinx-rtd-theme = "*"
 numpydoc = "1.1.0"
-proposal = "7.5.0"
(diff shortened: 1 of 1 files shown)


shallmann:
Yes, let's do that temporarily!
Go ahead, if tests run through....

shallmann (approved):

fschlueter (approved):
gogogo

PR #503, opened 2023-02-28
return useful debug message if .nur file is empty

lpyras:
Here is the cleaned up PR (before #476)

I added a file size check in NuRadioReco/modules/io/NuRadioRecoio.py and added an elif for version 0.0 in NuRadioReco/modules/io/event_parser_factory.py. Please contribute if you have more ideas.

@christophwelling wrote:
Could you make this more robust and also check if the file version is a reasonable number?
The reason is that if NuRadioReco tries to read in a corrupted file, it often interprets some random byte as the version, which often results in very weird version numbers.
So could you check if abs(self.__file_version) <= VERSION and give out a warning that the file is probably corrupted otherwise?


cg-laser (approved):
thanks Lilly. It is ok to be merged. 

There are some options to improve this further, by e.g., checking the filesize. But the current code is better than before, so we can also merge it now if you don't have time for further improvements.

PR #505, opened 2023-03-03
Minor updates to documentation

sjoerd-bouma:
Some small improvements to the documentation:
- Simplified/improved the error classification and error output of the `make_docs` script by switching to `re` string matching.
- Fixed a bug that meant custom css was only used locally, but not on the deployed documentation
- Changed the documentation for `parameters.py` from `#` (comments only visible in source code) to `#:`, which means their descriptions are now also visible in the deployed documentation.


fschlueter (approved):

cg-laser (approved):
thanks Sjoerd

PR #506, opened 2023-03-07
Only refactoring get_attenuation_along_path(): Renaming variables, co…

fschlueter:
…ding style.

I am working on an optimisation for `get_attenuation_along_path` when used with the GL3 ice model. This PR contains only cosmetic changes which I wanted to disentangle from the actual change in functionality.


cg-laser (approved):
looks good, thanks

PR #507, opened 2023-03-07
Optimise calculation of attenuation factor with GL3 model

fschlueter:
This PR tries to implement an optimization for `get_attenuation_along_path` when the `GL3` attenuation model is used. With this model the integration of `ds(z) / L(z, f)` (Eq. 14 in [paper](https://link.springer.com/article/10.1140/epjc/s10052-020-7612-8)) with `scipy.integration.quad` is very time consuming (due to details of a(z, f), @christophwelling can give more details). 

The idea is, instead of the numerical integration:
```
 attenuation_exp = np.array([integrate.quad(dt, x1[1], x2_mirrored[1], args=(
        C_0, f), epsrel=1e-2, points=points)[0] for f in freqs])
```
 to calculate the attenuation for discrete segments of `ds` and sum them up:  
```
attenuation_exp = np.array(
    [np.sum([dt(x, C_0, f) * dx_actual for x, dx_actual in zip(mid_points, dx_actuals)]) for f in freqs])
```
Tuning the (vertical) width of the segments (i.e., `dx_actual `) this gives decent accuracy with a great improvement in time (~ x10) for many geometries. However, in cases where the ray travels on horizontal trajectories this method becomes inaccurate (because we bin the path in segments of vertical distance). Hence, in those cases we fall back to a numerical integration (for this part of the ray path). And instead of integrating  `ds(z) / L(z, f)`, we only integrate `ds(z)` and divide by `L(<z>, f)` for performance reasons.

I also tested the new algorithm for the attenuation model `GL1`. However, for this model the new algorithm is not faster and can even be more slowly. Hence, it is only used for `GL3`

PS: This PR is based on PR #506. Once merged the cosmetic changes also included here should vanish from the diff.

In the following I enclose figures which show the accuracy between the original calculation and the new one for several different geometries with an antenna at an depth of -100m. Also the acceleration factor is shown. 

Emission at x=-100, varying depth:
attenuation_opt_dev_x-100 0m_dx10m_GL3_int20
attenuation_opt_hists_x-100 0m_dx10m_GL3_int20

Emission at x=-300, varying depth:
attenuation_opt_dev_x-300 0m_dx10m_GL3_int20
attenuation_opt_hists_x-300 0m_dx10m_GL3_int20

Emission at x=-500, varying depth:
attenuation_opt_dev_x-500 0m_dx10m_GL3_int20
attenuation_opt_hists_x-500 0m_dx10m_GL3_int20

Emission at x=-800, varying depth:
attenuation_opt_dev_x-800 0m_dx10m_GL3_int20
attenuation_opt_hists_x-800 0m_dx10m_GL3_int20

Emission at x=-1000, varying depth:
attenuation_opt_dev_x-1000 0m_dx10m_GL3_int20
attenuation_opt_hists_x-1000 0m_dx10m_GL3_int20

Emission at z=-300, varying horizontal distance 
attenuation_opt_dev_z-300 0m_dx10m_GL3_int20
attenuation_opt_hists_z-300 0m_dx10m_GL3_int20

Emission at z=-500, varying horizontal distance 
attenuation_opt_dev_z-500 0m_dx10m_GL3_int20
attenuation_opt_hists_z-500 0m_dx10m_GL3_int20

Emission at z=-800, varying horizontal distance 
attenuation_opt_dev_z-800 0m_dx10m_GL3_int20
attenuation_opt_hists_z-800 0m_dx10m_GL3_int20

Emission at z=-1000, varying horizontal distance 
attenuation_opt_dev_z-1000 0m_dx10m_GL3_int20
attenuation_opt_hists_z-1000 0m_dx10m_GL3_int20

### DIFF
--- a/NuRadioMC/SignalProp/analyticraytracing.py
+++ b/NuRadioMC/SignalProp/analyticraytracing.py
@@ -556,11 +556,19 @@ class ray_tracing_2D(ray_tracing_base):
             self.__logger.debug(f"calculating attenuation for frequencies {freqs}")
             return freqs
 
-    def get_attenuation_along_path(self, x1, x2, C_0, frequency, max_detector_freq, reflection=0, reflection_case=1):
-        tmp_attenuation = None
+
+    def get_attenuation_along_path(self, x1, x2, C_0, frequency, max_detector_freq, 
+                                   reflection=0, reflection_case=1):
+        
+        attenuation = np.ones_like(frequency)
+
         output = f"calculating attenuation for n_ref = {int(reflection):d}: "
-        for iS, segment in enumerate(self.get_path_segments(x1, x2, C_0, reflection, reflection_case)):
-            if(iS == 0 and reflection_case == 2):  # we can only integrate upward going rays, so if the ray starts downwardgoing, we need to mirror
+        for iS, segment in enumerate(self.get_path_segments(x1, x2, C_0, reflection, 
+                                                            reflection_case)):
+            
+            # we can only integrate upward going rays, so if the ray starts downwardgoing, 
+            # we need to mirror
+            if iS == 0 and reflection_case == 2: 
                 x11, x1, x22, x2, C_0, C_1 = segment
                 x1t = copy.copy(x11)
                 x2t = copy.copy(x2)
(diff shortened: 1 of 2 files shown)


fschlueter:
To summarise: The accuracy (w.r.t. the original calculation) is typically within 1% (which should be far below the uncertainty on the underlying ice model). The acceleration factor is typically > x10 when no fallback is needed and still above 2-3 with the fallback

cg-laser (changes requested):
thanks Felix,
I'm wondering about the interface. Shouldn't we make this an optional thing where the default is to use the faster method? And to prepare for now attenuation models,  maybe also make it configurable to which models the "speedup" is applied?

fschlueter:
> thanks Felix,
> I'm wondering about the interface. Shouldn't we make this an optional thing where the default is to use the faster method? And to prepare for now attenuation models, maybe also make it configurable to which models the "speedup" is applied?

Sounds good.

fschlueter:
> thanks Felix,
> I'm wondering about the interface. Shouldn't we make this an optional thing where the default is to use the faster method? And to prepare for now attenuation models, maybe also make it configurable to which models the "speedup" is applied?

done

fschlueter:
@cg-laser

fschlueter:
To summarise the discussion in the call. The overall opinion was, that this is good enough and can be merged. Potential improvements in the future are:
- Dave suggested an alternative calculation which avoids numerical integration and would be faster.
- It was suggested to smooth the GL3 attenuation length profile such that the original calculation does not take longer
- Tabulate the integration result and interpolate from this tabulation

fschlueter:
Just to add, this is the attenuation profile: https://arxiv.org/pdf/2201.07846.pdf

image

fschlueter:
@cg-laser want to approve the PR?

cg-laser (changes requested):

cg-laser (approved):

Issue #495, opened 2023-02-23
Measured noise importer


fschlueter:
Hi all,

I opened this issue to have a discussion about the various noise importer for measured noise we have. From what I see have we have:

- `io/noise/noiseImporter.py`
- `io/noise/noiseImporterROOT.py`
- `channelMeasuredNoiseAdder.py`

Furthermore, #293 implements a `io/noise/noiseImporterUproot.py` which (afaik) is suppose to add noise for/from the ARIANNA detector. I am currently working on a module to import noise from/for RNO-G. My question is now: Do we need the above modules, are some maybe redundant (in particular the `io/noise/noiseImporter.py` and `channelMeasuredNoiseAdder.py`) ? What I imagine could be necessary is the following: We have one generic noise adder module which uses `.nur` files to import noise and than modules for the specific detectors which use their specific data formats and are call appropriately with the detector name. Does this make sense?

PS: If we keep the module, we should move `channelMeasuredNoiseAdder.py` to `io/noise/`

cg-laser:
short answer is yes ;-)

all of them do slightly different things. The three modules in `io/noise` read noise files from different sources. 
The `channelMeasuredNoiseAdder.py` assumes (but @christophwelling should comment in case I got this wrong) that the station ids and channels ids between the simulated stations and datafiles are the same. Hence, this can directly be used e.g. for RNO-G data if the RNO-G data is available in nur files. 

But I agree, it is confusing that some are under `io/noise` and others are directly under `modules`. Normally, things in `io` should "yield" a new event and not modify an existing one. So all these modules add noise to existing channels. 

I don't think it is practical to generalize everything in one module, but we should rename the modules and improve the doc string to make clear what each module does.

fschlueter:
I agree, I would not generalise it neither. So should we create a new directory in `modules/` which is called `noise` or `measured_noise` and place the modules there? Furthermore should we rename the modules `noiseImporter.py` -> `noiseImporterARIANNA.py` and `noiseImporterROOT.py` -> `noiseImporterROOTARIANNA.py` (if that are appropiate names?)

cg-laser:
the suggestion sounds good to me! But I would use `noiseImporterARIANNAROOT` instead of `noiseImporterROOTARIANNA`

fschlueter:
I close this issue, it will be addressed very soon with the merging of #508

PR #508, opened 2023-03-07
Rename and relocate modules which add/import measured noise

fschlueter:
This PR addresses issue #495

We should also improve the doc strings of the modules?!


shallmann (commented):
Long term, might it be cleaner to have a structure:
```
modules/measured_noise/channelMeasuredNoiseAdder.py
modules/measured_noise/ARIANNA/noiseImporter.py
modules/measured_noise/ARIANNA/noiseImporterROOT.py
modules/measured_noise/RNO-G/noiseImporter.py
modules/measured_noise/RNO-G/noiseGenerator.py
...
```

@fschlueter What do you think?

fschlueter:
Well I do not have strong feelings about additional subdirectories. But should we keep the experiment in the name?

shallmann:
> Well I do not have strong feelings about additional subdirectories. But should we keep the experiment in the name?

Well if we have subdirectories there is no need to duplicate the experiment in the filename IMO.

fschlueter:
done

fschlueter:
@shallmann

cg-laser:
I'm a bit confused myself now about the ARIANNA noise reader and the general noise reader as both read noise traces from nur files. Does anyone know the difference of the two? @christophwelling ?

fschlueter:
> I quickly glanced through the files, and both look generic and not tied to ARIANNA .nur files at all. Details of the implementation are a bit different (one allows channel A -> B mapping, but requires same sampling / trace length in data and sim in one of the implementations vs. exact channel correspondance but resampling is used in the other. As of why there are two separate ones I have no clue.
> 
> For this PR: I would remove the ARIANNA from the docstring and put the noiseAdder to the directory above.

I am confused, are you suggesting to have 
```

modules/measured_noise/channelMeasuredNoiseAdder.py
modules/measured_noise/noiseImporter.py
modules/measured_noise/ARIANNA/noiseImporterROOT.py
modules/measured_noise/RNO-G/noiseImporter.py
modules/measured_noise/RNO-G/noiseGenerator.py
```
I would rather merge than `channelMeasuredNoiseAdder.py ` and `noiseImporter.py ` in one go here.

fschlueter:
> Yes that was what I was suggesting, since I at least did not want to volunteer cleaning up/merging the two. But if you have time for that this is probably even better ;)

Nope, I don't neither. @cg-laser had a reason why we should keep them? I think we should merge that and keep the issue open. Moving the file one directory up will be only more confusing for people fixing the issue.

cg-laser:
as far as I can see, it would be best to merge the two generic noise adder modules

fschlueter:
Okay I am on a train and have nothing better to do (apparently)... 
I started implementing features from /ARIANNA/noiseImporter.py into channelMeasuredNoiseAdder.py. 
A few things to discuss:

- Do we want to have the flexibility from noiseImporter to only provide a noise folder?
- Do we want to be able to specify the station id for the noise station (as in the noiseImporter)
- Do we want to subtract the mean (as currently copied from noiseImporter), recognise that I changed it a bit an subtract the mean from the noise trace and not from the combined trace.
- Do we need to make the clipping of the traces optional?
- Anything else?

cg-laser:
reading from a folder would be very useful to quickly read in a large library ;-)

fschlueter:
> reading from a folder would be very useful to quickly read in a large library ;-)

done

fschlueter:
Okay I actually have to test the code of the new module ... Besides that I was thinking we should add it to the CI if possible ...

fschlueter:
Alright @cg-laser @shallmann  I tested the basics of this module with a bunch of RNO-G noise (converted to nur files) and it works

fschlueter:
@shallmann or @cg-laser

shallmann (approved):
To me this looks good to go. Thanks for implementing the changes.

PR #511, opened 2023-03-14
 pass event time to the sim station to simulate time dependent detect…

pgiri4:
We have created a private package for the ARA detector that can be used in conjunction with NuRadioMC. This pull request adds a simple line to the simulation.py file, which passes event time information to the sim station, making the ARA package compatible with NuRadioMC.

changelog.txt has been updated


cg-laser (approved):
@christophwelling this is the line you need to merge into your refactored code ;-)

PR #513, opened 2023-03-15
512 feldman cousins list contains wrong number

shallmann:
fixes issue 512
Also added journal reference


lpyras (dismissed):
Thanks for finding and fixing this.

cg-laser (approved):
thanks for fixing this. Damn, these different comma vs. point conversions.

lpyras (approved):
Works for me :)

PR #514, opened 2023-03-20
Draft: Reorganize detector classes

fschlueter:
Disclaimer this is not working code! I share this branch at this very early stage to describe my idea and get feedback for it.

The context is, that I am currently writing an RNO-G detector class which should use the new mongo db as source for the detector description. Doing so, the current structure of classes confuses me and can be improved IMO. Let me know if my ideas are sensible and feasible. 

What I try to do here (started doing) is the following: 

The  DetectorBase class provides the (base) interface for NuRadioReco Modules to obtain the detector description but does not know about the source of information nor how to interface it. It provides all the internal buffer functionality including TinyDB and dict buffers, and the general getter functions (e.g., get_station_position, ...). However, this means that the class is never actually initialised directly but only through an inherited classes.

For each detector (Arianna, RNO-G, generic sql detector, generic json detector, ...) we implement an inherited class which implements the interface between the internal buffer (i.e. TinyDB with the virtual function `buffer_in_tiny_db`) and the source (e.g., sql db, mongo db, json files). Furthermore, in each of this classes we add Detector specific getter function or overwrite them from the DetectorBase if this detector needs more specific stuff to be done. One example for this would be `get_parallel_channels` for an ARIANNA detector.

The current file detector_sql.py implements not actual a detector class but rather an interface between a detector class and a sql db. I have renamed the file and class appropriately and started implementing an actual detector class which take the sql db as source and which inherits from DetectorBase. In the code that means having this one-liner:
```         
self._db = self. buffer_in_tiny_db()
```
Instead of this block which moves into the daughter classes:
```
        if source == 'sql':
            self._db = buffer_db(in_memory=True)
        elif source == 'dictionary':
            self._db = TinyDB(storage=MemoryStorage)
            self._db.truncate()
            stations_table = self._db.table('stations', cache_size=1000)
            for station in dictionary['stations'].values():
                stations_table.insert(station)
            channels_table = self._db.table('channels', cache_size=1000)
            for channel in dictionary['channels'].values():
                channels_table.insert(channel)
        else:
            self._db = TinyDB(
                json_filename,
                storage=self._serialization,
                sort_keys=True,
                indent=4,
                separators=(',', ': ')
            )
``` 
In case of `sql`, the new function `buffer_in_tiny_db` implements what `buffer_db` did. The same has to be written for the other cases and can be easily extended to other sources in the future.

### DIFF
--- a/NuRadioReco/detector/db_sql.py
+++ b/NuRadioReco/detector/db_sql.py
@@ -1,0 +1,551 @@
+import numpy as np
+from NuRadioReco.utilities import units
+import os
+import MySQLdb
+import inspect
+import datetime
+import json
+import logging
+
+logger = logging.getLogger('sql detector')
+
+
+class DatabaseInterface(object):
+    __instance = None
+
+    def __new__(cls):
+        if DatabaseInterface.__instance is None:
+            DatabaseInterface.__instance = object.__new__(cls)
+        return DatabaseInterface.__instance
+
+    def __init__(self):
+
+        dir_path = os.path.dirname(os.path.realpath(__file__))  # get the directory of this file
+        filename = os.path.join(dir_path, "detector_sql_auth.json")
+        fin = open(filename, 'r')
+        mysql_opts = json.load(fin)
+        fin.close()
+
+        self.__mysql = MySQLdb.connect(host=mysql_opts['host'],
+                                       user=mysql_opts['user'],
+                                       passwd=mysql_opts['pass'],
+                                       db=mysql_opts['DB'])
+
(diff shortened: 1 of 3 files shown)


cg-laser:
Hey Felix, before you invest more time into that, let's better discuss it in the next inicemc call. There were good reasons for why things have been done that way ;-)

fschlueter:
Okay, I take from the discussion that we want to rethink the buffering for the RNO-G detector and that we will implement an independent detector class, is that correct? After all I would like to rename the file (and class in) detector_sql.py because it is very confusing.

PR #517, opened 2023-03-24
fix bug that nur files are not closed properly

cg-laser:
on the Uppsala cluster, nur files were not closed properly. The io buffer was not written to disk at the end of the run. The last couple of events were always missing or the file was completely empty for small event numbers.  For some reason, the if statement always evaluated to false. This PR fixes it.


fschlueter (approved):

nilsphysics (approved):
I tested it and it seems to work, thanks for this fix!

christophwelling (approved):

PR #518, opened 2023-03-24
Rename file and class of prev. detector_sql.py to more appropiate names

fschlueter:


NuRadioReco/detector/detector_base.py_bck
sjoerd-bouma, line 1: I'm guessing you didn't mean to include this one?

fschlueter: Correct :D Should be removed now (forced push)

fschlueter:
Since you interact with the detector via the wrapper class detector.py I am not so worried about backwards comp., right? Regarding the name: I see, is there a more appropriate name? In my opinion this name I confusing, if other people think different we can also just close this PR

fschlueter:
What is the conclusion here @cg-laser

fschlueter:
ping @cg-laser

fschlueter:
I guess this is not getting approved.

PR #519, opened 2023-03-30
Add RNO-G data reader base on Mattak and a noise importer which utilises this data reader

fschlueter:
The data reader is based on code from @KelliMichaels. I added a few features such as a (optional) preliminary baseline correction, voltage calibration, and on the fly run selection. It is also possible to filter for specific trigger types (or any other information contained in the eventInfo class).

The noise reader is quite simple at the moment. The main issue I have with it right now is that is caches all noise event in memory at the beginning with can become impractical quite soon.


fschlueter:
Arg, this PR is based on #508. Once this is merged, I can make a rebase and the additional diff disappears...

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
```diff
@@ -0,0 +1,318 @@
...
+            self.__skipped += 1
+            continue
+                     
+         evt = NuRadioReco.framework.event.Event(event_info.run, event_info.eventNumber)
+         station = NuRadioReco.framework.station.Station(event_info.station)
+
+         trigger = NuRadioReco.framework.trigger.Trigger(event_info.triggerType)
+         trigger.set_triggered()
+         trigger.set_trigger_time(event_info.triggerTime)
+         station.set_station_time(datetime.datetime.fromtimestamp(event_info.triggerTime))
```
christophwelling, line 278: Little nitpick: We decided to use `astropy.time` in NuRadioMC instead of `datetime` .The reason is that astropy supports nanosecond precision. Doesn't matter too much here, but let's keep it consistent.

fschlueter: Good point, I will change that. I am currently wondering: What should I do with `event.__event_time`?

fschlueter: actually the setter takes care of this right?
```
def set_station_time(self, time):
        if time is None:
            self._station_time = None
            return
        if isinstance(time, datetime.datetime):
            time_strings = str(time).split(' ')
            self._station_time = astropy.time.Time('{}T{}'.format(time_strings[0], time_strings[1]), format='isot')
        else:
            if time.format == 'datetime':
                time_strings = str(time).split(' ')
                self._station_time = astropy.time.Time('{}T{}'.format(time_strings[0], time_strings[1]), format='isot')
            else:
                self._station_time = time
```

fschlueter: why
```
        if time is None:
            self._station_time = None
            return
```

fschlueter: Uh but when converting to `datetime.datetime` we loss the nanosecond ... This is bad ...

fschlueter: > why

okay now I get it xD

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
```diff
@@ -0,0 +1,568 @@
...
+   
+   time_offset: float
+      trace_start_time = trigger_time - time_offset 
+   
+   """
+   
+   time_offsets = {
+      "FORCE": 0,
+      "LT": 213 * units.ns,  # ~ 1 / 3 of trace @ 2048 sample with 3.2 GSa/s
+      "RADIANT": 320 * units.ns  # ~ 1 / 2 of trace @ 2048 sample with 3.2 GSa/s
```
shallmann, line 95: do we want to set accurate timings here as in the field or not?

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
```diff
@@ -0,0 +1,568 @@
...
+      apply_baseline_correction: bool
+         Only applies when non-calibrated data are read. If true, correct for DC offset.
+         (Default: True)
+
+      convert_to_voltage: bool
+         Only applies when non-calibrated data are read. If true, convert ADC to voltage.
+         (Default: True)
+         
+      select_runs: bool
+         Select runs
```
shallmann, line 151: can you add a bit more info what True/False means?

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
```diff
@@ -0,0 +1,568 @@
...
+
+class readRNOGData:
+
+   def begin(self, 
+             data_dirs, log_level=logging.INFO, selectors=None, 
+             read_calibrated_data=False,
+             apply_baseline_correction=True,
+             convert_to_voltage=True,
+             select_triggers=None,
+             select_runs=True,
```
shallmann, line 115: shouldn't the default be `False`? A frequent use case will be "I want to analyse run XX for station YY" ... i.e. the reader **should** work by just passing a file OR data dir (without any runtable)

shallmann (commented):
Hi Felix,
I think this is a great replacement for the current readers. Few small comments/questions.

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
fschlueter: sure, do we know them already? Those values are somewhat placeholders

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
fschlueter: We can change the default to false. However, just to say it: If the run_table is not available an error is logged but the program is not stoped and no selection applied

fschlueter:
@shallmann I would like to merge #508 before that one.

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
fschlueter: Any suggestions?

shallmann: for better place holders, maybe these plots are helpful:

shallmann: something of order 250ns for LT triggered, 450ns for surface

fschlueter:
Alright, I forced pushed only to get rid of commits which are are already in develop. Should be easier to see what this PR actually does. @cg-laser @sjoerd-bouma any comments, do you want to approve?

fschlueter:
@cg-laser @sjoerd-bouma, do you want to review this PR?

BTW, I implemented the trigger thresholds into mattaks python interface. So we can start thinking how to implement them into NuRadioReco (however I would like to do this in a different PR)

fschlueter:
It could be made optional.

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
```diff
@@ -0,0 +1,569 @@
...
```
sjoerd-bouma, line 117: While it's nice to have this  many options, I think this is also a bit overwhelming for someone that 'just' wants to read, say, a single combined.root file that I got from a more senior colleague. Some suggestions:
- Rearrange the parameters to have the more relevant/frequently used ones first. E.g. `log_level` shouldn't be the second parameter...
- (docstring) - move some of the more obscure parameters to an `Other Parameters` section
- (low priority) - maybe passing (a list of) combined.root files should also work? Though this may just be my bias because 'that's how we used to do it', I can see the advantages of the way you've implemented it.

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
```diff
@@ -0,0 +1,569 @@
...
+      event_info: mattak.Dataset.EventInfo
+         The event info object for one event.
+         
+      waveforms: np.array(n_channel, n_samples)
+         Typically what dataset.wfs() returns (for one event!)
+
+      Returns
+      -------
+      
+      evt: NuRadioReco.framework.event
```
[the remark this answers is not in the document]
fschlueter: I changed the name to _get_event

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
```diff
@@ -0,0 +1,569 @@
...
+      
+      event_index: int
+         The index of a particluar event. The index is the chronological number from 0 to 
+         number of total events (across all datasets).
+         
+      Returns
+      -------
+      
+      evt: NuRadioReco.framework.event
+      """
```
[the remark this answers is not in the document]
fschlueter: So I added another method `get_event` which returns the event of a specific `eventNumber`. However, while doing so I realised that, to get one specific event, one can use the run() method with a selector a la `lambda event_info: event_info.eventNumber == event_id`. And I think this is even a bit faster ...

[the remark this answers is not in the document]
fschlueter: Are you happy with the names? I am very against something like `get_event_i`

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
fschlueter: To your last point: I think this would require changes in Mattak. I opened an issue https://github.com/RNO-G/mattak/issues/7 to discuss this. FFTJ

fschlueter:
Maybe one comment (also to have it archived here): Mattak offers an `iterate()` method for datasets to iterate over all events in a dataset. The method also allows selectors. So we could make use of it (in an earlier implementation of the reader it was not so easy possible but now it is). However, right now we are not doing this. The only advantage of not doing is is logging information (which could be overcome).

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
fschlueter: any thoughts ? @shallmann

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
fschlueter: I rearranged it a bit. Please have a look

shallmann: ((in fact: the last point is more or less in-line with a previous comment I had... I would have been more radical and not made run-selection at all a part of the reader, but since you can just pass the directory without the other optional selection this should work great as is.))

Passing explicitly a filenames=["combined.root"] file like before I find useful ... not sure if the way the mattak class is called already would work now if you pass it a filename instead of a dir? Otherwise should be quick to include buy might needs some if/else checking depending on what is passed.

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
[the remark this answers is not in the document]
sjoerd-bouma: Re naming - I have no strong preferences, but wouldn't having `get_event` and `read_event` be a bit confusing? If you don't like `get_event_i`, something like `get_event_by_index` instead? Just to try to keep things somewhat intuitive.

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
```diff
@@ -0,0 +1,644 @@
...
+      dataset = self._datasets[dataset_idx]
+
+      event_idx_in_dataset = event_idx - self.__get_n_events_of_prev_datasets(dataset_idx)
+      dataset.setEntries(event_idx_in_dataset)  # increment iterator -> point to new event
+            
+      return dataset
+   
+   
+   def filter_event(self, evtinfo, event_idx=None):
+      """ Filter an event base on its EventInfo and the configured selectors.
```
sjoerd-bouma, line 374: Should this also be a private method? I'm again not quite sure about its use case. I think you don't currently expose any `mattak.Dataset.EventInfo` anywhere, in any case.

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
```diff
@@ -0,0 +1,644 @@
...
+                                 f"event number {evtinfo.eventNumber}) is skipped.")
+               self.__skipped += 1
+               return True
+      
+      return False
+   
+   
+   def get_event_informations(self, keys=["station", "run", "eventNumber"]):
+      """ Return information of all events from the EventInfo
+
```
sjoerd-bouma, line 404: pedantic, but maybe remove the s at the end (information is uncountable in English)

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
```diff
@@ -0,0 +1,644 @@
...
+         
+      run_time_range: tuple
+         Specify a time range to select runs (it is sufficient that runs cover the time range partially). 
+         Each value of the tuple has to be in a format which astropy.time.Time understands. A value can be None 
+         which means that the lower or upper bound is unconstrained. If run_time_range is None no time selection is
+         applied. (Default: None)
+         
+      max_trigger_rate: float
+         Used to select/reject runs from information in the RNO-G RunTable. Maximum allowed trigger rate (per run) in Hz.
+         If 0, no cut is applied. (Default: 1 Hz)
```
sjoerd-bouma, line 175: It seems I'm only leaving pedantic comments today, but you can of course ignore them. Here, wouldn't the default 0 (no cut) make more sense? E.g. say I want to look at a specific couple of runs (maybe I'm interested in wind events...), then some of them will by default be rejected, which might be easy to miss and tricky to troubleshoot (though I see at least a logger message is emitted at the default log level).

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
fschlueter: I know.. the proper name would be `get_events_information` but that sounded bad to me.

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
fschlueter: fine with me

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
fschlueter: yes mattak can not handle that atm.

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
```diff
@@ -0,0 +1,569 @@
...
+      evt = NuRadioReco.framework.event.Event(event_info.run, event_info.eventNumber)
+      station = NuRadioReco.framework.station.Station(event_info.station)
+      station.set_station_time(astropy.time.Time(event_info.triggerTime, format='unix'))
+
+      trigger = NuRadioReco.framework.trigger.Trigger(event_info.triggerType)
+      trigger.set_triggered()
+      trigger.set_trigger_time(event_info.triggerTime)
+      station.set_trigger(trigger)
+      
+      for channel_id, wf in enumerate(waveforms):
```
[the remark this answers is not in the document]
fschlueter: @cozzyd might be able to give a answer. I would guess it does not make much of a difference

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
fschlueter: ok

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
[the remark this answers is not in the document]
cozzyd: Right now the way it's serialized all waveforms are stored contiguously on disk, I'm pretty sure, so there is no benefit to only reading some of them (they get decompressed etc. regardless). This could in principle be changed in the future.   But, if expensive calibration is to be done, there may a still benefit to not doing that to all waveforms if you don't need them?

NuRadioReco/modules/measured_noise/RNO_G/noiseImporter.py
```diff
@@ -0,0 +1,176 @@
...
+              log_level=logging.INFO, reader_kwargs={}):
+        """
+        
+        Parameters
+        ----------
+        noise_folders: str or list(str)
+            Folder(s) containing noise file(s). Search in any subfolder as well.
+            
+        file_patters: str
+            File patters used to search for directories, (Default: "*", other examples might be "combined")
```
sjoerd-bouma, line 33: Typo (file_patters -> file_pattern)

NuRadioReco/examples/RNO_data/read_data_example/read_rnog.py
```diff
@@ -0,0 +1,65 @@
+from NuRadioReco.modules.io.rno_g import readRNOGDataMattak
```
sjoerd-bouma, line 1: rno_g -> RNO_G

NuRadioReco/modules/measured_noise/RNO_G/noiseImporter.py
```diff
@@ -0,0 +1,176 @@
...
+        for noise_folder in noise_folders:
+            noise_files += glob.glob(f"{noise_folder}/**/{file_pattern}root", recursive=True)
+        self.__noise_folders = np.unique([os.path.dirname(e) for e in noise_files])       
+        
+        self.logger.info(f"Found {len(self.__noise_folders)}")
+        if not len(self.__noise_folders):
+            raise ValueError
+                
+        if scramble_noise_file_order:
+            random.shuffle(self.__noise_folders)
```
sjoerd-bouma, line 84: We generally try to use seeded random generation (using either `np.random.RandomState` or `np.random.Philox`) to be able to make things reproducible. I guess alternatively (or maybe additionally) one could store which noise trace has been added to which channel. I don't know how important you think this is, it can also be postponed to a future PR so we can merge the new reader sooner.

NuRadioReco/modules/measured_noise/RNO_G/noiseImporter.py
```diff
@@ -0,0 +1,176 @@
...
+            station_mask = self.__station_id_list == station.get_id()
+            if not np.any(station_mask):
+                raise ValueError(f"No station with id {station.get_id()} in noise data.")
+                            
+        else:
+            # select all noise events
+            station_mask = np.full_like(self.__event_index_list, True)
+
+        # int(..) necessary because pyroot can not handle np.int64
+        i_noise = int(np.random.choice(self.__event_index_list[station_mask]))
```
sjoerd-bouma, line 129: See comment at l.84

sjoerd-bouma (commented):
Last couple of comments, hopefully
- while I'm in favour of only having a single reader module for RNO-G data, `mattak` is currently not yet pip installable, and does not have a fully functioning `uproot` backend yet (as far as I understand). Is it worth keeping one or both old reader modules with a deprecation warning until this has been implemented? I find it hard to judge how much of an additional challenge the installation of mattak + ROOT would present to, e.g., a starting master's or bachelor's student. If the consensus is it's doable, I'm also happy to only keep the new one.
- The old reader modules are still used in the eventbrowser, but I'm happy to leave the fix for that to a future PR

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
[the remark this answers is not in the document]
fschlueter: This can be a "nice to have" for the future

NuRadioReco/modules/io/rno_g/readRNOGDataMattak.py
fschlueter: done.

NuRadioReco/modules/measured_noise/RNO_G/noiseImporter.py
fschlueter: Implemented the generator. Book-keeping of the random seed and the "noise-trace add to which event" I will label "nice-to-have" for a future PR

fschlueter:
> while I'm in favour of only having a single reader module for RNO-G data, mattak is currently not yet pip installable, and does not have a fully functioning uproot backend yet (as far as I understand). Is it worth keeping one or both old reader modules with a deprecation warning until this has been implemented? I find it hard to judge how much of an additional challenge the installation of mattak + ROOT would present to, e.g., a starting master's or bachelor's student. If the consensus is it's doable, I'm also happy to only keep the new one.

Maybe @cozzyd  also wants to comment. AFAIK the uproot backend is fully functional. Yes mattak is not yet pip installed but I think it just have to be cloned and set in the Pythonpath for the uproot backend (correct @cozzyd). In that case it could be added into the install script (@sjoerd-bouma  would you volunteer for that, I think you will be x10 faster than me).

sjoerd-bouma:
> Maybe @cozzyd also wants to comment. AFAIK the uproot backend is fully functional. Yes mattak is not yet pip installed but I think it just have to be cloned and set in the Pythonpath for the uproot backend (correct @cozzyd). In that case it could be added into the install script (@sjoerd-bouma would you volunteer for that, I think you will be x10 faster than me).

Mattak still has to be compiled, unfortunately, and I think for the directory it chooses by default I needed to `sudo` it. I chatted to Steffen and I think we agree that for now (while mattak doesn't have a working `setup.py`) it would be nice to keep one old reader with a deprecation warning, would that be okay? Other than that I think this PR can be merged.

sjoerd-bouma (approved):

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -0,0 +1,642 @@
...
+   def get_event(self, event_id):
+      """ Allows to read a specific event identifed by its id
+
+      Parameters
+      ----------
+
+      event_id: int
+         Event Id
+         
+      Returns
```
sjoerd-bouma, line 595: Actually... :see_no_evil: can we change this to select by run number and event id? See https://github.com/nu-radio/NuRadioMC/blob/develop/NuRadioReco/modules/io/NuRadioRecoio.py#L277. If we use the reader to read multiple runs in one go, which seems to be an intended feature (?), selection by event id only is ambiguous.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -0,0 +1,642 @@
...
+
+      evt: NuRadioReco.framework.event
+      """
+
+      self.logger.debug(f"Processing event {event_id}")
+      t0 = time.time()
+
+      event_infos = self.get_events_information(keys=["eventNumber"])
+      event_idx_ids = np.array([[index, ele["eventNumber"]] for index, ele in event_infos.items()])
+      mask = event_idx_ids[:, 1] == event_id
```
sjoerd-bouma, line 606: Is it worth caching this (e.g. storing the output of get_events_information in a class variable)? Probably reading this information only is quick enough that it doesn't really matter.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -0,0 +1,642 @@
...
+         self.logger.info(f"Found {len(selectors)} selector(s)")
+
+      self._selectors = selectors
+            
+      if select_triggers is not None:
+         if isinstance(select_triggers, str):
+            selectors.append(lambda eventInfo: eventInfo.triggerType == select_triggers)
+         else:
+            for select_trigger in select_triggers:
+               selectors.append(lambda eventInfo: eventInfo.triggerType == select_trigger)
```
sjoerd-bouma, line 246: I'm noticing more stuff now that I'm implementing this for the data monitor... I think this order should be swapped, because it's `self._selectors` and not `selectors` that's used in the other methods, right?

I also wonder if it's worth implementing a `set_selectors` method (or changing where/how the selectors are implemented). E.g. in my use case I know in advance which runs I want to open, but I might then want to change only the selectors to look at different events. It feels slightly unintuitive to call `begin` every time in that case (+, on a mounted file system, the `os.path.exists` calls are not free either).

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
fschlueter: Yes it is worth it. Nope takes quite some time.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
fschlueter: I guess you are right. Do you want to do it? Or are you fine with having that in the next PR?

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
fschlueter: Sure seems useful.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
fschlueter: done

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
fschlueter: I resolved this comment. I fixed the bug. The set_selectors interface can be included in the next PR (which I am already working on)

sjoerd-bouma (approved):
Thanks Felix

shallmann:
Thanks Felix, this is awesome 🥇

PR #520, opened 2023-03-30
Change station time

fschlueter:
Do you like this change?

I think it makes the setter more readable and also allows to set unix time stamps. However, now the format with which it is stored is not fixed anymore. I do not see why this is a problem but maybe someone can correct me.

### DIFF
--- a/NuRadioReco/framework/base_station.py
+++ b/NuRadioReco/framework/base_station.py
@@ -89,9 +89,9 @@ class BaseStation():
 
         if isinstance(time, datetime.datetime):
             self._station_time = astropy.time.Time(time)
-        elif isinstance(time, astropy.time):
+        elif isinstance(time, astropy.time.Time):
             self._station_time = time
-        elif time == None:
+        elif time is None:
             self._station_time = None
         else:
             # time is interpreted as unix utc timestamp
--- a/NuRadioReco/framework/base_station.py
+++ b/NuRadioReco/framework/base_station.py
@@ -91,6 +91,8 @@ class BaseStation():
             self._station_time = astropy.time.Time(time)
         elif isinstance(time, astropy.time):
             self._station_time = time
+        elif time == None:
+            self._station_time = None
         else:
             # time is interpreted as unix utc timestamp
             self._station_time = astropy.time.Time(time, format='unix')
--- a/NuRadioReco/framework/base_station.py
+++ b/NuRadioReco/framework/base_station.py
@@ -114,6 +114,9 @@ class BaseStation():
         
         _station_time: astropy.time.Time
         """
+        if self._station_time is None:
+            return None
+        
         self._station_time.format = format
         return self._station_time
     
--- a/NuRadioReco/framework/base_station.py
+++ b/NuRadioReco/framework/base_station.py
@@ -76,20 +76,41 @@ class BaseStation():
         self._parameters.pop(key, None)
 
     def set_station_time(self, time):
-        if time is None:
-            self._station_time = None
-            return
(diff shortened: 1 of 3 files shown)


NuRadioReco/framework/base_station.py
```diff
@@ -76,20 +76,43 @@ def remove_parameter(self, key):
...
+        
+        time: astropy.time.Time or datetime.datetime or float
+            If time is a float it is interpreted as UTC unix timestamp.
+        """
+
         if isinstance(time, datetime.datetime):
-            time_strings = str(time).split(' ')
-            self._station_time = astropy.time.Time('{}T{}'.format(time_strings[0], time_strings[1]), format='isot')
+            self._station_time = astropy.time.Time(time)
+        elif isinstance(time, astropy.time):
```
sjoerd-bouma, line 92: `astropy.time` (module) -> `astropy.time.Time` (class)

NuRadioReco/framework/base_station.py
```diff
@@ -76,20 +76,43 @@ def remove_parameter(self, key):
...
+            self._station_time = time
+        elif time == None:
```
sjoerd-bouma, line 94: I guess this is fine? `a == None` (instead of `a is None`) raises an error if `a` is e.g. an array, but I guess there is no risk of that here.

NuRadioReco/framework/base_station.py
```diff
@@ -76,20 +76,43 @@ def remove_parameter(self, key):
...
+            self._station_time = None
         else:
-            if time.format == 'datetime':
-                time_strings = str(time).split(' ')
-                self._station_time = astropy.time.Time('{}T{}'.format(time_strings[0], time_strings[1]), format='isot')
-            else:
-                self._station_time = time
+            # time is interpreted as unix utc timestamp
+            self._station_time = astropy.time.Time(time, format='unix')
```
sjoerd-bouma, line 98: There are obviously a lot of float formats for time (e.g. 'mjd', ...). `astropy.time` therefore requires a format specification if time is given as a float. You do mention in the docstring that floats will be interpreted as unix by default, but even so, I would have a slight preference for copying the astropy convention (i.e. adding the argument `format=None` or `format='unix'`) to the `set_station_time` function, if you want to add this functionality.

sjoerd-bouma (commented):
Looks like an improvement to me. Once some minor comments have been addressed and the tests have completed I think I can approve this. Thanks Felix!

NuRadioReco/framework/base_station.py
fschlueter: Have a look now.

fschlueter:
I realised that the station_time is also stored in the event_header of a nur file but as astropy object (which is a bad idea and was already changed for the actual station serialisation). I fixed that but for reading allow backwards compatibility. I removed this weird set `date_hms` snippet because I think it is now obsolet. 

BTW, what is stored in the parameter station_time (21)? Is that deprecated? should we add an comment?

fschlueter:
Do we want to bump the nur file version for these changes?

fschlueter:
@sjoerd-bouma

sjoerd-bouma (approved):
I didn't realise that we already did things this way in `base_station.py`, so that this is basically just a fix for the event_header. I think this can be merged without further ado.

One thing I'm not entirely sure about is if we currently actually have ~ns level precision on the station time. E.g. if I set an astropy.time.Time object with a `np.float128` and then call `.value`, it seems to return a `np.float64` object with the associated loss in precision by default. But I think that is a separate issue.

PR #521, opened 2023-03-30
GL3 attenuation for C++ raytracer

christophwelling:
This fixes the GL3 attenuation model for the C++ raytracer. Instead of using the GSL integrator (which had problems dealing with the GL3 model), the raytracer uses `get_path` to get points on the ray path, and then uses those to calculate the attenuation factor.
This is not 100% the same, but I tested the difference by generating a transmitter at random depths from 0.5m and 150m, and a transmitter at random depths between 30 and 3000m and distances between 50 and 3000m. This plot shows the difference between the attenuation factor of the C++ and the python raytracer, so the difference is under 1%.
attenuation_test_random_points
To make it easier to compare, I also added the option to tell the raytracer to use python, even if the C++ implementation is available.

I also noticed another problem with the C++ raytracer: For frequencies above 1.4GHz or so, the interpolation can cause negative attenuation lenghts. Even though this is out of band, it can blow up the signal enough to overcome the filter. So I restricted the interpolation to 600MHz to prevent this.

### DIFF
--- a/NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp
+++ b/NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp
@@ -194,6 +194,9 @@ double ds (double t, void *p){
 }
 
 double get_path_length(double pos[2], double pos2[2], double C0, double n_ice, double delta_n, double z_0){
+
+
+
 	double x2_mirrored[2]={0.};
 	get_z_mirrored(pos,pos2,C0,x2_mirrored,n_ice, delta_n, z_0);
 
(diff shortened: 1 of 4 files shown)


cg-laser (changes requested):
thanks Christoph! Looks good. The feature to specify to use the python ray tracer is great. Did you change it att all occurrences? Maybe it would be better to only use `self._use_python_raytracer` which is set in the init function as 

`self._use_python_raytracer = use_python_raytracer or not cpp_available`

Could you also improve the log output so that it is logged that the python version is used? 

In this block in the beginning, could you change it to use the logger

```try:
    from NuRadioMC.SignalProp.CPPAnalyticRayTracing import wrapper
    cpp_available = True
    print("using CPP version of ray tracer")
except:
    print("trying to compile the CPP extension on-the-fly")
    try:
        import subprocess
        import os
        subprocess.call(os.path.join(os.path.dirname(os.path.abspath(__file__)),
                                 "install.sh"))
        from NuRadioMC.SignalProp.CPPAnalyticRayTracing import wrapper
        cpp_available = True
        print("compilation was successful, using CPP version of ray tracer")
    except:
        print("compilation was not successful, using python version of ray tracer")
        print("check NuRadioMC/NuRadioMC/SignalProp/CPPAnalyticRayTracing for manual compilation")
        cpp_available = False
```
Here I would change the log output to "CPP ray tracer available. It will be used if not specified otherwise".

christophwelling:
Okay, I changed the output to use the logger and give clearer feedback.
I changed it in all places where there is a check if the C++ raytracer is available, so that should be all of them.

christophwelling:
@cg-laser Can you approve this?

cg-laser (approved):

colemanalan:
The GL3 attenuation model is not working correctly. After running some simulations using `greenland_simple` as the icemodel and GL3. I get the following results for the distances as which triggers occur. Most importantly, once reaching 10^17eV, the distributions "saturate".

image

Changing only `GL3` --> `GL2` in my configuration file produced results that looked more like I would expect (no saturation).

image

christophwelling:
I finally found the time to look into this, and I have no idea how this can be happening. I compared the attenuation factors the C++ raytracer returns for the GL1, GL2 and GL3 models when raytracing between different points, and they were always relatively similar. Of course there were differences between them, since they are different models, but nothing that could explain Alan's plots (especially since GL3 usually returned a slightly larger signals than GL2). The only explanation I can think of is that there are events deeper than 3km, i.e. below the bedrock, where the models do return very different values. But Alan assured me this is not the case. So I cannot reproduce this! Is there a way I could see the simulations that showed this strange behavior?

fschlueter:
I merged develop and #918 into this branch to be able to compare the calculations.

fschlueter:
@christophwelling why did you implement this cap for the frequency? I am not getting any negative attenuation length with GL3:
image

fschlueter:
Same plots as in #918 but now with the optimised calculation for both the cpp and python wrapper. The cpp wrapper now takes again the crown for the fasters implementation thanks to @christophwelling. Results are comparable I would say
image

image

image

fschlueter:
next I think I will try to run neutrino simulations with both implementations to see whether they differ.

fschlueter:
@cg-laser, @sjoerd-bouma, @christophwelling  any comment yet? Do you think the shown accuracy is okay? Two things I still have in mind are:
- Checking whether the attenuation (as a function of the frequency) change for a given source but different receivers. 
- Running actual simulations and recalculating Veff.

fschlueter:
These are vertex distributions for triggered events with GL1 and GL3 (not the same events at generator level). I think they look fine. What @philippwindischhofer noticed is that for GL3 there are a events missing at very small distances to the antenna. I checked the attenuation calculation for very small distances but could not yet spot a problem.
distance_bins_GL1
distance_bins_GL3
vertex_distr_GL1
vertex_distr_GL3

fschlueter (dismissed):

anelles:
As far as I can see, the distributions look ok now. I think a lot has changed since Alan first reported this. So I would say we don't necessarily need to do archeology to retrieve the cause of the issue.

fschlueter:
This is the comparison between the calculation from the PR "cpp" and from #918 "cpp with GSL". This calculation is much faster! I guess we are sticking with this one?

image
distance_bins_GL3_cpp
vertex_distr_GL3_cpp
distance_bins_GL3_-_cpp_with_GSL
vertex_distr_GL3_-_cpp_with_GSL

fschlueter:
I merged PR #918 into this PR which makes the diff convoluted sorry. But in any case @christophwelling or @sjoerd-bouma are you happy with that and the cross-checks I made?

anelles:
In concur! Thanks for getting this implemented everyone!

sjoerd-bouma:
Actually... I've just remembered I've thought about this problem before, and I'm now wondering if the 'smarter' way to do this wouldn't be to normalize the result returned by `get_attenuation_integral_GL3` by multiplying it by `analytic_path_length / segmented_path_length`; this should be essentially instantaneous, and I'm not sure what the order of the error is in splitting up the path into 100 segments and computing the total path length... but I guess it could be of the order of 1% for some refracted/reflected paths with a lot of curvature?

Although I'm not sure if the analytic path length is in the C++ code already or it would first need to be ported from the Python version. If the latter, we can defer this to a future PR.

fschlueter:
Great, I will merge this PR asap and see what remains relevant in #918. @sjoerd-bouma I am not sure about the relevance of what you propose. But it sounds like it can be implemented easily. Nonetheless I would like to refer it to a future PR (if you have some capacity for it)

PR #529, opened 2023-04-24
class to access muon flux from MCEq

lpyras:
Here is a draft for a class to access the muon fluxes in MCEq.

### DIFF
--- a/NuRadioMC/utilities/muon_flux.py
+++ b/NuRadioMC/utilities/muon_flux.py
@@ -1,0 +1,175 @@
+from NuRadioReco.utilities import units
+import numpy as np
+import os
+import pickle
+from scipy.interpolate import interp1d
+from MCEq.core import MCEqRun
+import crflux.models as crf
+from functools import lru_cache
+
+class MuonFlux:
+    def __init__(self):
+        self.mc_cm = 1.
+        self.mc_m = 1e2
+        self.mc_km = 1e3 * self.mc_m
+
+        self.mc_deg = 1.
+        self.mc_rad = 180. / np.pi
+
+        self.mc_eV = 1.e-9
+        self.mc_GeV = 1
+        self.mc_TeV = 1.e3
+        self.mc_PeV = 1.e6
+        self.mc_EeV = 1.e9
+
+        self.mc_s = 1
+        self.mc_ns = 1e-9
+
+        self.__buffer = {}
+        self.file_buffer = "data/surface_muon_buffer.pkl"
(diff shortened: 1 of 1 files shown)


fschlueter:
Hi,
a few comments from my side:
- We have to add MCEq as dependency to the install toml file
- Can we remove the mc units which are not used from the init
- I noticed a few issues with coding style: using `if(...)` and leaving space between `Parameters` and `----` in the doc string

fschlueter (approved):

sjoerd-bouma (dismissed):

NuRadioMC/utilities/muon_flux.py
```diff
@@ -0,0 +1,180 @@
...
+        e_grid: array of floats
+            energy grid in eV
+        flux: array of floats
+            flux in NuRadioReco units 1/(area * time  * energy * steradian)
+        """
+         
+        altitude *= self.mc_m
+        mceq = MCEqRun(interaction_model=interaction_model,
+                       primary_model=primary_model,
+                       theta_deg=theta * self.mc_rad)
```
cg-laser, line 54: potentially dangerous, please also divide by `units.rad`. Or just directly divide by `units.deg` if you know that MCEq wants it in degrees.

NuRadioMC/utilities/muon_flux.py
```diff
@@ -0,0 +1,180 @@
...
+
+            flux = finterp(np.log10(energy))
+            self.__buffer[params] = flux
+            with open(self.file_buffer, "wb") as fout:
+                pickle.dump(self.__buffer, fout, protocol=4)
+
+        return self.__buffer[params]
+    
+
+    def get_e_grid(self, theta_deg=50, interaction_model='SIBYLL23C', primary_model=(crf.GlobalSplineFitBeta, None)):
```
cg-laser, line 160: the units seems to be wrong. All units should always be passed in the default units. If MCEq needsit in degree, just divide it by `units.deg` when you pass it to MCEq.

cg-laser (changes requested):
thanks Lilly! Useful utility. See small comments in line

NuRadioMC/utilities/muon_flux.py
lpyras: Thanks for catching, I resolved it now.

sjoerd-bouma (approved):
Thanks for fixing!

cg-laser (approved):
thanks Lilly!

PR #536, opened 2023-05-11
Allow trigger types 'UNKNOWN' to be read in by the reader

fschlueter:


fschlueter:
ping @shallmann

shallmann (approved):
Thanks for adding this so quickly (and sorry for not approving right away), looks great to merge :)

Issue #543, opened 2023-05-26
readRNOGDataMattak raises AttributeError when trying to access event info a second time


sjoerd-bouma:
Calling `get_events_information` twice raises an error. I think this is because the iterator of a dict returns its keys, not its items, as is assumed in lines 518-521.

MWE
```
from NuRadioReco.modules.io.RNO_G import readRNOGDataMattak
reader = readRNOGDataMattak.readRNOGData()
reader.begin('/path/to/.root/file')
event_info = reader.get_events_information()
event_info = reader.get_events_information()
```

Relatedly, I'm not sure there's any benefit to returning only a subset of the keys in `EventInfo`. Wouldn't it be easier/quicker to just return `mattak.Dataset.EventInfo` objects?

PR #544, opened 2023-05-30
Fix bug in get_events_information which appears when calling that fun…

fschlueter:
…ction twice

Fixes issue #543


sjoerd-bouma (approved):
Thanks Felix

PR #546, opened 2023-06-05
Expand the phased-array noise gen

colemanalan:
This is expanding the utility of the phased-array noise generation module. It simply allows the user full configuration of all the previously hard-coded values.

It also fixes a "bug" where waveforms weren't used because of where the trigger actually was instead of rolling. So this version is more efficient.

### DIFF
--- a/NuRadioReco/utilities/noise.py
+++ b/NuRadioReco/utilities/noise.py
@@ -328,6 +328,11 @@ class thermalNoiseGeneratorPhasedArray():
         self.window = int(16 * units.ns * self.sampling_rate * 2.0)
         self.step = int(8 * units.ns * self.sampling_rate * 2.0)
 
+        if self.window >= self.pre_trigger_bins:
+            logger.warning(f"Pre-trigger time ({pre_trigger_time / units.ns:0.2} ns, {self.pre_trigger_bins} bins)" +
+                           f" is within one window ({self.window} bins) of the beginning of the waveform" +
+                           " it is recommended to choose a larger value to avoid clipping effects")
+
         self.noise = channelGenericNoiseAdder.channelGenericNoiseAdder()
 
         # pre-calculate all parameters which will be used to simulate each triggered noise event to avoid re-calculation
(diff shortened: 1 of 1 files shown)


NuRadioReco/utilities/noise.py
```diff
@@ -248,34 +252,58 @@ def __init__(self, detector_filename, station_id, triggered_channels,
...
         self.pre_trigger_bins = int(pre_trigger_time * self.sampling_rate)
         self.n_samples_trigger = int(trace_length * self.sampling_rate)
-        det_channel = self.det.get_channel(station_id, triggered_channels[0])
-        self.adc_n_bits = det_channel["trigger_adc_nbits"]
-        self.adc_noise_n_bits = det_channel["trigger_adc_noise_nbits"]
+
+        if self.n_samples_trigger > self.n_samples:
+            logger.error(f"Requested `trace_length` of {trace_length/units.ns:.1f}ns ({self.n_samples_trigger} bins)" +
+                         f" is longer than the number of samples specified in the detector file ({self.n_samples} bins)")
+            exit()
```
sjoerd-bouma, line 300: See comment at line 352, replace with ValueError (or similar)

NuRadioReco/utilities/noise.py
```diff
@@ -508,15 +553,22 @@ def generate_noise(self, phasing_mode="slice", trigger_mode="binned_sum", debug=
...
+                # resample and use axis -1 since trace might be either shape (N) for analytic trace or shape (3,N) for E-field
+                self._traces = scipy.signal.resample(self._traces, np.shape(self._traces)[-1] // self.upsampling, axis=-1)
+
+                if (i_low >= 0) and (i_high < self.n_samples): # If range is directly a subset of the waveform
                     return self._traces[:, i_low:i_high], self._phased_traces
 
+                # Otherwise, roll the waveforms. Safe as long as noise is generated in the freq domain
+                self._phased_traces = np.roll(self._phased_traces, -i_low * self.upsampling, axis=-1)
+                self._traces = np.roll(self._traces, -i_low, axis=-1)
+                return self._traces[:, :self.n_samples_trigger], self._phased_traces, triggered_beam
```
sjoerd-bouma, line 570: These two `return` statements have different signatures (and neither corresponds to the one specified in the docstring, which should probably be updated?)
Re rolling - the idea is that the generated `self._traces` is periodic because the generation is in the frequency domain, and the resampling happens after the triggering step & doesn't introduce artifacts at the edges?

NuRadioReco/utilities/noise.py
```diff
@@ -508,15 +553,22 @@ def generate_noise(self, phasing_mode="slice", trigger_mode="binned_sum", debug=
             dt_triggering += time.process_time() - tstart
 
             if is_triggered:
-                triggered_bin = triggered_bin // 2  # the trace is cut in the downsampled version. Therefore, triggered bin is factor of two smaller. 
+                triggered_bin = triggered_bin // self.upsampling  # the trace is cut in the downsampled version. Therefore, triggered bin is factor of two smaller.
```
sjoerd-bouma, line 556: This is not introduced by this PR, but what happens if `trigger_mode == "stride"`? Wouldn't `triggered_bin` be undefined in that case?

sjoerd-bouma (commented):
Thanks Alan, looks like an upgrade. Just some minor comments (I think).

NuRadioReco/utilities/noise.py
colemanalan: Fixed the signature.

This is correct. If you generated, band-limited noise, then a roll is equivalent to a Fourier shift. The magnitude of the frequency content remains unchanged. So it still has the same band, phase-delays, etc

NuRadioReco/utilities/noise.py
colemanalan: Fixed this :)

sjoerd-bouma (approved):
Thanks for all the improvements and fixes! The error in the documentation test has already been fixed in the latest develop, so I think this should be good to go.

PR #548, opened 2023-06-23
fix positions/orientations of LPDAs

sjoerd-bouma:
Ran into this when looking at the RNO-G single station layout - some of the LPDA positions/orientations looked wonky. I think this should fix those.

Old:
Screenshot from 2023-06-23 12-53-59

New:
RNO_single_station_new

I also added the 'reference_station' specification to the station description, which is how the Detector class currently distinguishes a GenericDetector from a Detector, but this seems somewhat unintuitive for a detector with a single station.


sjoerd-bouma:
It seems my dodgy documentation building script has found a new way to raise error messages. This has nothing to do with this PR, and doesn't actually break the documentation, but I guess it will need fixing...

lpyras (approved):
Looks good to me - thanks for doing that.

sjoerd-bouma:
> > Sorry, Robin noticed yesterday that for some reason only channels 12-16 additionally had a cable delay specified (it's set equal to 0 for all other channels), I've removed these now. It probably makes sense to use reasonable guesses for the cable delays rather than 0, but at least this way it's consistent.
> 
> For the surface antennas it doesn't really matter, since all antennas have the same cable delay. so zeros is fine.

The problem was that they were specified only for 6 of the 9 surface channels, and 0 for the remaining 3 9 (as well as all deep channels). Now it's 0 for all of them.

lpyras (approved):

PR #549, opened 2023-07-06
Improve debug output direction fitter

cg-laser:
improves output of correlation DirectionFitter to calcualate angular differences better. 

Also fixed hardcoded RNO-G specific import

### DIFF
--- a/NuRadioReco/eventbrowser/dataprovider.py
+++ b/NuRadioReco/eventbrowser/dataprovider.py
@@ -1,4 +1,7 @@
-import NuRadioReco.eventbrowser.dataprovider_root
+try:
+    import NuRadioReco.eventbrowser.dataprovider_root
+except:
+    print("can't import RNO-G root reader")
 import NuRadioReco.eventbrowser.dataprovider_nur
 
 
--- a/NuRadioReco/modules/correlationDirectionFitter.py
+++ b/NuRadioReco/modules/correlationDirectionFitter.py
@@ -249,7 +249,10 @@ class correlationDirectionFitter:
                     ops += "{:.1f}, ".format(x / units.deg)
                 ops += ")"
                 self.logger.debug(ops)
-                ops = "average incident azimuth {:.1f} +- {:.1f}".format(np.mean(sim_az) / units.deg, np.std(sim_az) / units.deg)
+                temp_complex_representation = np.exp(1j * sim_az)
+                mean_sim_az = np.angle(np.mean(temp_complex_representation.real)+ 1j * np.mean(temp_complex_representation.imag))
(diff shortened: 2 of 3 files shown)


NuRadioReco/modules/correlationDirectionFitter.py
```diff
@@ -249,23 +249,28 @@ def ll_regular_station_fft(angles, corr_02_fft, corr_13_fft, sampling_rate, posi
                     ops += "{:.1f}, ".format(x / units.deg)
                 ops += ")"
                 self.logger.debug(ops)
-                ops = "average incident azimuth {:.1f} +- {:.1f}".format(np.mean(sim_az) / units.deg, np.std(sim_az) / units.deg)
+                temp_complex_representation = np.exp(1j * sim_az)
+                mean_sim_az = np.angle(np.mean(temp_complex_representation.real)+ 1j * np.mean(temp_complex_representation.imag))
+                std_sim_az = np.angle(np.std(temp_complex_representation.real)+ 1j * np.std(temp_complex_representation.imag))
```
sjoerd-bouma, line 254: First line could in principle be simplified (the two means are additive), and I'm not sure the second one is correct - [Wikipedia](https://en.wikipedia.org/wiki/Wrapped_normal_distribution#Moments) backed up my some toy Monte Carlo suggests
```
std_sim_az = np.sqrt(np.log(1/np.mean(temp_complex_representation)**2))
```

NuRadioReco/eventbrowser/dataprovider.py
```diff
@@ -1,4 +1,7 @@
-import NuRadioReco.eventbrowser.dataprovider_root
+try:
+    import NuRadioReco.eventbrowser.dataprovider_root
+except:
+    print("can't import RNO-G root reader")
```
sjoerd-bouma, line 4: The dataprovider_root has not yet been updated for the new RNO-G reader, and we already removed the old one, leading to this error. This is already fixed on `rnog_eventbrowser`, I can make a PR next week. But it's probably okay to leave it like this.

sjoerd-bouma (changes requested):
See comment regarding the standard deviation - I don't think this is correct right now. Other than that, this looks good to me. Documentation failing is due to the make_docs script, I'll look into that next week as well. Thanks Christian!

NuRadioReco/eventbrowser/dataprovider.py
fschlueter: Yes this class does not exist anymore.

fschlueter:
@cg-laser the requested changes seem minimal. Can we conclude on this PR soon?

cg-laser:
let's close this PR. It is likely anyway outdated.

PR #550, opened 2023-07-10
Feature/block offsets

sjoerd-bouma:
(Finally) - added a module that can add or remove 'block offsets' as seen in the voltage traces recorded by RNO-G. The fit is done by assuming a repeating rect shape, and fitting this in the Fourier domain, outside the band where the antennas/amplifiers are expected to be sensitive to real signals. 

I also added an additional `channelParameter` that stores the added/removed block offsets; maybe this can be used for the RNOG data reader in the future to ensure that the correction remains reversible. We can also consider adding (some version of) this module in the RNO-G data reader, while we're still waiting for the integration of the equivalent code on the DAQ side, although it's probably worth making sure first that the performance impact of the fitting and/or Fourier transforms is not too high.


cg-laser (commented):
looks good to me but someone with RNO-G data expertise should also have a look and check/test it.

shallmann (approved):
This is great! I think it can be merged, since it is standalone... not sure why the doc checks fail.

shallmann (dismissed):

fschlueter:
Damn, I did not know that I had to submit my comments/review. Sorry Sjoerd that those come so late (they were written a week ago)

NuRadioReco/modules/RNO_G/channelBlockOffsetFitter.py
```diff
@@ -0,0 +1,257 @@
...
+        ----------
+        event: NuRadioReco.framework.event.Event | None
+        station: NuRadioReco.framework.station.Station
+            The station to remove the block offsets from
+        mode: 'fit' | 'approximate' | 'stored'
+            
+            - 'fit' (default): fit the block offsets with a minimizer
+            - 'approximate' : use the first guess from the out-of-band component,
+              without any fitting (slightly faster)
+            - 'stored': use the block offsets already stored in the
```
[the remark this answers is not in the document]
sjoerd-bouma: The 'stored' option is probably superfluous, but I thought maybe one might want to undo the block offsets that have already been removed by e.g. the external calibration, or allow for some other fit to compute the block offsets.

In terms of simulation, right now we don't simulate the block offsets anyway, and because e.g. the timings are different I'm not sure how straightforward it would be to simulate them for the `sim_station` (rather than for the `station`, which is what I've been doing so far). Even if we end up doing this for the `sim_station`, I'd be inclined to leave it to the user to retrieve the simulated parameter from the `sim_station` to keep a cleaner separation between the two.

sjoerd-bouma:
I fixed (hopefully) Felix's comments, and additionally added the fitter to the mattak datareader. A quick comparison of how fast the different options are:
```
from NuRadioReco.modules.io.RNO_G.readRNOGDataMattak import readRNOGData
import time
reader = readRNOGData()
results = dict()
for mode in ['none', 'median', 'approximate', 'fit']:
    reader.begin('/home/sb/Python/scratch/data/inbox/station11/run1100/', apply_baseline_correction=mode)
    t0 = time.time()
    n_evts = 0
    max_n_evts = 1e4 if mode != 'fit' else 25

    for evt in reader.run():
        n_evts += 1
        if n_evts >= max_n_evts:
            break

    results[mode] = (time.time() - t0) / n_evts
    
for mode, t in results.items():
    print(f"{mode:12s} : {1e3*t:-9.3f} ms / event")
```
Giving 
```
none         :     1.888 ms / event
median       :     4.409 ms / event
approximate  :     4.824 ms / event
fit          :   768.660 ms / event
```

Both 'approximate' and 'median' (= the previous implementation in the reader) are probably good enough for most purposes.

fschlueter (dismissed):

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -57,8 +73,12 @@ def baseline_correction(wfs, n_bins=128, func=np.median):
     # np.moveaxis -> (n_events, n_channels, 2048)
     baseline_traces = np.moveaxis(baseline_traces, 0, -1)
      
+    if return_offsets:
+        return wfs - baseline_traces, baseline_values
+    
     return wfs - baseline_traces
 
+blockoffsetfitter = channelBlockOffsets()
```
fschlueter, line 81: That that here in between functions

fschlueter:
> > What happens if you relax the `tol` argument or introduce a max number of iterations within `curve_fit`? I assume the fit should converge stabil and fast
> 
> I'm definitely spending too much time on this now.... but of course it's a good idea, so I checked using toy Gaussian offsets of 10% on top of a (bandpass-filtered) Gaussian trace. image Basically, after 2 iterations the block offsets are already reduced by 2 orders of magnitude (RMS is normalized to the offset size), which is probably enough in most cases unless the offsets are huge, so I've made this the default `maxiter`.
> 
> I've also removed/deprecated the old baseline correction, and made the bandpass-filtered 'approximate' mode the default, seeing as it was about a factor 4 better without a noticeable decrease in performance. Let me know if you disagree.

Hi, you tested it with 10% amplitudes of what, a baseline ADC of ~ 1800? I am a bit worried that `maxiter=2` is a bit low. Not that an accuracy of `1e-2` is not sufficient but what happens if a fit does not converge as fast as in your test? Looking at your plot: Why don't we set it to ~5. It still gives us a time boost of over one order of magnitude right?

fschlueter:
Deprecating the old code is fine for me. However, I am still wondering a bit why it is not significant faster than the bandpass-filtered 'approximate' mode. After all this includes running an fft which should be quite time consuming compared to the other mathematical operations ?

fschlueter:
PS: An improvement with a factor of 100 in performance is never a waist of time :)

sjoerd-bouma:
Thanks Felix,
 I've increased the number of iterations to 5 as suggested - the 99th percentile is now ~2% (was 4%) 
image
The times are a bit different compared to the previous plot, I realize showing the time per channel after earlier providing times per event (= 24 channels) was a bit confusing. So the improvement is not a factor 100 but closer to a factor 10 in speed, unfortunately.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
sjoerd-bouma: I've made it a private attribute of the `readRNOGData` class, having `readRNOGDataMattak.blockoffsetfitter` showing up in IDE autocompletion is probably unnecessary clutter.

fschlueter (dismissed):

sjoerd-bouma:
@fschlueter Sorry, created a merge conflict for myself by removing all trailing whitespaces in #557, I've now turned that feature off again in my IDE. Nothing else should have changed.

fschlueter (approved):

fschlueter:
I just merged it. Did not feel like waiting longer ..

anelles:
Hmm, reading the history here, @fschlueter may just have avoided (but just barely :) ) the three strikes rule of only merging after 24 hours after final approval. It has been taking long enough, I concur.

PR #552, opened 2023-07-14
improve documentation script error parsing

sjoerd-bouma:
The documentation test currently fails because the `make_docs.py` still parses errors by string recognition, which is tricky for multi-line errors. This PR should split up the errors into errors emitted by sphinx-build (which are easier to separate and parse) and other errors during the build process, which the documentation script should now no longer fail on.

### DIFF
--- a/documentation/Makefile
+++ b/documentation/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -w sphinx-debug.log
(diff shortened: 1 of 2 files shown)


sjoerd-bouma:
For what it's worth, the exact reason the script was failing previously is that documentation references to the `aenum` package fail. While there was already a line in `conf.py` to suppress the resulting error, they recently changed the name of the class. Now errors for either class name are suppressed.

cg-laser (approved):

PR #554, opened 2023-07-24
Automatically make tag and github release on pushes to master

sjoerd-bouma:
This PR 
- fixes the version check - this test will now fail if a PR still has the same version number as the latest release on `master`, but does not perform a version check on the `master` branch itself.
- Automatically creates a tag and a github release for each push to master

An automatic changelog is included for each github release, which is just a list of all merged PRs since the last release (see the section(s) 'What's changed' in https://github.com/nu-radio/NuRadioMC/releases). I probably could write a small script to also include the changes in `changelog.txt`, if we think this is worth doing.


cg-laser:
Hey Sjoerd, thanks for further improving the workflow! Showing the relevant part of `changelog.txt` would be quite useful to get a quick overview of the relevant changes.

sjoerd-bouma:
@cg-laser 
I've added some regex that should in theory copy the part of the changelog that has changed since the latest github release into the github release notes. The [latest release](https://github.com/nu-radio/NuRadioMC/releases/tag/v2.2.1) was produced with the code I've added, so I'm hopeful it does what it should, but I've added it at the end of the publish_and_build workflow so that if it does break, the pip releases still get published as usual. 

Also, it's worth remembering that if it gets the text wrong on a github release that still can be edited afterwards (if we want to be very careful, in principle one can add `--draft` to the `gh release create` command so that manual confirmation is still required before the release is visible on github, but seeing as at that point the pip release is already out I think that would be overkill).

cg-laser (approved):
Cool, thanks a lot. This will make my life even easier in the future. I'm eager to try it out ;-)

PR #557, opened 2023-08-04
Fixes for eventbrowser

sjoerd-bouma:
This branch contains a couple of fixes for the eventbrowser:
- Implements the new RNO-G data reader (the old one has been deprecated and raises an ImportError)
- Correct the normalization of the shown frequency spectra (previously, the numpy convention rather than the NuRadio convention was used in some places)
- Add a toggle to display frequency spectra in log-scale rather than linear scale.

This last feature also requires an update to dash>=2.9.2

The branch is pretty old and seems to include more 'features' than I realized (e.g. an old RNO-G data example), so I probably have to clean this up before we can merge this.


MijnheerD:
Hey @sjoerd-bouma , any updates on this? :-) I am also trying to use the EventBrowser, and found the same issue with the old RNO-G reader. However, I am also encountering an issue where the stations are not drawn, even though I saved the detector instance using the EventWriter. Just like in #563 I am getting no errors whatsoever, so there might be a common issue.

fschlueter:
@sjoerd-bouma, yes your solution looks cleaner. I haven't checked if it works similar to my solution did you? How many times is the io class spawned when you read a single file (single user)? What happens if the user_id changes from None to a hash? We should also merge the 3 dataprovider files/classes into one! Can you do that? Should we sit together when I am in Erlangen?

fschlueter:
@sjoerd-bouma thanks for doing that! I noticed that the data handler spawns still two instances of the io class (once for user None and once for the user with a specific hash). Do you think we could fix that or not? Otherwise this looks good to me!

sjoerd-bouma:
@fschlueter Yeah, I guess that would be good to fix, but I'm not sure how straightforward that would be... the user_id is set by a callback, and they fire in parallel threads with no way of forcing that one to fire before all the others, I think.

fschlueter:
Okay, you could think about copying the object rather than initialising a new one once a none None user_id is passed. But I am fine with how it is right if you want to merge it.

fschlueter (approved):
I have to admit I did not look at all the changes in the "plotting" scripts. Only looked at the data provider and index.py

sjoerd-bouma:
Sorry @fschlueter, I created a merge conflict so you'll have to approve again. I also hopefully 'fixed' the user_id=None issue - if there is an entry for user None present this gets 'recycled' by the first user with an actual id.

fschlueter (approved):

PR #558, opened 2023-08-04
raytracing from ice to air

cg-laser:
I started implementing the raytracing from ice to air. In principle it is simple. If the receiver is above the surface, only consider rays that hit the surface, refract out of the ice and extend the path with a straight line in air. 
The minimization is also easier because there is always one unique solution and we know the boundaries of the launch angle, so an efficient root-finding algorithm can be used (instead of a minimization). 

The tricky part is to integrate it into all the other cases (including multiple reflections between surface and bottom layer before refracting into air) and extending all function to consider this special case. 

So far the ray tracing itself works (although not yet heavily tested). The functions to calculate the pathlength and propagation times still need to be extended. Also all changes were only done to the python version. 

It might also be tricky to integrate it into a full NuRadioMC simulation because we only have one RT solution (and not two).

### DIFF
--- a/NuRadioMC/SignalProp/analyticraytracing.py
+++ b/NuRadioMC/SignalProp/analyticraytracing.py
@@ -40,6 +40,8 @@ except:
         print("check NuRadioMC/NuRadioMC/SignalProp/CPPAnalyticRayTracing for manual compilation")
         cpp_available = False
 
+cpp_available = False
+
 """
 analytic ray tracing solution
 """
@@ -132,11 +134,10 @@ class ray_tracing_2D(ray_tracing_base):
         self.__logger.setLevel(log_level)
         self.__n_frequencies_integration = n_frequencies_integration
         self.__use_optimized_start_values = use_optimized_start_values
-        
+
         self._use_optimized_calculation = self.attenuation_model in speedup_attenuation_models
         if overwrite_speedup is not None:
             self._use_optimized_calculation = overwrite_speedup
-                    
 
     def n(self, z):
         """
@@ -590,19 +591,18 @@ class ray_tracing_2D(ray_tracing_base):
             self.__logger.debug(f"calculating attenuation for frequencies {freqs}")
             return freqs
 
-
-    def get_attenuation_along_path(self, x1, x2, C_0, frequency, max_detector_freq, 
+    def get_attenuation_along_path(self, x1, x2, C_0, frequency, max_detector_freq,
                                    reflection=0, reflection_case=1):
-        
+
         attenuation = np.ones_like(frequency)
 
         output = f"calculating attenuation for n_ref = {int(reflection):d}: "
-        for iS, segment in enumerate(self.get_path_segments(x1, x2, C_0, reflection, 
+        for iS, segment in enumerate(self.get_path_segments(x1, x2, C_0, reflection,
                                                             reflection_case)):
-            
-            # we can only integrate upward going rays, so if the ray starts downwardgoing, 
+
+            # we can only integrate upward going rays, so if the ray starts downwardgoing,
             # we need to mirror
-            if iS == 0 and reflection_case == 2: 
+            if iS == 0 and reflection_case == 2:
                 x11, x1, x22, x2, C_0, C_1 = segment
                 x1t = copy.copy(x11)
                 x2t = copy.copy(x2)
@@ -612,20 +612,20 @@ class ray_tracing_2D(ray_tracing_base):
                 x1 = x1t
             else:
                 x11, x1, x22, x2, C_0, C_1 = segment
-                
+
             if cpp_available:
                 mask = frequency > 0
                 freqs = self.__get_frequencies_for_attenuation(frequency, max_detector_freq)
                 tmp = np.zeros_like(freqs)
                 for i, f in enumerate(freqs):
                     tmp[i] = wrapper.get_attenuation_along_path(
-                        x1, x2, C_0, f, self.medium.n_ice, self.medium.delta_n, 
+                        x1, x2, C_0, f, self.medium.n_ice, self.medium.delta_n,
                         self.medium.z_0, self.attenuation_model_int)
-                
+
                 self.__logger.debug(tmp)
                 tmp_attenuation = np.ones_like(frequency)
                 tmp_attenuation[mask] = np.interp(frequency[mask], freqs, tmp)
-            
+
             else:
 
                 x2_mirrored = self.get_z_mirrored(x1, x2, C_0)
@@ -633,7 +633,7 @@ class ray_tracing_2D(ray_tracing_base):
                 def dt(t, C_0, frequency):
                     z = self.get_z_unmirrored(t, C_0)
                     return self.ds(t, C_0) / attenuation_util.get_attenuation_length(z, frequency, self.attenuation_model)
-                
+
                 # to speed up things we only calculate the attenuation for a few frequencies
                 # and interpolate linearly between them
                 mask = frequency > 0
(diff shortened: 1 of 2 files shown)


sjoerd-bouma:
Oops, broke everything - need to look into this z mirroring as well.

cg-laser:
Note from InIceMC meeting: 
* transmission coefficients at the boundary might depend on direction (ice to air or air to ice). So far the code assumes ice to air.

cg-laser:
I think I should have added the calculation of transmission coeffiiencts correctly but I noticed a few things:

The transmission coefficient is polarization dependent, so it's calculation needs to be added to the part of the code where the reflection coefficients are calculated. 

The transmission coefficient AND focusing factor depends on the direction (air-> ice or ice->air). All other properties are the same. The transmission coefficient can be larger than 1 which is expected for media with different index-of-refraction. 
Are we sure that we are not double counting?

Is the focusing factor being calculated correctly? Navively I would think that the focusing factor should be inverted if the propagation direction is changed. @sjoerd-bouma what do you think? 

```
finding solutions for [10000.     0.   100.] to [   0.    0. -149.]
     Solution 0, Type 3:
     Ray Distance 10071.793m 10071.793m and Travel Time 33978.540ns 33978.540ns
     Receiving Zenith 35.363 and Azimuth 0.000
     focusing factor = 0.08550910
[0.93755678 0.92127416]
WARNING:ray_tracing_analytic:propagating from air to ice: transmission coefficient is 0.77, 0.75
```

```
finding solutions for [   0.    0. -149.] to [10000.     0.   100.]
     Solution 0, Type 3:
     Ray Distance 10071.793m 10071.793m and Travel Time 33978.540ns 33978.540ns
     Receiving Zenith 89.420 and Azimuth 180.000
     focusing factor = 0.11248427
[0.93755678 0.92127416]
WARNING:ray_tracing_analytic:propagating from ice to air: transmission coefficient is 2.69, 1.99
```

sjoerd-bouma:
Ah, yes, that's all in the presentation I never gave...

So I ended up deriving an alternative (but equivalent) formula for the focussing factor. As it turns out I don't have it in neat LaTeX anywhere I'll just include the final result for the focussing factor during propagation from $x_1$ to $x_2$ with a path length $R=\int ds$:

$$1/f^2 = \frac{n'}{n}\frac{A'}{A} = \frac{n(x_2)}{n(x_1)} \frac{n(x_1)\cos\theta(x_1)\cos\theta(x_2)}{R} \int ds \frac{1}{n(\vec{x}) \cos^2 \theta(\vec{x})},$$

Some comments:
- Here $\theta$ is the angle to the normal, not the usual NuRadioMC convention
- If $n(\vec{x})=n(x_1)=n(x_2)$ is constant, this expression reduces to unity, as expected
- The first factor $\frac{n'}{n}$ is purely due to optics (as opposed to geometry), and due to energy conservation, though I realized this morning I'm a little bit confused on how this factor is derived - naively I would have thought the energy contained in the Efield, $\frac{1}{2}\epsilon|E|^2$ should be conserved, i.e. $|E|^2 \propto \frac{1}{\epsilon} \propto \frac{1}{n^2}$ (rather than $\propto 1/n$). Let me know if you have a clearer idea on where this factor originates, otherwise I'll see if Dave Seckel remembers... another potential issue here is probably double-counting; I'm not sure how this factor relates to the transmission coefficients you mention above.
- The rest is due to the geometry (diverging/converging ray bundles). Some things that I found unintuitive:
  - The integral is the same in both directions - if this dominates, (de)focussing will occur in both directions (in particular, the focussing factor is **not** inverted if the direction of propagation is inverted!)
  - The (de)focussing also continues in areas of constant $n$, i.e. for the same entry/exit angles at the ice/air interface the defocussing is stronger if the distance propagated in air is longer.

The short version is this formula turned out to give the same answer as the existing implementation (with the exception of refracted rays, as the integral above goes to infinity for horizontal propagation and thus becomes numerically unstable), so I'm fairly confident the $\frac{A'}{A}$ factor is correct. I took the first factor for granted originally but it might be good to verify once more it is correct, and to make sure we do not end up double-counting if we also take the transmission factor into account. I will try to follow up on this...

cozzyd:
I tried this out a little and noticed that there are exceptions thrown from the Brent solver for near-upward pointing trajectories. For example, if you replace one of the coordinates in the example with [0.5,0,2.5]. 

```
  File "/home/cozzyd/Work/NuRadioMC/NuRadioMC/SignalProp/examples/E02ToAir.py", line 39, in <module>
    r.find_solutions()
  File "/home/cozzyd/Work/NuRadioMC/NuRadioMC/SignalProp/analyticraytracing.py", line 1934, in find_solutions
    self._results = self._r2d.find_solutions(self._x1, self._x2)
  File "/home/cozzyd/Work/NuRadioMC/NuRadioMC/SignalProp/analyticraytracing.py", line 1275, in find_solutions
    result = optimize.brentq(self.obj_delta_y, logC0_start, logC0_stop, args=(x1, x2, reflection, reflection_case))
  File "/usr/lib64/python3.9/site-packages/scipy/optimize/zeros.py", line 776, in brentq
    r = _zeros._brentq(f, a, b, xtol, rtol, maxiter, args, full_output, disp)
ValueError: f(a) and f(b) must have different signs

```

sjoerd-bouma:
> I also don't think the focusing factor gets inverted when reversing the problem. In the ground-bounce measurements, for example, the focusing factor on the way down does NOT cancel the focusing factor on the way up.
> 
> But isn't the focusing factor sort of the same as applying the transmission equation over and over again on infinitesimal layers? So there is probably some double counting there....

My suspicion is that the transmission factor does already account for the factor $\frac{n'}{n}$. I don't think it accounts for the factor $\frac{A'}{A}$; e.g. in constant $n$ the transmission factor would be 1, whereas as I mentioned above (de)focussing continues to happen even in areas of constant $n$. In my mind, both of these effects are separate, and it's somewhat unfortunate we've grouped them together; I would have called only the second factor 'focussing factor' as this is purely due to ray bundle convergence or divergence.

cozzyd:
yes, I think I agree!

cg-laser:
Thanks for testing the code @cozzyd . Can we separate the transmission coefficient of the air/ice boundary from the rest of the focussing by only using a `n_of_firn_at_surface/n_at_receiver_posiiton` @sjoerd-bouma ? Then to first order, we should calculate things correctly.

The bug in the root finder should be simple to fix. For ice/air transmission, there is only one solution and the brentq method is only used to find the second solution which does not need to be done.

cg-laser:
I fixed the problem of vertical propagation. It was a bit more complicated but now it works for almost vertical directions. 

The problem is that the analytic function only becomes perfectly vertical if one of its parameters approaches infinity.... Therefore, the code does not support completely vertical directions, but it gets fairly close to vertical. I think that should be enough for most applications. 

The remaining todo is to get the n_air/n_firn factor out of the focussing correction to avoid double counting with the Fresnel transmission coefficient. @sjoerd-bouma would this be straight forward to implement?

cg-laser:
Thanks @sjoerd-bouma 
It seems that this PR would be ready to be merged now?

sjoerd-bouma (approved):
I hope all the tracking of in-air versus mirrorred internally is implemented correctly at the moment; I also have not thoroughly checked what happened in the case of (additional) bottom reflections (is that supported for ice-to-air right now?). And it would be good to validate that we're not double counting in the focussing and transmission coefficients. But seeing as the code emits a warning there I think we can merge this now and come back to it later if necessary.

sjoerd-bouma (approved):

fschlueter:
@cg-laser does this only works if the transmitter-receiver pair are in one plane? I am running some raytracing for an air shower and want to ray trace from Xmax to the showers in the ice. I get this warning (and hence no results) for all antennas which have `x!=0` 
```
WARNING - NuRadioMC.ray_tracing_2D - can't find a solution for ice/air propagation. The trajectory might be too vertical! This is currently not supported because of numerical instabilities.
```

fschlueter:
The full debug output for the raytracing I am running is:
```
--------------> Finding ray tracing between [ 123.5 -437.6  -80. ] and [ -0.  -0. 434.]
DEBUG - NuRadioMC.ray_tracing_analytic - X1 = [ 123.5 -437.6  -80. ], X2 = [ -0.  -0. 434.]
DEBUG - NuRadioMC.ray_tracing_analytic - dphi = -105.8
DEBUG - NuRadioMC.ray_tracing_analytic - X2 - X1 = [-123.5  437.6  514. ], X1r = [ 123.5 -437.6  -80. ], X2r = [ 578.19331422 -437.6         434.        ]
DEBUG - NuRadioMC.ray_tracing_analytic - 2D points [123.5 -80. ] [578.19331422 434.        ]
DEBUG - NuRadioMC.ray_tracing_2D - C_0 = 26881171418161356094253400435962903554686976.0000, C_1 = 123.5
DEBUG - NuRadioMC.ray_tracing_2D - turning points is z = 0.0, y =  123.5
DEBUG - NuRadioMC.ray_tracing_2D - not refracting into air
DEBUG - NuRadioMC.ray_tracing_2D - C_0 = 1.0000, C_1 = 248.2
DEBUG - NuRadioMC.ray_tracing_2D - turning points is z = 0.0, y =  191.4
DEBUG - NuRadioMC.ray_tracing_2D - reflecting off surface at y = 191.4m, reflection angle = 51.9deg
DEBUG - NuRadioMC.ray_tracing_2D - touching surface at 51.9deg -> 90.0deg -> x2 = [578.19331422 434.        ] diff 434.00
DEBUG - NuRadioMC.ray_tracing_2D - Looking for ice-air solutions between log(C0) (100, -0.8250747236024935) with delta_y (434.0, 433.99999423603856)
WARNING - NuRadioMC.ray_tracing_2D - can't find a solution for ice/air propagation. The trajectory might be too vertical! This is currently not supported because of numerical instabilities.
Event 1 1
```

PR #561, opened 2023-08-22
Fixed small comparison

colemanalan:
Simple fix for a bad comparison between an int and an iterable


fschlueter:
I don't want to be annoying but lets fix the coding style when doing those changes?

fschlueter (approved):
<3

anelles:
Sooo much love. 
And no need to apologize for being nit-picky -- this is how the code review works.

PR #563, opened 2023-08-28
Fix detector browser

fschlueter:
Fixes a bug which I believe was introduced with PR #418 when the `detector.__init__` method was replaced by the `detector.__new__` method. What was the reason for this? The other detector classes only implement the `__init__` function. 

The bug was caught by Mitja Desmet.

### DIFF
--- a/NuRadioReco/detector/detector_browser/detector_provider.py
+++ b/NuRadioReco/detector/detector_browser/detector_provider.py
@@ -29,8 +29,9 @@ class DetectorProvider(object):
         filename: string
             Path to the .json file containing the detector description
         """
-        self.__detector = NuRadioReco.detector.detector.Detector.__new__(NuRadioReco.detector.detector.Detector)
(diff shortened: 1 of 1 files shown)


fschlueter:
@christophwelling & @cg-laser is this app actively maintained? I just tried to open the RNO_G_season_2022.json file and can only see one station.

MijnheerD:
So it gets actually a bit more complicated. I made this test description based on the `detector_dummy.json` that is provided: [LOFAR_test.txt](https://github.com/nu-radio/NuRadioMC/files/12461766/LOFAR_test.txt) . Using the `visualize_detector.py` from the `NuRadioReco/detector` directory works fine and gives a (weird looking but) normal result. However, the detector browser shows nothing when trying to open this file (this is after I changed the call of the Detector `__new__` function to match the signature of the `__init__` function). It also throws no errors, so I have no idea what is going on here.

MijnheerD (approved):

PR #565, opened 2023-08-30
Fix GenericDetector reference channel checking

MijnheerD:
Issue being addressed: the GenericDetector class would complain if the reference channel would not be within the station, but in the reference station. The cause of this issue was simply the check implemented from line 187 in `generic_detector.py` onward. 

When commenting this check out the rest of the code worked fine, indicating that apart from that block the referencing works as intended. Therefore I have simply added a clause to the check inside the reference station in case the reference channel cannot be found in the current station.

Apart from that I have also reformatted the file (using my IDE) to fix PEP8 compliance and correct some typo's.


NuRadioReco/detector/generic_detector.py
```diff
@@ -155,40 +178,48 @@ def __init__(self, json_filename, default_station=None, default_channel=None, de
...
+                    # Check if reference channel sits in reference station
+                    reference_station_id = self.__lookuptable_reference_station[chan["station_id"]]
+                    ref = self._channels.get(
+                        (Channel.station_id == reference_station_id) &
+                        (Channel.channel_id == chan["reference_channel"])
+                    )
+                    if ref is None:
+                        raise ValueError(
+                            'The reference channel {} of station {} was not found in the detector description'.format(
+                                chan["reference_channel"], chan["station_id"]))
```
sjoerd-bouma, line 202: Sorry for weighing in so late and thanks for fixing this. However, I think the previous check (l.190-191) (at least in the current implementation) is simply wrong and should be removed - right now, if the reference_channel is present in the station but not in the reference_station, this will not raise an error here, but it will raise an error when actually trying to read in the reference_channel in l.269-271, because there we always (and only) check the reference_station, not the station.

MijnheerD: I see, but shouldn't this be an option, to specify a reference per station, even if the station has a reference station? Cause then we should actually change the logic of lines 269-271.

PR #567, opened 2023-09-05
This allows to specify root files to be read with the mattak reader i…

fschlueter:
…nstead of directories. Because the mattak library can not handle this so far we have to employ some diry hack. Create a temp directory and link the file in question to a the file combined.root within the directory. Simple and naive housekeeing included

This is not suppose to be a long term solution. But I feel that this feature is needed!


fschlueter:
> Fine with me.
> 
> I suppose you did not want to use e.g. the `tempfile` package to create temporary directories that already clean up after themselves because of the additional required package?

No that was not the reason. They are cleaning up once they run out of scope right? The mattak dataset, as far as I understand them, will only read data into memory once the data is "requested", i.e., when calling `dataset.wfs()`. Hence, that would not work when the tempfiles only "live" during the init function call

PR #570, opened 2023-09-13
upgrade proposal to version 7.6.2

### LINKED ISSUE #569: proposal error (undefined symbol)
The tests currently fail because of an issue with proposal 7.5.1. The installation of proposal does not raise an error, but subsequently importing anything relying on proposal fails with and `undefined symbol` error. Example below:

```
WARNING: autodoc: failed to import module 'NuRadioProposal' from module 'NuRadioMC.EvtGen'; the following exception was raised:
/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/site-packages/proposal.cpython-37m-x86_64-linux-gnu.so: undefined symbol: _ZN5boost10filesystem4pathdVERKS1_
```
It's not clear to me if this is a python 3.7 exclusive issue or due to some updated dependency of proposal, I can't reproduce this on Ubuntu 22.04 and python 3.10.12. I also do not know of a workaround.

Jean1995:
Upgrade PROPOSAL to version 7.6.2

Notably, this upgrades the package manager `conan` (which is used to install proposal dependencies) from version 1 to version 2. This should increase stability, and hopefully fix #569 

Full release notes of PROPOSAL can be found [here](https://github.com/tudo-astroparticlephysics/PROPOSAL/releases), although none of the other changes should affect NuRadioMC.

### DIFF
--- a/documentation/source/Introduction/pages/installation.rst
+++ b/documentation/source/Introduction/pages/installation.rst
@@ -189,7 +189,7 @@ These packages are recommended to be able to use all of NuRadioMC/NuRadioReco's 
 
   .. code-block:: bash
 
-    pip install proposal==7.5.1
+    pip install proposal==7.6.2
 
   Note that the pip installation for this version of proposal may not work on all systems, in particular:
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ numba = "*"
 Sphinx = "*"
 sphinx-rtd-theme = "*"
 numpydoc = "*"
-proposal = "7.5.1"
+proposal = "7.6.2"
 pygdsm = {git = "https://github.com/telegraphic/pygdsm"}
 nifty5 = {git = "https://gitlab.mpcdf.mpg.de/ift/nifty.git", branch="NIFTy_5"}
 pypocketfft = {git = "https://gitlab.mpcdf.mpg.de/mtr/pypocketfft"}


sjoerd-bouma (approved):
Thanks @Jean1995 for looking into this so quickly!
I'm not an active proposal user but if nothing relevant for us has changed and the tests pass I think this can be merged (@shallmann?)

Jean1995:
If there is a specific set of clusters, it might be a good idea to check whether the installation still works. Just to make sure that the switch to conan2 didn't introduce any unforeseen issues.

fschlueter:
Can this be merged?

shallmann:
Did not try if the install works on the cluster... try to do this asap

sjoerd-bouma:
Hi all,
I agree we should probably check that 7.6.2 works on all systems, but considering we know that 7.5.1 **doesn't** work, I would suggest merging this at this stage and creating a new issue should we find any with 7.6.2. It's a little bit of an annoyance that our CI constantly fails on proposal at the moment, obscuring any other issues with new code.

fschlueter:
> Hi all,
> I agree we should probably check that 7.6.2 works on all systems, but considering we know that 7.5.1 **doesn't** work, I would suggest merging this at this stage and creating a new issue should we find any with 7.6.2. It's a little bit of an annoyance that our CI constantly fails on proposal at the moment, obscuring any other issues with new code.

I second this!

PR #571, opened 2023-09-16
Bug fix for antennapattern.py

aj-nielsen:
Extremely minor bugfix. In antennapattern.py, class AntennaPatternAnalytic._get_antenna_response_vectorized_raw, line 1410, frequencies are constrained by the following mask for the "analytic_LPDA" antenna model:

`fmask = freq >= 0`

However, this allows for frequencies that are equal to 0, which causes issues later on, as other parameters are divided by the frequency. If freq = 0, of course, this returns NaN values for those parameters.

My fix is simply:
`fmask = freq > 0`
so that NaNs can not be produced if one uses this antenna model.

### DIFF
--- a/NuRadioReco/detector/antennapattern.py
+++ b/NuRadioReco/detector/antennapattern.py
@@ -64,8 +64,14 @@ def interpolate_linear_vectorized(x, x0, x1, y0, y1, interpolation_method='compl
     """
     x = np.array(x)
     mask = x0 != x1
-    result = np.zeros_like(x, dtype=complex)
+    result = np.zeros_like(x, dtype=np.complex)
     denominator = x1 - x0
+    
+    #debug test:
+    #if x0 == x1:
+        #return y0
+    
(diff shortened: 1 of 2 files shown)


fschlueter:
@aj-nielsen there seem to be a number of unintended changes in this PR.

- the data type changed from the native build in type to the corresponding numpy type
- Some comments / commented code
- Changes in the  `changelog.txt`

Please fix those first! I can imagine that those unintended changes come from the fact that you work on a forked repository and that you are basing your changes on a branch/version of the code which have been updated in the meantime. I would suggest to you to checkout a new branch within the main repository and reimplement your change there (as it seems to be minor as you say)

aj-nielsen:
@fschlueter Fixed (I think?), though I had to make this fix in an entirely separate branch on my fork, so I'm about to make a new pull request. See pull that I'm putting up in the next 15 minutes.

PR #572, opened 2023-09-18
Remove non built-in objects for the __modules_event list which gets s…

### LINKED ISSUE #568: Cannot write out event after using eventWriter with detector option
**Describe the issue**
When saving an `Event` with a detector description, the same `Event` cannot be saved again (to a different file for example). Attempting to do this will throw a `TypeError: cannot pickle '_io.TextIOWrapper' object` error.

This is due to how the `Event` object saves the modules applied to it. After running an `eventWriter` with both `Event` and `Detector` objects specified, the `Detector` object becomes a part of the `__modules_event` list of the `Event` which is serialized when writing out the `Event`. However, the `Detector` object cannot be pickled (I guess due to the TinyDB objects contained within it), so when attempting to `serialize` the `Event` again, it throws the error mentioned above.

**Expected behavior**
I should be able to save an `Event` twice in the same script, for example to store it before and after calibration.

I don't know what the cleanest solution would be, but my suggestion would be to **not** add `eventWriter` operations to the `__modules_event` list, as it seems pretty obvious this module was applied if there exists a .nur file :-)

**Additional context**
MWE:
```python
from NuRadioReco.detector import detector
from NuRadioReco.framework import event
from NuRadioReco.modules.io import eventWriter

evt = event.Event(0, 0)
det = detector.Detector('NuRadioReco/examples/example_data/arianna_station_32.json')

eventWriter = eventWriter.eventWriter()
eventWriter.begin(f'NuRadioReco/examples/LOFAR_data/test.nur')
eventWriter.run(evt, det=det)
eventWriter.end()

eventWriter.begin(f'NuRadioReco/examples/LOFAR_data/test2.nur')
eventWriter.run(evt, det=det)  # this line will throw a TypeError, even if the "det=det" argument is removed
eventWriter.end()
```

fschlueter:
…erialised together with the event. In this case the optional detector argument of the eventWriter caused the issue. Also adding entries to __modules_event before acutally running the module. In this case the eventWriter will already appear in that list the first time it is called.

This PR should fix #568. However, a few thoughts: 

- Currently there is no warning when an argument is dropped because it is not a built-in type. This might cause lost of needed information. On the other hand we might have frequently arguments which we want to drop. Do we need a white list for types to release a warning (or not to release a warning)
- As far as I understand we only store arguments passed to the function with a keyword. Hence we are ignoring any positional argument and any argument using its default value (the latter is problematic if defaults change). I think we should fix that, any opinions on that?

### DIFF
--- a/NuRadioReco/framework/event.py
+++ b/NuRadioReco/framework/event.py
@@ -8,13 +8,8 @@ import NuRadioReco.framework.parameters as parameters
 import NuRadioReco.utilities.version
 from six import itervalues
 import collections
-import copy
 import logging
 logger = logging.getLogger('Event')
-
-
-def is_serializable(obj):
-    return obj.__class__.__module__ == 'builtins' or obj.__class__.__module__ == "numpy"
 
 
 class Event:
@@ -46,15 +41,6 @@ class Event:
         kwargs:
             the key word arguments of the run method
         """
-        kwargs = copy.deepcopy(kwargs)
-        # Not every kwargs argument passed to a run(..) method is serializable. Example is a detector object. 
-        # Drop all arguments for which this is the case
-        keys_to_be_dropped = [key for key in kwargs 
-                             if not is_serializable(kwargs[key])]
-        
-        for key in keys_to_be_dropped:
-            logger.warn(f"Dropping \"{key}\" of type {type(kwargs[key])} from kwargs of __modules_event")
-            kwargs.pop(key)
                 
         self.__modules_event.append([name, instance, kwargs])
 
@@ -73,20 +59,9 @@ class Event:
         kwargs:
             the key word arguments of the run method
         """
-        kwargs = copy.deepcopy(kwargs)
-
         if station_id not in self.__modules_station:
             self.__modules_station[station_id] = []
             
-        # Not every kwargs argument passed to a run(..) method is serializable. Example is a detector object. 
-        # Drop all arguments for which this is the case
-        keys_to_be_dropped = [key for key in kwargs 
-                             if not is_serializable(kwargs[key])]
-        
-        for key in keys_to_be_dropped:
-            logger.warn(f"Dropping \"{key}\" of type {type(kwargs[key])} from kwargs of __modules_station")
-            kwargs.pop(key)
-        
         iE = len(self.__modules_event)
         self.__modules_station[station_id].append([iE, name, instance, kwargs])
 
--- a/NuRadioReco/framework/event.py
+++ b/NuRadioReco/framework/event.py
@@ -12,8 +12,8 @@ import logging
 logger = logging.getLogger('Event')
 
 
-def is_builtin_class_instance(obj):
-    return obj.__class__.__module__ == '__builtin__'
+def is_serializable(obj):
+    return obj.__class__.__module__ == 'builtins' or obj.__class__.__module__ == "numpy"
 
 
 class Event:
@@ -46,9 +46,10 @@ class Event:
             the key word arguments of the run method
         """
         key_to_be_dropped = [key for key in kwargs 
-                             if not is_builtin_class_instance(kwargs[key])]
+                             if not is_serializable(kwargs[key])]
         
         for key in key_to_be_dropped:
+            logger.warn(f"Dropping \"{key}\" of type {type(kwargs[key])} from kwargs of __modules_event")
             kwargs.pop(key)
                 
         self.__modules_event.append([name, instance, kwargs])
@@ -72,9 +73,10 @@ class Event:
             self.__modules_station[station_id] = []
             
         key_to_be_dropped = [key for key in kwargs 
-                             if not is_builtin_class_instance(kwargs[key])]
+                             if not is_serializable(kwargs[key])]
         
         for key in key_to_be_dropped:
+            logger.warn(f"Dropping \"{key}\" of type {type(kwargs[key])} from kwargs of __modules_station")
             kwargs.pop(key)
         
         iE = len(self.__modules_event)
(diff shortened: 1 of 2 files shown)


fschlueter:
Oh wow I did not know that this functionality is actually used somewhere explicitly for calculations... @sjoerd-bouma do you have a better idea how to select which objects we can/should drop? Should we work with white or black lists? That seems a bit hacky

fschlueter:
@sjoerd-bouma @shallmann

sjoerd-bouma:
I'd be inclined to use [inspect.signature](https://docs.python.org/3.8/library/inspect.html#inspect.Signature) to store both positional and kwargs, if I have time to look into it I can, but feel free to do this yourself if you want to get this functionality in quicker. If we ever want to store `*args` this may need some extra care (not sure that this is relevant anywhere right now).

fschlueter:
Hm, I just realised. If we want to "maintain" full reproducibility we would need to store also the arguments from the `begin`/`init` functions.... 
@sjoerd-bouma for the moment are you happy with the current solution?

NuRadioReco/framework/event.py
```diff
@@ -59,8 +72,18 @@ def register_module_station(self, station_id, instance, name, kwargs):
...
             self.__modules_station[station_id] = []
+            
+        # Not every kwargs argument passed to a run(..) method is serializable. Example is a detector object. 
+        # Drop all arguments for which this is the case
+        keys_to_be_dropped = [key for key in kwargs 
+                             if not is_serializable(kwargs[key])]
+        
+        for key in keys_to_be_dropped:
+            logger.warn(f"Dropping \"{key}\" of type {type(kwargs[key])} from kwargs of __modules_station")
+            kwargs.pop(key)
```
sjoerd-bouma, line 85: I'm never 100% sure with Python, but kwargs is a dict, so my intuition would be that this is **not** a local copy and you're modifying it in-place, i.e. when you later pass kwargs back to the `run` method it will have been modified. Can you test this?

sjoerd-bouma (commented):
You're probably right that it would make sense to also store arguments given to `begin`, though it still feels inconsistent to not store positional arguments. I think we can merge this to at least fix the detector storage issue if you prefer, once we've checked that we're not modifying kwargs in-place.

NuRadioReco/framework/event.py
fschlueter: damn, good point.

fschlueter: I think running a deep copy is anyway adequate? (I did not test it though)

fschlueter:
@sjoerd-bouma, fine with that?

sjoerd-bouma:
@fschlueter so I've changed it now to use `inspect.signature`, with the advantage (hopefully) that now all arguments (positional, keyword and defaults) should be stored, as long as they are serializable. I've dropped the warning for any keys that get dropped, mostly because I forgot to implement it (and it would most likely constantly be raised because there is almost always a `detector` object in the signature) - feel free to (ask me to) reimplement it if you think it's important enough.

fschlueter:
I was thinking maybe Christians suggestion is the right approach (just trying to serialize and catch it). However, then we might serialize large objects which takes a lot of time. On the other hand now, we might miss objects we want to serialize and would not realise

sjoerd-bouma:
> I was thinking maybe Christians suggestion is the right approach (just trying to serialize and catch it). However, then we might serialize large objects which takes a lot of time. On the other hand now, we might miss objects we want to serialize and would not realise

Sure, let me know if you think the new approach is better. I imagine it will lead to constant warnings about dropped detectors, but I have no strong preference either way.

fschlueter:
@sjoerd-bouma, ah because now we take all optional arguments (otherwise we would have gotten a lot of errors before which we did not). How about a whitelist for which types to not raise a warning. This is at least intentional ?

sjoerd-bouma:
@fschlueter I've just realised you can't actually review this PR, because it is yours, but let me know what you think before I self-approve it. Or maybe we can get @shallmann or @cg-laser to weigh in?

cg-laser (commented):
Hey Felix, Sjoerd, thanks for implementing this. 

A possible solution to the warning or not question: We need to print out a warning so that users know what is going on but we can only print this warning once? Maybe we can save if a warning for this (module, kwarg) pair was already issued.

fschlueter:
@sjoerd-bouma, the changes look good to me. One minor remark: You missed a whitespace when separating two variables with a "," several times (`x,y` -> `x, y`). I think if we exclude detector objects from raising a warning we will not encounter many of those?

sjoerd-bouma:
Thanks both. I think as long as we're okay with whitelisting detectors (i.e. the current implementation) @fschlueter is right and I would not expect a lot of warnings to be raised - scanning through the output of the unit tests I didn't see any, in any case. So I'd be inclined to leave it like this - feel free to let me know if you disagree @cg-laser.

> @sjoerd-bouma, the changes look good to me. One minor remark: You missed a whitespace when separating two variables with a "," several times (`x,y` -> `x, y`). I think if we exclude detector objects from raising a warning we will not encounter many of those?

Good spot! So it turns out this is a purely personal convention, but as usual with those I'm reluctant to change my ways. I find `for k,v in dict.items()` easier to parse than `for k, v in dict.items()`, because it's easier to see that they form a tuple together. I'd probably be happy with `for (k, v) in dict.items()`. But I think PEP agrees with you in this case, so feel free to correct it.

cg-laser (approved):
thanks for checking if warnings occur normally. As this doesn't see to be the case, I'm happy to approve it. We can still implement more checks if this becomes relevant later.

sjoerd-bouma:
@fschlueter lgtm?

fschlueter:
> @fschlueter lgtm?

I was not planing to fix the python style. Functional wise I think that PR is fine.

PR #573, opened 2023-09-19
Files without a detector would cause failures in combination with the event browser 

fschlueter:
Implements a proper logic with the argument 'parse_detector', avoid overwriting functions. Refactoring.

### DIFF
--- a/NuRadioReco/modules/io/NuRadioRecoio.py
+++ b/NuRadioReco/modules/io/NuRadioRecoio.py
@@ -3,8 +3,12 @@ import NuRadioReco.framework.event
 import NuRadioReco.detector.detector
 import NuRadioReco.detector.generic_detector
 import NuRadioReco.modules.io.event_parser_factory
+
 import numpy as np
+import astropy.time
+
 import logging
+
 import time
 import os
 
@@ -41,14 +45,16 @@ class NuRadioRecoio(object):
         buffer_size: int
             the size of the read buffer in bytes (default 100MB)
         """
-        if(not isinstance(filenames, list)):
+        if not isinstance(filenames, list):
             filenames = [filenames]
+        
         self.__file_scanned = False
         self.logger = logging.getLogger('NuRadioReco.NuRadioRecoio')
         self.logger.info("initializing NuRadioRecoio with file {}".format(filenames))
         t = time.time()
         if log_level is not None:
             self.logger.setLevel(log_level)
+        
         # Initialize attributes
         self._filenames = None
         self.__n_events = None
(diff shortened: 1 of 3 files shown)


fschlueter:
A question which came up. If `parse_detector=True` but no detector is in the file, should we raise
```
                if self._current_file_id not in self._detector_dicts:
                    raise AttributeError('The current file does not contain a detector description.')
```
or should we simply return a warning and `None`

fschlueter:
Can we maybe review and merge this and address the question which Mitja raised in a second PR?

NuRadioReco/modules/io/event_parser_factory.py
```diff
@@ -9,11 +9,16 @@ def scan_files_function(version_major, version_minor):
     specifying the first version that function is for
     """
     def scan_files_2_0(self, iF, current_byte):
+        
+        if self.__parse_detector:
```
sjoerd-bouma, line 13: Variables starting with `__` are private, so this will not work. We can promote `__parse_detector` to `_parse_detector` (single underscore) in NuRadioRecoio to get around this?

NuRadioReco/modules/io/event_parser_factory.py
```diff
@@ -80,18 +89,23 @@ def scan_files_2_2(self, iF, current_byte):
             bytes_to_read = int.from_bytes(bytes_to_read_hex, 'little')
             self._bytes_start[iF].append(current_byte)
             self._bytes_length[iF].append(bytes_to_read)
-        elif object_type == 1:  # object is detector info
+        
+        elif object_type == 1 and self.__parse_detector:  # object is detector info
```
sjoerd-bouma, line 93: See previous comment

sjoerd-bouma (commented):
I don't think the dependence on `__parse_detector` in the event_parser_factory works, because that variable will not be accessible outside the class - see relevant comments.

Other than that, I _think_ this will work as previously but without overwriting the `__scan_files` function, which is a nice improvement. With regards to the question of what `get_detector` should do, my preference is would be to raise a warning rather than an error.

NuRadioReco/modules/io/NuRadioRecoio.py
```diff
@@ -178,33 +190,43 @@ def _parse_event_header(self, evt_header):
...
             self._get_file(iF).seek(current_byte)
-            continue_loop, iF, current_byte = self.__scan_files(self, iF, current_byte)
+            continue_loop, iF, current_byte = self.__scan_files_versioned(self, iF, current_byte)
+
+            if iF_prev != iF:
+                self.logger.info(f"Finished scanning file {iF_prev}.")
+
             if not continue_loop:
                 break
+        self.logger.info(f"Finished scanning file {iF}. Finished all")
```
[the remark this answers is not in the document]
fschlueter: done

NuRadioReco/modules/io/event_parser_factory.py
fschlueter: done

NuRadioReco/modules/io/event_parser_factory.py
fschlueter: done

fschlueter:
>With regards to the question of what `get_detector` should do, my preference is would be to raise a warning rather than an error.
Done

fschlueter:
@sjoerd-bouma

sjoerd-bouma (approved):
This should work as far as I can tell : )

PR #574, opened 2023-09-19
autopep8 auto formatting

fschlueter:
formatting

### DIFF
--- a/NuRadioReco/modules/io/coreas/readCoREASShower.py
+++ b/NuRadioReco/modules/io/coreas/readCoREASShower.py
@@ -117,17 +117,17 @@ class readCoREASShower:
                         station_id, corsika, observer, channel_ids=[0, 1, 2])
                 else:
                     sim_station = coreas.make_sim_station(
-                        station_id, corsika, observer, channel_ids=self.__det.get_channel_ids(self.__det.get_default_station_id()))
+                        station_id, corsika, observer, 
+                        channel_ids=self.__det.get_channel_ids(self.__det.get_default_station_id()))
+                
                 station.set_sim_station(sim_station)
                 evt.set_station(station)
+                
                 if self.__det is not None:
                     position = observer.attrs['position']
-                    antenna_position = np.zeros(3)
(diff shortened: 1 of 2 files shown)


NuRadioReco/modules/io/coreas/readCoREASShower.py
```diff
@@ -103,20 +108,26 @@ def run(self):
...
                 station.set_sim_station(sim_station)
                 evt.set_station(station)
                 if self.__det is not None:
                     position = observer.attrs['position']
                     antenna_position = np.zeros(3)
-                    antenna_position[0], antenna_position[1], antenna_position[2] = -position[1] * units.cm, position[0] * units.cm, position[2] * units.cm
-                    antenna_position = cs.transform_from_magnetic_to_geographic(antenna_position)
+                    antenna_position[0], antenna_position[1], antenna_position[2] = - \
+                        position[1] * units.cm, position[0] * \
+                        units.cm, position[2] * units.cm
```
sjoerd-bouma, line 128: [No way this is PEP8](https://peps.python.org/pep-0008/#maximum-line-length)! I think the PEP8 way would be to either turn this into a tuple
```
... = (
    -position[1] * units.cm,
    position[0] * units.cm,
    position[2] * units.cm
)
```
or, alternatively, just turn this into three separate lines. `\` is ugly!

sjoerd-bouma (changes requested):

NuRadioReco/modules/io/coreas/readCoREASShower.py
fschlueter: better?

sjoerd-bouma: Much <3

sjoerd-bouma (approved):
Thanks Felix!

NuRadioReco/modules/io/coreas/readCoREASShower.py
fschlueter: I wonder why my formatter did not complain. I agree that was not good style at all.

PR #575, opened 2023-09-20
Wrap the loading of the ARA libraries with a try except block. This w…

fschlueter:
…ill fix the deployment of the docu

This is at least my attempt to fix the documentation (because the failing actions are annoying). Not sure if that is a practical solution, opinions? 

PS: Not sure who from ARA to tag.

### DIFF
--- a/NuRadioReco/modules/io/araroot/readARAData.py
+++ b/NuRadioReco/modules/io/araroot/readARAData.py
@@ -14,14 +14,15 @@ import ctypes
 
 
 sys.path.append(os.path.expandvars('${ARA_UTIL_INSTALL_DIR}/lib'))
-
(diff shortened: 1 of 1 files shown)


sjoerd-bouma:
The documentation test doesn't fail because of the failed ARA import (it has a fixed list of errors it will fail on, which currently does not include ImportErrors), but because of an issue with proposal 7.5.1 (#569 ) which is fixed in the latest version of proposal (fixed by #570, which hasn't been merged yet). This also causes the unit tests to fail.

We can still merge this regardless, if you like, it will make the output of the documentation test a bit tidier.

sjoerd-bouma (approved):

PR #579, opened 2023-09-25
Update highLowThreshold.py

martin01021207:
Trace digitization is added to the high-low-threshold trigger.

### DIFF
--- a/NuRadioReco/modules/trigger/highLowThreshold.py
+++ b/NuRadioReco/modules/trigger/highLowThreshold.py
@@ -2,6 +2,7 @@ from NuRadioReco.modules.base.module import register_run
 from NuRadioReco.utilities import units
 from NuRadioReco.framework.parameters import stationParameters as stnp
 from NuRadioReco.framework.trigger import HighLowTrigger
+from NuRadioReco.modules.analogToDigitalConverter import analogToDigitalConverter
 import numpy as np
 import time
 import logging
@@ -11,7 +12,7 @@ logger = logging.getLogger('HighLowTriggerSimulator')
 def get_high_low_triggers(trace, high_threshold, low_threshold,
                           time_coincidence=5 * units.ns, dt=1 * units.ns):
     """
-    calculats a high low trigger in a time coincidence window
+    calculates a high low trigger in a time coincidence window
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -106,7 +107,12 @@ def run(self, evt, station, det,
             number_concidences=2,
             triggered_channels=None,
             trigger_name="default_high_low",
-            set_not_triggered=False):
+            set_not_triggered=False,
+            Vrms=None,
+            trigger_adc=False,  # by default, assumes the trigger ADC is the same as the channels ADC
```
colemanalan, line 112: Since this is a trigger, we would want this to be `True`.

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -140,6 +146,10 @@ def run(self, evt, station, det,
         """
```
colemanalan, line 146: You need to add docstrings here for your new parameters

colemanalan (approved):

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -136,10 +142,27 @@ def run(self, evt, station, det,
...
+            that start with `'trigger_'`
+        clock_offset: float
+
+        adc_output: string
+            Options:
+            * 'voltage' to store the ADC output as discretised voltage trace
+            * 'counts' to store the ADC output in ADC counts
+
+        use_digitization: bool
+            If True, traces will be digitized
```
colemanalan, line 158: Sorry to be picky, but I think just to be more helpful to users we should reorder the options and set `use_digitization` first and then add a "(Only active if `use_digitization` is set to `True`)" for the other options.

colemanalan (changes requested):

NuRadioReco/modules/trigger/highLowThreshold.py
martin01021207: I've just reordered them and added a comment right next to the `use_digitization` argument.

colemanalan (approved):

PR #580, opened 2023-09-27
Does not allow to specifty the sampling rate to unrealisticly high va…

fschlueter:
…lues > 100 Gs. However, this often happens when one missinterpret the unit

I use the function sometimes outside of NuRadioReco. I keep forgetting that one have to give the sampling rate in units of Gsamples. Hence this small check.

### DIFF
--- a/NuRadioReco/modules/channelGenericNoiseAdder.py
+++ b/NuRadioReco/modules/channelGenericNoiseAdder.py
@@ -106,6 +106,11 @@ class channelGenericNoiseAdder:
         *   Add 'multi_white' noise option on 20-Sept-2018 (RL)
 
         """
+        if sampling_rate > 100:
+            raise ValueError(f"The sampling rate can not be set to values larger than 100 (is set to {sampling_rate}). "
+                             "The sampling rate is interpreted as giga samples per second (since the base unit for "
+                             "time in NuRadioReco is nano second).")
+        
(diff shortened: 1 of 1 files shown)


sjoerd-bouma:
I'm not so sure about this - in principle, it's the user's responsibility to provide the correct units, and we already provide the [units module](https://nu-radio.github.io/NuRadioMC/NuRadioReco/pages/utilities.html#unit-system) that should be used for all dimensional quantities to facilitate this. I.e. one can (should) always pass `sampling_rate=100*units.MHz` or similar.

So while a sanity check here doesn't hurt, there are probably hundreds of similar cases where the user runs into the same problem if they forget to correctly specify the units. We can still merge this if you like, but that's my take on it.

fschlueter:
True, I have to admit for some reason it does not file naturally for me to use Hertz as unit and we do not have something like giga sample as unit. Although I totally admit that is probably my own quirk. But would it harm?

cg-laser:
I agree with Sjoerd here. All code should be written with the unit package. We should be able to change our base units at any time without breaking the code. Therefore the check is not implemented correctly as a baseunit of GHz is implicitly assumed. If we want to test, then it should be `> 100 * units.GHz`. 
I don't think we should implement this check at all.

fschlueter:
@cg-laser you are right with explicitly multiplying the unit. However, I will close that PR since you both share the same feeling.

PR #585, opened 2023-10-09
Feature/arbitrary noise spectra

sjoerd-bouma:
New feature - rather than simulating only flat noise spectra with the channelGenericNoiseAdder, I added the option to specify the desired frequency spectrum as a function of frequency. Note that this (somewhat by design) is not compatible with specifying a desired Vrms. 

Just sharing in case someone else thinks this might be useful - happy to refine the implementation if desired.

### DIFF
--- a/NuRadioReco/modules/channelGenericNoiseAdder.py
+++ b/NuRadioReco/modules/channelGenericNoiseAdder.py
@@ -84,8 +84,11 @@ class channelGenericNoiseAdder:
             number of samples in the time domain
         sampling_rate: float
             desired sampling rate of data
-        amplitude: float
-            desired voltage of noise as V_rms (only roughly, since bandpass limited)
+        amplitude: float | function
+            desired voltage of noise as V_rms (only roughly, since bandpass limited).
+            Can also be given as a function; in this case, it should take the frequency as an argument
+            and return the desired amplitude for each frequency bin. This allows to simulate arbitrary
+            (non-flat) noise spectra.
         type: string
             perfect_white: flat frequency spectrum
             rayleigh: Amplitude of each frequency bin is drawn from a Rayleigh distribution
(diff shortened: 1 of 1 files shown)


cg-laser:
I think this is useful for standalone applications. However, it should still be able to define the desired Vrms. We do that all the time. One just needs to calculate the normalization, which is the integral over the square of the filter function within the fmin, fmax interval. 
I suggest to change the interface to add an additional optional parameter which defines the frequency filter. Either directly an array with an amplitude for each frequency bin, or as a function. I would prefer the former as it allows for more flexibility, e.g., one can directly load in the measured S21 parameters of a signal chain.

NuRadioReco/modules/channelGenericNoiseAdder.py
```diff
@@ -130,18 +134,24 @@ def bandlimited_noise(self, min_freq, max_freq, n_samples, sampling_rate, amplit
...
+            ampl[selection] = amplitude / (bandwidth / (sampling_bandwidth)) ** 0.5  # normalize noise level to the bandwidth its generated for
+        else:
+            ampl[selection] = amplitude
 
-        ampl = np.zeros(n_samples_freq)
-        sigscale = (1. * n_samples) / np.sqrt(nbinsactive)
+        if callable(amplitude):
+            sigscale = sampling_rate
+        else:    
+            sigscale = (1. * n_samples) / np.sqrt(nbinsactive)
```
MartinRavn, line 149: These four lines can maybe be included in the if statements above (line 137 to 144) as this is just another scaling. It should maybe also be applied there.

MartinRavn (commented):
I agree with Christan, that it would be nice to have the custom filter as an additional parameter that can both be an array of amplitudes or a function that returns the amplitudes. Then maybe the `amplitude` can be taken into account.

Otherwise I only have two minor comments.

sjoerd-bouma:
I think this was implemented by someone else as [channelGenericNoiseAdder.bandlimited_noise_from_spectrum](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.modules.channelGenericNoiseAdder.html#NuRadioReco.modules.channelGenericNoiseAdder.channelGenericNoiseAdder.bandlimited_noise_from_spectrum), so I'm closing this PR.

PR #587, opened 2023-10-27
RNO-G 2023

christophwelling:
This adds an updated detector JSON for RNO-G:
- GPS coordinates are based on the 2023 survey campaign.
- Cable delays account for the extra jumper cable and splitter in the PA signal chain


sjoerd-bouma:
Thanks Christoph, this is great! Just a couple of comments:
- Station 23 is missing channel 12;
- For some (especially stations 11, 12 and 21?) the LPDA orientations look a bit odd, as in not aligned with their trenches - but maybe this is something for future calibrations to figure out correctly?
- Would we want to also include solar panel and wind turbine positions in these or should we/I get those from the RNO-G repo somehow?

christophwelling:
Hi,
I added channel 12 for station 23 (it was just a stupid typo in the script to generate the detector).
About the LPDA orientations: In 2021 we measured the orientation for some trenches, but I forgot to correct for magnetic declination. That is fixed now (stations 11 and 21). For the others I just used the nominal orientations they should have. None of these are probably very accurate, but it is the best information we have right now.
I could add solar panels and wind turbines under "devices" but I don't know if that is really a good place for them. I could also just add it to the Desy databse once it is ready.

sjoerd-bouma (dismissed):
Thanks Christoph! I imagine also swapping channels 6 and 7 for station 23 was intentional?

christophwelling:
No it wasn't. Channels 6 and 7 are switched on station 23 and I always forget to change them. Thanks for catching!

PR #588, opened 2023-11-01
Cleanup coreas reader

lpyras:
This is a small PR improving the docstrings of the coreas readers (including when to use which reader) and adds a detector description for the rno-g surface array with upward facing antennas.


sjoerd-bouma (approved):
Looks fine to me, though I'm not so sure I understand what the RNO_cr_array.json is.

lpyras:
Thanks, it's just the upward LPDAs for the cosmic ray simulations

PR #591, opened 2023-11-15
Feature/galactic noise adder upgrade channel dts

sjoerd-bouma:
Previously, the galactic noise adder generated random galactic noise for each channel. This PR expands the selection of available galactic sky models (thanks @karenterveer) and uses consistent galactic noise between all channels (thanks @MoritzReichert), taking into account the phase differences due to the different antenna positions.

TODO:

- [ ] fix imports of optional sky models; probably wrap them in `try:... except ImportError:...` clauses
- [ ] check that the phase adjustment is correct, or at least a good approximation for antennas at different depths
- [ ] update docstrings
- [ ] is it correct to have random polarizations for each channel still?
- [ ] Replace print statements with logger

### DIFF
--- a/NuRadioReco/modules/channelGalacticNoiseAdder.py
+++ b/NuRadioReco/modules/channelGalacticNoiseAdder.py
@@ -1,13 +1,22 @@
 from NuRadioReco.utilities import units, fft, ice, geometryUtilities
 from NuRadioReco.modules.base.module import register_run
 import NuRadioReco.framework.channel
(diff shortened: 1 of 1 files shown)


sjoerd-bouma:
This PR has been superseded by #740 (sorry @MoritzReichert that your commit credit has disappeared!), closing this PR for now. Some open questions remain but they can be addressed in future issues/PRs.

Issue #593, opened 2023-11-22
Calculation of bandwidth in simulation.py does not work for RNO-G detector


fschlueter:
https://github.com/nu-radio/NuRadioMC/blob/15044001786cdb1856651987f536657364b076fd/NuRadioMC/simulation/simulation.py#L314-L317

The code in question
```
                self._amplification_per_channel[self._station_id][channel_id] = np.abs(filt).max()
                bandwidth = np.trapz(np.abs(filt) ** 2, ff)
                self._bandwidth_per_channel[self._station_id][channel_id] = bandwidth
                logger.status(f"bandwidth of station {self._station_id} channel {channel_id} is {bandwidth/units.MHz:.1f}MHz")
```
should calculate the bandwidth of the simulated detector (to calculate the amplitude of the thermal noise). However, as far as I understand the code, it only works when `np.abs(filt) ** 2` is 1 within the bandwidth and 0 outside. However, for the RNO-G detector (i.e., what comes our of NuRadioReco/detector/RNO_G/analog_components.py) this is not true.

fschlueter:
Solution could look like that:
```
                self.__amplification_per_channel[station_id][channel_id] = np.abs(filt).max()
                limit = np.abs(filt).max() / 100
                freqs_in_bandwidth = ff[np.abs(filt) > limit]
                bandwidth = freqs_in_bandwidth[-1] - freqs_in_bandwidth[0]
                self.__bandwidth_per_channel[station_id][channel_id] = bandwidth
                print(f"bandwidth of station channel is {bandwidth / units.MHz:.1f}MHz")
```

fschlueter:
When I am running the example `T02RunSimulation.py` with the example detector description in the same folder and plot the filter (i.e., `np.abs(file)`) I get the plot below (which explains why it worked before)

![image](https://github.com/nu-radio/NuRadioMC/assets/30903175/193e4373-1e8f-4fcc-a889-d2b5508e6e54)

sjoerd-bouma:
I think the existing code is actually the correct thing to do? At least in terms of the formula
`Vrms = \sqrt(k bw R T)` this expression should correctly convert from temperature to Vrms, as this also scales with the amplification. I agree the term bandwidth may be somewhat confusing though.

cg-laser:
[...]

PR #594, opened 2023-11-22
Renaming bandwidth variable in simulation

fschlueter:
Addresses #593

### DIFF
--- a/documentation/source/NuRadioMC/pages/HDF5_structure.rst
+++ b/documentation/source/NuRadioMC/pages/HDF5_structure.rst
@@ -45,7 +45,7 @@ The HDF5 files can be thought of as a structured dictionary:
 
 - The top level :ref:`attributes <NuRadioMC/pages/HDF5_structure:HDF5 file attributes>`, which can be accessed through ``f.attrs``, contain some top-level information about the simulation.
 - The :ref:`individual keys <NuRadioMC/pages/HDF5_structure:HDF5 file contents>` contain some properties (energy, vertex, ...) for each stored event or shower.
-- Finally, the ``station_<station_id>`` key contains slightly more detailed information (triggers, propagation times, amplitudes...) at the level of individual channels :ref:`for each station <NuRadioMC/pages/HDF5_structure:Station data>`.
+- Finally, the ``station_<station_id>`` key contains slightly more detailed information (triggers, propagation times, amplitudes...) at the level of individual channels :ref:`for each station <NuRadioMC/pages/HDF5_structure:Station data>`. Each station group has its own attributes (``f[station_<station_id>].attrs``)
 
 HDF5 file attributes
 ____________________
@@ -72,13 +72,25 @@ The top-level attributes can be accessed using ``f.attrs``. These contain:
             ``start_event_id`` | ``event_id`` of the first event in the file
             ``trigger_names`` | List of the names of the different triggers simulated
             ``Tnoise`` | (explicit) noise temperature used in simulation
-            ``Vrms`` |  RMS of the voltage used as thermal noise floor. Determine from ``Tnoise`` and ``bandwidth``
-            ``bandwidth`` | Bandwidth of the antennas/detector (for triggering)
             ``n_samples`` | Samples of the to-be generated antenna signals
             ``config`` | The (yaml-style) config file used for the simulation
             ``deposited`` |
             ``detector`` | The (json-format) detector description used for the simulation
             ``dt`` | The time resolution, i.e. the inverse of the sampling rate used for the simulation. This is not necessarily the same as the sampling rate of the simulated channels!
+
+
+The station-level attributes can be accessed using ``f[station_<station_id>].attrs``. The first two attributes ``Vrms`` and ``bandwidth`` also exist on the top-level and refer to the corresponding to the first station/channel pair.
+
+    .. _hdf5-station-attrs-table:
+
+    .. csv-table:: HDF5 station attributes
+            :header: "Key", "Description"
+            :widths: auto
+            :delim: |
+
+            ``Vrms`` | RMS of the voltage used as thermal noise floor :math:`v_{n} = (k_{B} \, R \, T \, \Delta f) ^ {0.5}`. See the relevant section "Noise voltage and power" in this `wiki article <https://en.wikipedia.org/wiki/Johnson%E2%80%93Nyquist_noise>`_ (last two equations). Determine from ``Tnoise`` and ``bandwidth`` (see below).
+            ``bandwidth`` | Bandwidth is above equation. Calculated as the integral over the simulated filter response (`filt`) squared: :math:`\Delta f = np.trapz(np.abs(filt) ** 2, ff)`.
+            ``antenna_positions`` | Relative position of all simulated antennas (channels)
 
 HDF5 file contents
 __________________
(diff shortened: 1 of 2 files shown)


cg-laser:
the calculation is correct. It is just the variable name that could be improved. See discussion in #593

fschlueter:
@cg-laser, I am not sure that I fully understand this. The calculation of the v_rms is (from Christoph branch):
```
                    self.__v_rms_per_channel[station_id][channel_id] = (noise_temp_channel * 50 * scipy.constants.k *
                           self.__bandwidth_per_channel[station_id][channel_id] / units.Hz) ** 0.5  # from elog:1566 and https://en.wikipedia.org/wiki/Johnson%E2%80%93Nyquist_noise (last Eq. in "noise voltage and power" section
```
Looking at the wikipage, I think the appropriate formular is:
image
What happened to the factor "4"? I assume the factor "50" in the equation above refers to 50 Ohm? I still do not understand why integrating over the absolute amplifier response is the correct thing. What I can imagine is that one would need to amplify the noise power calculated as proposed in this PR with the amplifier gain and that this is equivalent with integrating over the amplifier response?

fschlueter:
I think to remember that there was such a discussion one on the NuRadio slack. It's a shame that we do not have the history there.

fschlueter:
this logging is certainly very confusing `logger.status(f"bandwidth of station {self._station_id} channel {channel_id} is {bandwidth/units.MHz:.1f}MHz") `

fschlueter:
> With your approach, frequencies that are attenuated by the filter by a factor 100 contribute the same amount as frequencies that are not attenuated at all,
Hm, I agree that with "my approach" you would need to multiply again with the amplifier gain. But when you average this over the exact same bandwidth, does that really result in some other result?

fschlueter:
@cg-laser @sjoerd-bouma. I reverted the calculation and just renamed the variable. However, now things are a bit inconsistent: We still have the variable `self._bandwidth` and we store everything still with this name. Also should the square  only enter the calculation in lines ~348?

cg-laser (commented):
I don't think I prefer the new name. `filter_response` sounds more like the S21 parameter as a function of frequency. 
Wouldn't the better option be to keep the name "bandwidth" but add comments to the code that specifies "bandwidth refers to the detector response (filter, amplifiers, etc.) integrated over frequency"
and I would also improve the logger output to also mention that.

fschlueter:
So it's the integrated filter/detector response. I can change `filter_response` to `integrated_filter_response` but `bandwidth` is more confusing I think. I can improve the logger output. What about the keywords in the hdf5 files?

fschlueter:
> and I would also improve the logger output to also mention that.

Did you notice that I added a normalisation to the logger output such that it gives you a reasonable value for the bandwidth:
```
logger.debug(f"Estimated bandwidth of station {self._station_id} channel {channel_id} is "
                             f"{integrated_channel_response / mean_integrated_response / units.MHz:.1f} MHz")

```
With that I get something like this `STATUS:NuRadioMC:Estimated bandwidth of station 101 channel 0 is 784.6 MHz` (however I switched `status` to `debug`)

fschlueter:
@cg-laser is that fine? Any other suggestions? I also updated the documentation.

cg-laser (commented):
please don't switch to debug output. This is a useful output that let's the user check if the channels and noise temperatures and RMS values were calculated correctly. 

I don't like the normalization because it will give wrong answers for filters. 
Can you rename `mean_integrated_response` to `mean_amplification` 
And I would also set `mean_amplification` to 1 if the maximum amplification was 1. This allows to correctly detect if there is any amplification at all in the system. 

Why don't output 
"integrated channel response = X MHz, average amplification = X, maximum amplification = X, approximate bandwidth = X"

Or even better, if the `mean_amplification` is one, output what was outputted before. Only if we have amplification, output what I wrote above. 

For the documentation, I would rather not change the name of the field due to backward compatibility, but could you update the documentation saying that this is the integrated response which is equal to the bandwidth in case of no amplification

documentation/source/NuRadioMC/pages/HDF5_structure.rst
```diff
@@ -72,14 +72,26 @@ The top-level attributes can be accessed using ``f.attrs``. These contain:
...
+The station-level attributes can be accessed using ``f[station_<station_id>].attrs``. The first two attributes ``Vrms`` and ``bandwidth`` also exist on the top-level and refer to the corresponding to the first station/channel pair.
+
+    .. _hdf5-station-attrs-table:
+
+    .. csv-table:: HDF5 station attributes
+            :header: "Key", "Description"
+            :widths: auto
+            :delim: |
+
+            ``Vrms`` | RMS of the voltage used as thermal noise floor :math:`v_{n} = (k_{B} \, R \, T \, \Delta f) ^ {0.5}`. See the relevant section "Noise voltage and power" in this `wiki article <https://en.wikipedia.org/wiki/Johnson%E2%80%93Nyquist_noise>`_ (last two equations). Determine from ``Tnoise`` and ``bandwidth`` (see below).
```
cg-laser, line 91: I missed that you already updated the documentation, sorry about that. Looks good.

fschlueter:
> please don't switch to debug output. This is a useful output that let's the user check if the channels and noise temperatures and RMS values were calculated correctly.

Later the value is reported for one channel. Dont you think that is sufficient?

cg-laser:
no, in many applications, different channels have different bandwidths

fschlueter:
> no, in many applications, different channels have different bandwidths

What is the than of logging and storing the bandwidth of the first channel?

fschlueter:
@cg-laser I improved the logging. Now, it should also be more obvious what the settings are in the config, i.e., whether the `Vrms` is calculated per channel based on a noise temperature and response or fixed to a certain value. 

For example this is the output for a couple of RNO-G channels:
```
STATUS:NuRadioMC:Station.channel 11.16: noise temperature = 300.0 K, est. bandwidth = 742.33 MHz, max. filter amplification = 8.10e+02 -> Vrms = 7.74 mV -> efield Vrms = 9.55muV/m (VEL = 1m)
STATUS:NuRadioMC:Station.channel 11.17: noise temperature = 300.0 K, est. bandwidth = 742.33 MHz, max. filter amplification = 8.10e+02 -> Vrms = 7.74 mV -> efield Vrms = 9.55muV/m (VEL = 1m)
STATUS:NuRadioMC:Station.channel 11.18: noise temperature = 300.0 K, est. bandwidth = 742.33 MHz, max. filter amplification = 8.10e+02 -> Vrms = 7.74 mV -> efield Vrms = 9.55muV/m (VEL = 1m)
STATUS:NuRadioMC:Station.channel 11.19: noise temperature = 300.0 K, est. bandwidth = 742.33 MHz, max. filter amplification = 8.10e+02 -> Vrms = 7.74 mV -> efield Vrms = 9.55muV/m (VEL = 1m)
STATUS:NuRadioMC:Station.channel 11.20: noise temperature = 300.0 K, est. bandwidth = 742.33 MHz, max. filter amplification = 8.10e+02 -> Vrms = 7.74 mV -> efield Vrms = 9.55muV/m (VEL = 1m)
STATUS:NuRadioMC:Station.channel 11.21: noise temperature = 300.0 K, est. bandwidth = 722.57 MHz, max. filter amplification = 1.77e+03 -> Vrms = 16.64 mV -> efield Vrms = 9.41muV/m (VEL = 1m)
STATUS:NuRadioMC:Station.channel 11.22: noise temperature = 300.0 K, est. bandwidth = 722.57 MHz, max. filter amplification = 1.77e+03 -> Vrms = 16.64 mV -> efield Vrms = 9.41muV/m (VEL = 1m)
STATUS:NuRadioMC:Station.channel 11.23: noise temperature = 300.0 K, est. bandwidth = 722.57 MHz, max. filter amplification = 1.77e+03 -> Vrms = 16.64 mV -> efield Vrms = 9.41muV/m (VEL = 1m)
STATUS:NuRadioMC:All signals with less then 2.0 x Vrms_efield will be skipped.
```

fschlueter:
For my understanding the `trigger` settings in the config are not affecting the actual trigger simulations but only the speedup cuts? Is the name historically grown? Should we change that to avoid confusion?

cg-laser:
Thanks Felix. I added the int. response to the output. Looks good to me now. 

About the `trigger` group in the config file. The noise temperature and Vrms influence the trigger threshold. So I think the naming still makes sense. And we shouldn't change it due to backward compatibility. You can always add comments to the default config file to clarify things.

fschlueter:
@cg-laser a approval is needed here

cg-laser (approved):

PR #586, opened 2023-10-24
Simulation refactor

christophwelling:
This is the refactor of the simulation.py script:
- Simulations are split up between classes that handle the simulation at station, event and shower level
- Writing the output is handled by its own class
- Passing information around using member variables is gone. Only exceptions are the methods `set_event_group` and `set_event`.
I'm going to write some documentation about how this is supposed to work once the documentation builds properly again.

PR #595, opened 2023-11-24
Release v2.2.0

### LINKED ISSUE #582: Time adjuster creates invalid waveforms
When the wavefom is assigned in `triggerTimeAdjuster`:
https://github.com/nu-radio/NuRadioMC/blob/7b5b259c2b06684b32f49c5610a6021e767cdadb/NuRadioReco/modules/triggerTimeAdjuster.py#L96

there is an explicit check when `set_trace` is performed to ensure that waveforms cannot have an odd length. However, this module can generate waveforms with an odd length when cutting when the pulse is too close to the "right-most" edge or more generally when a pre-trigger time is requested that is larger than the waveform. This causes the whole program to crash.

Example:
If this function is called, for a desired waveform length of 1024 bins at 2.4 GS/s but is currently sampled at 5 GS/s, then this module will try to make waveforms that are: 1024 * 5 / 2.4 = 2133 bins long
https://github.com/nu-radio/NuRadioMC/blob/7b5b259c2b06684b32f49c5610a6021e767cdadb/NuRadioReco/modules/triggerTimeAdjuster.py#L63

sjoerd-bouma:
There have been a large number of bug fixes and improvements added to NuRadioMC since the last release, so it's time for the next one. We discussed yesterday which pull requests we would still like to merge before this release; this PR can serve as a place to track and discuss this.

**Include in 2.2.0**
- [x] #557 (the eventbrowser does not work on the current develop)
~~- [ ] #556 (add initial LOFAR reader to enable LOFAR students to get started)~~ the meaning of group_id will be changed. We should not make a release with that as this will be confusing and a non-backward compatible change. 
- [x] #526 (?) (fixes #582, but also contains changes to the way we store trace start times)
- [x] #558 (?) (unlikely to be fully finished soon, but maybe sufficient for some projects that use this feature to get started)
- [x] #587 

**For 2.3.0?**
- #586 unless the review & writing documentation go a lot quicker than expected
- #525


MijnheerD:
@sjoerd-bouma or @cg-laser , could you have a look at the LOFAR branch and approve it today? I would like to start fixing some things and working on the code with our master students, but I think it would be better to begin working with smaller branches rather than than keep adding to this large one.

sjoerd-bouma:
I guess before we release this, we should update the version number in pyproject.toml and in the changelog to `2.2.0` - see #605. Once that is merged, I think we're good to go?

PR #596, opened 2023-11-27
Update simulate_rnog_2021.py

martin01021207:
Updated the cylinder depth table. Saw that there were cutoffs in the shower vertex distributions of events at higher energy levels.

### DIFF
--- a/NuRadioReco/examples/RNO_data/station_simulation/simulate_rnog_2021.py
+++ b/NuRadioReco/examples/RNO_data/station_simulation/simulate_rnog_2021.py
@@ -75,7 +75,7 @@ def task(q, iSim, energy_min, energy_max, detectordescription, config, output_fi
         elif E <= 10**19.0 * units.eV:
             return 5.6 * units.km
         else:
-            return 6.1 * units.km
+            return 7.0 * units.km
 
     def get_min_z_shallow(E):  # estimated min z-pos from noiseless 1 sigma triggers (for PA trigger), E = shower energy
         if E <= 10**16.25 * units.eV:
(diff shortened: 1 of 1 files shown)


sjoerd-bouma:
Hi Martin, 
The ice sheet at Summit is only 3 km deep (https://arxiv.org/abs/2201.07846), so I wouldn't make the cylinder depth larger than that. If you happen to have some plots of the cutoffs you mention, you can attach them to the PR as well.

Also, it would be good if you could update to the latest `develop` branch (you can merge it into your branch using `git pull origin develop`) so the tests work as well.

martin01021207:
Hi, Sjoerd @sjoerd-bouma,

Thanks for the reply and the information you provided. I've generated events for different energy levels and when it's greater than 10^19 eV, the depth can go deeper than 3 km in the shower vertex distribution, so that's why I wanted to expand the depth table in script, but now I know there's the 3 km limit.
 
This is the shower vertex distribution for 10^20 eV, it is deeper than 3 km:
[vertex_e_cc_20.00eV.pdf](https://github.com/nu-radio/NuRadioMC/files/13492297/vertex_e_cc_20.00eV.pdf)

sjoerd-bouma:
> @sjoerd-bouma, I was just wondering. We might see muons/taus escaping the bedrock? Okay those muons/tau would need a lot of energy which makes their survival prob very low. What is with radio emission escaping from the rock from quite horizont slightly downgoing neutrinos?

I'm not sure if we're covering those corner cases correctly, but I imagine this is the sort of scenario where one would want to use PROPOSAL to properly simulate the lepton propagation and interactions, though I don't know how the bedrock is treated in that case. 
For a particle shower starting at the edge of the bedrock but continuing to develop in the ice, I'm sure we do not simulate this at the moment, but this would be an even more obscure case (both geometrically and, judging from Martin's plots, requiring an energy of at least 1e20 eV) and obviously extremely difficult to get right - certainly beyond the Askaryan models part of NuRadioMC.

PR #597, opened 2023-11-29
Hotfix/none trace

### LINKED ISSUE #369: re-writing a .nur file with missing sim traces creates broken .nur file
Reading a `.nur` file with 'missing' `sim_channel_traces` (e.g., because they were not saved in the simulation) using the `eventReader` module, and then writing the read event using the `eventWriter` module results in an error when reading the resulting `.nur` file:

```
(...)
~/Python/software/nu-radio/NuRadioMC/NuRadioReco/framework/electric_field.py in deserialize(self, data_pkl)
    141         data = pickle.loads(data_pkl)
    142         if(data['base_trace'] is not None):
--> 143             NuRadioReco.framework.base_trace.BaseTrace.deserialize(self, data['base_trace'])
    144         if 'position' in data:  # for backward compatibility
    145             self._position = data['position']

~/Python/software/nu-radio/NuRadioMC/NuRadioReco/framework/base_trace.py in deserialize(self, data_pkl)
    193     def deserialize(self, data_pkl):
    194         data = pickle.loads(data_pkl)
--> 195         self.set_trace(data['time_trace'], data['sampling_rate'])
    196         if('trace_start_time' in data.keys()):
    197             self.set_trace_start_time(data['trace_start_time'])

~/Python/software/nu-radio/NuRadioMC/NuRadioReco/framework/base_trace.py in set_trace(self, trace, sampling_rate)
     78         """
     79         if trace is not None:
---> 80             if trace.shape[trace.ndim - 1] % 2 != 0:
     81                 raise ValueError('Attempted to set trace with an uneven number ({}) of samples. Only traces with an even number of samples are allowed.'.format(trace.shape[trace.ndim - 1]))
     82         self.__time_domain_up_to_date = True

IndexError: tuple index out of range
```
The trace is something like `np.array(None)`, i.e. returns `True` on `trace is not None`, but has shape `()`, leading to the above error.
A workaround is to replace the check `if trace is not None` by `if np.any(trace != None)` (suggested by @shallmann), though I'm not sure if it's a fix or a workaround for some bug in `eventWriter` that we're missing.

sjoerd-bouma:
Fixes #369 

Thanks to Tingwei Liu, I was finally able to reproduce this bug. It turns out that if an Event contains empty traces (e.g. electric field objects with no trace), but they are saved (which is the default setting in the eventWriter), the trace stored is `np.copy(None)` rather than `None`, which then raises an error during deserialization.

This fix instead does not store the time trace if there isn't one.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,7 +4,7 @@ please update the categories "new features" and "bugfixes" before a pull request
 
 version 2.2.0-dev
 new features:
-- expand values stored in SimplePhasedTrigger 
+- expand values stored in SimplePhasedTrigger
 - added getting to query ARZ charge-excess profiles
 - upgrade to proposal v 7.5.0 with new NuRadioProposal interface and improved LPM treatment
 - add script to generate / download pre-calculated proposal tables for known detector configurations
@@ -19,6 +19,7 @@ bugfixes:
 - fixed/improved C++ raytracer not finding solutions for some near-horizontal or near-shadowzone vertices
 - fixed wrong number in Feldman-Cousins upper limit
 - in antennapattern.py, fixed line 1398; was masking frequencies >= 0 instead of frequencies > 0, causing NaN errors
+- Fixed issue where saving empty traces (channels or Efields) created unreadable .nur files
 
 
 version 2.1.8


cg-laser (approved):
thanks for fixing this! This is very subtle. 
I was coincidentally running into a similar problem when improving the ray tracing code yesterday. @sjoerd-bouma Is checking for `if not array.shape` the best method for checking for numpy arrays of type None?

sjoerd-bouma:
> thanks for fixing this! This is very subtle. I was coincidentally running into a similar problem when improving the ray tracing code yesterday. @sjoerd-bouma Is checking for `if not array.shape` the best method for checking for numpy arrays of type None?

I think so - `np.copy()` always returns an array, so `.shape` should always work (return a tuple), and a tuple will evaluate to `True` unless it's `()`. So I think it returns `False` if and only if the object is `array(None)`.

PR #609, opened 2024-01-23
Update doc: not refering to a fixed python version

fschlueter:


sjoerd-bouma:
We need to explicitly mention the minimum supported Python version in order to make pip releases; this is done in the pyproject.toml. I would vote to keep the version explicit in the documentation as well.

If we do want to deprecate python 3.6, you can update it in both places, but while I don't advocate fixing things to keep it supported I'm not actually aware of anything that broke? (Adding CI tests for multiple Python versions is still on my to-do list...)

cg-laser:
I agree, we should only update the minimum supported python version if we use new features that require a newer version. On clusters one has often the problem of fairly old python versions. 

On the other hand, it might be good to test both on the lowest supported version and the most recent one to make sure we keep compatible with the newest python version. Thanks @sjoerd-bouma for looking into that!

fschlueter:
Fine with me. I do think there are already issues with python 3.6 not working anymore for NuRadio. @sjoerd-bouma are you doing this? Should we do this in this PR or should we close this?

sjoerd-bouma:
I think since we probably don't want to keep the changes, PR title or branch name, I'd suggest closing this. I can make a PR to implement tests for multiple python versions, should be relatively quick.

fschlueter:
Okay but please update the min. python version also

PR #611, opened 2024-01-25
Tools for the shallow CR analysis

jakobhenrichs:
With this PR I want to merge the new files that I created to perform the RNO-G shallow CR analysis. The first file (NuRadioReco/modules/RNO_G/crRNOGTemplateCreator.py) is RNO-G specific and can be used to create templates from a Gaussian electric field. The second file (NuRadioReco/utilities/variableWindowSizeCorrelation.py) can be used to calculate the correlation of two traces, where only a small window of the trace is used for the correlation (sliding window). The function is not RNO-G specific.


NuRadioReco/modules/RNO_G/crRNOGTemplateCreator.py
```diff
@@ -0,0 +1,248 @@
...
+                                antenna_rotation=[160, 160, 160], Efield_amplitudes=[-0.2, 0.8], cr_zenith=[55, 55, 55], cr_azimuth=[0, 0, 0],
+                                sampling_rate=3.2 * units.GHz, number_of_samples=2048):
+        """
+        set_parameter_templates method
+
+        sets the parameter to create the template set
+
+        Parameters
+        ----------
+        template_run_id: list of int
```
anelles, line 87: I think the documentation should comment on the default values. In particular the ones where it is != 0

NuRadioReco/modules/RNO_G/crRNOGTemplateCreator.py
```diff
@@ -0,0 +1,248 @@
...
+            event ids of the artificial templates
+        template_station_id:
+            station ids of the artificial templates
+        template_channel_id:
+            channel ids of the artificial templates
+        Efield_width: list of int
+            width (in samples) of the gaussian function used to create the Efield
+        antenna_rotation: list of int
+            rotation angle of the LPDA
+        Efield_ratio: list
```
anelles, line 99: Why does come with capital E?

NuRadioReco/modules/RNO_G/crRNOGTemplateCreator.py
```diff
@@ -0,0 +1,248 @@
...
+        self.__Efield_amplitudes = Efield_amplitudes
+        self.__antenna_rotation = antenna_rotation
+
+        self.__sampling_rate = sampling_rate
+        self.__template_sample_number = number_of_samples
+
+        self.__cr_zenith = cr_zenith
+        self.__cr_azimuth = cr_azimuth
+
+    def run(self, template_filename='templates_cr_station_101.p', include_hardware_response=True, return_templates=False):
```
anelles, line 126: Are you sure that .p is a useful extension?

NuRadioReco/modules/RNO_G/crRNOGTemplateCreator.py
```diff
@@ -0,0 +1,248 @@
...
+                logger_level: string or logging variable
+                    Set verbosity level for logger (default: logging.NOTSET)
+                """
+
+        self.__detector_file = detector_file
+
+        self.logger.setLevel(logger_level)
+
+        # set up the efield to voltage converter
+        self.__efieldToVoltageConverter.begin(debug=False)
```
anelles, line 71: Shouldn't this use the same  debug level?

anelles: It might also be useful to actually have this module spit out a bit of logging text, i.e. where the file is save, if it is save etc. But that is mostly cosmetics.

changelog.txt
```diff
@@ -4,6 +4,8 @@ please update the categories "new features" and "bugfixes" before a pull request
 
 version 2.3.0-dev
 new features:
+- added module to create CR templates for shallow RNO-G channels
```
anelles, line 7: General question: Why can this only make templates for shallow? Shouldn't deep templates work in the same way? Shouldn't we make something more general?

anelles (changes requested):
It has been a while, but pre-summer cleaning 😇

anelles:
PR that are open forever, really are not a good idea. @jakobhenrichs do you want to merge this and then open a new one for future changes?

NuRadioReco/modules/RNO_G/crRNOGTemplateCreator.py
jakobhenrichs: I added the default values

NuRadioReco/modules/RNO_G/crRNOGTemplateCreator.py
jakobhenrichs: changed it to a lower e

NuRadioReco/modules/RNO_G/crRNOGTemplateCreator.py
jakobhenrichs: I added some logging text

changelog.txt
jakobhenrichs: I don't see why it should not work for the deep channels. However, I never tested it. I removed "shallow" from the changelog.

jakobhenrichs:
Sorry, I missed your comments from August. Yes, I would like to merge the PR. I tried to address all your comments. Please let me know now if there are more changes needed.

NuRadioReco/modules/RNO_G/crRNOGTemplateCreator.py
jakobhenrichs: I changed it to .pickle

anelles (approved):
Thanks for addressing this. Let's indeed merge this.

fschlueter:
Why does this PR adds a module to utilities? Looking quickly over it, the module does not have the standard interface. I would either implement the functionality as an actual function and/or add a prober module (with the standard interface) to run the correlation. I think that would be very useful also for other people!

PR #613, opened 2024-01-26
revert variable change to assure backward compatibility with Gen2 TDR…

cg-laser:
… simulation scripts

The recent change renaming the _bandwidth variables in simulation.py https://github.com/nu-radio/NuRadioMC/pull/594/files breaks the Gen2 TDR simulations https://github.com/nu-radio/analysis-scripts/blob/gen2-tdr-2021/gen2-tdr-2021/detsim/D01detector_sim.py#L144 This is quite annoying and super confusing for all users. In principle, it is a simple fix, renaming _bandwidth_per_channel to _integrated_channel_response, but only if you know exactly what to do, which only a few people know. 

This PR reverts the variable renaming and adds additional documentation to the code to define the meaning of the variable.

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -41,6 +41,7 @@ from NuRadioReco.framework.parameters import particleParameters as simp
 from NuRadioReco.framework.parameters import generatorAttributes as genattrs
 import datetime
 import logging
+import warnings
 from six import iteritems
 import yaml
 import os
@@ -263,11 +264,12 @@ class simulation:
             logger.status(f"input file {self._inputfilename} is empty")
             return
 
-        ################################
-        # perfom a dummy detector simulation to determine how the signals are filtered
-        self._bandwidth_per_channel = {}  # this variable stores the integrated channel response for each channel, i.e.
-        # the integral of the squared signal chain response over all frequencies, int S21^2 df. For a system without
-        # amplification, it is equivalent to the bandwidth of the system. 
+        # Perfom a dummy detector simulation to determine how the signals are filtered.
+        # This variable stores the integrated channel response for each channel, i.e.
+        # the integral of the squared signal chain response over all frequencies, int S21^2 df.
+        # For a system without amplification, it is equivalent to the bandwidth of the system.
+        self._integrated_channel_response = {}
+
         self._integrated_channel_response_normalization = {}
         self._max_amplification_per_channel = {}
         self.__noise_adder_normalization = {}
@@ -307,7 +309,7 @@ class simulation:
             # run detector simulation
             self._detector_simulation_filter_amp(self._evt, self._station, self._det)
 
-            self._bandwidth_per_channel[self._station_id] = {}
+            self._integrated_channel_response[self._station_id] = {}
             self._integrated_channel_response_normalization[self._station_id] = {}
             self._max_amplification_per_channel[self._station_id] = {}
 
@@ -325,14 +327,14 @@ class simulation:
                 self._integrated_channel_response_normalization[self._station_id][channel_id] = mean_integrated_response
 
                 integrated_channel_response = np.trapz(np.abs(filt) ** 2, ff)
-                self._bandwidth_per_channel[self._station_id][channel_id] = integrated_channel_response
+                self._integrated_channel_response[self._station_id][channel_id] = integrated_channel_response
 
                 logger.debug(f"Station.channel {self._station_id}.{channel_id} estimated bandwidth is "
                              f"{integrated_channel_response / mean_integrated_response / units.MHz:.1f} MHz")
 
         ################################
 
-        self._bandwidth = next(iter(next(iter(self._bandwidth_per_channel.values())).values()))  # get value of first station/channel key pair
+        self._bandwidth = next(iter(next(iter(self._integrated_channel_response.values())).values()))  # get value of first station/channel key pair
         norm = next(iter(next(iter(self._integrated_channel_response_normalization.values())).values()))
         amplification = next(iter(next(iter(self._max_amplification_per_channel.values())).values()))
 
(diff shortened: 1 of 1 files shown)


fschlueter:
I am fine reverting this. The comments, updated documentation and logging should help preventing confusion in the future. However, I am a bit confused by what our policy on this is. Can we never break user code? Are we responsible for code outside of the repository. (I agree thought that renaming a variable is not the best reason to break compatibility).

fschlueter (approved):

sjoerd-bouma:
Let me echo what I put on the slack - I think we should try to make sure that public variables/methods/classes do not get removed/changed without at least one major release including a deprecation warning, but anything starting with `_` should ideally not be something user code (has to) rely on. If this is not currently the case this is something we should maybe think about.

fschlueter:
I just read the discussion on slack. And I second Brian, Sjoerd and Christoph opinions/suggestions. An elegant solution might be to implement a `@property` for the deprecated variable:
```
@property
def _bandwidth_per_channel(self):
    self.logger.warning("Deprecated... Please change to integrated_channel_response")
    return self._integrated_channel_response
```
This should keep the code running because it effectively replaces `_bandwidth_per_channel` with `_integrated_channel_response` and alerts users to change the code. (And we should make this variable public or give it a public interface)

sjoerd-bouma:
> I just read the discussion on slack. And I second Brian, Sjoerd and Christoph opinions/suggestions. An elegant solution might be to implement a `@property` for the deprecated variable:
> 
> ```
> @property
> def _bandwidth_per_channel(self):
>     self.logger.warning("Deprecated... Please change to integrated_channel_response")
>     return self._integrated_channel_response
> ```
> 
> This should keep the code running because it effectively replaces `_bandwidth_per_channel` with `_integrated_channel_response` and alerts users to change the code. (And we should make this variable public or give it a public interface)

Ah, that's neat, I was not aware of the @property decorator. One more pedantic comment before I go to sleep - it would be nice to use the `DeprecationWarning` from the built-in [warnings](https://python.readthedocs.io/en/stable/library/warnings.html#warning-categories), so that we can eventually make the CI system raise errors for these.

fschlueter:
I think implementing it like this could work:
```
import warnings

class A():

    def __init__(self):

        self._integrated_channel_response = {"dummy": 2}

    @property
    def _bandwidth_per_channel(self):
        warnings.warn("This variable `_bandwidth_per_channel` is deprecated. "
                      "Please use `integrated_channel_response` instead.", DeprecationWarning)
        return self._integrated_channel_response

    @_bandwidth_per_channel.setter
    def _bandwidth_per_channel(self, value):
        warnings.warn("This variable `_bandwidth_per_channel` is deprecated. "
                        "Please use `integrated_channel_response` instead.", DeprecationWarning)
        self._integrated_channel_response = value

    @property
    def integrated_channel_response(self):
        return self._integrated_channel_response

    @integrated_channel_response.setter
    def integrated_channel_response(self, value):
        self._integrated_channel_response = value


a = A()

# This still works but logs a warning
print(a._bandwidth_per_channel)

# This still works but logs a warning
a._bandwidth_per_channel = {"dummy": 3}

# this also works
print(a._integrated_channel_response)
a.integrated_channel_response = {"dummy": 4}

# and this also works
print(a.integrated_channel_response)
```

cg-laser:
This is indeed a very neat way of fixing it! Thanks Felix for suggesting. @fschlueter Can you implement this as you suggested and also use `DeprecationWarning` class as Sjoerd suggested? 

Re general policy: NuRadioMC is not (and might never be) a software that can only be used via fixed interfaces. This would reduce its usability a lot. A lot of the things we are doing are still exploratory, where we want to be able to quickly test new ideas without first defining interfaces for a functionality we might never use again. So we need to find a good compromise between well defined interfaces and still allowing users to do advanced stuff. 
Things that prove useful should eventually get a proper interface, but this is sometimes not the case. Therefore, we should be extra careful not to break user code. 
I consider NuRadioMC stable in its core features. We implemented most aspects to the best of our current knowledge. The only main area of improvement will be better ice propagation, but progress here will be slow as we are limited by our understanding and/or the unknown properties of the ice. Therefore, e.g., the results we obtained for the extensive Gen2 studies will be valid for many years and the NuRadioMC steering scrips are the baseline for many new analyses. So it is very important that we keep them usable. 

So there isn't a clear answer to what should or should not be changed but we need to make pragmatic decisions and I would put the bar of changing an interface fairly high and only do it if we need it for new functionality, and not for rather cosmetic changes.

sjoerd-bouma (approved):

colemanalan (approved):

clark2668:
@fschlueter  i am less unhappy with `integrated_channel_response`. I just think it's useful if the word indicates the behavior of the object. If we wanted to be more detailed, perhaps "integral_of_channel_s21" or something like that?

fschlueter:
> @fschlueter i am less unhappy with `integrated_channel_response`. I just think it's useful if the word indicates the behavior of the object. If we wanted to be more detailed, perhaps "integral_of_channel_s21" or something like that?

`integrated_channel_gain` ?

clark2668:
I like that one even better. I guess the thing I'm getting hung up on is the word "integrated." Somehow it sounds like "integrated" from "integrated system" so in the RNO-G context it might mean like IGLU + DRAB + blah blah. Where in this sense you mean "We have done an integral." But probably I'm just word-smithing it to death, and you should move forward.

fschlueter:
> I like that one even better. I guess the thing I'm getting hung up on is the word "integrated." Somehow it sounds like "integrated" from "integrated system" so in the RNO-G context it might mean like IGLU + DRAB + blah blah. Where in this sense you mean "We have done an integral." Hence my preference for something like "integral of." But probably I'm just word-smithing it to death, and you should move forward.

Yes but also this variable describes the combined response/gain of all components applied to the signal (for RNO-G this is DRAB, IGLU, ...) @cg-laser please correct me if I am wrong here.

cg-laser:
I think we should not overthink this here. It is just an internal variable of a well-working code. Only if you want to go through the source code and want to understand it, you will notice it. And with the additional comment explaining what the variable contains, this should be crystal clear, independent of the variable's name.

clark2668:
I agree not to overthink. Just go with `integrated_channel_gain` imo.

fschlueter:
Damn, I should have changed the name of the PR  in the merge message ...

PR #614, opened 2024-01-31
Automatically check compatibility with different python versions

sjoerd-bouma:
Minor update to the unit tests:
- It's now possible to specify the python version and os to run the tests on;
- There is a new action that runs the unit tests on 'all' python versions (3.6 - latest 3.x). For now, I've made it such that this test runs only on push to develop (i.e. after a PR has been merged into develop) or manually, as running it on every pull request / push would probably cause our runner time to balloon very quickly. Happy to take suggestions if someone prefers an alternative strategy.

I've already used this to check our Python 3.6 compatibility [here](https://github.com/nu-radio/NuRadioMC/actions/runs/7728595096). It fails, but the errors are (as far as I can see) all from test output deviating from the reference files. Would be good if someone had time to investigate this...

### DIFF
--- a/.github/workflows/python_version_compatibility.yaml
+++ b/.github/workflows/python_version_compatibility.yaml
@@ -11,7 +11,11 @@ jobs:
   pythonversionsmatrix:
     strategy:
       matrix:
-        python_version: [3.6, 3.8, 3.9, 3.10, 3.11, 3.12, 3.x]
+        python_version: [3.8, 3.9, 3.10, 3.11, 3.12, 3.x]
+        os: [ubuntu-latest]
+        include:
+          - os: ubuntu-20.04
+            python_version: 3.6
     uses: ./.github/workflows/run_tests.yaml
     with:
       python_version: ${{ matrix.python_version }}
--- a/.github/workflows/python_version_compatibility.yaml
+++ b/.github/workflows/python_version_compatibility.yaml
@@ -1,0 +1,17 @@
+name: Check python version compatibility
+
+on:
+  push:
+    branches:
+      - develop
+  workflow_dispatch:
+
+
+jobs:
+  pythonversionsmatrix:
+    strategy:
+      matrix:
+        python_version: [3.6, 3.8, 3.9, 3.10, 3.11, 3.12, 3.x]
+    uses: ./.github/workflows/run_tests.yaml
+    with:
+      python_version: ${{ matrix.python_version }}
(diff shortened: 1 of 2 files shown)


fschlueter:
so by default the code runs on `3.7` unless otherwise specified? And the new action tests all others? Fine! I agree, we should understand why this fails for 3.6 ...

fschlueter (approved):

.github/workflows/python_version_compatibility.yaml
```diff
@@ -0,0 +1,21 @@
...
+    branches:
+      - develop
+  workflow_dispatch:
+
+
+jobs:
+  pythonversionsmatrix:
+    strategy:
+      matrix:
+        python_version: [3.8, 3.9, 3.10, 3.11, 3.12, 3.x]
```
cg-laser, line 14: don't you only select 3.8 upwards? Where is 3.6 selected? And what does 3.x do here? Is this selecting the latest version? 

I'm just trying to understand the syntax.

cg-laser:
wouldn't it be sufficient to test on the lowest and highest version? This will reduce the testing time a lot and would probably also catch everything? Or could there be something that worked in 3.7 and 3.12 but not in 3.9? 

Or alternatively, only run a subset of the tests on all python version? Although I'm not sure if this will then really test all relevant things.

.github/workflows/python_version_compatibility.yaml
sjoerd-bouma: See [here](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) for what 'matrix' does - essentially it takes all possible combinations of python_version and os, plus those given by `include` (python 3.6 is not available via github actions on ubuntu 22.04, so I've made it run on 20.04 instead).

sjoerd-bouma:
> wouldn't it be sufficient to test on the lowest and highest version? This will reduce the testing time a lot and would probably also catch everything? Or could there be something that worked in 3.7 and 3.12 but not in 3.9?
> 
> Or alternatively, only run a subset of the tests on all python version? Although I'm not sure if this will then really test all relevant things.

Yes, I suppose we could change it to check only 3.6 and 3.x (in addition to the 'default' 3.7). My thinking was that as long as we only run on pushes to develop this won't trigger very frequently, so it doesn't hurt to test all versions, but I can change this.

cg-laser:
@sjoerd-bouma I think we should just give it a try with whatever setting you think is best. Please go ahead and merge it.

PR #617, opened 2024-02-05
Add helper function to repsonse.py

fschlueter:


cg-laser (approved):
looks good to me. The only comment I have that we have a similar function to time shift traces : delay_trace in trace_utilities.py. 

Howeer, the interface is different enough that they can remain as two separate functions.

fschlueter:
@cg-laser thanks for pointing that function out. I was not aware of it (I could have made a better effort in looking for it). I just noticed that the signs to subtract/add the time delay are different in both function. It sounds correct to me as both functions sever a slightly different propose (e.g. subtracting and adding a time delay). However, tracking down bugs like that can be extremely hard so I thought I bring that to your attention (if we did not notice already) such that at least 2 heads thought about it.

fschlueter:
@cg-laser any comment? Otherwise I will go on with the merge

cg-laser:
yes, please go ahead. I agree that in this case it is better to have two different functions.

PR #618, opened 2024-02-05
Refactor bandpass_filter.py

fschlueter:
Only cosmetics

### DIFF
--- a/NuRadioReco/utilities/bandpass_filter.py
+++ b/NuRadioReco/utilities/bandpass_filter.py
@@ -35,6 +35,9 @@ def get_filter_response(frequencies, passband, filter_type, order, rp=None):
 
     """
 
+    if filter_type == 'rectangular':
+        return np.where(passband[0] <= frequencies <= passband[1], 1, 0)
+
     # we need to specify if we have a lowpass filter
     # otherwise scipy>=1.8.0 raises an error
     if passband[0] == 0:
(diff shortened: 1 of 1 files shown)


fschlueter:
The tests fail with
```
ValueError: operands could not be broadcast together with shapes (10000,) (9999,) () 
```
To reproduce the only behaviour I think we can do `np.where(mask[:-1], h, 0`. Is that correct?

sjoerd-bouma:
I don't think so - you want to return something the same shape as `frequencies`. Your suggestion would return something that's the shape of `h`, which excludes the DC component so is one smaller than `frequencies.shape`.

Feel free to keep it the way it was, I don't think `np.where` necessarily makes things that much easier to parse... though that's just my opinion.

fschlueter:
okay fair

fschlueter:
fine like this?

sjoerd-bouma (dismissed):
:+1:

cg-laser:
I don't really see an improvement here. And as the tests revealed, it was not only cosmetics, and it is often difficult to see (especially when reviewing) if the functionality is really exactly the same. In this case we were lucky that the tests caught it, but there are many parts of the code that are not tested. Therefore, I want to say again that we should be very careful when doing seemingly only cosmetic changes. Especially if if is often subjective if the new implementation is better. Here I don't see any improvement. The variables `filt` and `mask` are defined once and not in each if block individually which decreases the lines of code slightly but not performance. Depending on what each filter implements, this change could also make if more difficult in the future to implement new things. 
Therefore, I think we should stop doing subjective cosmetic changes, especially when we can't be 100% sure that we ensured the exact same behavior. 
I think the threshold should be that the code gets (significantly) more readable (not only personal preference), to better follow our coding standards (pep8 etc) or to improve performance or functionality.

fschlueter:
So of course it is hard for me to argue against it here after I need two commits to fix the first one. And I am fine to revert most of what I charged because I agree that the improvement here is questionable. However, I would say this is why we have unit testing and when you say that the coverage of those is insufficient we should make an effort to change that. And I think I am allowed to say that lines like this
```
        f = np.ones_like(frequencies)
        f[np.where(frequencies < passband[0])] = 0.
        f[np.where(frequencies > passband[1])] = 0.
        return f
  ```
  are actually harder to understand. Thinking about it I am not actually sure I understand them right now. What is `np.where` returning when you only pass a condition? It is not returning a mask (i.e. booking array) is it. 
So I agree that in this case for the rest I probably "shot over the goal line" (probably does not make any sense in Englisch but I guess you understand :)), but as a general theme I do think it is legit if we try to improve code the readability of the code (I agree though it should be beyond only personal preference, however, before think about that to long I rather do it and I am happy if it gets rejected)

fschlueter:
So I reverted most changes. Is that acceptable?

cg-laser:
The problem is that we are designing a software for often unknown usecases. We come up with new use cases every day and NuRadio should be flexible enough to allow studying them. I think this means as a consequence, that we will never be able to fully test all potential use cases, or that it would at least take a long time and resources we don't have. 

I'm not saying that we should not improve the codebase. In principle, I am for "refactor early and refactor often" and it is good if we work on new features. But at the same time we need guarantee that we remain backward compatible and more important, don't introduce new bugs. 

I'm not disagreeing with you that some parts of the code can be improved for readability and speed, but the person who implemented it will have very likely tested the code extensively before it was merged into develop. So even if it is not the best code, the likelihood that it is bugfree is very high. In the end, this is what counts in physics. And given that many developers are early grad students where programming skills are still developing, we will always have non-optimal code, I'm afraid. 
It is probably better to continue this discussion in one of our next calls.

cg-laser (approved):

PR #619, opened 2024-02-05
Numba ray tracing

elariekerboull:
The python 2D ray-tracing code is currently out-performed in terms of computation time by a C++ wrapper.
This C++ wrapper however fails to find certain solutions where the python version does.

This PR creates a new helper class that includes modified methods from the `ray_tracing_2D` class that are used in the root-finding process so they can be compiled using [Numba](https://numba.pydata.org) and used if Numba is available in the user's environment. If Numba is not instealled, the C++ or Python versions can still be used instead.
mean_values_plot_all
mean_values_plot_cpp_numba

The comparion of the different runtimes seems to indicate that the Numba version has a runtime of the same order as the C++ version. This was obtained by running the following sample script and varying some parameters: [long running tests](https://github.com/elariekerboull/NuRadioMC/blob/first_time/NuRadioMC/SignalProp/examples/long_prop_test.py)

Checking whether the solutions that C++ could not find were found with Numba was verified [here](https://github.com/elariekerboull/NuRadioMC/blob/first_time/NuRadioMC/SignalProp/examples/prop_error_test.py)


elariekerboull:
Please don't hesitate to ask for clarifications on some choices if needed.

sjoerd-bouma:
I don't have time to look at this in detail right now, but I've enabled the tests, and they don't seem to complete without errors yet - can you take a look?
Other than that, this looks really nice, thanks @elariekerboull! I am a little bit apprehensive about building this on a feature in `numba` that is called 'experimental' (jitclass), but I don't know to what extent we should be worried about that. Two more thoughts: 
1) as you mention, compiling the code the first time takes a while - I know for functions the compiled code can be cached, if this is the same for `jitclass`, it would be nice if you could enable this here too
2) it would be good to add a test to the unit tests that tests all three codes (C++, Python and numba)

cg-laser (changes requested):
Nice work! This will be very useful!

Your testing script does not yet test all relevant functions, as you only calculate the path between two points. The other time critical computation is the attenuation. To do that add something like
```
frequencies = np.linspace(0, 1000, 200) * units.MHz
for iS in range(2):
      attenuation = rays.get_attenuation(iS, frequencies, max_detector_freq=500*units.MHz)
```
to your test script. 
You will probably find that you will also need to optimize the `get_attenuation_along_path` function. 

Here a few thoughts about the implementation: 

It seems that you replicated/copy-pased a few functions to the new helper class. Can we avoid that? It would be much better if you could avoid code duplication. The current implementation is error prone as soon as we want to change anything as we would need to consistently change it in both locations. 

Do you need a class? Or would it be easier to just implement the relevant functions as non-member functions? Then you could even leave them in the original file and you would only need to add a @numba decorator. Maybe then you also don't need the experimental "jitclass" feature? This is just an idea, you probably have better ideas. 
If there is no way around having them in a separate class, maybe the default python implementation could just use these functions(?).

Can you allow turning numba on and off via an argument to the ray_tracing class? Either add a `use_numba` argument (similar to the `use_cpp` argument).

sjoerd-bouma:
Closing this as merged via #660

PR #624, opened 2024-02-10
hotfix for python compatibility workflow

sjoerd-bouma:
- added missing os variable to workflow (so Python 3.6 is also tested)
- fixed python version specifications (3.10 was interpreted as 3.1 -> "3.10"); removed 3.12 for now as 3.x (latest) = 3.12
- added `fail-fast: false` - check now runs on all specified python versions instead of cancelling as soon as any one of them fails.

I should probably have tested this more thoroughly before #614, but hopefully it now works as intended.

### DIFF
--- a/.github/workflows/python_version_compatibility.yaml
+++ b/.github/workflows/python_version_compatibility.yaml
@@ -1,4 +1,4 @@
-name: Check python version compatibility
+name: Check compatibility
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


cg-laser (approved):

fschlueter:
So to me it looks like that we are actually incompatible with python3.6 since we merged the new RNO-G detector class. This is the error meg:
```  
File "/home/runner/work/NuRadioMC/NuRadioMC/NuRadioReco/detector/RNO_G/rnog_detector.py", line 268, in <listcomp>
    [datetime.datetime.fromisoformat(v) for v in value["modification_timestamps"]]}
AttributeError: type object 'datetime.datetime' has no attribute 'fromisoformat'
Error: Process completed with exit code 1.
```
The following [thread](https://stackoverflow.com/questions/60266554/type-object-datetime-datetime-has-no-attribute-fromisoformat) explains it. I think it would be possible to store and read the modification time stamps in a different format, or we drop the compatibility?

fschlueter:
Regarding the other failing tests. Do we understand that? Can we fix a random seed somewhere to fix that in the tests?

PR #625, opened 2024-02-10
Documentation/add event structure image source

sjoerd-bouma:
Adds the source file for the image with the NuRadioReco event structure, so it can be updated more easily.

@cg-laser I've set this to merge directly into `emitter`, feel free to merge it yourself if you think this is helpful.


sjoerd-bouma:
I've just noticed this is branched off `develop` rather than `emitter` so it will also merge the latest `develop` into `emitter`, sorry. The long changelog should disappear if you merge `develop` first, but the only new change in here is in the very last commit.

sjoerd-bouma:
@cg-laser has already added the source file in #319, closing this.

PR #627, opened 2024-02-13
Fix example file after interface change. Improved logging and comment…

fschlueter:
…s. Fix tab vs ws.

This also reminds me that we have no unit test for the RNO-G reader. @sjoerd-bouma we should add one. How would we handle the rnog-data file? We could download it on the fly to not pollute the repo?

### DIFF
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -600,7 +600,7 @@ class readRNOGData:
             for selector in self._selectors:
                 if not selector(evtinfo):
                     self.logger.debug(f"Event {event_idx} (station {evtinfo.station}, run {evtinfo.run}, "
-                                      f"event number {evtinfo.eventNumber}) is skipped.")
+                                      f"event number {evtinfo.eventNumber}) did not pass a filter. Skip it ...")
                     self.__skipped += 1
                     return True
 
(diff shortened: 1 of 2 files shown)


fschlueter:
The unit test will be implemented with #632. This PR already fixes issues with the example script which is used in the unit test. @sjoerd-bouma, that should be a quick review!

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -897,14 +897,16 @@ def get_event(self, run_nr, event_id):
...
                 f"\n\tTime to read all events       : {self._time_run:.2f}s"
                 f"\n\tTime to per event             : {self._time_run / self.__counter:.2f}s"
                 f"\n\tRead {self.__n_runs} runs, skipped {self.__skipped_runs} runs.")
         else:
-            self.logger.info(
-                f"\n\tRead {self.__counter} events   (skipped {self.__skipped} events, {self.__invalid} invalid events)")
+            self.logger.warning(
+                f"\n\tRead {self.__counter} events   (skipped {self.__skipped} events, {self.__invalid} invalid events)"
+                f"\n\tTime to initialize data sets  : {self._time_begin:.2f}s"
+                f"\n\tTime to read all events       : {self._time_run:.2f}s")
```
sjoerd-bouma, line 909: This would probably be a good fit for the new 'status' logging level implemented in #602. But seeing as that's not merged yet, and this output only occurs on calling `end`, I think this is fine.

sjoerd-bouma (approved):

PR #632, opened 2024-02-20
Remove hack from readRNOGDataMattak. Add unit test

fschlueter:
PLEASE EDIT AND DELETE THIS TEXT WHEN MAKING A PR.
THIS IS A REMINDER OF WHAT THE PR SHOULD CONTAIN.

Issue being addressed:

Did you remember to: 
* include doc strings in all commits ?
* edit the changelog.txt file describing your changes?


fschlueter:
@sjoerd-bouma @Cosmin, the new test which uses mattak fails because (afaiu) of typings like this `starting_window : float | int) ` where a type can be either or `|`. Apparently that was just introduced with py3.10. What should we do here?

fschlueter:
See this action which runs through beccause it uses py3.10 https://github.com/nu-radio/NuRadioMC/actions/runs/7978652834/job/21784236727

fschlueter:
Okay, after discussing with Cosmin I found a solution. However, we might want to use this PR to discuss how we manage the version managment between mattak an nuRadio?

fschlueter:
Fine, lets keep it this way and see if it works.

fschlueter:
If we merge #627 first I might have to fix a little merge conflict. However, I think this PR is good to be reviewed already! @sjoerd-bouma

sjoerd-bouma (dismissed):
Thanks for tidying up and adding a unit test, Felix! 
The unit tests seem to run successfully with the current version of mattak (see [here](https://github.com/nu-radio/NuRadioMC/actions/runs/7978794214/job/21784685357)).

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -413,28 +413,9 @@ def begin(self,
                 if not _all_files_in_directory(dir_file):
                     self.logger.error(f"Incomplete directory: {dir_file}. Skip ...")
                     continue
-            else:
-                # Providing direct paths to a Mattak combined.root file is not supported by the mattak library yet.
-                # It only accepts directry paths in which it will look for a file called `combined.root` (or waveforms.root if
-                # it is not a combined file). To work around this: Create a tmp directory under `/tmp/`, link the file you want to
-                # read into this directory with the the name `combined.root`, use this path to read the run.
-
-                path = _create_random_directory_path()
```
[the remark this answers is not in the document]
fschlueter: Jep, good catch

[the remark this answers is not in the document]
fschlueter: done

fschlueter:
sorry @sjoerd we need your approval again

sjoerd-bouma:
I'm happy with the changes, but whereas the one failed test seems to (IIRC) fail sometimes at random, it has now failed twice in a row... I don't see how this is related to your changes, fundamentally, other than that maybe by changing the order of the tests you've changed some random numbers. This is not a huge issue - we could update the reference files - but maybe the easiest solution would be if you just move the tests back to the end of run_tests.yaml?

fschlueter:
> I'm happy with the changes, but whereas the one failed test seems to (IIRC) fail sometimes at random, it has now failed twice in a row... I don't see how this is related to your changes, fundamentally, other than that maybe by changing the order of the tests you've changed some random numbers. This is not a huge issue - we could update the reference files - but maybe the easiest solution would be if you just move the tests back to the end of run_tests.yaml?

Hm why would that just affect one test and not all? Non of those tests call a random generator, or am I missing something here.

fschlueter:
In general I thought it make sense to have short tests running at the beginning.

NuRadioReco/examples/RNO_data/read_data_example/read_rnog.py
```diff
@@ -58,7 +58,10 @@ class EventInfo:
     # Only use runs of a certain run type
     run_types=["physics"],
     # Only use runs with a maximum trigger rate of 1 Hz
-    max_trigger_rate=1 * units.Hz)
+    max_trigger_rate=1 * units.Hz,
+    # This might be necessary to read older files which have an invalid value stored as sampling rate
+    overwrite_sampling_rate=3.2
```
fschlueter, line 63: @sjoerd-bouma what do you say to this line? I actually do not like it at all. The overwrite is necessary for the test file. I can think of two solutions: Rewrite the interface of the example script which allows to provide a overwrite_sampling_rate via the command line (to be specified in the CI script). Or that we change the reader class and only allow the overwrite to be happening when the sampling rate from the data is actually invalid (or both). What do you think?

sjoerd-bouma:
Actually a lot of the tests rely on random number generation, but most of them are seeded to ensure they produce the same result every time. For some reason the Aeff one isn't https://github.com/nu-radio/NuRadioMC/blob/dcaaeeb275349569ca01dcd299ad4a16b613f9bc/NuRadioMC/test/atmospheric_Aeff/1e18eV/T01generate_event_list.py#L34
(the `seed` parameter is not defined)
That's probably an oversight, actually, and one that should be fixed.

NuRadioReco/examples/RNO_data/read_data_example/read_rnog.py
sjoerd-bouma: I had optimistically assumed that was already what the reader did (i.e. overwrite/set the sampling rate only if it's not defined) - if it isn't, I agree that is probably what it should do.

fschlueter: Well, I am wondering if there is a reason where we really want to overwrite the sampling rate but I dont know

fschlueter: see #641

fschlueter:
@sjoerd-bouma what is your conclusion? Can we merge this? I would assume we should rather fix the seed of the failing test or make it otherwise independent of the test order.

sjoerd-bouma:
I think I just fooled myself into thinking changing the order of the tests does anything, I guess if the Aeff test is not seeded and the random number generation works properly it should be random each time - maybe the only thing that's wrong is its idea of sigma (which is definitely wrong). Anyway, it seems to have run successfully now a couple of times again :shrug:

sjoerd-bouma (approved):

PR #633, opened 2024-02-20
hotfix: detectorbrowser

sjoerd-bouma:
...more of a hotfix, really. Wanted to look at a detector and noticed the detectorbrowser didn't work with the updated interface for `detector.Detector`. Still needs to be updated to account for the automatic detection of detector vs generic_detector descriptions, if anyone fancies doing this. But at least it should now be somewhat functional again.

### DIFF
--- a/NuRadioReco/detector/detector_browser/open_file.py
+++ b/NuRadioReco/detector/detector_browser/open_file.py
@@ -99,10 +99,11 @@ layout = html.Div([
                 html.Div([
                     dcc.Slider(
                         id='detector-time-slider',
-                        value=10000,
+                        value=0,
                         min=0,
-                        max=1551092200,
-                        step=1000
+                        max=1000,
+                        step=1000,
+                        marks = {}
                     )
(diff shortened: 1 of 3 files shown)


NuRadioReco/detector/detector_browser/open_file.py
```diff
@@ -342,7 +343,7 @@ def toggle_open_button_active(filename, detector_type, default_station, need_def
 
 @app.callback(
     Output('detector-time-div', 'style'),
-    [Input('load-detector-button', 'n_clicks'),
+    [Input('output-dummy', 'children'),
```
fschlueter, line 346: dont understand this change but okay.

sjoerd-bouma: The previous trigger led to a race condition - it triggered on the same trigger that loads the detector, so if it was quicker than the other callback, detector_provider.get_detector() would still return None. Now it triggers on the output of the callback that loads the detector.

PR #634, opened 2024-02-21
Added tolerance to parallel channels

rricesmith:
Modifications to function get_parallel_channels:
-Fixed bug where it attempted to save differently sized numpy arrays into a single array, causing an error
-Removed repeats of parallel or individual channels from being returned
-Added a tolerance parameter, which allows for there to be deviation in parallel channels and still be counted as parallel. Necessary for detector descriptions of real experiments whose antennas may deviate by a few degrees.

### DIFF
--- a/NuRadioReco/detector/detector_base.py
+++ b/NuRadioReco/detector/detector_base.py
@@ -607,7 +607,7 @@ class DetectorBase(object):
             channel_ids.append(channel['channel_id'])
         return sorted(channel_ids)
 
-    def get_parallel_channels(self, station_id, tol=1*units.deg):
+    def get_parallel_channels(self, station_id, tol=1*units.deg, anti_parallel=False):
             """
             get a list of parallel antennas
 
@@ -617,9 +617,16 @@ class DetectorBase(object):
                 the station id
             tol:        float
                 tolerance to difference in angles between parallel channels
+            atni-parallel: bool
+                if True, also returns anti-parallel channels
                 
             Returns list of list of ints
             """
+            def angle_diff(a, b, anti_parallel=False):
+                if anti_parallel:
+                    return np.abs(hp.get_normalized_angle(a - b, interval=np.deg2rad([0, 90])))
+                return np.abs(hp.get_normalized_angle(a - b, interval=np.deg2rad([0, 180])))
+
             res = self.__get_channels(station_id)
             orientations = np.zeros((len(res), 4))
             antenna_types = []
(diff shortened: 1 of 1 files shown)


cg-laser (changes requested):
it is an improvement but not yet doing what it should. 

E.g. if one angle is 0.1deg and the other 359.9, the difference is 359.8. Or in other words, the periodicity needs to be taken into account. 

The other thing is the restriction to 180deg interval for the rotation angle. Antennas can be anti-parallel which makes a difference (the trace is inverted, i.e. multiplied with -1). I would make it optional to return also anti-parallel channels or not.

rricesmith:
Updated to account for the periodicity of the angles. Also included an option for anti-parallel.

Example using a sample detector, with 0-3 standard downward LPDAs, 4-6 are standard upward with the 4th LPDA removed, and ch7 is a bicone.:
tol=1deg: [array([7]), array([0, 2]), array([1, 3]), array([4, 6]), array([5])]
tol=91deg: [array([7]), array([0, 2]), array([0, 1, 2, 3]), array([4, 6]), array([4, 5, 6])]

There are repeated channels still in some cases. This existed in the old version of this function, and I believe causes no issues with modules that use this. It might be preferred, as I believe this happens because we have 4 axis of rotation, and its possible that given a set of 3 orientations (a, b & c), that [a,b] can be parallel within tolerance, and [b, c] can be, but [a, c] cannot be.

NuRadioReco/detector/detector_base.py
```diff
@@ -607,42 +607,55 @@ def get_channel_ids(self, station_id):
...
+                tolerance to difference in angles between parallel channels
+            atni-parallel: bool
+                if True, also returns anti-parallel channels
+                
+            Returns list of list of ints
+            """
+            def angle_diff(a, b, anti_parallel=False):
+                if anti_parallel:
+                    return np.abs(hp.get_normalized_angle(a - b, interval=np.deg2rad([0, 90])))
+                return np.abs(hp.get_normalized_angle(a - b, interval=np.deg2rad([0, 180])))
```
cg-laser, line 628: I think the intervals need to be [-90, 90] and [-180,180].

fschlueter:
I created a branch `parallel-channels` with the changes introduced here + some modifications + the suggestion by Christian. I will close this PR. Feel free to reopen it or continue developing this feature with the new branch. (I had to make a new branch because this PR uses a fork).

PR #635, opened 2024-02-21
Copy a base_trace object before multiplying with a response

fschlueter:
After an operation like 
```
ef2 = fake_response * ef
```
`ef` and `ef2` are equal. I think this is not the behaviour we want. I added a `copy.deepcopy`. I am not sure whether this is overdoing it or if there is a better way. Feel free to jump in! I am also wondering if it is necessary to add a copy for the case when we multiply responses.... 

This is a little script illustrating the issue:
```
import numpy as np
from NuRadioReco.framework import electric_field
from NuRadioReco.utilities import analytic_pulse
from NuRadioReco.detector.response import subtract_time_delay_from_response, Response
import signal
signal.signal(signal.SIGINT, signal.SIG_DFL)

ef = electric_field.ElectricField(channel_ids=[0])
ef.set_frequency_spectrum(analytic_pulse.get_analytic_pulse_freq(1, -0.5, 1 * np.pi, 2048, 1, 2.4, bandpass=(0.1, 0.7)), sampling_rate=2.4)

ef.set_trace(np.roll(ef.get_trace(), 480), "same")

fig, ax = plt.subplots()
ax.grid()
ax.plot(ef.get_times(), ef.get_trace(), label="orig")

n = 5000
resp = np.ones(n) + 1j * np.zeros(n)
freq = np.linspace(0, 1.2, len(resp))
resp_shifted = subtract_time_delay_from_response(freq, resp, time_delay=-50)
fake_response = Response(freq, [np.abs(resp_shifted), np.angle(resp_shifted)], ["mag", "rad"])
ef2 = fake_response * ef
ax.plot(ef2.get_times(),  ef2.get_trace(), label=f"Shifted + {50}")

ax.plot(ef.get_times(), ef.get_trace(), label="orig")

ax.legend()
plt.show()
```


cg-laser (changes requested):
I agree, we want to copy the object. But wouldn't a shallow copy, i.e. `copy.copy` be sufficient?

fschlueter:
@cg-laser from the description [here](https://arc.net/l/quote/whoanayz), I conclude that a deep copy is necessary because a base_trace objects contains mutable objects (the numpy arrays) which might still reference the old one. Do you agree? 

> The difference between shallow and deep copying is only relevant for compound objects (objects that contain other objects, like lists or class instances):

> 
> A shallow copy constructs a new compound object and then (to the extent possible) inserts references into it to the objects found in the original.
> 
> A deep copy constructs a new compound object and then, recursively, inserts copies into it of the objects found in the original.
> 
> Two problems often exist with deep copy operations that don’t exist with shallow copy operations:
> 
> Recursive objects (compound objects that, directly or indirectly, contain a reference to themselves) may cause a recursive loop.
> 
> Because deep copy copies everything it may copy too much, such as data which is intended to be shared between copies

fschlueter:
I went ahead and added the copy statement also when we copy other objects of type `Response`.

cg-laser:
can you just try it out with your test script? I'm think `copy.copy()` should be sufficient and would be much faster then deep copy

fschlueter:
Hi @cg-laser, was sensibel to check it😉 indeed one can shallow copy base_trace safely. However, a response object needs deep copy because it stores everything in lists. And one have to copy self and not other when multiplying responses!

fschlueter:
Maybe we should add a unit test, to make sure that stays that way?

cg-laser:
thanks for digging deeper. I agree, a unit test for this behaviour would be very good as this could be a very subtle bug

fschlueter:
I added a issue (#642) for the unit tests. I do not have so much time atm. Can we merge this PR without the unit test?

cg-laser (approved):

PR #637, opened 2024-02-22
Feature/robust corsika reader

cg-laser:
this change will automatically skip broken input files. An error is reported but the program continues.


sjoerd-bouma (approved):
Looks good. As far as I understand the idea is still to eventually get rid of  some (or at least make it more straightforward to pick the right one) of our 4 (!) CoREAS readers, but I guess ideally this fix would be implemented for all of them until then?

PR #638, opened 2024-02-22
Add sanity check to subtract_time_delay_from_response

fschlueter:
Hi all, i noticed that the subtraction of a time delay can go wrong if the response from which the time delay is subracted is to corsely sampled, i.e., the frequency binning is insufficient such that `time_delay * Delta frequencies * 2 * pi > pi` which caused issues. 

I used this little program to convince myself of it.

```
import numpy as np
from NuRadioReco.framework import electric_field
from NuRadioReco.utilities import analytic_pulse
from NuRadioReco.detector.response import subtract_time_delay_from_response, Response
import signal
signal.signal(signal.SIGINT, signal.SIG_DFL)

ef = electric_field.ElectricField(channel_ids=[0])
ef.set_frequency_spectrum(
    analytic_pulse.get_analytic_pulse_freq(
        1, -0.5, 1 * np.pi, 2048, 1, 2.4, bandpass=(0.1, 0.7)), sampling_rate=2.4)

ef.set_trace(np.roll(ef.get_trace(), 480), "same")

fig, ax = plt.subplots()
ax.grid()

dt = []
df = []
for n in np.arange(80, 150, 1):
    resp = np.ones(n) + 1j * np.zeros(n)
    freq = np.linspace(0, 1.2, len(resp))
    df.append(np.diff(2  * 50 * freq)[0])

    resp_shifted = subtract_time_delay_from_response(freq, resp, time_delay=-50)
    fake_response = Response(freq, [np.abs(resp_shifted), np.angle(resp_shifted)], ["mag", "rad"])
    dt.append(fake_response._calculate_time_delay())

ax.plot(df, dt)
ax.axhline(50, color="k", lw=1, ls="--")
ax.legend()
plt.show()
```


sjoerd-bouma:
> If you are fine with that we should also implement it in the function in `trace_utilities`

Have you checked this is actually a problem for the function in `trace_utilities`? I'm fairly confident this is an issue that only arises because of your interpolation; it might therefore even be preferable to implement some clever fix in the interpolation (if that's possible).

fschlueter:
More or less yes - the subtraction comes before any interpolation and I think there is a physical reason for that. However, I have not the time to make a code example and give a better answer atm. For `trace_utilities` is should not be a serious issue because it shifts traces which are typically recorded with a much higher sampling rate.

sjoerd-bouma:
What am I missing?
```
import numpy as np

trace = np.zeros(2048)
trace[1024] = 1
plt.plot(trace, color='k', label='original')

prop_cycle = plt.rcParams['axes.prop_cycle']
colors = list(prop_cycle.by_key()['color'])

shift_in_ns = 500
for i, sampling_rate in enumerate([2.4, 1.2, 0.6, 0.3, 0.1]): # GHz
    delayed_trace = trace_utilities.delay_trace(trace, sampling_rate, shift_in_ns)

    plt.plot(delayed_trace, color=colors[i]) # shifted trace
    plt.axvline((1024 + shift_in_ns * sampling_rate) % 2048, ls=':', color=colors[i])

plt.ylim(0, 1.5)
plt.legend()
plt.xlabel('sample')
```

image

fschlueter:
@sjoerd-bouma, I am fine with that being only added to the response class. But there I think we should add it soon?

sjoerd-bouma:
This is too long ago so I don't remember. But the reason I was such a stickler about it is that it seems like you're implementing a workaround for a bug, rather than fixing the bug, as a simple Fourier shift as in `trace_utilities.delay_trace` is fine regardless of sampling rate.

So I guess my question is/was: is it clear why this is an issue in `response` but not in `delay_trace`? And is it clear that this is not due to an issue with the implementation, but really should be solved in the way that you have?

fschlueter:
> This is too long ago so I don't remember. But the reason I was such a stickler about it is that it seems like you're implementing a workaround for a bug, rather than fixing the bug, as a simple Fourier shift as in `trace_utilities.delay_trace` is fine regardless of sampling rate.
> 
> So I guess my question is/was: is it clear why this is an issue in `response` but not in `delay_trace`? And is it clear that this is not due to an issue with the implementation, but really should be solved in the way that you have?

I am not saying that the sampling rate is causing an issue when Fourier shifting. But the frequency resolution or trace length might be an issue. What I meant is that it is probably not a practical issue in `delay_trace` because it is typically used only for small shits in time. But in `response` we subtract phase shifts of several hundred nano seconds.

sjoerd-bouma:
> I am not saying that the sampling rate is causing an issue when Fourier shifting. But the frequency resolution or trace length might be an issue. What I meant is that it is probably not a practical issue in `delay_trace` because it is typically used only for small shits in time. But in `response` we subtract phase shifts of several hundred nano seconds.

But as my plot shows... a Fourier shift should give you the correct result _regardless_ of the frequency resolution or trace length. A 500 ns shift shows up as a 500 ns shift, regardless of the trace length or frequency resolution. So I still don't understand what the problem is!

fschlueter:
Great but this is not what I mean.

sjoerd-bouma:
You'll either have to explain it such that I understand or get someone else to approve, I'm afraid. The problem is that `Response._calculate_time_delay` doesn't work if the frequency binning is too sparse?

fschlueter:
> You'll either have to explain it such that I understand or get someone else to approve, I'm afraid. The problem is that `Response._calculate_time_delay` doesn't work if the frequency binning is too sparse?

Fair, maybe I will find time for it eventually (did not work well the last half year though).

sjoerd-bouma:
Okay, having spent some more time looking into this, I'm guessing what happens is that the phase wraps in between frequency bins. This is indeed a problem if you subsequently try to interpolate the response. While upsampling works, you could also just store the unwrapped phase (seeing as you internally store a phase and gain anyway), or store a separate float to account for large time delays (which if I understand correctly also happens anyway).

I probably like both of those approaches more than just upsampling (which inflates the size of the response object) but maybe there's something I'm missing still. Anyway, if you insist on this solution I think it's fine too. Just that the issue is not so much applying the time delay in the Fourier domain (which is done 'correctly' without upsampling) but retrieving or interpolating it afterwards. (This is also why this problem does not apply to the function in trace_utilities).

cozzyd:
Indeed, if you subtract (or add) a time delay that's too large from an S21, then eventually you end up with more than a cycle jump between phase in frequencies, which results in ambiguity (aliasing). 


I don't think upsampling helps in trace_utilities for the analogous problem (too few frequency points... upsampling just adds 0's at the end in frequency, not intermediate frequencies). The way to think about the problem for a time-domain trace is if you have a delay that's so long that you wrap around more than once, causing an aliasing problem. The solution is upsampling in the frequency domain (zero-padding in the time domain). 

A similar procedure can be done for the S21's, assuming they're not already aliased, by transforming to the time domain (easier said than done, if interpolation is required), and zero-padding, then transforming back.

sjoerd-bouma:
> Indeed, if you subtract (or add) a time delay that's too large from an S21, then eventually you end up with more than a cycle jump between phase in frequencies, which results in ambiguity (aliasing).
> 
> I don't think upsampling helps in trace_utilities for the analogous problem (too few frequency points... upsampling just adds 0's at the end in frequency, not intermediate frequencies). The way to think about the problem for a time-domain trace is if you have a delay that's so long that you wrap around more than once, causing an aliasing problem. The solution is upsampling in the frequency domain (zero-padding in the time domain).
> 
> A similar procedure can be done for the S21's, assuming they're not already aliased, by transforming to the time domain (easier said than done, if interpolation is required), and zero-padding, then transforming back.

I'm not sure if the time domain analogy applies? E.g. for Felix's example actually none of the cases wrap in the time domain. Also in general the width of the frequency bins does not determine the trace length (=maximum shift without wrapping) without also specifying the sampling rate.

(I suppose this point is academic - it's clear what the problem is, and that upsampling in frequency space resolves it. But I think it's important to point out there's no 'problem' with the function in trace_utilities (unless you're worried about unphysical wrapping of traces, but this is just a consequence of working in the Fourier domain and applies regardless of the time delay or frequency bin width))

cozzyd:
A waveform with a length T and N samples will have dt = T/N and therefore a max frequency of N/T and a frequency spacing of  (1 / 2T). You get a phase ambiguity when the phase jumps by more than half a cycle between frequency steps. The group delay is - d\phi/d\omega. In a finite difference, d\omega \approx  2\pi df  =pi/T, and you want a max d\phi of \pi to avoid ambiguity, implying a maximum unambiguous group delay of T ? (Maybe I'm off by a factor of 2 or something).  So I think the same delay that causes phase ambiguity also causes multiple wraparounds in the time domain?   If you set your time delay to 1000 ns in your script, you'll see what I mean!

cozzyd:
(but yes, if you know what roughly the total delay should be, and the dispersion is not greater than T, you can disambiguate, I think, sort of like working in higher Nyquist zones...).

cozzyd:
Commenting on @fschlueter implementation, you want to interpolate in re/im, not mag/phase. And linear interpolation is also not desirable (for the same reason it's not desirable in the time domain). You want to use a method similar to what you would use to interpolate in the time domain (fourier transforming and zero padding, or sinc interpolation, or zero-stuff then apply low-pass filter) for it to be "correct." (How much this matters, I'm not sure...obviously if the values are slowly varying enough, linear interpolation on them, perhaps even in mag/phase, will be fine). 

Note also that this assumes that you don't already have an aliased response, which you can't easily correct without knowing what the total group delay ought to be.

fschlueter:
> Commenting on @fschlueter implementation, you want to interpolate in re/im, not mag/phase. And linear interpolation is also not desirable (for the same reason it's not desirable in the time domain). You want to use a method similar to what you would use to interpolate in the time domain (fourier transforming and zero padding, or sinc interpolation, or zero-stuff then apply low-pass filter) for it to be "correct." (How much this matters, I'm not sure...obviously if the values are slowly varying enough, linear interpolation on them, perhaps even in mag/phase, will be fine).
> 
> Note also that this assumes that you don't already have an aliased response, which you can't easily correct without knowing what the total group delay ought to be.

I agree the interpolation can definitely improved. But this deserves its own PR. Regrding the anti-aliasing, do you think the upsampling is okay or what is the better method for this.

cozzyd:
Looking at the application more closely, if I'm understanding correctly, I'm not sure what the right thing to do in this case. It looks to me (please tell me I'm wrong!) that we're trying to subtract a time delay from an already aliased response, and so you would want to upsample with the right number of additional cycles in between points.  

Perhaps this means zero-padding in the time domain before rather than after, or something like that? But if there's a partial wraparound, that won't work correctly (this would be analogous to crossing between nyquist zones, I think). In other words, I'm not sure there's always an unambiguous solution if the number of wraparounds isn't the same for every step, though there may be some way to recover it based on knowing that the group delay can't jump too much from sample to sample.  Or perhaps some kind of decimation and upsampling can get you in a consistent number of wraparounds regime?

sjoerd-bouma:
@cozzyd thanks for the explanation - apart from a factor of 2 (frequency spacing is 1/T, not 1/2T, I think, and the maximum unambiguous shift is T/2, not T, which is why there was no wrapping in Felix' example but there already is aliasing) you were of course correct. It's probably a sign I spend too much time with Python that I had thought of the trace length as N rather than T...

Regarding the issue addressed in this PR - I thought we're not trying to antialias an already aliased `Response` (which can in principle be initialized with an arbitrary frequency spacing, so it depends on the input whether or not it's aliased, right?) but to avoid it becoming aliased after the addition (subtraction) of an additional large time delay. If that interpretation is correct, I still think my suggestion (storing the phase unwrapped, or storing an overall time delay as a float) should be sufficient, no?

cozzyd:
yes you are of course correct about the factor of 2! (And that makes it nice and symmetric between the two domains). It makes sense that it's T/2 rather than T because the shift can have both signs, and -T/2 aliases T/2.

sjoerd-bouma:
> Whether or not the response is aliased depends on if the measurement is aliased or not, I think? I guess I don't see why you would subtract a large delay if it was not already in there in the form of a cable delay, but I suppose things like that can happen in simulation, so maybe that's the justification. In that case, if the response is not aliased to start with, I think your suggestions work. Storing an overall time delay as a float seems like the simplest? I haven't reasoned through if storing unwrapped phase is sufficient or not. That information gets lost when transforming to the time domain, but perhaps it tells you how to properly later upsample.

Yes, maybe @fschlueter can comment as to whether the original response is already aliased or not. I've thought of one other reason to prefer a float over the unwrapped phase, which is that if long-term we want to use re/im interpolation instead of mag/phase, we may want to also move away from storing mag/phase internally also.

fschlueter:
Thanks guys to get so engaged here! However, I am not sure I can completely follow you. What I am proposing is to upsample the response in the frequency domain. Upsampling in the frequency domain is equivalent to zero-padding in the time domain which means also making the "trace" longer against possible cycling behaviour (I put trace in quotes because I have a hard time thinking about a response in the time domain). It does not mean changing the sampling rate of the response in the time domain. Hence, I am not sure I understand what you mean with aliasing (which for me is the effect that you try to measure signals while violating the nyquist limit). 

But to the situation. The S21 complex vectors are stored with the full phase in the DB. Along side the S21 a float in nanoseconds for the constant phase delay is stored. This time delay is computed in the webinterface but can be manually changed. 

Now when we read in the S21 vector into a `response` object we want to remove this constant phase shift from the complex vector which gets multiplied with the data. And for this subtraction I am upsampling the vector (prior the subtraction and prior the interpolation). I think that upsampling can be interpreted and increasing the trace length in the time domain to avoid cycling?

Did that help? Can such a problem (insufficient sampling of the frequency domain, i.e., insufficient frequency resolution - which is not controlled by the sampling rate) be also described under the term aliasing?

fschlueter:
@sjoerd-bouma and @cozzyd, do we want to quickly meet today - might be more efficient.

fschlueter:
@cozzyd

sjoerd-bouma (dismissed):
I think this should be uncontroversial - raise an error if there is an issue, and leave it up to the user to fix it :) happy to revisit if you come up with a different/better automatic solution in the future.

cozzyd:
sorry, I just did some spelling fixes on it!

sjoerd-bouma (approved):

Issue #642, opened 2024-02-27
Add unit test for detector.response.Response class


fschlueter:
See discussion in #635. It would be good to add a unit tests which checks whether the multiplication operate works as expected (copy logic)

PR #643, opened 2024-02-28
Add unit test

fschlueter:
This PR adds unit tests for the Response class. So far only the multiply operator functionality is tests ....

Based on PR #635. Solves #642


fschlueter:
@sjoerd-bouma

fschlueter:
ping

NuRadioReco/detector/test/tests.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
 set -e
 
-python3 NuRadioReco/detector/test/test_rnog_detector.py
\ No newline at end of file
+python3 NuRadioReco/detector/test/test_rnog_detector.py
+python3 NuRadioReco/detector/test/test_response.py
```
sjoerd-bouma, line 5: I think this won't work reliably - if the first python script raises an error, the second one will probably still run, and the github test will only consider the output of the second test. Feel free to put this into a new test, I think.

sjoerd-bouma (commented):
See comment, otherwise this is probably fine : )

NuRadioReco/detector/test/tests.sh
fschlueter: I see thanks! I fixed it in the workflow yaml. Would that work?

fschlueter:
ping

fschlueter:
@cg-laser maybe you can approve this?

sjoerd-bouma (approved):
Oops! Should have approved this a long time ago. Thank Felix!

PR #644, opened 2024-03-01
Create beam_pattern.py

sanyukta13:
Functions to correct for distance and azimuthal signal attenuation, uses .json file of the season to find relative distances and angles to correct for them

### DIFF
--- a/NuRadioReco/modules/RNO_G/beam_pattern.py
+++ b/NuRadioReco/modules/RNO_G/beam_pattern.py
@@ -1,0 +1,71 @@
+import matplotlib.pyplot as plt
+import numpy as np
+from NuRadioReco.utilities import units
+from NuRadioReco.modules import channelAddCableDelay, channelBandPassFilter, sphericalWaveFitter
+from NuRadioReco.detector import detector
+from NuRadioReco.modules.io.RNO_G import readRNOGDataMattak
+from datetime import datetime
+import NuRadioReco.detector.detector
+import astropy.time
+import logging
+from NuRadioReco.utilities import fft
+from os import listdir
+from os.path import isfile, join
+from scipy.signal import correlate
+from NuRadioReco.utilities.fft import time2freq
+from NuRadioReco.utilities.fft import freq2time
+from scipy.interpolate import interp1d
+import pandas as pd
+import json
+
+#define json file to take positions from
+distance_params_df = pd.DataFrame(list(json.load(open('/home/sanyukta/astropart/data/json/RNO_season_2023.json'))['channels'].values()))
+
+def get_channel_positions(station_number):
+    """
+    To find the position of each channel in a given year (using json file)
+    Input parameters: 
+    -station number (should be a valid station in the json file)
+    Returns:
+    - 3 arrays of x, y, z coordinates of each channel
+    """
+    x, y, z = {}, {}, {}
+    for ch in [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 21, 22, 23]:
+        x[ch] = distance_params_df.loc[(distance_params_df['station_id'] == station_number) & (distance_params_df['channel_id'] == ch), 'ant_position_x'].iloc[0]
+        y[ch] = distance_params_df.loc[(distance_params_df['station_id'] == station_number) & (distance_params_df['channel_id'] == ch), 'ant_position_y'].iloc[0]
+        z[ch] = distance_params_df.loc[(distance_params_df['station_id'] == station_number) & (distance_params_df['channel_id'] == ch), 'ant_position_z'].iloc[0]
+    return x, y, z
+
+def correct_for_position(x, y, z, ref, raw_volts):
+    """
+    Given the ref channel (pulsing channel), corrects for distance and azimuthal attenuation in the signal
+    Input Parameters:
+    - x, y, z: arrays of x, y and z coordinate of channels (obtained from get_channel_positions)
+    - ref: ref channel (cal pulser)
+    - raw volts: voltage traces of channels
+    Returns:
+    -corrected voltage trances for each channel
+    """
+    volts_corr = {}
+    for ch in [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 21, 22, 23]:
+        x_val = np.abs(x[ch]-x[ref])
+        y_val = np.abs(y[ch]-y[ref])
+        z_val = np.abs(z[ch]-z[ref])
+        r = np.sqrt(x_val*x_val + y_val*y_val + z_val*z_val + 1e-6)
+        sine = np.sqrt(x_val*x_val + y_val*y_val + 1e-6)/r
+        volts_corr[ch] = np.array(raw_volts[ch])*r/(sine*sine)
+    return volts_corr
+
+"""
+Example: 
+det = detector.Detector(json_filename = 'RNO_G/RNO_season_2023.json')
+det.update(datetime.now())
+station_number = 11
+fiber_number = 0
+run_number = 1785
+ref_ch = 21
+factor = 1
+events, times, volts_raw = read_raw_traces(station_number, fiber_number, run_number, det)
+x, y, z = get_channel_positions(station_number)
+volts_corr_for_pos = correct_for_position(x, y, z, ref_ch, volts_raw)
+"""


NuRadioReco/modules/RNO_G/beam_pattern.py
```diff
@@ -0,0 +1,71 @@
...
+from scipy.interpolate import interp1d
+import pandas as pd
+import json
+
+#define json file to take positions from
+distance_params_df = pd.DataFrame(list(json.load(open('/home/sanyukta/astropart/data/json/RNO_season_2023.json'))['channels'].values()))
+
+def get_channel_positions(station_number):
+    """
+    To find the position of each channel in a given year (using json file)
```
cozzyd, line 26: This suggests it might be year dependent but here the json is hard coded for 2023? 

Also are we sure this functionality (getting channel position) I'd not duplicated elsewhere?

NuRadioReco/modules/RNO_G/beam_pattern.py
```diff
@@ -0,0 +1,71 @@
...
+    Returns:
+    -corrected voltage trances for each channel
+    """
+    volts_corr = {}
+    for ch in [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 21, 22, 23]:
+        x_val = np.abs(x[ch]-x[ref])
+        y_val = np.abs(y[ch]-y[ref])
+        z_val = np.abs(z[ch]-z[ref])
+        r = np.sqrt(x_val*x_val + y_val*y_val + z_val*z_val + 1e-6)
+        sine = np.sqrt(x_val*x_val + y_val*y_val + 1e-6)/r
```
cozzyd, line 55: What does that is 1e-6 do?

NuRadioReco/modules/RNO_G/beam_pattern.py
```diff
@@ -0,0 +1,71 @@
...
+
+def correct_for_position(x, y, z, ref, raw_volts):
+    """
+    Given the ref channel (pulsing channel), corrects for distance and azimuthal attenuation in the signal
+    Input Parameters:
+    - x, y, z: arrays of x, y and z coordinate of channels (obtained from get_channel_positions)
+    - ref: ref channel (cal pulser)
+    - raw volts: voltage traces of channels
+    Returns:
+    -corrected voltage trances for each channel
```
cozzyd, line 47: Typo here

NuRadioReco/modules/RNO_G/beam_pattern.py
```diff
@@ -0,0 +1,71 @@
...
+def get_channel_positions(station_number):
+    """
+    To find the position of each channel in a given year (using json file)
+    Input parameters: 
+    -station number (should be a valid station in the json file)
+    Returns:
+    - 3 arrays of x, y, z coordinates of each channel
+    """
+    x, y, z = {}, {}, {}
+    for ch in [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 21, 22, 23]:
```
cozzyd, line 33: I would not necessarily expect this function to return only a subset of channels

NuRadioReco/modules/RNO_G/beam_pattern.py
```diff
@@ -0,0 +1,71 @@
...
+    Input Parameters:
+    - x, y, z: arrays of x, y and z coordinate of channels (obtained from get_channel_positions)
+    - ref: ref channel (cal pulser)
+    - raw volts: voltage traces of channels
+    Returns:
+    -corrected voltage trances for each channel
+    """
+    volts_corr = {}
+    for ch in [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 21, 22, 23]:
+        x_val = np.abs(x[ch]-x[ref])
```
cozzyd, line 51: I'm confused, the calpulser location does not correspond to any channel.

NuRadioReco/modules/RNO_G/beam_pattern.py
```diff
@@ -0,0 +1,71 @@
...
+        y_val = np.abs(y[ch]-y[ref])
+        z_val = np.abs(z[ch]-z[ref])
+        r = np.sqrt(x_val*x_val + y_val*y_val + z_val*z_val + 1e-6)
+        sine = np.sqrt(x_val*x_val + y_val*y_val + 1e-6)/r
+        volts_corr[ch] = np.array(raw_volts[ch])*r/(sine*sine)
```
cozzyd, line 56: So here  are two assumptions:

1) the beam pattern is sin(zenith). This is approximately fine but we have access to the real antenna beam patterns.

2) that the beam pattern is reciprocal. This will be true for straight line propagation and elevation symmetry, but not in general.

While these simplifying assumptions can still be useful, it should be clearly documented that they are happening, and in the future you can try to add raytracing  and real beam patterns to make this more generally applicable.

cozzyd (changes requested):
So thinking about what you're trying to do, I wonder if it makes more sense to have a function that corrects a single channel given a reference channel, based on positions (and in the future, beam pattern and ice model). This will allow you to do all channels (rather than a subset) and properly specify the calpulser position.

NuRadioReco/modules/RNO_G/beam_pattern.py
fschlueter: The detector class should be used to read the channel position

NuRadioReco/modules/RNO_G/beam_pattern.py
sanyukta13: Yes I did it only for the deep channels

NuRadioReco/modules/RNO_G/beam_pattern.py
sanyukta13: 1- Where can I find the real antenna beam patterns? I was not aware it was present
2- I am currently trying to work out the focusing factor corrections and will keep updating as we go on

NuRadioReco/modules/RNO_G/beam_pattern.py
sanyukta13: Would the helper string cal pulsers be a part of the channels? I don't think so and in that case how do I use the detector class to get the cal pulsers' position?

fschlueter: That is a good question. Probably as device. But I think they are not in atm. @jakobhenrichs  what do you think?

jakobhenrichs: Yes, the idea is to store the cal pulsers as devices (including its positions) in the database. Unfortunately, they are not added to the DB, yet.

fschlueter: Would it be hard to add those, I assume it would also need some tuning on the nuradio end

NuRadioReco/modules/RNO_G/beam_pattern.py
```diff
@@ -0,0 +1,71 @@
...
+import astropy.time
+import logging
+from NuRadioReco.utilities import fft
+from os import listdir
+from os.path import isfile, join
+from scipy.signal import correlate
+from NuRadioReco.utilities.fft import time2freq
+from NuRadioReco.utilities.fft import freq2time
+from scipy.interpolate import interp1d
+import pandas as pd
```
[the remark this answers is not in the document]
fschlueter: Would be possible to clean up the imports? It looks like there are a lot of un used imports here, and you can also order them in build-in, external and internal libraries.

NuRadioReco/modules/RNO_G/beam_pattern.py
fschlueter: Nit-picking comment, would be nice to add spaced around each math sign.

NuRadioReco/modules/RNO_G/beam_pattern.py
```diff
@@ -0,0 +1,71 @@
...
+"""
+Example: 
+det = detector.Detector(json_filename = 'RNO_G/RNO_season_2023.json')
+det.update(datetime.now())
+station_number = 11
+fiber_number = 0
+run_number = 1785
+ref_ch = 21
+factor = 1
+events, times, volts_raw = read_raw_traces(station_number, fiber_number, run_number, det)
```
fschlueter, line 68: `read_raw_traces` is not a method in NuRadioReco right?

fschlueter:
Right now this file looks more like a utility and not like module. If it should be a module you have to implement it as class with begin, run and finish method. If it is more a selection of utilities it might be got to move it to a folder `utilities/RNO_G/` ?

NuRadioReco/modules/RNO_G/beam_pattern.py
sjoerd-bouma: @sanyukta13 - someone pointed out to me that you said 'focusing factor', I happen to have worked on this recently - see #648 (if this is what you meant by focusing factor).

NuRadioReco/modules/RNO_G/beam_pattern.py
jakobhenrichs: In principle, everything should work on the DB side to add those. But I think it would be good if I double-check first.

NuRadioReco/modules/RNO_G/beam_pattern.py
sanyukta13: No, it's a function I defined for ease

Issue #648, opened 2024-03-05
Missing factor in focusing factor & errors in analytic path length / travel time equations


sjoerd-bouma:
While writing this up for my thesis (draft attached), I found

1. mistakes in the analytic formulas for the path length and travel time:
Original:
![image](https://github.com/nu-radio/NuRadioMC/assets/78698543/3c393543-12cb-4078-b4b4-b6fb98ceb8c1)
Corrected:
![image](https://github.com/nu-radio/NuRadioMC/assets/78698543/c099ddaf-459e-4f0c-8960-dae87075b07a)
These were checked against numerical integration:
[check_new_analytic_traveltimes.pdf](https://github.com/nu-radio/NuRadioMC/files/14493186/check_new_analytic_traveltimes.pdf)
[check_new_analytic_pathlength.pdf](https://github.com/nu-radio/NuRadioMC/files/14493184/check_new_analytic_pathlength.pdf)
and observed to agree within the 1e-4 numerical integration precision. Note that the previous formulas 
 - used a separate ad-hoc formula for the deep ice
 - regularly raised ArithmeticErrors and fell back on numerical integration
I see no reason why any issues should have arisen in these (clearly) finite integrals in the first place, but someone else can plug the equations into WolframAlpha again if they fancy.

sjoerd-bouma:
2. The focusing factor was missing a factor in the $\phi$ direction. 
Previously, it assumed the focusing only takes place in the $\theta$ direction. But focusing also takes place in the $\phi$-direction, where the angular distance changes as $\sin \theta d\phi$. A straightforward example to show that the previous version was wrong is the ground-bounce experiment (vertical propagation) - previously, this would give a focusing factor of $n_{ice}/n_{surface}$ instead of $n_{ice}^2/n_{surface}^2$. However, for most neutrino trajectories the $\phi$-factor is very close to 1 as $n\approx n_{ice}$ for most of the ray path.
Derivation in pdf attached
![image](https://github.com/nu-radio/NuRadioMC/assets/78698543/d417a8e2-7080-4c1a-821e-90c1f7f1dbfb)
$\phi$-factor only:
![image](https://github.com/nu-radio/NuRadioMC/assets/78698543/4ac0e48e-003b-41c0-9be7-486352ab9d96)

sjoerd-bouma:
3. Other minor corrections:
For some reason, a separate expression was derived for $dy/dz$ and used in the code that looks really complicated:
[...]

PR #649, opened 2024-03-05
Focusing factor

sjoerd-bouma:
I found a missing factor in the focusing factor used in NuRadioMC. This PR also fixed a couple of incorrect analytic formulas for the path length and travel time in the analytic ray-tracing case, and fixes the double-counting of the focusing factor for ice-to-air (or air-to-ice) ray-tracing. For more details, see #648


anelles:
You should get Dave Seckel to review this, but I am not sure that he does code review :D
If he is ok with your concept, I would accept this as a review though ... ;)

fschlueter:
I can confirm that with the changes in here I am able to get a ray-tracing solutions for all my ice-to-air geometries. For my air-to-ice ray tracing the zenith angle of the receiving vector looks normal. If I flip the start and end position and simulate an ice-to-air ray-tracing the receiving vector zenith angles are all below 90deg which I would expect them to be +90 deg

anelles:
@sjoerd-bouma can you fix the conflicts? I think if this is the version from your thesis, we should merge this. It is documented and better than what we had before.

fschlueter:
What is the status of this PR?

fschlueter:
@sjoerd-bouma ?

sjoerd-bouma:
So I've been over all the differences with the tests:

-**Single event test (SP)**: the only difference is in the max_amp_shower_and_ray attribute. This difference is <~ 1% for all but two events; the one event where the difference is more than 5% has a vertex at -133 m. I've confirmed that the differences for these two events is due only to the change in the focusing factor. (See also https://github.com/nu-radio/NuRadioMC/issues/648#issuecomment-2122888452 for the expected difference.)
- **Single event test (ARZ)**: Here, in addition to the difference in amplitude, there is one event fewer than in the reference file. This event turns out to have an amplitude that is exactly at the trigger threshold, so presumably the ~1% decrease in amplitude from the focusing factor made it no longer trigger
- **Veff test with noise**: the decrease in Veff is ~1%, this seems plausible as the change in the focusing factor tends to slightly decrease the signal amplitude;
- **Trigger tests**: the trigger times are shifted by less than 0.05 ns. This is due to the previously incorrect travel times.

The test files have now all been updated.

One other change I noticed to the output is that the `max_amp_shower_and_ray` output in the HDF5 files is now `np.nan` rather than 0 if no signal is present. This is however not a change due to this PR.

sjoerd-bouma:
One final comment - I reverted two commits that were workarounds for the air-to-ice raytracing. As this feature isn't currently working anyway (see #823 ) I removed them to reduce the potential differences with the reference files. Fixing/reimplementing air-to-ice raytracing should be a separate PR.

sjoerd-bouma:
Ping @fschlueter or anyone who would like to approve this.

fschlueter (dismissed):

cg-laser (changes requested):
Thanks Sjoerd!

It all looks good to me, but it would be great if you could add additional documentation to the analytic travel time and distance functions with a link to the derivation in your thesis. An additional note that these equations are (slightly) different than originally published in the NuRadioMC paper. 
 
The np.nan for max_amp_showre_and_ray is likely from the NuRadioMC 3.0 changes. And in general, it is much better to have a nan when there is no RT solution, than zero. 

And super annoying that the air-to-ice ray tracing was removed... I did not catch this... However, it shouldn't be too difficult to reimplement it. And maybe we should do this soonish so that we don't forgot about it. I'll try to make a PR in the next days.

fschlueter:
> And super annoying that the air-to-ice ray tracing was removed... I did not catch this... However, it shouldn't be too difficult to reimplement it. And maybe we should do this soonish so that we don't forgot about it. I'll try to make a PR in the next days.

@cg-laser, yes please for the FEARIE simulations we need that.

sjoerd-bouma:
> Thanks Sjoerd!
> 
> It all looks good to me, but it would be great if you could add additional documentation to the analytic travel time and distance functions with a link to the derivation in your thesis. An additional note that these equations are (slightly) different than originally published in the NuRadioMC paper.
> 
> The np.nan for max_amp_showre_and_ray is likely from the NuRadioMC 3.0 changes. And in general, it is much better to have a nan when there is no RT solution, than zero.
> 
> And super annoying that the air-to-ice ray tracing was removed... I did not catch this... However, it shouldn't be too difficult to reimplement it. And maybe we should do this soonish so that we don't forgot about it. I'll try to make a PR in the next days.

My thesis won't be online until May (I'm defending 30th April), so I can't add an actual link, but I can refer to it in words? There's no derivation as such (I just put the integral into Wolfram Alpha), but I suppose it has the formulas (which I could also copy into the docstrings, if that's useful).

sjoerd-bouma:
@cg-laser I've added the equations and referred to my PhD thesis, which it felt very weird to do in the third person about myself.

cg-laser (approved):
thanks Sjoerd!

sjoerd-bouma:
@fschlueter #768 caused a merge conflict, which I've fixed; the difference to the reference Veff was a bit larger than I would have expected (~6%), but it looks like #768 changes the noise in the Veff with noise test (it looks the same for the first 1000 or so simulated events, then different after that), and the simulated trigger is a phased trigger, so... I'm assuming everything is fine.

fschlueter (approved):

PR #655, opened 2024-03-13
group_id for SimChannel

twybo:
**Issue being addressed**
Added the option to pass a group id to a SimChannel, which was not yet implemented.

Will quickly add a commit for changelog

### DIFF
--- a/NuRadioReco/framework/sim_channel.py
+++ b/NuRadioReco/framework/sim_channel.py
@@ -31,7 +31,8 @@ class SimChannel(NuRadioReco.framework.channel.Channel):
         ray_tracing_id: int or None
             the id of the corresponding ray tracing solution
         channel_group_id: int (default: None)
-            group_id of channel
+            optionally, several channels can belong to a "channel group". Use case is to identify
+            the channels of a single dual or triple polarized antenna as common in air shower arrays. 
         """
         NuRadioReco.framework.channel.Channel.__init__(self, channel_id, channel_group_id=channel_group_id)
         self._shower_id = shower_id
--- a/NuRadioReco/framework/sim_channel.py
+++ b/NuRadioReco/framework/sim_channel.py
@@ -18,7 +18,7 @@ class SimChannel(NuRadioReco.framework.channel.Channel):
     a shower and ray tracing solution id
     """
 
-    def __init__(self, channel_id, shower_id, ray_tracing_id, channel_group_id):
+    def __init__(self, channel_id, shower_id, ray_tracing_id, channel_group_id=None):
         """
         Initializes SimChannel object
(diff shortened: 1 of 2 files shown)


NuRadioReco/framework/sim_channel.py
```diff
@@ -14,11 +14,11 @@ class SimChannel(NuRadioReco.framework.channel.Channel):
     """
     Object to store simulated channels.
 
-    This class is the same as the regular channel trace but has apart from the channel id also
+    This class is the same as the regular channel trace but has apart from the channel id (and possible group id) also
     a shower and ray tracing solution id
     """
 
-    def __init__(self, channel_id, shower_id, ray_tracing_id):
+    def __init__(self, channel_id, shower_id, ray_tracing_id, channel_group_id):
```
sjoerd-bouma, line 21: Please set the default argument to None (just copy what is done in Channel on the LOFAR branch). An remember to update the docstring accordingly (you can also copy that from the LOFAR branch docstring if you like)

changelog.txt
```diff
@@ -11,6 +11,7 @@ It comes with a class to handle responses of the different signal chain componen
 added new STATUS logging level by default, added logging documentation and moved logging
 module to NuRadioReco/utilities folder. Logging functionality from NuRadioReco/modules/base
 is now deprecated.
+- SimChannel can now take a group id
```
sjoerd-bouma, line 14: It probably makes more sense to update the changelog once, on the LOFAR branch, so I think this can be removed for now : )

sjoerd-bouma (commented):
Thanks! Can you please change the target branch to `LOFAR` (instead of `develop`)? This feature won't work before it's implemented for the `Channel` too, which is done on that branch.

NuRadioReco/framework/sim_channel.py
twybo: I set the default argument and copied the docstring from the LOFAR branch, thanks.

MijnheerD:
So there are really only 5 commits here that differ from the LOFAR branch... Not sure why the PR is not updating, on my local version git shows this correctly. I think this can be merged into LOFAR without any issues, or should start a new PR just to be safe?

MijnheerD:
I think what I am going to do is to merge this locally into LOFAR, push the changes, and close this PR. I don't understand why Github says there are 53 commits to be merged into LOFAR, this should not be the case. Is there anything against this procedure?

MijnheerD:
Ah damn, I forgot to merge this into LOFAR before merging that branch in develop. Is it okay to merge this into develop now @sjoerd-bouma ?

cg-laser (approved):

PR #658, opened 2024-03-13
Hotfix/detector from nur

### LINKED ISSUE #657: Reading normal Detector from nur file not working
**Describe the issue**
When trying to load a normal Detector (i.e. not a Generic one) from a `nur` file, an error is thrown. The cause is line 389 in NuRadioRecoio.py, where the `__new__` method is called. At some point (I believe this happened in commit 844f3aa9be081da98b62bd593a40da9cfaa52114 , where the concept of GenericDetector was introduced) the `json_filename` parameter in this method changed from an optional to required parameter. So now the call on line 389 is missing a required parameter.

Interestingly, the GenericDetector does not produce this error because its `__new__` method is not overwritten.

**Additional context**
MWE to reproduce the error:
```python
import NuRadioReco.modules.io.eventReader.

reader = NuRadioReco.modules.io.eventReader.eventReader()
reader.begin('path/to/nur', read_detector=True)
detector = reader.get_detector()
```

fschlueter:
Fixes #657.

### DIFF
--- a/NuRadioReco/detector/detector.py
+++ b/NuRadioReco/detector/detector.py
@@ -29,12 +29,12 @@ def find_reference_entry(station_dict):
     """
 
     for station in station_dict['stations']:
-        if 'reference_station' in station_dict['stations'][station].keys():
+        if 'reference_station' in station_dict['stations'][station]:
             return True
 
     for channel in station_dict['channels']:
-        if 'reference_channel' in station_dict['channels'][channel].keys() or 'reference_station' in \
-                station_dict['channels'][channel].keys():
+        if 'reference_channel' in station_dict['channels'][channel] or 'reference_station' in \
+                station_dict['channels'][channel]:
             return True
 
     return False
@@ -112,10 +112,13 @@ def Detector(*args, **kwargs):
 
         elif source == "dictionary":
 
-            if "dictionary" not in kwargs:
-                raise ValueError("Argument \"dictionary\" is not passed to Detector() while source=\"dictionary\" is set.")
+            if not isinstance(dictionary, dict):
+                raise ValueError("Argument \"dictionary\" is not correct while source=\"dictionary\" is set.")
 
-            station_dict = kwargs["dictionary"]
+            station_dict = dictionary
+
+            # in this case, `json_filname` not need and should not be passed twice
+            kwargs.pop("json_filename", None)
             filename = ''
 
         elif source == 'json':
--- a/NuRadioReco/detector/generic_detector.py
+++ b/NuRadioReco/detector/generic_detector.py
@@ -94,6 +94,7 @@ class GenericDetector(NuRadioReco.detector.detector_base.DetectorBase):
                 # json when updating the table to include reference_station/channel/device
                 with open(json_filename, "r") as json_input:
                     dictionary = json.load(json_input)
+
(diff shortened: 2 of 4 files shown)


sjoerd-bouma (approved):
Looks good to me, although it may be worth @MijnheerD validating that it works once more : )

MijnheerD:
Yes, this solves #657 , thanks!

MijnheerD (approved):

MijnheerD:
Actually, I am afraid I have to come back to this. I think I only received the change from commit 633cc151b193674aa3bc54adeac5c636befaece4 now (because I merged develop back into our lofar/development branch), so I was always working with the Detector class instead of the function variant. 

Now I am having a bunch of problems because stuff is not correctly imported from the `.nur` file. The Detector time is not longer set and the `self._antenna_by_depth` keyword is reset (which is very annoying when trying to load antenna models). Not sure how this can be addressed?

sjoerd-bouma:
@MijnheerD I am slightly surprised if that is a new issue, because I don't think that has changed. It is however a good point - it would be much more convenient if the detector time and the antenna_by_depth kw would also be stored when storing the detector in the nur file.

MijnheerD:
@sjoerd-bouma The reason this is new to me, is because I was testing the hotfix on our lofar branch, which was not up to date with develop. Therefore, I was still working with the class implementation of Detector instead of the function variant. I am not sure what changed, but somehow with the class variant before I was not encountering the issues with the keyword not being set (this is most notable when loading antenna patterns, because when the `antenna_by_depth` kw is kept to True, there is something append to the name which makes it incompatible with our LOFAR antenna pattern name).

fschlueter:
Hi both, I will try to have a look at this. This is indeed a problem (at least for the antenna depth key word). I think the detector time should be actually set to None which would require to call an `det.update`

fschlueter:
I have know I idea why it worked before though. Are you sure it worked before @MijnheerD ?

sjoerd-bouma:
@fschlueter @MijnheerD I would be somewhat surprised if there was no need to call `update` before. The `antenna_by_depth` keyword was also not set previously, but would have defaulted to `False` instead of the current `True`, so that would have 'worked'. 
Obviously it would still be preferable to have both stored correctly, but as far as I'm concerned that does not need to happen in this PR, as it is a separate issue (if you do decide to include it in this PR that's also fine by me).

fschlueter:
Arg, I did change the default behaviour? This is bad of course.... @sjoerd-bouma  do we store the keyword in the existing nur files (does not look like it on a first glance)

sjoerd-bouma:
> Arg, I did change the default behaviour? This is bad of course.... @sjoerd-bouma do we store the keyword in the existing nur files (does not look like it on a first glance)

No, this change predates you - originally the Detector and GenericDetector classes were separate, and had different default arguments for `antenna_by_depth`. At some point the Detector class was moved to detector_base and Detector would automatically be able to decide if an imported detector was a generic detector or not, which simplified life a lot, but this also meant that their default arguments were unified.

As far as I can tell the initialization arguments and detector time have never been stored in the nur files. I agree with @MijnheerD that they should be, though.

fschlueter:
Fine. I suggest we merge this and open an new PR for this. @MijnheerD would you mind creating a short Issue and assign me?

PR #661, opened 2024-03-19
Utility for gain, position and focusing factor correction

sanyukta13:
Closed the pull request on modules, put in in utilities. Stopped using json files for positions

### DIFF
--- a/NuRadioReco/utilities/Gain&Focusingfactor.py
+++ b/NuRadioReco/utilities/Gain&Focusingfactor.py
@@ -1,0 +1,426 @@
+import matplotlib.pyplot as plt
+import numpy as np
+from NuRadioReco.utilities import units
+from NuRadioReco.modules import channelAddCableDelay, channelBandPassFilter, sphericalWaveFitter
+from NuRadioReco.detector import detector
+from NuRadioReco.modules.io.RNO_G import readRNOGDataMattak
+from datetime import datetime
+import NuRadioReco.detector.detector
+import astropy.time
+import logging
+from NuRadioReco.utilities import fft
+from os import listdir
+from os.path import isfile, join
+from scipy.signal import correlate
+from NuRadioReco.utilities.fft import time2freq
+from NuRadioReco.utilities.fft import freq2time
+from scipy.interpolate import interp1d
+import pandas as pd
+import json
+from scipy.interpolate import griddata
+
+AddCableDelay = channelAddCableDelay.channelAddCableDelay()
+
+def read_root_file(station_number, fiber, run, selectors=[], sampling_rate=3.2):
+    """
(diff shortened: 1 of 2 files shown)


cg-laser (changes requested):
I'm not quite sure what this utility is supposed to do, so please correct me if some of my comments are wrong. 

This seems to be very RNO-G specific. If this should be a general NuRadioReco utility, it needs to be generatlized a bit. Also some/many of the functions implement functionality which is already present in NuRadioReco, e.g. to calculate SNRs. This is handled by the channelSignalReconstructor. Or the bandapss filter. 

The utility also needs very specific data files. Without these data files, the utility is useless. 

Some of the functionality seems to be rather suited to be an RNO-G specific reconstruction module that corrects the data for certain effects. 

Our way of taking antenna responses into account is through the antenna pattern class, where antenna responses are centrally defined. I suggest to use this interface. 

In case this should stay a very RNO-G specific utility, I suggest to move it to the RNO-G github instead.

sanyukta13:
> I'm not quite sure what this utility is supposed to do, so please correct me if some of my comments are wrong.
> 
> This seems to be very RNO-G specific. 
- It is RNO-G specific yes, do you have suggestions on alternate folders/place to push it?
If this should be a general NuRadioReco utility, it needs to be generatlized a bit. Also some/many of the functions implement functionality which is already present in NuRadioReco, e.g. to calculate SNRs. This is handled by the channelSignalReconstructor. Or the bandapss filter.
- SNR I was not aware of, I will check if it does the same task. I was using the noise_rms for SNR calculation, the part of the waveform before the pulse. I will check if it does a similar thing.
> The utility also needs very specific data files. Without these data files, the utility is useless.
- Yes, one needs gain and density files, I am not sure then if I should put it on the utility
> Some of the functionality seems to be rather suited to be an RNO-G specific reconstruction module that corrects the data for certain effects.
- Yes I agree it is RNO-G specific, please let me know if you can think of alternate and more suitable file paths
> Our way of taking antenna responses into account is through the antenna pattern class, where antenna responses are centrally defined. I suggest to use this interface.
- I don't exactly know of antenna pattern class, can you refer me to the same?
> In case this should stay a very RNO-G specific utility, I suggest to move it to the RNO-G github instead.
- Okay

anelles:
I agree with @cg-laser that this is a wild mix of detector parameters and other functions. I don't think they are useful in utilities either. 
If you want to move further with this, can I suggest that you attend one of the NuRadio calls once and discuss with the developers. Please reach out to @fschlueter to be added to the schedule.

anelles:
Otherwise, we should close this PR as is.

fschlueter:
@sanyukta13 feel free to reach out to us. Or address our comments. But for the moment I will close this PR.

PR #662, opened 2024-03-26
Restore backward-compatbility with correct default arguments for Gene…

fschlueter:
…ricDetector objects

@MijnheerD

### DIFF
--- a/NuRadioReco/detector/detector.py
+++ b/NuRadioReco/detector/detector.py
@@ -99,13 +99,14 @@ def Detector(*args, **kwargs):
         if len(args) >= 5:
             antenna_by_depth = args[4]
         else:
-            antenna_by_depth = kwargs.pop("antenna_by_depth", True)
+            # None because the default argument for GenericDetector and DetectorBase are different
+            antenna_by_depth = kwargs.pop("antenna_by_depth", None)
 
 
         if source == "sql":
             return detector_base.DetectorBase(
                 json_filename=None, source=source, dictionary=dictionary,
-                assume_inf=assume_inf, antenna_by_depth=antenna_by_depth)
+                assume_inf=assume_inf, antenna_by_depth=antenna_by_depth or True)
 
         elif source == "rnog_mongo":
             return rnog_detector.Detector(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


MijnheerD:
This still does not work for me unfortunately. When reading in the LOFAR detector description like this:
```python
# Load the LOFAR detector description
detector = NuRadioReco.detector.detector.Detector(
    'LOFAR/LOFAR.json',
    source='json',
    antenna_by_depth=False
)
```
the `antenna_by_depth` is not parsed properly. The cause seems to be this syntax:
```python
antenna_by_depth or True
```
in line 164. This expression always evaluates to `True` , so it is impossible to create a normal Detector with `antenna_by_depth = False`.

fschlueter:
`antenna_by_depth or True` should return `False` if `antenna_by_depth == False`

fschlueter:
I thought the lofar detector is of type `GenericDetector`. Hence the expression should be `antenna_by_depth or False`

fschlueter:
> `antenna_by_depth or True` should return `False` if `antenna_by_depth == False`

Grrr okay I tested it but apparently it does not work. Sorry, not sure what I tested ...

fschlueter:
@MijnheerD can you give it another try?

MijnheerD:
Jups, this works now. Though the `antenna_by_depth` keyword is still not saved in the .nur file, is this intentional?

I do have a suggestion to clean up the code though. I would change the Detector and GenericDetector `__init__` functions to have `antenna_by_depth=None` and then implement a check that sets them to their default values when that keyword is left `None`. That would make it much easier in the `Detector()` function to create them, as it removes the necessity to check whether the values are `None`.

Also, the docstring for the GenericDetector is inconsistent. It says the default value for `antenna_by_depth` is `True`.

fschlueter:
@MijnheerD, it is kind of intentional. I do not have the time to implement it. Feel free to do it, it is certainly the best option. This PR only aims in bringing back the status-quo where we do not save it but correctly use the two different default values in the factory function.

Do you think my solution is so messy? I would keep it as it is. I would fix the doc-string in the `GenericDetector` class and keep the default value as it is.

MijnheerD:
Sure, let's discuss the serialization of the parameters further in #663 

And your solution is not really messy, it's just that it requires the custom `if_not_None()` function which - while being very clear - it just some extra code that _could_ be avoided. But on the other hand, it avoids the user having to go look at the (Generic)Detector class to find out what the default values are, which might be better (although could introduce inconsistencies like that docstring).

MijnheerD (approved):

fschlueter:
@MijnheerD didnt you implement all changes from here also in #665? Maybe it is easier to just merge 665?

MijnheerD:
Indeed, I merged this branch into #665 because I needed the changes. Though I do think the docstring update was not there yet, so we would need to merge that change into #665 as well.

For me it's fine to drop this PR and do everything in #665 :)

PR #664, opened 2024-03-26
Improve logging for NuRadioReco and NuRadioMC

MijnheerD:
Adding a FileHandler to the logger (to for example capture more verbose output) does not work as intended. Because the logging level is currently controlled by the logger itself, the different handlers cannot decide on what output to capture.

Additionally, logging for scripts which use a mix of NRR and NRMC modules is convoluted. As a potential solution, we could try initializing loggers for both packages when they are important, to remove the need to call `setup_logger()` for each one separately.


MijnheerD:
Other suggestions that were discussed on Slack:

- Define the our custom logging level STATUS in the NRR __init__ file, to ensure it's existence
- Create our own logger class, to avoid potential conflicts with other Python packages which use their own STATUS log level
- Let `setup_logger()` always create both loggers by default
- Let `setup_logger()` do nothing in case loggers already exist
- New logging format, which includes time (@sjoerd-bouma do you already have a nice format for this?)

MijnheerD:
I changed the implementation of the logging, by adding a custom logging class instead of modifying the core `logging` library of Python. This custom class, called `NuRadioLogger`, has the `status()` method and adds the STATUS name to the internal logging dictionary which is used to properly format the messages. I also modified the `addHandler()` method to ensure the logger level is always lower or equal to the lowest handler level. This is for convenience, as otherwise one would have to set the logger level manually after adding a handler.

This should avoid any clashes with other packages with also have their own STATUS log level. The only thing I can foresee, is that multiple levels would be associated to the STATUS in the internal dictionary. Though I don't think that would be such an issue? If we really want to avoid this, we could call the log level NURADIO STATUS or something, but that seems a bit unnecessary to me.

Additionally, I have also changed the format to include time (with timezone) and date - adding some colors along the way. I am sure this can be made much more fancy, with colors per severity level, but I think this is overkill (furthermore, there are plugins for code editors which can highlight the severity levels in different colors for you).

Lastly, I have added calls to `setup_logger()` in the NRR and NRMC **init** files. The function now does nothing if the logger in question has already been created, which is detected by checking if there are already handlers present. This should make it such that there is no harm in accidentally calling the function twice on the same name.

What do you think? If you are happy with these changes, I can update the documentation and start converting some of the test scripts. Although the latter should still work, as the a second call to the `setup_logger()` function should simply do nothing.

cg-laser:
Your solution seems very good! Thanks a lot! To avoid any confusion, can you post an example of how this logger needs to be used? 
I didn't immediately see how a file handler can be set up. Can you add this to your example code? 

Thanks

MijnheerD:
@cg-laser I added some example code to the logging section of the documentation. Is this clear enough? Or do you want an example somewhere else as well?

MijnheerD:
Oh, and maybe to clarify how one would use this in a script: if during the script you don't want to log anything (except for the messages from the modules), you don't need to do anything. 

If you want to change the general logging level, you can use the `set_general_logging_level()` from the NRR logging module. 

In case you want to make logging statement during the script, you should get a logger using the standard logging function:
```python
logger = logging.getLogger("NuRadioReco.my_script")
```
You could also get the NRR logger itself, but I think it is conceptually cleaner to stick to the naming scheme for scripts as well :-)

MijnheerD:
I took some time to go through the Python logging library, and found basically two methods we can choose to handle our logging. They both have some (dis)advantages and I'll try to detail them both here.

But perhaps first a bit of context about how logging works in the Python standard library. In Python, a `Logger` object is not able to write a log message to the output. This is done by one or more `Handler` objects which we add to a logger. By default, a logger does not have a handler. Also by default, loggers propagate log records to their parent. This behavior is set up in this way because the root logger is the only one equipped with a handler out of the box. Because all log records are propagated up to the root logger, we get logging statement into our output.

To decide what logs are actually supposed to be printed, there are log levels. They are the lowest level that will be processed (ie log records with a level lower than the log level will be thrown away). Interestingly both loggers and handlers can have a log level, though the logger's level takes priority. If the record has a higher level than the logger, the record is passed on to each handler which can still decide whether to process it or not.

Option A: global log level
-------------------------------------

To achieve a truly global logging level, with one setting we can control the logging level of all modules, we can make use of the fact that by default a logger does not have a logging level (ie its level is `logging.NOTSET`). Instead, they inherit it from their parent. If we ensure that all module loggers (so all loggers with a name like `NuRadioReco.my_module`) do not have a logging level set, we can control all their outputs by simply changing the logging level of the `NuRadioReco` logger.

The disadvantage is that debugging a single module will become difficult, as there is no easy way to only set the level of that module. What I mean with this, is the following:

- Imagine setting the NRR logger to WARNING, but the module you are testing you explicitly set to DEBUG
- At this point all module loggers inherit WARNING from the parent logger, but your module has DEBUG logging active
- While the module will send DEBUG logs to the NRR logger, the latter's level is WARNING, which means that the records will be rejected. And as the NRR is the one which has a handler and is thus responsible for outputting the records, the module will not actually have any statements printed to the ouput.

There are of course ways around this. You could for example simply add a handler to the module logger, making it capable of outputting its own records. Another way that I think should be possible, is to install a `Filter` on the NRR logger which should be capable of intercepting the DEBUG statement even if the log level is higher. But these are more involved and would require a user to mess with the loggers themselves.

Another downside is that we would need to change almost all the modules, as they typically set the logging level in their `begin()` function right now.

Option B: module log level
---------------------------------------

The other way of doing it, is to stick with the current scheme where every module has its own log level and therefore decides itself what records to send to the NRR logger. The latter one then should have the lowest level possible, to ensure all records are processed.

The downside here is that it is not easy to globally control the logging, at least not "downwards". What I mean by that is that we still set the NRR logger to a higher log level to block lower level messages. So even if modules have DEBUG level set, if the NRR logger has WARNING those messages will not be outputted. 

But we cannot do it the other way around. If we need the DEBUG logs from some modules, we will need to explicitly set the level for each of them, as well as make sure the NRR logger has the DEBUG level as well (which I think should be the default in this scenario, but if users start messing with it like in the previous paragraph, it can become confusing why the messages are not appearing).

MijnheerD:
@sjoerd-bouma , @cg-laser , @fschlueter and other interested people, please let me know what you think :) Both options are relatively easy to implement, so we should really just decide what behavior is most desirable.

MijnheerD:
Actually, I have another, much easier solution to the problem in Option A:

> There are of course ways around this. You could for example simply add a handler to the module logger, making it capable of outputting its own records. Another way that I think should be possible, is to install a Filter on the NRR logger which should be capable of intercepting the DEBUG statement even if the log level is higher. But these are more involved and would require a user to mess with the loggers themselves.

Apparently the parent loggers level is skipped when propagating log records (looking at [this diagram](https://docs.python.org/3/howto/logging.html#logging-flow)). This means that as long as the handlers level is low enough, setting a single module log level lower will result in the messages being outputted. 

So the only real downside that remains for Option A is that we have to edit all `begin()` functions.

NuRadioMC/SignalGen/ARZ/ARZ.py
```diff
@@ -15,8 +15,7 @@
 except ImportError:
     numba_available = False
 
-logger = logging.getLogger("SignalGen.ARZ")
-logging.basicConfig()
+logger = logging.getLogger("NuRadioMC.SignalGen.ARZ")
 # logger.setLevel(logging.INFO)
```
fschlueter, line 19: should we remove the comment?

NuRadioMC/SignalProp/examples/E02ToAir.py
```diff
@@ -8,8 +8,8 @@
 import logging
 from radiotools import helper as hp
 from radiotools import plthelpers as php
-logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger('raytracing')
+logger = logging.getLogger('NuRadioMC.SignalProp.raytracing')
+logger.setLevel(logging.INFO)
```
fschlueter, line 12: should we remove this line?

NuRadioMC/SignalProp/examples/example_3d.py
```diff
@@ -7,8 +7,8 @@
 import logging
 from radiotools import helper as hp
 from radiotools import plthelpers as php
-logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger('raytracing')
+logger = logging.getLogger('NuRadioMC.SignalProp.raytracing')
+logger.setLevel(logging.INFO)
```
fschlueter, line 11: same here? Do we want to set here logging.INFO?

NuRadioMC/examples/03_station_coincidences/E06RunSimPreprocess.py
```diff
@@ -10,8 +10,8 @@
 from NuRadioReco.utilities import units
 from NuRadioMC.simulation import simulation
 import logging
-logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger("runstrawman")
+logger = logging.getLogger("NuRadioMC.runstrawman")
+logger.setLevel(logging.INFO)
```
fschlueter, line 14: .

NuRadioMC/test/SignalProp/G01generate_data_set_for_unit_test.py
```diff
@@ -6,8 +6,8 @@
 from NuRadioReco.utilities import units
 import logging
 from numpy import testing
-logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger('test_raytracing')
+logger = logging.getLogger('NuRadioMC.test_raytracing')
+logger.setLevel(logging.INFO)
```
fschlueter, line 10: .

NuRadioMC/test/SignalProp/G02generate_data_set_for_unit_test_mooresbay.py
```diff
@@ -6,8 +6,8 @@
 from NuRadioReco.utilities import units
 import logging
 from numpy import testing
-logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger('test_raytracing')
+logger = logging.getLogger('NuRadioMC.test_raytracing')
+logger.setLevel(logging.INFO)
```
fschlueter, line 10: .

NuRadioMC/test/SignalProp/T01test_python_vs_cpp.py
```diff
@@ -6,8 +6,8 @@
 from NuRadioReco.utilities import units
 import logging
 from numpy import testing
-logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger('test_raytracing')
+logger = logging.getLogger('NuRadioMC.test_raytracing')
+logger.setLevel(logging.INFO)
```
fschlueter, line 10: .

NuRadioMC/test/SignalProp/T02test_analytic_D_T.py
```diff
@@ -6,8 +6,8 @@
 from NuRadioMC.utilities import medium
 from NuRadioReco.utilities import units
 import logging
-logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger('test_raytracing')
+logger = logging.getLogger('NuRadioMC.test_raytracing')
+logger.setLevel(logging.INFO)
```
fschlueter, line 10: .

MijnheerD:
Hi @fschlueter , thanks for your review. I tried to not touch the logging levels of scripts, as I suspected those might be desired by the user :) But if you think they can go, then I would happily do so

fschlueter:
> Hi @fschlueter , thanks for your review. I tried to not touch the logging levels of scripts, as I suspected those might be desired by the user :) But if you think they can go, then I would happily do so

I think it would be fair to change all with are set to `INFO` but keep the one which are set to `WARNING` or `DEBUG` but I am happy if you skip the scripts.

NuRadioMC/test/SignalProp/T02test_analytic_D_T.py
fschlueter: I found many more examples but stopped marking them

MijnheerD:
In general this PR now fully implements option A from above. I tested it with a LOFAR script I had, and I can set a general logging level as well as set one module to a different level. Please run your own scripts as well, to see if I missed anything in one of the other directories.

I also tried to update all logger names to follow the naming scheme. I think modules related to specific experiments should try to use the `NuRadioReco.Experiment.Module` scheme, as this could allow one to modify the `NuRadioReco.Experiment` logger to immediately touch the logging of all sub-loggers. But for some modules I was not sure whether they were general or not, so sometimes I had to guess.

The last things that need to be done are to update the documentation and the `changelog.txt` . If you are happy with the changes, I will do these last things and then the branch can be merged.

MijnheerD:
Also, the default logging level is STATUS (25). Do you think the unit tests should be set to a lower value, to have more output in case of failure?

fschlueter:
> Also, the default logging level is STATUS (25). Do you think the unit tests should be set to a lower value, to have more output in case of failure?

Hm I can see pros and cons here. I would keep it status for the time being, one can run it locally as default once a test fails.

MijnheerD:
> > Also, the default logging level is STATUS (25). Do you think the unit tests should be set to a lower value, to have more output in case of failure?
> 
> Oh so you implemented `STATUS` now "repository wide" - I assume only very few modules actually use the difference between `INFO` and `STATUS`

Probably, because it was not available yet 😝 But currently the default setting of STATUS makes it such that INFO messages are not outputted. From my perspective this is normal, as the default from the Python `logging` library is WARNING, so we are consistent with that. I'll try to makes this also clear in the documentation.

MijnheerD:
Any updates on this PR? Did anyone else try to run a script with these changes?

fschlueter (approved):

fschlueter:
I approved the changes. All tests are passing. I am currently busy can not test it but I will it in the future (this should not delay this PR).

Issue #663, opened 2024-03-26
Store arguments with which a detector is initialised in  nur files.


fschlueter:
Currently we are not storing arguments such as `antenna_by_depth` in the nur files. Hence, when reading a detector from a nur file the default values are used.

MijnheerD:
This issue comes up when loading a normal Detector from a .nur file for an in-air experiment. Because the `antenna_by_depth` keyword defaults to `True`, the naming for the antenna pattern files is messed up when using the Detector which is read in.

PR #665, opened 2024-03-27
Save detector keywords in NUR file

MijnheerD:
Issue being addressed: #663 

The Detector parameters `assume_inf` and `antenna_by_depth` were not saved in the .nur file when writing out the Detector object. This means that these values are reset to their defaults when the Detector is being read in from the .nur file, which can cause issues with for example antenna model naming.

In this PR, I have made `assume_inf` and `antenna_by_depth` properties of the DetectorBase class, with getters and setters to easily access them (though I am not sure that a setter is really necessary - we can also remove it to protect these attributes from being overwritten).  These attributes are then added to the `det_dict` in the eventWriter, and are also read in by the event_parse_factory.py code. The NuradioRecoio class then check if these values are present, and sets them to `None` if they are not. I have changed the Detector factory function of @fschlueter to use `None` as the default for `assume_inf` (this was already done for `antenna_by_depth`) and change this to the default value of `True` when calling the appropriate Detector class using the `if_not_None()` function.

### DIFF
--- a/NuRadioReco/detector/detector.py
+++ b/NuRadioReco/detector/detector.py
@@ -99,7 +99,7 @@ def Detector(*args, **kwargs):
         if len(args) >= 4:
             assume_inf = args[3]
         else:
-            assume_inf = kwargs.pop("assume_inf", True)
+            assume_inf = kwargs.pop("assume_inf", None)
 
         if len(args) >= 5:
             antenna_by_depth = args[4]
@@ -110,7 +110,8 @@ def Detector(*args, **kwargs):
         if source == "sql":
             return detector_base.DetectorBase(
                 json_filename=None, source=source, dictionary=dictionary,
-                assume_inf=assume_inf, antenna_by_depth=if_not_None(antenna_by_depth, True))
+                assume_inf=if_not_None(assume_inf, True),
+                antenna_by_depth=if_not_None(antenna_by_depth, True))
 
         elif source == "rnog_mongo":
             return rnog_detector.Detector(*args, **kwargs)
@@ -165,7 +166,8 @@ def Detector(*args, **kwargs):
 
             return generic_detector.GenericDetector(
                 json_filename=filename, source=source, dictionary=dictionary,
-                assume_inf=assume_inf, antenna_by_depth=if_not_None(antenna_by_depth, False), **kwargs)
+                assume_inf=if_not_None(assume_inf, True),
+                antenna_by_depth=if_not_None(antenna_by_depth, False), **kwargs)
         else:
             # Keys might be present (but should be None). Keys are deprecated, keep them for backwards compatibility
             for key in ["default_station", "default_channel", "default_device"]:
@@ -173,4 +175,5 @@ def Detector(*args, **kwargs):
 
             return detector_base.DetectorBase(
                 json_filename=filename, source=source, dictionary=dictionary,
-                assume_inf=assume_inf, antenna_by_depth=if_not_None(antenna_by_depth, True), **kwargs)
+                assume_inf=if_not_None(assume_inf, True),
+                antenna_by_depth=if_not_None(antenna_by_depth, True), **kwargs)
--- a/NuRadioReco/detector/detector_base.py
+++ b/NuRadioReco/detector/detector_base.py
@@ -208,6 +208,38 @@ class DetectorBase(object):
             logger.info("the correct antenna model will be determined automatically based on the depth of the antenna")
         self._antenna_by_depth = antenna_by_depth
 
+    @property
+    def assume_inf(self):
+        """
+        Getter function for the `assume_inf` attribute
+        """
+        return self.__assume_inf
+
+    @assume_inf.setter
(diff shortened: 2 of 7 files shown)


MijnheerD:
So far, I have tested this with a normal Detector (the LOFAR one) and it works. I still need to verify that it also works with a GenericDetector, and that this change is backwards compatible.

MijnheerD:
And then another note to self: I still need to update the changelog once we are happy with the changes :-)

NuRadioReco/detector/detector.py
```diff
@@ -94,18 +99,19 @@ def Detector(*args, **kwargs):
         if len(args) >= 4:
             assume_inf = args[3]
         else:
-            assume_inf = kwargs.pop("assume_inf", True)
+            assume_inf = kwargs.pop("assume_inf", None)
```
fschlueter, line 102: is that necessary here? If both classes have the same default for this argument?

NuRadioReco/detector/detector_base.py
```diff
@@ -208,6 +208,38 @@ def __init__(self, source='json', json_filename='ARIANNA/arianna_detector_db.jso
...
 
+    @property
+    def assume_inf(self):
+        """
+        Getter function for the `assume_inf` attribute
+        """
+        return self.__assume_inf
+
+    @assume_inf.setter
+    def assume_inf(self, value):
```
fschlueter, line 219: should that function than not raise a warning if the value is not set?

NuRadioReco/modules/io/NuRadioRecoio.py
```diff
@@ -368,6 +378,7 @@ def get_detector(self):
                         source='dictionary', dictionary=detector_dict,
                         default_station=detector_dict.get('default_station', None),
                         default_channel=detector_dict.get('default_channel', None),
+                        assume_inf=assume_inf, antenna_by_depth=antenna_by_depth,
```
fschlueter, line 381: instead of the code block from 360-369 you can only write `assume_inf=detector_dict['detector_parameters'].get('assume_inf', None)`

fschlueter: ah, in the old files the nested dict does not exist neither. In that case I would get rid of the nested dict to simplify the code.

NuRadioReco/detector/detector.py
MijnheerD: True, but this change is to accommodate the default values from NuradioRecoio.py . If the value for assume_inf is not in the .nur file, it defaults to None. Rather then setting a default value there as well (making it hardcoded in that place else as well), I opted to change this here.

NuRadioReco/detector/detector_base.py
MijnheerD: Are you talking about the getter or the setter?

fschlueter:
@MijnheerD thanks for doing this. Can you have a look at my comments please?

NuRadioReco/detector/detector_base.py
fschlueter: setter

NuRadioReco/detector/detector.py
fschlueter: okay

NuRadioReco/modules/io/NuRadioRecoio.py
MijnheerD: With the change in NuradioRecoio.py, the 'detector_parameters' key will always exist. If it's not present in the .nur file, it will just be an empty dictionary, so your solution works perfectly 👍🏽

NuRadioReco/detector/detector_base.py
MijnheerD: Why would a warning be necessary? In any case, these values are set in the `__init__` so I don't see how the value could ever not be set.

NuRadioReco/modules/io/event_parser_factory.py
```diff
@@ -98,10 +98,16 @@ def scan_files_2_2(self, iF, current_byte):
                 is_generic_detector = False
             else:
                 is_generic_detector = detector_dict['generic_detector']
+
+            if 'detector_parameters' not in detector_dict.keys():
```
fschlueter, line 102: no`.keys()` needed

fschlueter: actually you can use `dict.get(`detector_parameters`, {})` here as well

NuRadioReco/detector/detector_base.py
fschlueter: It might be over doing it

NuRadioReco/modules/io/event_parser_factory.py
MijnheerD: I copied that syntax from the generic_detector code just above. I'll change that as well then ;)

fschlueter (dismissed):

fschlueter (dismissed):

fschlueter (approved):

Issue #666, opened 2024-04-02
Bug in get_travel_time in directRayTracing module


MartinRavn:
**Describe the issue**
The [get_travel_time](https://github.com/nu-radio/NuRadioMC/blob/9b5c3aeab959c3576bfbef234c56c1ec0b9afd7b/NuRadioMC/SignalProp/directRayTracing.py#L48) function in the directRayTracing propagation module gives wrong propagation times. The function [get_path](https://github.com/nu-radio/NuRadioMC/blob/9b5c3aeab959c3576bfbef234c56c1ec0b9afd7b/NuRadioMC/SignalProp/directRayTracing.py#L35) returns an array with the path in segments and has dimensions [n_segments,3]. However, get_travel_time uses the array as if it had dimensions [3,n_segments] and thus returns wrong travel times.

PR #667, opened 2024-04-02
666 bug in get travel time in directraytracing module

MartinRavn:
Issue being addressed:
#666

### DIFF
--- a/NuRadioMC/SignalProp/directRayTracing.py
+++ b/NuRadioMC/SignalProp/directRayTracing.py
@@ -46,16 +46,11 @@ class direct_ray_tracing(ray_tracing_base):
         return path_length 
     
     def get_travel_time(self, iS):
-        traveltime = 0
         path = self.get_path(iS)
-        segment = [path[1][0]-path[0][0],path[1][1]-path[0][1],path[1][2]-path[0][2]]
-        r = np.linalg.norm(segment)
-        for i in range(len(path)-1):
-            xx = np.array([path[i][0], path[i][1], path[i][2]])
-            yy = np.array([path[i+1][0], path[i+1][1], path[i+1][2]])
-            center_of_segment = (xx+yy)/2
-            n = self._medium.get_index_of_refraction(center_of_segment)
-            traveltime += r/(speed_of_light/n)
+        segment_length = np.linalg.norm(path[1,:]-path[0,:])
+        segment_centers = (path[:-1]+path[1:])/2
+        n = self._medium.get_index_of_refraction(segment_centers)
+        traveltime = np.sum(segment_length/(speed_of_light/n))
         return traveltime
     
     
--- a/NuRadioMC/SignalProp/directRayTracing.py
+++ b/NuRadioMC/SignalProp/directRayTracing.py
@@ -47,12 +47,11 @@ class direct_ray_tracing(ray_tracing_base):
     
     def get_travel_time(self, iS):
         path = self.get_path(iS)
-        segment_length = np.linalg.norm(path[1,:]-path[0,:])
+        segment_length = np.linalg.norm(path[1]-path[0])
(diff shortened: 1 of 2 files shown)


NuRadioMC/SignalProp/directRayTracing.py
```diff
@@ -48,13 +48,13 @@ def get_path_length(self, iS):
     def get_travel_time(self, iS):
         traveltime = 0
         path = self.get_path(iS)
-        segment = [path[0][1]-path[0][0],path[1][1]-path[1][0],path[2][1]-path[2][0]]
+        segment = [path[1][0]-path[0][0],path[1][1]-path[0][1],path[1][2]-path[0][2]]
```
colemanalan, line 51: Just to clarify, what are the indices of `path`?

NuRadioMC/SignalProp/directRayTracing.py
```diff
@@ -48,13 +48,13 @@ def get_path_length(self, iS):
...
-        for i in range(len(path[0])-1):
-            xx = np.array([path[0][i], path[1][i], path[2][i]])
-            yy = np.array([path[0][i+1], path[1][i+1], path[2][i+1]])
-            x = (xx+yy)/2
-            n = self._medium.get_index_of_refraction(x)
+        for i in range(len(path)-1):
+            xx = np.array([path[i][0], path[i][1], path[i][2]])
+            yy = np.array([path[i+1][0], path[i+1][1], path[i+1][2]])
+            center_of_segment = (xx+yy)/2
+            n = self._medium.get_index_of_refraction(center_of_segment)
```
colemanalan, line 57: I think it is always nice to have the iterator variables give some indication of what they are iterating over.
so like ieng or isol or icos. Whatever is appicable here

NuRadioMC/SignalProp/directRayTracing.py
MartinRavn: The function `get_path` makes a straight line from the shower to the observer and splits it into 1000 segments (points). The dimension of the returned `path` is [n_points,n_dimensions]. So in this line, we calculate the vector representing a segment (first point minus zeroth point).

MartinRavn: It would be better just to write `segment = path[1,:]-path[0,:]`, but I tried to keep the changes minimal to focus on the bugfix.

NuRadioMC/SignalProp/directRayTracing.py
MartinRavn: I tried to keep the changes minimal to only cover the bug. There are a few things I would like to change. Maybe I should just do it.

colemanalan: I think there is some precident to update the code to make it more readable too.

Maybe change `i` to `istep`.

I really don't see why this code does the casing to an array when you could just do `xx = path[istep]` and `yy=path[istep+1]`

I guess we can leave it for now.

MartinRavn: Exactly. I will push a more readable version of the code now.

MartinRavn: I got rid of the loop too.

colemanalan: I realize that no one really uses this, but this should actually make the code a lot faster :D

colemanalan (approved):

fschlueter:
Since you are cleaning up the code, would you mind adding some doc-string? Personally I think code get much more readable when adding some whitespaces where they belong bug I guess that might be just me.

colemanalan (approved):

colemanalan:
Looks good @MartinRavn. Once the checks pass (I assume they will) I will merge this

PR #668, opened 2024-04-04
Skip TBB files from remote stations on read-in

MijnheerD:
*Issue being addressed*

When globbing over all TBB files in the directory, it is possible for a bunch of remote station files to be included in the `all_tbb_files` list. When doing the pattern matching
```python
re.findall(r"CS\d\d\d", tbb_filename)
```
on those files, the result is an empty list. When then calling the first element, the reader crashes.

*Proposed solution*

I changed the code to either take the first element of the resulting list, or default to None in case the list is empty. If the `station_name` is `None`, the file is assumed to be from a remote station and skipped. Is this desired behavior? Should we explicitly check that the pattern "RS\d\d\d" does appear in the file name and do something else with it?


MijnheerD:
The changes from this PR have been included in #679

PR #669, opened 2024-04-04
Add LOFAR site longitude and latitude to Detector

MijnheerD:
Add the coordinates of LOFAR to the `get_site_coordinates()` function of the Detector


MijnheerD:
@cg-laser Is this the correct place to put these coordinates?

anelles (approved):
As long as you know that LOFAR is really there, all good 😉

PR #670, opened 2024-04-10
Add support for other sky models in channelGalaticNoiseAdder

MijnheerD:
Issue being addressed:

The default channelGalacticNoiseAdder only supports `pygdsm` . This PR wraps the [radiocalibrationtoolkit](https://github.com/F-Tomas/radiocalibrationtoolkit) from T. Fodran such that all the sky models from that package can be used.

### DIFF
--- a/NuRadioReco/modules/channelGalacticNoiseAdder.py
+++ b/NuRadioReco/modules/channelGalacticNoiseAdder.py
@@ -1,13 +1,22 @@
 from NuRadioReco.utilities import units, fft, ice, geometryUtilities
 from NuRadioReco.modules.base.module import register_run
 import NuRadioReco.framework.channel
+import NuRadioReco.framework.sim_station
 import NuRadioReco.detector.antennapattern
 import logging
 import numpy as np
 import matplotlib.pyplot as plt
 import scipy.constants
 import scipy.interpolate
-import pygdsm
+from radiocalibrationtoolkit import * #Documentation: https://github.com/F-Tomas/radiocalibrationtoolkit/tree/main contains SSM, GMOSS, ULSA
+from pygdsm import (
+    GlobalSkyModel2016,
+    GlobalSkyModel,
+    LowFrequencySkyModel,
+    HaslamSkyModel,
+)
+
+from pylfmap import LFmap #Documentation: https://github.com/F-Tomas/pylfmap needs cfitsio installation
 import healpy
 import astropy.coordinates
 import astropy.units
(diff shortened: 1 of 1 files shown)


MijnheerD:
@karenterveer Some of the skymodels are not imported right now, because the line is commented out. Maybe you can try to import `radiocalibrationtoolkit` in a `try` block, with a warning if it fails, to make the module completely usable out of the box?

karenterveer:
@MijnheerD added try-except statements. The module should work without additional installations now. It defaults back to using the standard sky model in case a model that is unavailable is chosen.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -8,20 +8,31 @@
 import matplotlib.pyplot as plt
 import scipy.constants
 import scipy.interpolate
-#from radiocalibrationtoolkit import * #Documentation: https://github.com/F-Tomas/radiocalibrationtoolkit/tree/main contains SSM, GMOSS, ULSA
+
+logger = logging.getLogger('channelGalacticNoiseAdder')
```
MijnheerD, line 12: Please use the NRR logger naming scheme: "NuRadioReco.channelGalacticNoiseAdder"

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -87,8 +98,13 @@ def begin(
         
         # initialise sky model
         if skymodel == 'lfmap':
-            sky_model = LFmap()
-            print("Using LFmap as sky model")
+            try:
+                sky_model = LFmap()
+                print("Using LFmap as sky model")
```
MijnheerD, line 103: Since you initialised a logger, why not replace all print statements with a proper logger.error() call? ;)

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -87,8 +98,13 @@ def begin(
...
+            except:
+                sky_model = GlobalSkyModel(freq_unit="MHz")
+                print(f"{skymodel} import not found. Defaulting to GSM2008 as sky model.")
+
         if skymodel == 'lfss':
```
MijnheerD, line 108: Also, if there is a reason for not using `elif`s here? This speeds up performance (currently every if statement is checked, but this is unnecessary I believe)

MijnheerD (commented):
One last suggestion: instead of having a try-except block in every if statement, I think it would be cleaner to `try` the if-block and have one `except` block. This will make it easier to change the fallback in the future.

karenterveer (approved):
Looks good to me!

PR #672, opened 2024-04-16
Extend functionality of __call__ method of response class

fschlueter:


sjoerd-bouma (approved):
I think sometimes in your design philosophy you are closer to the 'more features' end of the scale than 'simplicity of use and understanding' (see also: [unix](https://en.wikipedia.org/wiki/Unix_philosophy)) than I would be. But it is not hard to see the use cases for this (I might have gone with `invert_selection` rather than `blacklist` for the new kwarg), so I am happy to be overruled on this one.

anelles (approved):
Seems like a useful addition. I didn't test this, but code looks good.

PR #673, opened 2024-04-17
Unify download for antenna models / proposal tables / shower library with additional fallback servers

shallmann:
* removed code duplication by adding download function to `NuRadioReco.utilities.dataservers.py`
* Added Chicago server as backup. Should simplify changing/adding/removing servers in the future
* Option to rank servers, not enabled by default
* Improves exception handling printout (in the past, people asked if the antenna model was wrongly spelled / server down, when in fact they didn't have internet connection)

Things one could discuss:
* Do we want to rank servers by proximity/response time by default? I added two options, although the ranking by timezone requires additional python packages? (@cozzyd am I missing a much simpler solution than the function I added in `dataservers.py`?
* Are pretabulated ARZ tables still used/useful: ```./NuRadioMC/SignalGen/ARZ/ARZ.py:            URL = 'http://arianna.ps.uci.edu/~arianna/data/ce_shower_library/ARZ_library_v{:d}.{:d}.pkl'.format(*self._version)``` I think this server does not exist any more?

### DIFF
--- a/NuRadioReco/utilities/dataservers.py
+++ b/NuRadioReco/utilities/dataservers.py
@@ -1,7 +1,27 @@
 import requests
 import os
-
+import filelock
 import logging
+  
+  lock = filelock.FileLock("my_lock_file.lock")
+  
+  # This is the easiest way to use the file lock. Note, that the FileLock
+  # object blocks until the lock can be acquired.
+  
+  if not os.path.exists("my_lock_file"):
+      print("waiting for my_lock_file download")
+      with lock:
+          if not os.path.exists("my_lock_file"):
+              time.sleep(20)
+--            f = open("my_lock_file", "w")
+              f.write("done")
+              f.close()
+              print("Download complete")
+  
+          else:
+              print("Already downloaded")
+  print("using file...")
+
 logger = logging.getLogger('NuRadioReco.dataservers')
 
 dataservers = ["https://rnog-data.zeuthen.desy.de", "https://rno-g.uchicago.edu/data/desy-mirror"]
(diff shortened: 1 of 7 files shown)


NuRadioReco/utilities/dataservers.py
```diff
@@ -0,0 +1,115 @@
...
+logger = logging.getLogger('NuRadioReco.dataservers')
+
+dataservers = ["https://rnog-data.zeuthen.desy.de", "https://rno-g.uchicago.edu/data/desy-mirror"]
+
+def get_available_dataservers_by_responsetime(dataservers=dataservers):
+    """ requests a small index file from the list of dataservers and returns a list of responsive ones ordered by elapsed time """
+    response_times = []
+    available_dataservers = []
+
+    for dataserver in dataservers:
```
cozzyd, line 17: it could make sense to parallelize this, though in practice it probably doesn't matter very much with just two servers...

NuRadioReco/utilities/dataservers.py
```diff
@@ -0,0 +1,115 @@
+import requests
+import os
+import filelock
```
cozzyd, line 3: filelock claims it requires python >= 3.8, is that a problem?

NuRadioReco/utilities/dataservers.py
```diff
@@ -0,0 +1,115 @@
...
+
+    lockfile = target_path+".lock"
+    lock = filelock.FileLock(lockfile)
+   
+    logger.warning(f"Assuring no other process is downloading. Will wait until {lockfile} is unlocked.")
+    with lock:
+        if os.path.isfile(target_path):
+            logger.warning(f"{target_path} already exists. Maybe download was already completed by another instance?")
+            return
+        elif unpack_tarball and (len(glob(os.path.dirname(target_path) + "/*.dat")) > 0): #just check if any .dat files present (similar to NuRadioProposal.py)
```
cozzyd, line 68: are the data files ever updated? Or will the name get changed in that case?

NuRadioReco/utilities/dataservers.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if requests_status != requests.codes["ok"]:
+            logger.error(f"error in download of file {target_path}. Tried all servers in {dataservers} without success.")
+            raise IOError
+
+        with open(target_path, "wb") as code:
+            code.write(r.content)
+        logger.warning("...download finished.")
+
+        if unpack_tarball and target_path.endswith(".tar.gz"):
```
cozzyd, line 111: FYI unpack archive supports additional archive formats. Not sure if it would make sense to support .tar.xz to save some bandwidth in some cases?

NuRadioReco/utilities/dataservers.py
shallmann: The download function just gets called if the file needs download (either because the hash sum is not the latest version or the file is not there. The `if`/`elif` here just catches the case when 1000 jobs are started and the first that aquired the lock did already perform the download...

NuRadioReco/utilities/dataservers.py
shallmann: Good, but I don't think this is really needed. the proposal tarballs are the only tarballed files, and we chose to have them as tar.gz... they are not so big, and you only need to download once for a NuRadioMC installation.

NuRadioReco/utilities/dataservers.py
shallmann: Well the tests download without problems. We require 3.6 for NuRadioMC currently. Do you know alternatives? Or would you fix filelock to v3.4.1, when they dropped python 3.6 support?

NuRadioReco/utilities/dataservers.py
```diff
@@ -0,0 +1,115 @@
...
+    return ranked_dataservers
+
+def get_available_dataservers_by_timezone(dataservers=dataservers):
+    """ uses the server locations' timezones from the list of dataservers and returns the list of dataservers ordered by proximity """
+    import socket
+    import pytz
+    from datetime import datetime
+    from geolite2 import geolite2
+
+    geo = geolite2.reader()
```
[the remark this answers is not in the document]
shallmann: Ok, I removed the function entirely. Just having response time around is enough, if we are not enabling the option for now anyways (we agreed with @cg-laser that since download is only once for new files, we'd not request ordering, but leave DESY the "master")

NuRadioReco/utilities/dataservers.py
shallmann: yeah... right.. I don't expect that we will ever have >> 2 servers. I suggest to leave it as is.

shallmann:
@cozzyd @cg-laser I think the only remaining thing is if we need to change anything about the filelock version we are requiring, since they did drop python 3.6 support a while ago. But tests are running fine.

Otherwise I would like to merge this

NuRadioReco/utilities/dataservers.py
cozzyd: I guess pip should do the right thing and automatically download an older version if needed (and it looks like the old versions still exist on pypi), so I think this should be fine.

cozzyd:
If tests are running fine pip must be resolving the correct version automatically so I think it's fine to merge?

cg-laser (approved):

PR #674, opened 2024-04-18
Add LOFAR signal chain

karenterveer:
Added new module LOFAR/hardwareResponseIncorporator and detector/analog_components, along with data for LOFAR signal chain.


MijnheerD (commented):
Looks good! I am wondering however if you could easily add some headers to the txt files, to explain the contents a bit? Like what the columns stand for, and what the difference is between the files? If that messes to much with the reading in, we could also add a README file in the directory with the explanations.

karenterveer:
> Looks good! I am wondering however if you could easily add some headers to the txt files, to explain the contents a bit? Like what the columns stand for, and what the difference is between the files? If that messes to much with the reading in, we could also add a README file in the directory with the explanations.

I would have to change in the read-in for that, and these files are directly from coma. I suggest we add a README file.

MijnheerD (approved):

PR #675, opened 2024-04-27
Model of the diode trigger used for RNO-G

lpyras:
This PR contains a model of the diode trigger behaviour by filtering the trace in the bandpass of 80-200 MHz and then taking the power of the pulse and integrating it over 11ns. Since the trigger threshold is supposed to be noise riding, a multiple of the SNR value of the trace is used as a threshold. Usually the trigger operates in a 2/3 or 2/6 coincidence mode with a time window of 60 ns.


NuRadioReco/framework/trigger.py
```diff
@@ -394,9 +395,8 @@ def __init__(self, name, passband, order, threshold, number_of_coincidences=2,
         self._coinc_window = channel_coincidence_window
 
 class RNOGSurfaceTrigger(Trigger):
-    from NuRadioReco.utilities import units
     def __init__(self, name, threshold, number_of_coincidences=1,
-                 channel_coincidence_window=60*units.ns, channels=[13, 16, 19], temperature=250*units.kelvin, Vbias=2*units.volt):
+                 channel_coincidence_window=60, channels=[13, 16, 19], temperature=250, Vbias=2):
```
anelles, line 399: Why does this no longer have units?

NuRadioReco/framework/trigger.py
```diff
@@ -425,3 +425,30 @@ def __init__(self, name, threshold, number_of_coincidences=1,
...
+        name: string
+            unique name of the trigger
+        threshold_sigma: float
+            the threshold as a multiple of the noise standard deviation
+        int_window: float
+            the integration window
+        number_of_coincidences: int
+            the number of channels that need to fulfill the trigger condition
+            default: 1
+        channel_coincidence_window: float or None (default)
```
anelles, line 444: The doc here suggests for this to have units.

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,173 @@
...
+    def get_power_int_trace(self, trace, int_window):
+        """
+        calculats a power integration over a given time window
+
+        Parameters
+        ----------
+        trace: array of floats
+            the signal trace
+        window: float
+            the integration window
```
anelles, line 52: In what units? Time? Samples?

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,173 @@
...
+        filtered_trace = self.get_filtered_trace_trigger(trace)
+
+        power_int_trace = self.get_power_int_trace(filtered_trace, int_window)
+        threshold = self.get_threshold(filtered_trace, threshold_sigma, int_window)
+        triggered_bins = np.array(power_int_trace) > threshold
+        return (triggered_bins, threshold)
+
+    @register_run()
+    def run(self, evt, station, det, threshold_sigma=30, int_window=11, coinc_window=60, number_coincidences=2, triggered_channels=[13, 16, 19], trigger_name='radian_aux_trigger'):
+        """
```
anelles, line 95: Really "radian_aux_trigger"? No "t"?

anelles (changes requested):
Please check why this fails and take a look at my documentation comments.

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,173 @@
...
+                logger.warning('Channel has a trace_start_time that differs from '
+                               '        the other channels. The trigger simulator may not work properly')
+
+            (triggered_bins, threshold) = self.get_trigger_response(channel.get_trace(), threshold_sigma, int_window)
+            triggered_bins_channels.append(triggered_bins)
+
+            if True in triggered_bins:
+                channels_that_passed_trigger.append(channel.get_id())
+            
+            channel.set_parameter(chp.radiant_aux_trigger_threshold, threshold)
```
colemanalan, line 148: Again, I would set this in the `RadiantAUXTrigger` object unless there is a good reason to put this here

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,173 @@
...
+        
+        has_triggered, triggered_bins, triggered_times = get_majority_logic(triggered_bins_channels,
+                                                                            number_coincidences, coinc_window, dt)
+        trigger = RadiantAUXTrigger(trigger_name, threshold_sigma, int_window, number_coincidences, coinc_window, triggered_channels)
+        trigger.set_triggered_channels(channels_that_passed_trigger)
```
colemanalan, line 153: If you give `None` to your module, then this trigger also gets `None`. Maybe it would be better to build a list of channels that do consider for the trigger during this script (maybe you can do it in the for loop at L132)

colemanalan: Maybe at the beginning, you can right away do

```
considered_channels = copy(triggered_channels)
if triggered_channels is None:
    considered_channels = list(station.get_channel_ids())
```
and then use `considered_channels` everywhere here

anelles:
With 8 days on the new job, @lpyras now of course has time to fix this, right? 
Please don't wait too long. This is an essential module.

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,173 @@
...
+        self.begin()
+
+    def begin(self):
+        return
+    
+    def get_filtered_trace_trigger(self, trace):
+        n_samples = len(trace)
+        spectrum = fft.time2freq(trace, self.sampling_rate)
+        freqs = np.fft.rfftfreq(n_samples, 1 / self.sampling_rate)
+        filtering_out = np.where( (freqs < 0.08) | (freqs > 0.2)) # or!
```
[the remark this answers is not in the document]
lpyras: In general I would agree. Here, this is the trigger path and the trace is not stored for analysis. I parametrised it this way, so I would prefer to take the hard cut, otherwise I have to fit the trigger curves again.

NuRadioReco/framework/trigger.py
lpyras: Since no trigger here had units I thought I continue doing this. But I agree, with units is better!

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,173 @@
...
+from NuRadioReco.utilities import fft
+import matplotlib.pyplot as plt
+logger = logging.getLogger('radiant_aux_trigger')
+
+def tanh_func(x, b, c):
+    return 0.5*(np.tanh((x-b)/c) + 1)
+
+class triggerSimulator:
+    """
+    """
```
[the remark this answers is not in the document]
lpyras: will do!

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def __init__(self):
+        self.__t = 0
+        self.begin()
+
+    def begin(self):
+        return
+    
+    def get_filtered_trace_trigger(self, trace):
```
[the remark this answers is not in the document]
lpyras: added. :)

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,173 @@
...
+        filtered_spectrum [filtering_out] *= 0.0
+        filtered_trace = fft.freq2time(filtered_spectrum, self.sampling_rate)
+
+        return filtered_trace
+
+    def get_threshold(self, trace, threshold_sigma, int_window):
+        noise_samples = int(130 * self.sampling_rate)
+        noise = self.get_power_int_trace(trace[10:noise_samples], int_window)
+        threshold = threshold_sigma * np.std(noise)
+        return threshold
```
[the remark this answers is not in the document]
lpyras: it is within the first half of our trace, so that's the number. I agree, it need explanation. :)

NuRadioReco/modules/trigger/radiant_aux_trigger.py
lpyras: added :)

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,173 @@
...
+
+        Parameters
+        ----------
+        trace: array of floats
+            the signal trace
+
+        threshold: float
+            the threshold as value of sigma (a factor of snr)
+
+        window: float
```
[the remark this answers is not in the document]
lpyras: sorry, time :) I added it.

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,173 @@
...
+            channels ids that are triggered on, if None trigger will run on all channels
+        coinc_window: float
+            time window in which number_coincidences channels need to trigger
+        trigger_name: string
+            a unique name of this particular trigger
+        """
+        t = time.time()  # absolute time of system
+
+        self.sampling_rate = station.get_channel(det.get_channel_ids(station.get_id())[0]).get_sampling_rate()
+        dt = 1. / self.sampling_rate
```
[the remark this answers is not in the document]
lpyras: True, thanks! :)

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,173 @@
...
+                continue
+
+            if channel.get_trace_start_time() != channel_trace_start_time:
+                logger.warning('Channel has a trace_start_time that differs from '
+                               '        the other channels. The trigger simulator may not work properly')
+
+            (triggered_bins, threshold) = self.get_trigger_response(channel.get_trace(), threshold_sigma, int_window)
+            triggered_bins_channels.append(triggered_bins)
+
+            if True in triggered_bins:
```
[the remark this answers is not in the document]
lpyras: thanks, I change that :)

NuRadioReco/framework/parameters.py
```diff
@@ -58,6 +58,7 @@ class channelParameters(Enum):
     signal_ray_type = 16        #: type of the ray propagation path of the signal received by this channel. Options are direct, reflected and refracted
     signal_receiving_azimuth = 17   #: the azimuth angle of direction at which the radio signal arrived at the antenna
     block_offsets = 18 #: 'block' or pedestal offsets. See `NuRadioReco.modules.RNO_G.channelBlockOffsetFitter`
+    radiant_aux_trigger_threshold = 19  #: the threshold of the radiant aux trigger
```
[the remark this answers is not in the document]
lpyras: Since the trigger threshold is adjusted for each trace (mimicking the threshold servo in the field) the threshold differs for each channel and each event. I don't really know how to store that. This was one idea, but maybe that's overdoing it and we dont need to save it.

NuRadioReco/modules/trigger/radiant_aux_trigger.py
lpyras: Thanks, that is nice 👍

lpyras:
I finally cleaned this up and added units and docstrings. 
Two more questions: 

1. Should we keep calculating the threshold on single channel and event level since this is how the servoing in the field works, or should we calculate the threshold once and use the same for all channels, which is probably faster.

2. The threshold is given as a multiple of the noise rms. Should the voltage value (here power integrated, so (V^2/ns)) be stored? What would be the best option?

Thanks!

NuRadioReco/modules/trigger/radiant_aux_trigger.py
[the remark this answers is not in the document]
colemanalan: Sorry you still didn't address the first fixing. The parameter is `int_winow` instead of `window`. I am looking now and this also applies to the `get_power_int_trace` function

NuRadioReco/modules/trigger/radiant_aux_trigger.py
```diff
@@ -0,0 +1,204 @@
...
+        det: Detector
+            The detector description
+        number_coincidences: int
+            number of channels that are required in coincidence to trigger a station
+        triggered_channels: array of ints or None
+            channels ids that are triggered on, if None trigger will run on all channels
+        coinc_window: float
+            time window in which number_coincidences channels need to trigger
+        trigger_name: string
+            a unique name of this particular trigger
```
colemanalan, line 151: I think you are missing `threshold_sigma` and `int_window`. Also, `coinc_window` is out of order. Double check this and make sure that everything is included, and it is in the same order as the parameters in the function signature.

colemanalan (changes requested):
Just two more documentation things. But thanks for the updates already.
Functionally, this looks good

NuRadioReco/modules/trigger/radiant_aux_trigger.py
lpyras: Thanks for checking so carefully. I was really in a rush when I wrote this....

lpyras:
> I finally cleaned this up and added units and docstrings. Two more questions:
> 
> 1. Should we keep calculating the threshold on single channel and event level since this is how the servoing in the field works, or should we calculate the threshold once and use the same for all channels, which is probably faster.
> 2. The threshold is given as a multiple of the noise rms. Should the voltage value (here power integrated, so (V^2/ns)) be stored? What would be the best option?
> 
> Thanks!

@cg-laser any opinions on this?

colemanalan (approved):

lpyras:
@cg-laser , @fschlueter and me discussed that we abandon this PR, since the trigger is basically a power integrated trigger with a passband filter and a set threshold. I will change to power integration trigger such that it can take a frequency range as input. In the future, trigger and signal channels will be simulated differently which solves the storage problem. Also, a fixed threshold is assumed rather than a noise riding one.

PR #677, opened 2024-05-07
Fix for rnog_detector singleton

sjoerd-bouma:
The rnog_detector uses a singleton `Database`, meaning that re-initializing it (e.g. to create another database connection) does not work, as there was no way to pass the argument `create_new` to the Database. This is added by this PR.

### DIFF
--- a/NuRadioReco/detector/RNO_G/rnog_detector.py
+++ b/NuRadioReco/detector/RNO_G/rnog_detector.py
@@ -61,7 +61,7 @@ def _check_detector_time(method):
 class Detector():
     def __init__(self, database_connection='RNOG_test_public', log_level=logging.INFO, over_write_handset_values={},
                  database_time=None, always_query_entire_description=True, detector_file=None,
-                 select_stations=None):
+                 select_stations=None, create_new=False):
         """
 
         Parameters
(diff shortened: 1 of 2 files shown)


fschlueter:
What is the use-case for this. Why can't the database not be a real singelton?

sjoerd-bouma:
> What is the use-case for this. Why can't the database not be a real singelton?

The database is still a real singleton? I'm not sure if I understand your question. The problem is that it is currently impossible to create an rnog_detector with a different database connection once an rnog_detector has been initialized, because all arguments passed on to the database `__init__` are ignored (it is a singleton).

So in my use case, I created an rnog_detector with the default arguments (which goes to the test database rather than the live one), and couldn't create an rnog_detector with the correct database connection without restarting the Python kernel. 

Alternative solutions would be to make the rnog_detector a singleton (but then the database can not be a singleton, for the same reason as given above), or to allow the database to change the connection after initialization and implement the necessary logic inside rnog_detector.

fschlueter:
Maybe I do not understand what a singleton is. But once you have created a object of the class you should not be able to create another one within the same program right?

fschlueter:
You can not create now objects across projects?

sjoerd-bouma:
This does not work
```
rnog_detector = Detector(source='rnog_mongo') # oops, called the detector with the 'wrong' arguments...

# now I want to try again - this does not work, because `database_connection` is ignored
rnog_detector = Detector(source='rnog_mongo', database_connection='RNOG_public')
```

The only change I made is to add the `create_new` argument, which is passed on to `self.__db`, so I could add this to the last line above if I know I want to open a detector with a different database connection (e.g. in an interactive ipython session) without having to restart the kernel.

fschlueter:
Well the point of a singleton is to prevent exactly this right? It should not be possible

fschlueter:
Than it seems all singletons in NuRadio are actual no real singletons. Singletons are designed to be unique.I guess the issue is that you are using jupyter notebooks. That being said, I do not mind adding that since I now understand why we need fake singletons :D

fschlueter (approved):

PR #678, opened 2024-05-14
Update signalgen tests

anelles:
The tests in SignalGen were broken, because we changed the interfaces a long time ago and no one updated the examples. I don't claim that they are pretty now, but they work again. 
Apart from T03, which is a detailed comparison with HCRB, which we don't use often and which has a different interface. If anyone feels like fixing that, please go ahead.


fschlueter (approved):
I dont have time to check it and trust Anna here.

PR #679, opened 2024-05-16
LOFAR reader hotfixes

MijnheerD:
This PR contains a bunch of small hotfixes for the LOFAR branch:

- The TBB reader now raises an IndexError when the endpoint is off the end of a file
- TBB files from remote stations are now skipped when globbing over the directory (solves crashes involving events with only remote TBB files)
- The channel group IDs are now int instead of str (for consistency with Detector description)
- After reading in the channels, we now check whether both channels from the flagged group IDs are removed from the station
- The calibration delays can now be retrieved from the readLOFARData object as a dict, ensuring the calibration delays are associated to the correct channel ID

### DIFF
--- a/NuRadioReco/modules/LOFAR/stationGalacticCalibrator.py
+++ b/NuRadioReco/modules/LOFAR/stationGalacticCalibrator.py
@@ -96,7 +96,7 @@ class stationGalacticCalibrator:
             ),
         )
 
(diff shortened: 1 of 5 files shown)


anelles (approved):
Code looks good from my side, but I haven't tested this. 
I assume you tested that it works. 
Feel free to wait for LOFAR core team to review in independent cross-check, but you the ok to merge already.

PR #680, opened 2024-05-21
NuRadioMC 3.0

cg-laser:
This PR is a complete refactor of the NuRadioMC core (simulation.py), enabling long-awaited features such as:
* separate simulation of trigger channels
* non-consecutive channel ids
* enable resimulation from nur files
* adding additional antennas on-the-fly to existing simulations
* Exposing functions to calculate voltage traces from showers for stand-alone use

The key was to change the internal looping structure from 
image

to 

image

Changes will be discussed in upcoming InIceMC calls. More information will be provided to the PR later.


fschlueter:
@cg-laser it looks like the diff includes the changes of the birefringent2 branch which was merged. Can you try to remove it from the PR? Maybe by either merging or rebasing the current development branch in this one?

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -104,7 +104,9 @@ def run(self, evt, station, det):
         # for different cable delays
         times_min = []
         times_max = []
-        for iCh in det.get_channel_ids(sim_station_id):
+        if channel_ids is None:
+            channel_ids = det.get_channel_ids(sim_station_id)
```
fschlueter, line 108: this could be:
```
channel_ids = channel_ids or det.get_channel_ids(sim_station_id)
```
Since I know the syntax I start liking it a lot. But not so many people do so it could make the code harder to read for those.

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -104,7 +104,9 @@ def run(self, evt, station, det):
...
+        for iCh in channel_ids:
```
fschlueter, line 109: you could use the same variable as in the previous loop (`channel_id`)

NuRadioReco/modules/efieldToVoltageConverterPerEfield.py
```diff
@@ -8,6 +8,7 @@
 from NuRadioReco.utilities import units, ice, geometryUtilities
 from NuRadioReco.utilities import trace_utilities
```
fschlueter, line 9: lines could be merged

sjoerd-bouma:
One more question with regards to the new looping structure - how would you treat e.g. coherent noise sources (like the galaxy)? This is currently still WIP (see PR #591), but eventually we will probably want to have galactic noise that is coherent between different antennas (I'm not sure about deep detectors, but certainly for e.g. LOFAR this will be useful for calibration). How would that fit into the new looping structure?

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1031 @@
...
+channelAddCableDelay = NuRadioReco.modules.channelAddCableDelay.channelAddCableDelay()
+electricFieldResampler = NuRadioReco.modules.electricFieldResampler.electricFieldResampler()
+efieldToVoltageConverterPerEfield = NuRadioReco.modules.efieldToVoltageConverterPerEfield.efieldToVoltageConverterPerEfield()
+efieldToVoltageConverter = NuRadioReco.modules.efieldToVoltageConverter.efieldToVoltageConverter()
+channelSignalReconstructor = NuRadioReco.modules.channelSignalReconstructor.channelSignalReconstructor()
+channelResampler = NuRadioReco.modules.channelResampler.channelResampler()
+channelGenericNoiseAdder = NuRadioReco.modules.channelGenericNoiseAdder.channelGenericNoiseAdder()
+channelResampler = NuRadioReco.modules.channelResampler.channelResampler()
+eventWriter = NuRadioReco.modules.io.eventWriter.eventWriter()
+triggerTimeAdjuster = NuRadioReco.modules.triggerTimeAdjuster.triggerTimeAdjuster()
```
fschlueter, line 63: Personally I am not a fan of calling the object the same as the module/class. If we would change that we could also simplify the module imports.

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1031 @@
...
-from NuRadioReco.utilities import geometryUtilities as geo_utl
-from NuRadioReco.framework.parameters import channelParameters as chp
 from NuRadioReco.framework.parameters import electricFieldParameters as efp
 from NuRadioReco.framework.parameters import showerParameters as shp
+from NuRadioReco.framework.parameters import channelParameters as chp
 # parameters describing simulated Monte Carlo particles
 from NuRadioReco.framework.parameters import particleParameters as simp
 from NuRadioReco.framework.parameters import emitterParameters as ep
 # parameters set in the event generator
 from NuRadioReco.framework.parameters import generatorAttributes as genattrs
```
fschlueter, line 47: those lines could be merged.

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1031 @@
...
+    evt : NuRadioReco.framework.event.Event
+        Event object containing all the showers/emitters and electric fields
+    det : Detector object
+        the detector description defining all channels.
+    config : dict
+        the NuRadioMC configuration dictionary (from the yaml file)
+    detector_simulation_filter_amp: function (optional)
+        a function that applies the filter and amplifier response to the electric field
+        the arguments to the function are (event, station, detector)
+        if not provided, the function `detector_simulation_part1` needs to be provided.
```
fschlueter, line 535: here and in the following it should be `_part2` and not `_part1`. Also these names are a bit confusing. Can we find better names?

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1031 @@
...
+                                    det.get_relative_position(sim_station.get_id(), channel_id))
+        trace = np.zeros(n_samples)
+        trace[n_samples // 2] = 100 * units.V  # set a signal that should satisfy any trigger and speedup cuts
+        trace[n_samples // 2 + 1] = -100 * units.V
+        electric_field.set_trace(np.array([np.zeros(n_samples), trace, trace]), 1. / dt)
+        electric_field.set_trace_start_time(0)
+        electric_field[efp.azimuth] = 0
+        electric_field[efp.zenith] = 100 * units.deg
+        electric_field[efp.ray_path_type] = 0
+        sim_station.add_electric_field(electric_field)
```
fschlueter, line 625: Why is that necessary?

fschlueter:
One general comment / question. I appreciated that Christoph's solution spread the code over several files. In my opinion that increases readability (of course there can be a to much files), instead of having one file with +1500 lines.

NuRadioReco/detector/antennapattern.py
```diff
@@ -575,11 +575,11 @@ def get_pickle_antenna_response(path):
             antenna_hashs = json.load(fin)
             if os.path.basename(path) in antenna_hashs.keys():
                 if sha1.hexdigest() != antenna_hashs[os.path.basename(path)]:
-                    logger.warning("antenna model {} has changed on the server. downloading newest version...".format(
+                    logger.status("antenna model {} has changed on the server. downloading newest version...".format(
                         os.path.basename(path)))
                     download_file = True
             else:
-                logger.warning("no hash sum of {} available, skipping up-to-date check".format(os.path.basename(path)))
+                logger.status("no hash sum of {} available, skipping up-to-date check".format(os.path.basename(path)))
```
fschlueter, line 582: I think this is actually worth a warning?

NuRadioReco/framework/base_station.py
```diff
@@ -198,6 +198,16 @@ def set_electric_fields(self, electric_fields):
...
         return self._electric_fields
 
+    def get_electric_field_ids(self):
+        """
+        returns a list with the electric field IDs of all simElectricFields of the simStation
+        """
+        efield_ids = []
+        for efield in self._electric_fields:
+            efield_ids.append(efield.get_unique_identifier())
+        efield_ids.sort()
```
fschlueter, line 208: I would not necessarily expect this sort here. Maybe add it to the information to the doc string

NuRadioReco/framework/base_station.py
```diff
@@ -322,3 +332,19 @@ def deserialize(self, data_pkl):
                 self.set_station_time(data['_station_time'])
 
         self._particle_type = data['_particle_type']
+
+
+    def __add__(self, x):
```
fschlueter, line 337: doc string missing. Is it intuitive to implement such a operator? Maybe we should implement a `merge` or `combine` function

NuRadioReco/framework/event.py
```diff
@@ -376,8 +382,9 @@ def get_sim_showers(self):
         """
         Get an iterator over all simulated showers in the event
         """
-        for shower in self.__sim_showers.values():
-            yield shower
+        return self.__sim_showers.values()
+        # for shower in self.__sim_showers.values():
+        #     yield shower
```
fschlueter, line 387: change the doc-string. Typically this kind of functions returns an iterator right? Why changing that for this one?

NuRadioMC/simulation/simulation.py
```diff
@@ -147,112 +1094,103 @@ def __init__(self, inputfilename,
...
         if det is None:
-            logger.status("Detectorfile {}".format(os.path.abspath(self._detectorfile)))
-            kwargs = dict(json_filename=self._detectorfile, default_station=default_detector_station,
+            logger.status("Detectorfile {}".format(os.path.abspath(detectorfile)))
+            kwargs = dict(json_filename=detectorfile, default_station=default_detector_station,
                               default_channel=default_detector_channel, antenna_by_depth=False)
             kwargs.update(det_kwargs)
             self._det = detector.Detector(**kwargs)
         else:
             self._det = det
```
fschlueter, line 1151: Can we break backwards compatibility? If so I am opting for removing the initialisation of the detector object from this class. The object should be passed to the class and initialized in the top level script. This also forces the user to think about it a bit more.

NuRadioMC/simulation/simulation.py
cg-laser: what is your suggestion?

fschlueter: same names but in snake_case? `channelAddCableDelay` -> `channel_add_cable_delay` or something similar like `channel_cable_delay_adder`

cg-laser:
> One more question with regards to the new looping structure - how would you treat e.g. coherent noise sources (like the galaxy)? This is currently still WIP (see PR #591), but eventually we will probably want to have galactic noise that is coherent between different antennas (I'm not sure about deep detectors, but certainly for e.g. LOFAR this will be useful for calibration). How would that fit into the new looping structure?

Good question. This is important to consider. I think I would implement an internal buffer of the genericNoiseAdder. As long as the `event_group_id` is the same, the same noise is being used and just delayed to accommodate the difference in propagation times to the antennas. (Although I'm not sure how you would do this in practice as the emission comes from all directions but you must have figured this out. )

NuRadioReco/modules/efieldToVoltageConverter.py
cg-laser: fixed

NuRadioMC/simulation/simulation.py
cg-laser: I resolve this for now. This is a general question that is not directly related to the refactor branch. This can be changed everywhere in a new PR.

NuRadioMC/simulation/simulation.py
cg-laser: fixed

sjoerd-bouma (commented):
Minor comments from a while ago that I clearly forgot to submit - if/once I have time to have a closer look at the code I will.

NuRadioMC/simulation/simulation.py
fschlueter: Than it gets never done.

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+        vertex_positions = []
+        shower_energies = []
+        for i, shower in enumerate(showers):
+            vertex_positions.append(shower.get_parameter(shp.vertex))
+            shower_energies.append(shower.get_parameter(shp.energy))
+        vertex_positions = np.array(vertex_positions)
+        shower_energies = np.array(shower_energies)
+        vertex_distances = np.linalg.norm(vertex_positions - vertex_positions[0], axis=1)
+        time_logger.stop_time('distance cut')
+    logger.debug("Calculating electric field for station %d , channel %d from list of showers", sid, cid)
```
fschlueter, line 138: I would prefer modern string formatting for more readability

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+    p = propagator # shorthand for more compact coding
```
fschlueter, line 139: this is a bad idea

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+    propagator : propagator object
+        the propagator that should be used to calculate the electric field from emitter to receiver, typically a ray tracer
+    medium: medium object
+        the medium in which the electric field is propagating, typically ice
+    config : dict
+        the NuRadioMC configuration dictionary (from the yaml file)
+    time_logger: time_logger object
+        the time logger to be used for the simulation
+    min_efield_amplitude: float
+        speedup cut: the minimum electric field amplitude, if all efields from all showers are belwo this threshold value
```
fschlueter, line 118: typo

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+    sim_station.set_candidate(False)
+    if min_efield_amplitude is None:
+        sim_station.set_candidate(True)
+    sim_station.set_is_neutrino()  # naming not ideal, but this function defines in-ice emission (compared to in-air emission from air showers)
+
+    x2 = det.get_relative_position(sid, cid) + det.get_absolute_position(sid)
+    dt = 1. / config['sampling_rate']
+    # rescale the number of samples to the internal (higher) sampling rate used in the simulation
+    n_samples = det.get_number_of_samples(sid, cid) / det.get_sampling_frequency(sid, cid) / dt
+    n_samples = int(np.ceil(n_samples / 2.) * 2)  # round to nearest even integer
```
fschlueter, line 151: [`numpy.rint`](https://numpy.org/doc/stable/reference/generated/numpy.rint.html#numpy-rint) can be used?

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+            shower_energy_sum = np.sum(shower_energies[mask_shower_sum])
+            if np.linalg.norm(x1 - x2) > distance_cut(shower_energy_sum):
+                # logger.warning(f"shower {shower.get_id()} is too far away ({np.linalg.norm(x1 - x2)/units.km:.2f} > {distance_cut(shower_energy_sum)/units.km:.2f}) from station {sid} channel {cid}, skipping shower")
+                time_logger.stop_time('distance cut')
+                continue
+            # logger.warning(f"shower {shower.get_id()} is close enough ({np.linalg.norm(x1 - x2)/units.km:.2f} < {distance_cut(shower_energy_sum)/units.km:.2f}) to station {sid} channel {cid}, continuing with shower")
+            time_logger.stop_time('distance cut')
+        time_logger.start_time('ray tracing')
+        logger.debug(f"Calculating electric field for shower {shower.get_id()} and station {sid}, channel {cid}")
+        shower_axis = -1 * shower.get_axis() # We need the propagation direction here, so we multiply the shower axis with '-1'
```
fschlueter, line 167: I would rename the variable to what it is `shower_direction`

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+        n_index = medium.get_index_of_refraction(x1)
+        cherenkov_angle = np.arccos(1. / n_index)
+
+        p.set_start_and_end_point(x1, x2)
+        p.use_optional_function('set_shower_axis', shower_axis)
+        if config['speedup']['redo_raytracing']:  # check if raytracing was already performed
+            pass
+            # TODO: initiatlize ray tracer with existing results if available
```
fschlueter, line 175: raise a NotImplementedError ?

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+        p.find_solutions()
+        if not p.has_solution():
+            logger.debug(f"shower {shower.get_id()} and station {sid}, channel {cid} from {x1} to {x2} does not have any ray tracing solution")
+            continue
+
+        n = p.get_number_of_solutions()
```
fschlueter, line 181: I am not a fan of single letter variables.

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+        logger.debug(f"found {n} solutions for shower {shower.get_id()} and station {sid}, channel {cid} from {x1} to {x2}")
+        delta_Cs = np.zeros(n)
+        viewing_angles = np.zeros(n)
+        # loop through all ray tracing solution
+        for iS in range(p.get_number_of_solutions()):
```
fschlueter, line 186: `range(n)`

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+        for iS in range(n): # loop through all ray tracing solution
+            time_logger.start_time('ray tracing (time)')
+            # skip individual channels where the viewing angle difference is too large
+            # discard event if delta_C (angle off cherenkov cone) is too large
+            if np.abs(delta_Cs[iS]) > config['speedup']['delta_C_cut']:
+                logger.debug('delta_C too large, ray tracing solution unlikely to be observed, skipping ray tracing solution')
+                continue
+            # TODO: Fill with previous values if RT was already performed
+            R = p.get_path_length(iS)  # calculate path length
+            T = p.get_travel_time(iS)  # calculate travel time
```
fschlueter, line 203: no capital letters for variables. no single letter variables

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+        this function gives the user the full flexibility to implement all processing
+        arguments to the function are (event, station, detector)
+    time_logger: time_logger object
+        the time logger to be used for the simulation
+    channel_ids: list of ints
+        the channel ids for which the detector response should be calculated. If None, all channels are used.
+
+    Returns nothing. The Channels are added to the Station object.
+    """
+    if time_logger is not None: time_logger.start_time('detector response')
```
fschlueter, line 556: please not this syntax

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+        the time of the event to be simulated
+    detector_simulation_part1: function (optional)
+        this function gives the user the full flexibility to implement all processing
+        arguments to the function are (event, station, detector)
+
+    Returns nothing. The SimChannels are added to the SimStation object.
+    """
+    if time_logger is not None: time_logger.start_time('detector response (sim)')
+    if evt is None:
+        evt = NuRadioReco.framework.event.Event(0, 0)
```
fschlueter, line 487: `evt = evt or NuRadioReco.framework.event.Event(0, 0)`

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+    config_file_default = os.path.join(os.path.dirname(__file__), 'config_default.yaml')
+    logger.status('reading default config from %s', config_file_default)
+    with open(config_file_default, 'r', encoding="utf-8") as ymlfile:
+        cfg = yaml.load(ymlfile, Loader=yaml.FullLoader)
+    if config_file is not None:
+        logger.status('reading local config overrides from %s', config_file)
+        with open(config_file, 'r', encoding="utf-8") as ymlfile:
+            local_config = yaml.load(ymlfile, Loader=yaml.FullLoader)
+            new_cfg = merge_config(local_config, cfg)
+            cfg = new_cfg
```
fschlueter, line 761: is the second line necessary

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+    if time_logger is not None: time_logger.start_time('group into events')
+    event_group_id = event_group.get_run_number()
+    start_times = []
+    channel_identifiers = []
+    for channel in station.get_sim_station().iter_channels():
+        channel_identifiers.append(channel.get_unique_identifier())
+        start_times.append(channel.get_trace_start_time())
+    start_times = np.array(start_times)
+    start_times_sort = np.argsort(start_times)
+    delta_start_times = start_times[start_times_sort][1:] - start_times[start_times_sort][:-1]  # this array is sorted in time
```
fschlueter, line 901: `np.diff`

NuRadioMC/simulation/simulation.py
```diff
@@ -1,85 +1,1032 @@
...
+    fin = {}
+    fin_stations = {}
+    fin_attrs = {}
+    for key, value in fin_hdf5.items():
+        if isinstance(value, h5py._hl.group.Group):
+            fin_stations[key] = {}
+            for key2, value2 in value.items():
+                fin_stations[key][key2] = np.array(value2)
+        else:
+            if len(value) and type(value[0]) == bytes:
```
fschlueter, line 996: isinstance()

NuRadioMC/simulation/simulation.py
```diff
@@ -147,112 +1095,134 @@ def __init__(self, inputfilename,
...
+        if hasattr(self, '_detector_simulation_trigger'):
+            self.detector_simulation_trigger = self._detector_simulation_trigger
+        if hasattr(self, '_detector_simulation_part1'):
+            self.detector_simulation_part1 = self._detector_simulation_part1
+        if hasattr(self, '_detector_simulation_part2'):
+            self.detector_simulation_part2 = self._detector_simulation_part2
+        if((self.detector_simulation_filter_amp is None) ^ (self.detector_simulation_trigger is None)):
+            logger.error("Please provide both detector_simulation_filter_amp and detector_simulation_trigger or detector_simulation_part1 and detector_simulation_part2")
+            raise ValueError("Please provide both detector_simulation_filter_amp and detector_simulation_trigger or detector_simulation_part1 and detector_simulation_part2")
+        if((self.detector_simulation_part1 is None) ^ (self.detector_simulation_part2 is None)):
```
fschlueter, line 1160: first time this syntax of using `(..)` in if conditions is used.

NuRadioMC/simulation/simulation.py
```diff
@@ -147,112 +1095,134 @@ def __init__(self, inputfilename,
...
+            n_frequencies_integration=int(self._config['propagation']['n_freq']),
+            n_reflections=int(self._config['propagation']['n_reflections']),
+            config=self._config,
+            detector=self._det
+        )
 
         self._station_ids = self._det.get_station_ids()
         self._event_ids_counter = {}
         for station_id in self._station_ids:
             self._event_ids_counter[station_id] = -1  # we initialize with -1 becaue we increment the counter before we use it the first time
```
fschlueter, line 1201: `self._event_ids_counter = {station_id: -1 for station_id in self._station_ids}`

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
-                            self._raytracer.set_solution(ray_tracing_solution)
+                # group events into events based on signal arrival times
+                events = group_into_events(station, event_group, particle_mode, self._config['split_event_time_diff'], time_logger=self.__time_logger)
+
+                evt_group_triggered = False
+                for evt in events:
+                    station = evt.get_station()
+                    apply_det_response(evt, self._det, self._config, self.detector_simulation_filter_amp,
+                                       bool(self._config['noise']),
+                                       self._Vrms_efield_per_channel, self._integrated_channel_response,
```
fschlueter, line 1509: I am not sure what is correct but in the previous version we used here `self._Vrms_per_channel` instead of `self._Vrms_efield_per_channel`?

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
         # calculate bary centers of station
-        self._station_barycenter = np.zeros((len(self._station_ids), 3))
+        station_barycenter = np.zeros((len(self._station_ids), 3))
         for iSt, station_id in enumerate(self._station_ids):
             pos = []
-            for channel_id in range(self._det.get_number_of_channels(station_id)):
+            for channel_id in self._det.get_channel_ids(station_id):
                 pos.append(self._det.get_relative_position(station_id, channel_id))
-            self._station_barycenter[iSt] = np.mean(np.array(pos), axis=0) + self._det.get_absolute_position(station_id)
+            station_barycenter[iSt] = np.mean(np.array(pos), axis=0) + self._det.get_absolute_position(station_id)
```
fschlueter, line 1394: comprehensive lists?

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
-                    if self._cfg['signal']['shower_type'] == "had":
-                        if self._fin['shower_type'][self._shower_index] != "had":
-                            continue
+                    # applies the detector response to the electric fields (the antennas are defined
+                    # in the json detector description file)
+                    apply_det_response_sim(sim_station, self._det, self._config, self.detector_simulation_filter_amp,
+                                           event_time=self._evt_time, time_logger=self.__time_logger,
+                                           detector_simulation_part1=self.detector_simulation_part1)
+                    logger.debug(f"adding sim_station to station {sid} for event group {event_group.get_run_number()}, channel {channel_id}")
+                    station.add_sim_station(sim_station)  # this will add the channels and efields to the existing sim_station object
```
fschlueter, line 1491: I do not really like this "interface". It is vert unintuitive and actually needed changes in the framework. Why don't we return a list of efields from `calculate_sim_efield` and add it to the sim_station without creating this "dummy" sim_station objects?

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
+                # end channel loop, skip to next event group if all signals are empty (due to speedup cuts)
+                sim_station = station.get_sim_station()  # needed to get sim_station object containing all channels and not just the last one.
+                if len(sim_station.get_electric_fields()) == 0:
```
fschlueter, line 1494: if not len(...) ?

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
+                event_group.set_station(station)
+
+                # we allow to first only simualte trigger channels. As the trigger channels might be different per station, 
+                # we need to determine the channels to simulate first per station
+                channel_ids = self._det.get_channel_ids(sid)
+                if self.__trigger_channels is not None:
+                    if isinstance(self.__trigger_channels, dict):
+                        channel_ids = self.__trigger_channels[sid]
+                    else:
+                        channel_ids = self.__trigger_channels
```
fschlueter, line 1458: I would add a warning here if trigger_channels is `None`. Because this is a new feature people might forget to specify it and will run everything as trigger channel. To avoid the warning (if it is intentional) one can pass `[]`.

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
+                        continue
+                    triggerTimeAdjuster.run(evt, station, self._det)
+                    evt_group_triggered = True
+                    output_buffer[sid][evt.get_id()] = evt
+                # end event loop
+                # now simulate non-trigger channels
+                # we loop through all non-trigger channels and simulate the electric fields for all showers. 
+                # then we apply the detector response to the electric fields and find the event in which they will be visible in the readout window
+                non_trigger_channels = list(set(self._det.get_channel_ids(sid)) - set(channel_ids))
+                if (len(non_trigger_channels) > 0):
```
fschlueter, line 1530: `if len(non_trigger_channels)`

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
-                            # discard event if delta_C (angle off cherenkov cone) is too large
-                            if np.abs(delta_Cs[iS]) > self._cfg['speedup']['delta_C_cut']:
-                                logger.debug('delta_C too large, ray tracing solution unlikely to be observed, skipping event')
+                        # applies the detector response to the electric fields (the antennas are defined
+                        # in the json detector description file)
+                        apply_det_response_sim(sim_station, self._det, self._config, self.detector_simulation_filter_amp,
+                                            event_time=self._evt_time, time_logger=self.__time_logger,
+                                            detector_simulation_part1=self.detector_simulation_part1)
+                        logger.debug(f"adding sim_station to station {sid} for event group {event_group.get_run_number()}, channel {channel_id}")
+                        station.add_sim_station(sim_station)  # this will add the channels and efields to the existing sim_station object
```
fschlueter, line 1559: One could make this block optional (for the non-trigger-channels) (opt-out) to further reduce the computing time. Although I do not know if that is significant?

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
+                            for trigger in evt.get_station().get_triggers().values():
+                                if trigger.get_pre_trigger_times() is not None:
+                                    pre_trigger_time = trigger.get_pre_trigger_times()
+                                    trigger_time = trigger.get_trigger_time()
+                                    trigger_name = trigger.get_name()
+                                    break
+
+                            assert(trigger_time is not None)
+                            assert(trigger_name is not None)
+                            assert(trace_start_time == trigger_time)
```
fschlueter, line 1575: Not sure what exactly does this block but we should make sure in the trigger class that a trigger which triggered can never have a None name or None time

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
+
+                            for sim_channel in sim_station.get_channels_by_channel_id(channel_id):
+                                tt = sim_channel.get_times()
+                                t0 = tt[0]
+                                t1 = tt[-1]
+                                pre_trigger_time = triggerTimeAdjuster.get_pre_trigger_times(sim_channel.get_id())
```
fschlueter, line 1581: This overwrites the pre_trigger_time from the trigger object

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
+                            # determine the trigger that was used to determine the readout window
+                            trigger_name = None
+                            trigger_time = None
+                            pre_trigger_time = None
+                            for trigger in evt.get_station().get_triggers().values():
+                                if trigger.get_pre_trigger_times() is not None:
+                                    pre_trigger_time = trigger.get_pre_trigger_times()
+                                    trigger_time = trigger.get_trigger_time()
+                                    trigger_name = trigger.get_name()
+                                    break
```
fschlueter, line 1571: don't we have to select the min trigger time and not the trigger time of the first trigger with a not None pre trigger time?

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
+                                pre_trigger_time = triggerTimeAdjuster.get_pre_trigger_times(sim_channel.get_id())
+                                t0_readout = trigger_time - pre_trigger_time
+                                t1_readout = t0_readout + self._det.get_number_of_samples(sid, sim_channel.get_id()) * self._det.get_sampling_frequency(sid, sim_channel.get_id())
+
+                                # determine if the two intervals have any overlap
+                                if max(t0, t0_readout) < min(t1, t1_readout):
+                                    tmp_channel = NuRadioReco.framework.channel.Channel(sim_channel.get_id())
+                                    tmp_channel.set_trace(sim_channel.get_trace(), sim_channel.get_sampling_rate())
+                                    tmp_channel.set_trace_start_time(sim_channel.get_trace_start_time())
+                                    if(station.has_channel(sim_channel.get_id())):
```
fschlueter, line 1590: no brackets

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
+                                        channel = station.get_channel(sim_channel.get_id())
+                                        # logger.status(f"channel type {type(channel)}, sim_channel type {type(sim_channel)}")
+                                        combined_trace = channel + tmp_channel
+                                        channel.set_trace(combined_trace.get_trace(), combined_trace.get_sampling_rate())
+                                        channel.set_trace_start_time(combined_trace.get_trace_start_time())
+
+                                        # logger.status(f"channel type {type(channel)}")
```
fschlueter, line 1597: we could extent the "add_channel" interface or add a "combine_channel" interface.

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
+                            for sim_channel in sim_station.get_channels_by_channel_id(channel_id):
+                                tt = sim_channel.get_times()
+                                t0 = tt[0]
+                                t1 = tt[-1]
+                                pre_trigger_time = triggerTimeAdjuster.get_pre_trigger_times(sim_channel.get_id())
+                                t0_readout = trigger_time - pre_trigger_time
+                                t1_readout = t0_readout + self._det.get_number_of_samples(sid, sim_channel.get_id()) * self._det.get_sampling_frequency(sid, sim_channel.get_id())
+
+                                # determine if the two intervals have any overlap
+                                if max(t0, t0_readout) < min(t1, t1_readout):
```
fschlueter, line 1586: wont the triggerTimeAdjuster take care of this (later)

fschlueter: resp. should we just add the sim_channels and run the tirggerTimeAdjuster?

NuRadioMC/simulation/simulation.py
fschlueter: Also this function is called in the channel loop. Is that correct - efficient? One could first simulate the efields for all channels and than apply to detector response

NuRadioMC/simulation/simulation.py
```diff
@@ -147,112 +1095,134 @@ def __init__(self, inputfilename,
...
+        self.detector_simulation_filter_amp = None
+        self.detector_simulation_trigger = None
+        if hasattr(self, '_detector_simulation_filter_amp'):
+            self.detector_simulation_filter_amp = self._detector_simulation_filter_amp
+        if hasattr(self, '_detector_simulation_trigger'):
+            self.detector_simulation_trigger = self._detector_simulation_trigger
+        if hasattr(self, '_detector_simulation_part1'):
+            self.detector_simulation_part1 = self._detector_simulation_part1
+        if hasattr(self, '_detector_simulation_part2'):
+            self.detector_simulation_part2 = self._detector_simulation_part2
```
fschlueter, line 1156: more compact would be this oneliner:
```
self.detector_simulation_part1 = getattr(self, "_detector_simulation_part1", None)
```

fschlueter:
I get the following error when I run the new simulation:
```
(24, 4, 0) []
(24, 4, 1) []
(24, 5, 0) []
(24, 5, 1) []
(25, 4, 0) [24]
  File "/Volumes/software/rnog/Followup_Paper_100PeV/Followup_100PeV/C02_event_sim/simulate_rnog_100PeV.py", line 472, in <module>
    task({}, args.index, args.energy, det, args.config, output_filename, args.flavor, args.interaction_type, x0=args.x0, y0=args.y0, n_events_per_file=args.n_events_per_file)
  File "/Volumes/software/rnog/Followup_Paper_100PeV/Followup_100PeV/C02_event_sim/simulate_rnog_100PeV.py", line 378, in task
    n_trig = sim.run()
             ^^^^^^^^^
  File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1496, in run
    station.add_sim_station(sim_station)  # this will add the channels and efields to the existing sim_station object
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/felix/software/nuRadioMC/NuRadioReco/framework/station.py", line 24, in add_sim_station
    self.__sim_station = self.__sim_station + sim_station
                         ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
  File "/Users/felix/software/nuRadioMC/NuRadioReco/framework/sim_station.py", line 176, in __add__
    if channel.get_unique_identifier() in channel_ids:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
``` 
The first few lines is a print I added: 
```
            print(channel.get_unique_identifier(), channel_ids)
            if channel.get_unique_identifier() in channel_ids:
```
I am running a simulation with trigger channels 24 - 27 which are at the same position as non-trigger channels 0-3.

fschlueter:
Also I get this status:
```
STATUS - NuRadioMC.simulation - Simulating non-trigger channels for station 13
```
which make me thing I got a trigger (which I expect for at 1e20 event). But than in the end:
```
WARNING:NuRadioMC.HDF5OutputWriter:no events to save, not writing output file
WARNING - NuRadioMC.simulation - No events were triggered. Writing empty HDF5 output file.
simulation pass 994 with None events
```

NuRadioReco/framework/sim_station.py
```diff
@@ -129,3 +160,24 @@ def deserialize(self, data_pkl):
                 channel = NuRadioReco.framework.sim_channel.SimChannel(0, 0, 0)
                 channel.deserialize(channel_pkl)
                 self.add_channel(channel)
+
+    def __add__(self, x):
```
fschlueter, line 164: typically one passes the argument `other` to an operator. Makes it easier to read the code

fschlueter:
> I get the following error when I run the new simulation:
> 
> ```
> (24, 4, 0) []
> (24, 4, 1) []
> (24, 5, 0) []
> (24, 5, 1) []
> (25, 4, 0) [24]
> Traceback (most recent call last):
>   File "/Volumes/software/rnog/Followup_Paper_100PeV/Followup_100PeV/C02_event_sim/simulate_rnog_100PeV.py", line 472, in <module>
>     task({}, args.index, args.energy, det, args.config, output_filename, args.flavor, args.interaction_type, x0=args.x0, y0=args.y0, n_events_per_file=args.n_events_per_file)
>   File "/Volumes/software/rnog/Followup_Paper_100PeV/Followup_100PeV/C02_event_sim/simulate_rnog_100PeV.py", line 378, in task
>     n_trig = sim.run()
>              ^^^^^^^^^
>   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1496, in run
>     station.add_sim_station(sim_station)  # this will add the channels and efields to the existing sim_station object
>     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/felix/software/nuRadioMC/NuRadioReco/framework/station.py", line 24, in add_sim_station
>     self.__sim_station = self.__sim_station + sim_station
>                          ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
>   File "/Users/felix/software/nuRadioMC/NuRadioReco/framework/sim_station.py", line 176, in __add__
>     if channel.get_unique_identifier() in channel_ids:
>        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
> ```
> 
> The first few lines is a print I added:
> 
> ```
>             print(channel.get_unique_identifier(), channel_ids)
>             if channel.get_unique_identifier() in channel_ids:
> ```
> 
> I am running a simulation with trigger channels 24 - 27 which are at the same position as non-trigger channels 0-3.

one can fix the issue by using `get_unique_identifier()[0]` to get the channel_id. That should be relatively save because the unique identifier is also checked in the `add_channel` method

fschlueter:
A comment to the (time) logging. So far we also print the number of triggered events which I find very useful. Can that be implemented in the new time_logger class?

NuRadioMC/simulation/output_writer_hdf5.py
```diff
@@ -0,0 +1,528 @@
...
+            else:
+                dict_to_fill[key].append(value)
+        else:
+            if first_event:
+                dict_to_fill[key] = [value]
+            else:
+                if key not in dict_to_fill:
+                    logger.error(f"key {key} not in dict but not first event")
+                    raise KeyError(f"key {key} not in dict but not first event")
+                dict_to_fill[key].append(value)
```
fschlueter, line 92: Maybe using `defaultdict(list)` makes the code simpler and faster?

fschlueter:
I get the following error:
```
  File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1617, in run
    channel = station.get_channel(channel_id)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/felix/software/nuRadioMC/NuRadioReco/framework/station.py", line 41, in get_channel
    return self.__channels[channel_id]
           ~~~~~~~~~~~~~~~^^^^^^^^^^^^
KeyError: 6
```
This is before adding noise. My suspicion is that the channel does not exist because no ray_traceing_solution exist for this channel.

fschlueter:
> I get the following error:
> 
> ```
>   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1617, in run
>     channel = station.get_channel(channel_id)
>               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/felix/software/nuRadioMC/NuRadioReco/framework/station.py", line 41, in get_channel
>     return self.__channels[channel_id]
>            ~~~~~~~~~~~~~~~^^^^^^^^^^^^
> KeyError: 6
> ```
> 
> This is before adding noise. My suspicion is that the channel does not exist because no ray_traceing_solution exist for this channel.

that might help:
```
                            if station.has_channel(channel_id):
                                channel = station.get_channel(channel_id)
                            else:
                                channel = NuRadioReco.framework.channel.Channel(channel_id)
                                channel.set_trace(np.zeros(self._det.get_number_of_samples(sid, channel_id)), 1. / (self._config['sampling_rate']))
                                channel.set_trace_start_time(trigger_time)
```

NuRadioMC/simulation/simulation.py
fschlueter, line 1530: By default we only want to simulate non-triggered channels, right? How is that enforced? Looks like we simulate them anyway but than just not add them to the events/stations.

fschlueter:
Also getting this:
```
  File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1667, in run
    self._output_writer_hdf5.add_event_group(output_buffer)
  File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/output_writer_hdf5.py", line 126, in add_event_group
    assert all(np.atleast_1d(self._mout_groups_attributes[sid][key_hdf5] == np.array(channel_values))), f"station {sid} key {key_hdf5} is {self._mout_groups_attributes[sid][key_hdf5]}, but current channel is {np.array(channel_values)}"
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: operands could not be broadcast together with shapes (19,) (18,)
```

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
-                                                                                         pretty_time_delta(t_total), 1.e3 * t_total / self._n_showers,
-                                                                                         100 * input_time / t_total,
-                                                                                         100 * (rayTracingTime - askaryan_time) / t_total,
-                                                                                         100 * askaryan_time / t_total,
-                                                                                         100 * detSimTime / t_total,
-                                                                                         100 * outputTime / t_total,
-                                                                                         100 * weightTime / t_total))
-        triggered = remove_duplicate_triggers(self._mout['triggered'], self._fin['event_group_ids'])
-        n_triggered = np.sum(triggered)
-        return n_triggered
```
fschlueter, line 1134: the new code does not return the number of triggered events.

fschlueter:
> Also getting this:
> 
> ```
>   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1667, in run
>     self._output_writer_hdf5.add_event_group(output_buffer)
>   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/output_writer_hdf5.py", line 126, in add_event_group
>     assert all(np.atleast_1d(self._mout_groups_attributes[sid][key_hdf5] == np.array(channel_values))), f"station {sid} key {key_hdf5} is {self._mout_groups_attributes[sid][key_hdf5]}, but current channel is {np.array(channel_values)}"
>                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ValueError: operands could not be broadcast together with shapes (19,) (18,)
> ```

This does not happen always though.

fschlueter:
> > Also getting this:
> > ```
> >   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1667, in run
> >     self._output_writer_hdf5.add_event_group(output_buffer)
> >   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/output_writer_hdf5.py", line 126, in add_event_group
> >     assert all(np.atleast_1d(self._mout_groups_attributes[sid][key_hdf5] == np.array(channel_values))), f"station {sid} key {key_hdf5} is {self._mout_groups_attributes[sid][key_hdf5]}, but current channel is {np.array(channel_values)}"
> >                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> > ValueError: operands could not be broadcast together with shapes (19,) (18,)
> > ```
> 
> This does not happen always though.

I think the issue is again the same, that sometimes a `station` does not have the channel because no ray_tracing_solution

fschlueter:
> > I get the following error:
> > ```
> >   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1617, in run
> >     channel = station.get_channel(channel_id)
> >               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> >   File "/Users/felix/software/nuRadioMC/NuRadioReco/framework/station.py", line 41, in get_channel
> >     return self.__channels[channel_id]
> >            ~~~~~~~~~~~~~~~^^^^^^^^^^^^
> > KeyError: 6
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is before adding noise. My suspicion is that the channel does not exist because no ray_traceing_solution exist for this channel.
> 
> that might help:
> 
> ```
>                             if station.has_channel(channel_id):
>                                 channel = station.get_channel(channel_id)
>                             else:
>                                 channel = NuRadioReco.framework.channel.Channel(channel_id)
>                                 channel.set_trace(np.zeros(self._det.get_number_of_samples(sid, channel_id)), 1. / (self._config['sampling_rate']))
>                                 channel.set_trace_start_time(trigger_time)
> ```

ah here a `station.add_channel(channel)` is missing in the else block. This should also fix the other problem in the hdf5_wirter

NuRadioMC/simulation/simulation.py
cg-laser: this is actually the correct way to do it. WIth this notation, the string is only compiled if the message is logged. f-strings will always be compiled.

NuRadioMC/simulation/simulation.py
cg-laser: I don't understand this comment.

NuRadioMC/simulation/simulation.py
cg-laser: Thanks for noticing that. I think this was actually a bug but wasn't caught by any unit test because we never have any amplification and therefore the two variables were the same.

NuRadioMC/simulation/simulation.py
cg-laser: what do you mean?

NuRadioMC/simulation/simulation.py
fschlueter: Interesting I did not know that. Thanks

NuRadioMC/simulation/simulation.py
fschlueter: Before, the syntax `if statement:` is used. This and in the following sometimes this syntax is used `if(statement):` is used.

NuRadioMC/simulation/simulation.py
fschlueter: You can compress the code into:
```
        station_barycenter = np.array([
            np.mean([self._det.get_relative_position(station_id, channel_id) for channel_id in self._det.get_channel_ids(station_id)], axis=0) + self._det.get_absolute_position(station_id)
            for station_id in self._station_ids])
```
or 
```
        # calculate bary centers of station
        station_barycenter = np.zeros((len(self._station_ids), 3))
        for iSt, station_id in enumerate(self._station_ids):
            station_barycenter[iSt] = np.mean([self._det.get_relative_position(station_id, channel_id) for channel_id in self._det.get_channel_ids(station_id)], axis=0) + self._det.get_absolute_position(station_id)
```

NuRadioMC/simulation/simulation.py
cg-laser: I first implemented it that way but I ran into some problems with that. 
E.g.  we need additional meta attributes such as "event time" and if it is an "inince" shower or a "cosmic ray". The sim_station is the natural place to store these. 
In addition, this is exactly how the NuRadio event structure is set up. The `sim_station` is the container object of simulated electric field.

NuRadioMC/simulation/simulation.py
cg-laser: ok, I renamed and added a warning output to the init function.

NuRadioMC/simulation/simulation.py
cg-laser: the order shouldn't matter. And we often want the SimChannels for all channels. I think it is even used later to obtain the real channels.

NuRadioMC/simulation/simulation.py
cg-laser: the trigger modules don't define a pre_trigger_time. This is defined, set and applied in the triggerTimeAdjuster module. Or I misunderstand your question

NuRadioMC/simulation/simulation.py
sjoerd-bouma: I'm not a big fan of this - you save one line of code at the cost of making it less straightforward to parse for a human.

NuRadioMC/simulation/simulation.py
fschlueter: Both functions (`calculate_sim_efield*`) set `sim_station.set_is_neutrino()` and `sim_station.set_candidate(True or False)`. The latter could be done later? What am I missing?

If it is necessary to return a sim_station can we rename `add_sim_station` to `merge_sim_station`?

NuRadioMC/simulation/simulation.py
fschlueter: maybe I got so used to this syntax but for me it makes it easier to read code ...

NuRadioMC/simulation/simulation.py
fschlueter:
```
  for evt in output_buffer[sid].values():
                            trace_start_time = evt.get_station().get_channel(channel_ids[0]).get_trace_start_time()
                            # determine the trigger that was used to determine the readout window
                            trigger_name = None
                            trigger_time = None
                            pre_trigger_time = None
                            for trigger in evt.get_station().get_triggers().values():
                                if trigger.get_pre_trigger_times() is not None:
                                    pre_trigger_time = trigger.get_pre_trigger_times()
                                    trigger_time = trigger.get_trigger_time()
                                    trigger_name = trigger.get_name()
                                    break

                            assert(trigger_time is not None)
                            assert(trigger_name is not None)
                            assert(trace_start_time == trigger_time)

                                tt = sim_channel.get_times()
                                t0 = tt[0]
                                t1 = tt[-1]
                                pre_trigger_time = triggerTimeAdjuster.get_pre_trigger_times(sim_channel.get_id())
                                t0_readout = trigger_time - pre_trigger_time
                                t1_readout = t0_readout + self._det.get_number_of_samples(sid, sim_channel.get_id()) * self._det.get_sampling_frequency(sid, sim_channel.get_id())
```

NuRadioMC/simulation/simulation.py
cg-laser: it is doing that. 
# we need to select the trigger that was used to determine the readout window, which is the one that has the
                                # `pre_trigger_times` set. All other triggers will return None. We can double check that by testing that the
                                #  trigger_time is the same as start time as the trace
I added a comment to clarify that. This is maybe not ideal but it is that way how we do things at the moment. If we want to change that, we also need to change the way the triggerTimeAdjuster works.

NuRadioMC/simulation/simulation.py
cg-laser: this is (much) more efficient. We have potentially many many event where we try to add new channels to. The timing will only overlap with one event (or no event at all).

cg-laser:
> > I get the following error when I run the new simulation:
> > ```
> > (24, 4, 0) []
> > (24, 4, 1) []
> > (24, 5, 0) []
> > (24, 5, 1) []
> > (25, 4, 0) [24]
> > Traceback (most recent call last):
> >   File "/Volumes/software/rnog/Followup_Paper_100PeV/Followup_100PeV/C02_event_sim/simulate_rnog_100PeV.py", line 472, in <module>
> >     task({}, args.index, args.energy, det, args.config, output_filename, args.flavor, args.interaction_type, x0=args.x0, y0=args.y0, n_events_per_file=args.n_events_per_file)
> >   File "/Volumes/software/rnog/Followup_Paper_100PeV/Followup_100PeV/C02_event_sim/simulate_rnog_100PeV.py", line 378, in task
> >     n_trig = sim.run()
> >              ^^^^^^^^^
> >   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1496, in run
> >     station.add_sim_station(sim_station)  # this will add the channels and efields to the existing sim_station object
> >     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> >   File "/Users/felix/software/nuRadioMC/NuRadioReco/framework/station.py", line 24, in add_sim_station
> >     self.__sim_station = self.__sim_station + sim_station
> >                          ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
> >   File "/Users/felix/software/nuRadioMC/NuRadioReco/framework/sim_station.py", line 176, in __add__
> >     if channel.get_unique_identifier() in channel_ids:
> >        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> > ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The first few lines is a print I added:
> > ```
> >             print(channel.get_unique_identifier(), channel_ids)
> >             if channel.get_unique_identifier() in channel_ids:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am running a simulation with trigger channels 24 - 27 which are at the same position as non-trigger channels 0-3.
> 
> one can fix the issue by using `get_unique_identifier()[0]` to get the channel_id. That should be relatively save because the unique identifier is also checked in the `add_channel` method

Good catch. The correct check should be 
```
if channel.get_unique_identifier() in self.__channels:
```
as in the `add_channel` function. I mistakenly thought that `get_channel_ids()` would return a list of the unique identifiers.

NuRadioMC/simulation/simulation.py
fschlueter: > we also need to change the way the triggerTimeAdjuster works

I think we should do that :)

cg-laser:
> A comment to the (time) logging. So far we also print the number of triggered events which I find very useful. Can that be implemented in the new time_logger class?

I reused Christoph's code and I think there was a reason for why this was not easily possible but I forgot what it was. Otherwise, yes, this information would be useful

NuRadioMC/simulation/simulation.py
cg-laser: we always simulate them if we have a trigger. The big difference to before is that we don't calculated all channels for events that don't trigger.

NuRadioMC/simulation/simulation.py
cg-laser: is this a problem?

NuRadioMC/simulation/simulation.py
fschlueter: sure but this is not the case at the moment

cg-laser:
> station.add_channel(channel)

Thanks! I fixed it in the. most recent commit.

NuRadioMC/simulation/simulation.py
fschlueter: with the `continue`
```
                if not evt.get_station().has_triggered():
                        continue

                    triggerTimeAdjuster.run(evt, station, self._det)
                    evt_group_triggered = True
                    output_buffer[sid][evt.get_id()] = evt

```
we only skip the next 3 lines in the for-loop `for evt in events:`. The block of `if (len(non_trigger_channels) > 0):` is still executed

NuRadioMC/simulation/simulation.py
fschlueter: no not a big problem. my simulation script did use the return value. So we might break user code (in my case it was only to print it)

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1339,325 @@ def __init__(self, inputfilename,
...
+                                       detector_simulation_part2=self.detector_simulation_part2)
+
+                    # calculate trigger
+                    self.__time_logger.start_time("trigger")
+                    self.detector_simulation_trigger(evt, station, self._det)
+                    logger.debug(f"event {evt.get_run_number()},{evt.get_id()} tiggered: {evt.get_station().has_triggered()}")
+                    self.__time_logger.stop_time("trigger")
+                    if not evt.get_station().has_triggered():
+                        continue
+                    triggerTimeAdjuster.run(evt, station, self._det)
```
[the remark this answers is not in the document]
fschlueter: Also this module should probably run at the end for all channels and not only for the trigger channels.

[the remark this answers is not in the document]
fschlueter: Why can't we run the efieldToVoltageConverter and later triggerTimeAdjuster?

NuRadioMC/simulation/simulation.py
fschlueter: I do not understand what you mean. Maybe we can discuss during the call.

NuRadioMC/simulation/simulation.py
fschlueter: And, cant we move this calculation into the init function?

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1342,370 @@ def __init__(self, inputfilename,
...
-                        sim_station.add_channel(tmp_sim_station.get_channel(ch_uid))
-                        efield_uid = ([ch_uid[0]], ch_uid[1], ch_uid[2])  # the efield unique identifier has as first parameter an array of the channels it is valid for
-                        for efield in tmp_sim_station.get_electric_fields():
-                            if efield.get_unique_identifier() == efield_uid:
-                                sim_station.add_electric_field(efield)
+                                channel = NuRadioReco.framework.channel.Channel(channel_id)
+                                n_samples = int(round(self._det.get_number_of_samples(sid, channel_id)) * self._config['sampling_rate'] / self._det.get_sampling_frequency(sid, channel_id))
+                                channel.set_trace(np.zeros(n_samples), self._config['sampling_rate'])
+                                # we need to use any other channel to get the correct trace_start_time. All channels have the same start time at the end
+                                # of the simulation.
```
fschlueter, line 1655: Is that still necessary? You are adding empty channels in the previous for loop (might be did not test it).

fschlueter:
I put that here so we do not forget: `channelParameter.signal_ray_type` is not set in simulations.py

fschlueter:
@cg-laser can you briefly think about the last two comments I made? Otherwise here are a lot of comment from myself and sjoerd which have not been addressed. I agree that those are mostly of cosmetic nature (for example: writing better doc-strings and using better function and variable names are not crucial for code to run fast and correct but they are crucial for a successful project with a larger developer base) and thus can be (but also should be!) addressed in separate PR.

NuRadioMC/simulation/simulation.py
cg-laser: I'm pretty sure that this is still needed because the previous loop will skip channels that don't have RT solutions for example.

cg-laser:
> > > Also getting this:
> > > ```
> > >   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1667, in run
> > >     self._output_writer_hdf5.add_event_group(output_buffer)
> > >   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/output_writer_hdf5.py", line 126, in add_event_group
> > >     assert all(np.atleast_1d(self._mout_groups_attributes[sid][key_hdf5] == np.array(channel_values))), f"station {sid} key {key_hdf5} is {self._mout_groups_attributes[sid][key_hdf5]}, but current channel is {np.array(channel_values)}"
> > >                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> > > ValueError: operands could not be broadcast together with shapes (19,) (18,)
> > > ```
> > 
> > 
> > This does not happen always though.
> 
> I think the issue is again the same, that sometimes a `station` does not have the channel because no ray_tracing_solution

this should be fixed now, as we always fill all channels.

cg-laser:
> I put that here so we do not forget: `channelParameter.signal_ray_type` is not set in simulations.py

fixed

NuRadioReco/framework/event.py
fschlueter: Ping - wrong doc-string

NuRadioReco/framework/base_station.py
fschlueter: missing doc string

NuRadioMC/simulation/simulation.py
fschlueter: This is an remnant from the old code but I really thing that these single letter variables hurt here (also the properator `p`). They are do not serve to make the code more readable.

fschlueter (dismissed):

NuRadioMC/simulation/simulation.py
```diff
@@ -401,752 +1342,370 @@ def __init__(self, inputfilename,
...
+                                channel.set_trace_start_time(station.get_channel(station.get_channel_ids()[0]).get_trace_start_time())
+                                station.add_channel(channel)
+                            # logger.status(f"norm  = {norm}, Vrms = {Vrms[channel_id]}, max_freq = {max_freq}")
+                            ff = channel.get_frequencies()
+                            filt = np.ones_like(ff, dtype=complex)
+                            for i, (name, instance, kwargs) in enumerate(evt.iter_modules(sid)):
+                                if hasattr(instance, "get_filter"):
+                                    filt *= instance.get_filter(ff, sid, channel_id, self._det, **kwargs)
+                            noise = channelGenericNoiseAdder.bandlimited_noise_from_spectrum(len(channel.get_trace()), channel.get_sampling_rate(),
```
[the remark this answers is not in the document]
cg-laser: in principle you are right, we could do that, but implementing it will be error prone and I rather now want to change a "running system".

NuRadioMC/simulation/simulation.py
```diff
@@ -401,757 +1342,327 @@ def __init__(self, inputfilename,
...
-                        self._mout_groups[self._station_id] = {}
-                    for key in sg:
-                        if key not in self._mout_groups[self._station_id]:
-                            self._mout_groups[self._station_id][key] = list(sg[key])
-                        else:
-                            self._mout_groups[self._station_id][key].extend(sg[key])
-
-                detSimTime += time.time() - t1
+                            eventWriter.run(evt, mode=output_mode)
+                    remove_all_traces(evt)  # remove all traces to save memory
```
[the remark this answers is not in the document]
cg-laser: this is called after the eventWriter. So the full event has already been written to file. 
We need to keep all events (of one event group) in memory to be able to write the hdf5 file which is done after each event group. But we don't need any trace information for he hdf5 files, so we already remove it from memory.

NuRadioReco/framework/trigger.py
```diff
@@ -124,35 +144,54 @@ def set_pre_trigger_times(self, pre_trigger_times):
...
+        channel_id: int
+            the channel id
+
+        Returns
+        -------
+        trigger_time: float
+            the trigger time for the channel
+        """
+        if self._pre_trigger_times is None:
+            return self._pre_trigger_times
```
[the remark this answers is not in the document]
cg-laser: yes it can be removed. Done.

NuRadioMC/simulation/simulation.py
cg-laser: so that it passed potential speedup cuts

documentation/source/NuRadioReco/pages/times.rst
```diff
@@ -0,0 +1,52 @@
+overview of times
```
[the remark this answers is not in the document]
cg-laser: I will table this for the next PR where we change this

fschlueter (approved):

fschlueter (approved):

PR #681, opened 2024-05-27
change get_channel_group_id to return channel id if no group_id is set

lpyras:
This PR changes the function get_channel_group_id to return the channel id if no group id is set. The group_ids describe antennas are set for antennas with the same positions. Looping over the group_ids instead of the channel_ids and assign a quantity to several station ids can therefore be faster. The channel_id is more specific, therefore returning the channel id if no group id is set should not cause problems.


NuRadioReco/detector/detector_base.py
```diff
@@ -967,10 +967,10 @@ def get_channel_group_id(self, station_id, channel_id):
         """
         res = self.__get_channel(station_id, channel_id)
         if 'channel_group_id' not in res.keys():
```
fschlueter, line 969: `.keys()` is not necessary

NuRadioReco/detector/detector_base.py
```diff
@@ -967,10 +967,10 @@ def get_channel_group_id(self, station_id, channel_id):
...
-            logger.warning(
-                'Channel group ID not set for channel {} in station {}, returning -1'.format(
+            logger.info(
```
MijnheerD, line 970: I agree with you that warning is a bit much, but maybe info is too little? Not sure. I think STATUS would be optimal actually, as that would still show up in the logs by default in the new system. Though for that we need to wait until the new logging is done :)

cg-laser (dismissed):
looks good. I thought we already had this implemented. (Maybe I only did it on the `refactor` branch)

I would not even log an info method. This is a robust way for writing modules that work with antenna group and without, so we don't want to get flushed with log messages for an expected behaviour.

fschlueter (approved):

PR #682, opened 2024-05-29
Hotfix and cosmic ray energy reconstruction parameterization for summit

sjoerd-bouma:
This PR adds the parameterization for Summit station for the cosmic ray energy to the cosmicRayEnergyReconstructor module. This parameterization was determined in a similar way as described in [the original paper](https://arxiv.org/abs/1905.11185) by @christophwelling; I used the cosmic ray simulations by @cg-laser (see [MSc thesis](https://bib-pubdb1.desy.de/record/456326/files/Pyras_Msc.pdf?version=1) by @lpyras for more details) and fit the parameterization to noiseless cosmic ray signals using a 4-LPDA layout. The fit was performed directly over all zeniths and energies, weighted by the simulation weights and using an $E^-2$ flux, and minimizing $\Delta \log E$.

In addition, I noticed a bug in the code - the third parameter of the `scale_parameterization` was never used, instead the first parameter was used twice. I fixed this, but I am actually not sure if the parameterizations are otherwise correct; also in the original paper, I'm worried the order of the parameters might occasionally be swapped, because e.g. numpy uses the convention `p[0] + p[1] x + p[2] x**2 + ...` (i.e. constant term first) rather than putting the highest-order term first. It would maybe be good if @christophwelling could double check this : )

### DIFF
--- a/NuRadioReco/modules/cosmicRayEnergyReconstructor.py
+++ b/NuRadioReco/modules/cosmicRayEnergyReconstructor.py
@@ -13,6 +13,7 @@ import radiotools.atmosphere.models
 class cosmicRayEnergyReconstructor:
     """
     Reconstructs the energy of an air shower from its radio signal
+
     Requires the following modules to be run beforehand:
     
     * a 10th order Butterworth bandpass filter with passband 80-300 MHz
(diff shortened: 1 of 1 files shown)


NuRadioReco/modules/cosmicRayEnergyReconstructor.py
```diff
@@ -65,30 +105,45 @@ def run(self, event, station, detector):
...
-        cs = radiotools.coordinatesystems.cstrafo(zenith, azimuth, site=self.__site)
+        alpha = hp.get_angle_to_magnetic_field_vector(zenith, azimuth, site)
+        cs = radiotools.coordinatesystems.cstrafo(zenith, azimuth, site=site)
         efield_trace_vxB_vxvxB = cs.transform_to_vxB_vxvxB(cs.transform_from_onsky_to_ground(electric_field.get_trace()))
         efield_trace_vxB_vxvxB[0] /= np.sin(alpha)  # correct energy fluence for effect of angle to magnetic field
         energy_fluence = NuRadioReco.utilities.trace_utilities.get_electric_field_energy_fluence(efield_trace_vxB_vxvxB, electric_field.get_times())
         energy_fluence = np.abs(energy_fluence[0]) + np.abs(energy_fluence[1])
-        xmax_distance = self.__atmosphere.get_distance_xmax_geometric(zenith, 750., self.__elevation)  # parametrization is for Xmax of 750g/cm^2
+        xmax_distance = self.__atmosphere.get_distance_xmax_geometric(zenith, 750., elevation)  # parametrization is for Xmax of 750g/cm^2
+        xmax_distance = np.abs(xmax_distance) #np.max([xmax_distance, 200*units.m]) # for some zeniths and altitudes the parameterization can become negative
```
anelles, line 136: Should we not print out a warning, if this happens at least?

anelles:
If you change my request for a warning @sjoerd-bouma , I am happy to move forward here. It is not that @christophwelling will remember any better what happened here, the longer we wait.

NuRadioReco/modules/cosmicRayEnergyReconstructor.py
sjoerd-bouma: Good suggestion, thanks.

anelles (approved):
Looks good to go from my side.

PR #685, opened 2024-05-31
Refactoring coreas to include interpolation

lpyras:
Major refactoring of the coreas interface to read the electric field from coreas hdf5 files. 

A new class  readCoREASDetector.py can be used to read an array and get the interpolated position. 

Deleting 
- readCoREAS.py which reads a random position 
- readCoREASStationGrid.py which takes the closest observer in an array. Now readCoREASDetector.py can be used to read an array and get the interpolated position.


cg-laser:
I implemented the changes that I suggested in #659 https://github.com/nu-radio/NuRadioMC/pull/659#pullrequestreview-2079129240 In particular to have a single function `read_CORSIKA7` which parses the coreas output and returns everything as NuRadioReco event. This future proofs this implementation for CORSIKA8 which is coming soon. 

There was also a bug in requesting the antenna position (the channel_group_id instead of the channel_id was used). This is fixed now. I seem to be getting reasonable result. 

One important thing that is not yet implemented is the start time of the electric field. I don't know how this is handled in the interpolator class. Maybe the current code is already correct, but the experts should check. Currently, the start time is set to zero (which was equivalent to the previous implementation). 

Further cleanup is likely needed.

lpyras:
Hi Christian, thanks for taking the time to clean this up. 

> There was also a bug in requesting the antenna position (the channel_group_id instead of the channel_id was used). This is fixed now. I seem to be getting reasonable result.

I thought the channels with the same group id share a location? therefore I used the position of the channel group id. But maybe this is wrong.

lpyras:
> I don't think it's a good idea to delete readCoREAS.py. This is still the most accurate module for estimating detection efficiencies. The new module completely relies on the interpolator working accurately enough, and it only works with star-shape pattern simulations that fullfil certain criteria. There don't seem to be any checks that guarantee a certain accuracy, so this is potentially dangerous. I also saw in the code, that the interpolator does not work for small geomagnetic angles? As the magnetic field at the South Pole is very vertical, it might limit the usability and we need to other modules instead.

I dont insist on keeping readCoREAS.py. I thought it doubles readCoREASStation.py which will take all observers and convert them into events with different core positions, which should be more accurate than taking a random core position and choosing the closes observer (as in readCoREAS.py). I have seen though, that readCoREAS.py is part of the unit test.

About the geomagnetic angle. At small geomagnetic angles the geomagnetic emission is weak, and the rotational symmetry on which part of the interpolation relies is no longer true. For Greenland this is the case for very vertical showers. The radio emission is anyways small, so taking the closest observer should not be a problem. I assume it should be similar for Southpole. But I agree that we definitely should have something besides the interpolation module.

lpyras:
@cg-laser wrote:
It would be great if we could also add an example of a "Lofar style Xmax fit":

read in a coreas star shape simulation and fill the efields of an array of dual-polarized antennas (maybe just a square grid)
run a full detector simulation, i.e., to voltage level + noise + and back to efield
calculate the fluence and fit the core position using the fluence interpolation of the input coreas shower.

I will also add an example for effective area calculation for multiple stations

Issue #616, opened 2024-02-05
Interface to modify RNO-G detector description


fschlueter:
This was mentioned in one of the RNO-G calls. Several analyses (I guess mostly calibration related) will profit / make it necessary to change the detector description "on-the-fly" to validate the effect on the changed description. I can think of a few approaches how to realise that. This issue is here to collect ideas and as a discussion platform. 

I have two ideas:
- Implement a setter for the `total_response` of a channel. Keep in mind this key is buffered internally. Also this key is not exported to file (because one can/should not serialise custom classes and the data are already exported). 
- Implement a setter for the signal chain list which specifies the individual response which constitute the `total_reponse` (those would than need to be defined in the database)

fschlueter:
A other possibility is to modify the json file. If that is the case we should add some utility scripts to NuRadio for that.

cg-laser:
what about a new class "ModifiedDetector" that inherits from the RNO-G detector but allows the user to overwrite any parameters/response functions? This detector class could also implement other useful functions such as systematically changing response functions to study systematic uncertianties, etc. 
It might be good to separate this functionality from the RNO-G detector class to let users unintentionally mess up the reference detector description.

cg-laser:
and as always, I would take a pragmatic approach and implement only the functionality that someone needs for their analysis right now. If more functionality is needed at a later stage, this can be added later.

PR #686, opened 2024-06-04
Enable modifications of the RNO-G detector description 

fschlueter:
I all, 

with this PR I want to start a discussion on how we will support the feature to modify the RNO-G detector description "on-the-fly". The feature is useful for systematic studies and in our efforts to improve the detector description. On example: We want to rotate the antenna orientations to see the effect of this rotation to the simulation of the galactic emission. In the future it will come in handy to allow to replace the response function of a channel with something different. 

This PR adds a very simple function to the detector class which should (untested!) allow you to replace a certain value in the detector description. It is not very user friendly in the sense that you really need to know what you have to do. We could think of more specific user friendly interfaces. It could be a good possibility to create an inherent class which hosts all these modification interfaces we envision. 

Any suggestions / comments / opinions?


philippwindischhofer:
Hi @fschlueter thanks for starting the discussion on this topic! Below are a few quick comments and questions from my side. If I can think of anything else, I will post another comment.

**Station geometry storage and bookkeeping**

1. Is the long-term plan to store the station / channel geometry information also in the hardware DB? (I can already see some geometry-related collections filled, so I guess the answer is "yes"?) If so, I would advocate for removing the current `json` files as soon as the DB-based solution is fully available.
2. The DB has the nice "interval of validity" (IOV) mechanism to allow multiple geometries to be available for non-overlapping times. When we iterate on the geometry calibration later this year (and in the longer-term future), we might have to update the geometry as new calibration data becomes available. If our best estimate for the station geometry changes over time, I wonder if the IOV is the best mechanism to handle that? E.g. we might want to keep multiple past calibration results around to easily check how reconstruction / simulation results change. One way to implement this would be to build a poor-man's commit-tree by storing for each station geometry also a "parent" that points to the geometry that it corrects / updates. (I don't know if having multiple overlapping IOVs would break things elsewhere.)

**On-the-fly updates of station geometry**

3.  I guess we might want to support two separate interfaces: an "expert" one that allows the user to modify all degrees of freedom in an unconstrained manner as they please; and a "user" one that provides a more high-level interface. In the latter case, only a few "knobs" would be exposed that internally map onto antenna positions. For example, if we decide to go for the eigenvector parametrization of geometry calibration results that I presented at the virtual collaboration meeting, the exposed DOFs would be the nuisance parameters corresponding to the eigenmodes.

**Visualization**

4. Do we already have some centralized code to visualize a given station geometry? I believe this could be worthwhile to build intuition for how a particular update to the geometry actually changes the situation in the ice. Right now, I think people write their own scripts (I have one, the calibration code has one, and I think @MatthewMarsee also has one), and it might be worth harmonizing these.

fschlueter:
Hey @philippwindischhofer some quick answers:
To 1): Yes and Yes :)
To 2): We have anticipated for what you describe and the DB / NuRadio class already accommodates for this. A longer explanation is [here](https://nu-radio.github.io/NuRadioMC/NuRadioReco/pages/detector/rnog.html#detector-time-and-database-time) but the short version is: With the `database_time` you can select a `measurement` of e.g. the channel position which was primary at that specific time while in the mean time with have specified a new, supposedly better measurement, as the primary one. 
To 3): Sounds reasonable and I guess we can add the user-friendly interface along the road as we need it. Should be but that in a dedicated class though?
To 4): I am not terrible aware of all the different visualisation codes neither (maybe @sjoerd-bouma  can help us out here). But I remember that we have a detector browser viewer?

philippwindischhofer:
Hi @fschlueter, thanks for your rapid answer!

2. Brilliant!! I wasn't aware of this.
3. Yes, I would put all that into dedicated classes to keep it nice and separate. A future `GeometryCalibration` class would then take as inputs the `database_time` and values for the nuisance parameters, and return the properly calibrated detector geometry.

fschlueter:
> Yes, I would put all that into dedicated classes to keep it nice and separate. A future GeometryCalibration class would then take as inputs the database_time and values for the nuisance parameters, and return the properly calibrated detector geometry.

Do you plan to perform calibrations "on-the-fly". I thought the idea is to run a calibration analysis and put the results of this in the DB

philippwindischhofer:
Yes, the calibration would be run once, recorded in the DB, and a particular calibration retrieved via the `database_time` mechanism. 

As part of the calibration we will get uncertainties on the station geometry. If we want to propagate the geometry uncertainties into a simulation / reconstruction (especially at the beginning, where the uncertainties will be large), we need a mechanism to sample the space of allowed station geometries. That part, I think, would profit from being done "on the fly" (since it is fast), instead of storing many explicit variations of the geometry in the DB.

sjoerd-bouma:
@philippwindischhofer Re 4) - there's a nice interactive 3D visualization of NuRadio detectors (see [the documentation](https://nu-radio.github.io/NuRadioMC/NuRadioReco/pages/detector/detector.html#detector-viewer) for some more information on how to use it). It probably needs an update in order to also support the RNO-G detector, though; right now it only works with jsons, and the interface is slightly outdated.

fschlueter:
> Yes, the calibration would be run once, recorded in the DB, and a particular calibration retrieved via the `database_time` mechanism.
> 
> As part of the calibration we will get uncertainties on the station geometry. If we want to propagate the geometry uncertainties into a simulation / reconstruction (especially at the beginning, where the uncertainties will be large), we need a mechanism to sample the space of allowed station geometries. That part, I think, would profit from being done "on the fly" (since it is fast), instead of storing many explicit variations of the geometry in the DB.

I see. Probably we should think of a smart way to allow shifting of positions and orientations.

shallmann:
Thanks for starting this, Felix!
- The detector validity is not affected by this, i.e. running a `detector.update()` would not re-set the values to the original. Maybe we need to keep this in mind (and might want to introduce a `force_update` or a `is_modified` somewhere  to check if the detector is modified or not and be able to get back to the original if need be.

- Having the possibility to modify detector descriptions will be very useful for sure. In addition to the general `modify_channel_description` function we can add more specialised functions to shift/rotate individual channels as we need them.
- I like the suggestion @philippwindischhofer to be able to automatically randomly shift detector components by their allowed uncertainties. I guess not but: Are there already fields for uncertainty in the database?

- If we have no other volunteers, I can look into updating the detector visualisation next week.

sjoerd-bouma:
Tangentially to Steffen's comment and as was suggested by Christian in #616, I think it's important to be able to distinguish a modified detector from an unmodified detector. If we don't introduce a separate 'ModifiedDetector' class I think it would be good to at least have a (one-time) warning issued when a user first modifies the detector, and maybe an `is_modified` property as Steffen suggests to keep track of this. Then in terms of being able to store/keep track of modified values I guess exporting to json still works?

philippwindischhofer:
Hey @fschlueter and @shallmann,

> I see. Probably we should think of a smart way to allow shifting of positions and orientations.

> I like the suggestion @philippwindischhofer to be able to automatically randomly shift detector components by their allowed uncertainties.

For at least some cases (calibration being one of them) a proper treatment of the uncertainties will be a bit more complicated than simply "randomly wiggle antennas around independently". Rather, each uncertainty component will move around a set of antennas in a particular way. This is what I called "eigenmodes" earlier.

>  I guess not but: Are there already fields for uncertainty in the database?

I don't think there currently are---but @fschlueter or @jakobhenrichs would know for sure. Given the above, I think an important aspect will be to decide on the format we choose to store the "modifications" in. If we want this to go into the DB, it should be sufficiently general to cover all our future use cases.

Hey @sjoerd-bouma,

> Tangentially to Steffen's comment and as was suggested by Christian in https://github.com/nu-radio/NuRadioMC/issues/616, I think it's important to be able to distinguish a modified detector from an unmodified detector. 

I would even go one step further and, instead of just tagging a detector geometry as "unmodified" and "modified", give them more informative "tags". There would be a single "nominal" detector geometry that represents the best-estimate of the actual as-deployed instrument, and then there would be several named variations around that nominal case.

> If we don't introduce a separate 'ModifiedDetector' class I think it would be good to at least have a (one-time) warning issued when a user first modifies the detector, and maybe an is_modified property as Steffen suggests to keep track of this. Then in terms of being able to store/keep track of modified values I guess exporting to json still works?

I don't have any opinion on whether a `ModifiedDetector` class is the way to go, or whether the current `Detector` class should hold this functionality. I would just advocate to make sure that the interface is the same for both, e.g. querying channel positions etc. should work in exactly the same way for both, so that we can use a modified detector in all locations where we can use an unmodified one. That's the essential feature that will make rigorous propagation of geometry uncertainties possible.

shallmann:
> For at least some cases (calibration being one of them) a proper treatment of the uncertainties will be a bit more complicated than simply "randomly wiggle antennas around independently". Rather, each uncertainty component will move around a set of antennas in a particular way. This is what I called "eigenmodes" earlier.

I didn't fully appreciate the meaning of 'eigenmodes' earlier, yes you are right. We need to agree on which uncertainties and correlations / eigenmodes we need in the database. Independent per-channel uncertainties would have been a no-brainer to add to the database, eigenmodes may need a bit more thought.

> Given the above, I think an important aspect will be to decide on the format we choose to store the "modifications" in. If we want this to go into the DB, it should be sufficiently general to cover all our future use cases.

> I would even go one step further and, instead of just tagging a detector geometry as "unmodified" and "modified", give them more informative "tags". There would be a single "nominal" detector geometry that represents the best-estimate of the actual as-deployed instrument, and then there would be several named variations around that nominal case.

1.  We intended the database to hold new "best fit" positions, and (eventually) uncertainties/eigenmodes after each iteration of calibrating the detector (each iteration retrievable by specfying a `database_time` when initialising the detector object). I think this is what we definitely want in the official database.
2. Another use case is to deliberately shift/delay/rotate a set of channels for individual studies. We should have a functionality to allow for that by passing a modification dict or applying some modification functions. I think this is what Felix already had in mind when starting this thread. 
3. We can discuss on wether (and how) we need official fixed named modification sets stored centrally (which cannot be regenerated by a function using the official best fit database values and uncertainties).
4. A simple variable to tell if a detector was manipulated will be useful irrespective of that.

> I don't have any opinion on whether a `ModifiedDetector` class is the way to go, or whether the current `Detector` class should hold this functionality. I would just advocate to make sure that the interface is the same for both, e.g. querying channel positions etc. should work in exactly the same way for both, so that we can use a modified detector in all locations where we can use an unmodified one. That's the essential feature that will make rigorous propagation of geometry uncertainties possible.

No strong opinion either, but if we opt for a separate `ModifiedDetector` class this should derive from the unmodified detector of course to ensure all queries still work the same. Keeping both a modified and an unmodified detector alive (which afaik may be not possible with two normal detectors since they are a singleton?!?) would allow for direct comparisons between the two, if there is a use case

philippwindischhofer:
>  Independent per-channel uncertainties would have been a no-brainer to add to the database, eigenmodes may need a bit more thought.

Yep! Although, if we find a good way of keeping the eigenmodes in the DB, then independent per-antenna uncertainties are just a special case :-)

I fully agree with your points 1-4!

> Keeping both a modified and an unmodified detector alive (which afaik may be not possible with two normal detectors since they are a singleton?!?) would allow for direct comparisons between the two, if there is a use case

Indeed! I think this is a good argument in favor of a `ModifiedDetector`. And if we generate the `ModifiedDetector` through a call to the original `Detector` like `Detector.modify(mod)`, then we can also easily loop over all variations (including the nominal case).

philippwindischhofer:
Hi again,

I wanted to come back to this and open up another topic for discussion. If we are going to have the detector class support alternative geometries, it might be useful to be able to read those in from a file, similar to how detector descriptions may currently be specified in a `json` file.

I had a go at a possible schema [here](https://github.com/RNO-G/antenna-positioning/blob/phwindis_eigendeformations/AntPosCal/position_fitting/global_fit/calib_result.schema.json). I can provide an example calibration output if interested. As you can see, this follows the current NuRadioMC schema quite closely (is this formally specified somewhere?).

Sticking with `json` should also allow us to keep this inside the DB if we want.

What do you think?

fschlueter:
Hi @philippwindischhofer, 
in the future (and actually the future is now :) ) we will use a detector description queried from the hardware database with the new rnog_detector class. The position of each channel is stored in the DB and the idea was to store results of a calibration there too. So far there are no fields for uncertainties but this could be added easily.

fschlueter:
Hi all, I will need to go through that entire thread at some point (sorry for not having done so already and coming back to this). However, I wanted to let you know that I started an implementation of a `ModDetector` class in the `rnog_trigger` branch.

shallmann:
I started taking the `ModDetector` class Felix started with from the `rnog_trigger` branch in order to develop that further.

NuRadioReco/detector/RNO_G/rnog_detector_mod.py
```diff
@@ -0,0 +1,237 @@
...
+                                        name=component['name'],
+                                        station_id=station_id,
+                                        channel_id=channel_id)
+
+        orig_response = self.get_signal_chain_response(station_id, channel_id)
+        signal_chain_dict = self.get_channel_signal_chain(station_id, channel_id)
+
+        # modify the signal chain
+        signal_chain_dict["total_response"] = orig_response * component_response
+        signal_chain_dict["response_chain"][component['name']] = component
```
shallmann, line 186: To me the way the `response_chain` and `total_response` are stored do not look ideal, it seems that things are stored twice and it is obscure which functions internally use which, and not straightforward to keep them in sync.

In an ideal world, a function like the `add_response` above should be enough, but `det.get_time_delay()` does not look at the total response, so the modifications seem necessary at two places.

Maybe I am missing something obvious how this was supposed to work?

shallmann: This is a demo of what the (still work in progress) added functions do...

Screenshot 2024-10-09 at 15 51 18

fschlueter: Hi @shallmann, sorry to reply so late. I am not sure I understand your point here: with the `add_response` you modify both the `total_reponse` object and the dict `response_chain`. Furthermore, by now `get_time_delay` does make use of the `total_reponse`. But maybe I am missing something?

fschlueter:
@philippwindischhofer just to comment on this point:

> The DB has the nice "interval of validity" (IOV) mechanism to allow multiple geometries to be available for non-overlapping times. When we iterate on the geometry calibration later this year (and in the longer-term future), we might have to update the geometry as new calibration data becomes available. If our best estimate for the station geometry changes over time, I wonder if the IOV is the best mechanism to handle that? E.g. we might want to keep multiple past calibration results around to easily check how reconstruction / simulation results change. One way to implement this would be to build a poor-man's commit-tree by storing for each station geometry also a "parent" that points to the geometry that it corrects / updates. (I don't know if having multiple overlapping IOVs would break things elsewhere.)

The DB is able to store multiple measurements of the same thing for overlapping time periods. The detector class can easily extended to get those measurements if needed.

fschlueter:
@shallmann not sure I understand this point:

> The detector validity is not affected by this, i.e. running a detector.update() would not re-set the values to the original. Maybe we need to keep this in mind (and might want to introduce a force_update or a is_modified somewhere to check if the detector is modified or not and be able to get back to the original if need be.

fschlueter:
> @shallmann not sure I understand this point:
> 
> > The detector validity is not affected by this, i.e. running a detector.update() would not re-set the values to the original. Maybe we need to keep this in mind (and might want to introduce a force_update or a is_modified somewhere to check if the detector is modified or not and be able to get back to the original if need be.

Ah now I got it.

fschlueter:
Hey guys, going through this PR I identified a couple of action items:

- [ ] Implement a `is_modified` tag
- [ ] Implementing a more detailed status variable which gives information how the detector was modified
- [ ] Implementing a function which removes all modifications. 
- [ ] Deciding whether this should live in its own class or not

The question I have. Do we need this class sooner or later? Should we merge it soon and work on the open items in the future?

CamphynR:
Hi all,
While testing, I ran into the issue that not specifying all the kwargs in the ModDetector init seems to cause problems with kwarg inheritance from the parent class. For example only specifying select_stations overwrites the inherited kwargs and only saves that variable in the new kwarg dictionary. I include some testing code to make this clearer.

```
import argparse
import datetime
import logging
import numpy as np

from NuRadioReco.detector.RNO_G.rnog_detector_mod import ModDetector


def get_antenna_type(ch_id):
    if ch_id in [0, 1, 2, 3]:
        return "VPol"
    elif ch_id in [4, 8]:
        return "HPol"
    else:
        return "LPDA"


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--station", default=23, type=int)
    args = parser.parse_args()

    logging.getLogger().setLevel(logging.DEBUG)

    station_id = args.station
    include_channels = [0, 1, 2, 3, 4, 8]


    antenna_models = {"VPol" : "RNOG_vpol_v3_5inch_center_n1.74",
                      "HPol" : "RNOG_hpol_v4_8inch_center_n1.74"}

# all kwargs explicitly defined
#    detector = ModDetector(database_connection='RNOG_public', log_level=logging.NOTSET, over_write_handset_values=None,
#                 database_time=None, always_query_entire_description=False, detector_file=None,
#                 select_stations=station_id, create_new=False)

# only subset of kwargs defined
    detector = ModDetector(database_connection='RNOG_public', select_stations=station_id)
    detector_time = datetime.datetime(2022, 8, 1)
    detector.update(detector_time)

    for channel_id in [0, 1, 2, 3]:
        antenna_model = "RNOG_vpol_v3_5inch_center_n1.74"
        detector.modify_channel_description(station_id, channel_id, ["signal_chain","VEL"], antenna_model)
        print("model getter yields: " + detector.get_antenna_model(station_id, channel_id))
        ch_info = detector.get_channel(23, channel_id)
        print("channel info yields: " + ch_info["signal_chain"]["VEL"])


    for channel_id in [4, 8]:
        antenna_model = "RNOG_hpol_v4_8inch_center_n1.74"
        detector.modify_channel_description(station_id, channel_id, ["signal_chain","VEL"], antenna_model)
```

running this yields the error
in __init__
self.logger.setLevel(kwargs'log_level"])
KeyError: 'log_level'

fschlueter:
This is because the specific key `log_level` is used in the init function of the inherited class:
```
class ModDetector(Detector):
    def __init__(self, *args, **kwargs):
        super(ModDetector, self).__init__(*args, **kwargs)
        self.logger = logging.getLogger("NuRadioReco.RNOGDetectorMod")
        self.logger.setLevel(kwargs["log_level"])
```
I am not sure what the best solution is. Maybe we can do something like `self.logger = logging.getLogger("NuRadioReco.RNOGDetector.Mod")`? @MijnheerD with that will the logger `NuRadioReco.RNOGDetector.Mod` inherit the log level from `NuRadioReco.RNOGDetector`?

MijnheerD:
> I am not sure what the best solution is. Maybe we can do something like `self.logger = logging.getLogger("NuRadioReco.RNOGDetector.Mod")`? @MijnheerD with that will the logger `NuRadioReco.RNOGDetector.Mod` inherit the log level from `NuRadioReco.RNOGDetector`?

This is the intend behaviour yes. The idea is that the child loggers do not have a handler, hence they pass on their messages to the parent to be handled there. Though they can have a log level set, to restrict what logs are actually sent up the chain. If the `NuRadioReco.RNOGDetector.Mod` has no logging level set (ie its value is `logging.NOTSET`) it should pass on all the messages to ``NuRadioReco.RNOGDetector`, effectively inheriting its level. If this would not be the case, let me know cause then there is probably a bug somewhere 😓

NuRadioReco/detector/RNO_G/rnog_detector_mod.py
fschlueter: Okay I understand now your concern better. Yes the function `add_response` is problematic because it keeps the `signal_chain_dict = self.get_channel_signal_chain(station_id, channel_id)` untouched. I think by now the function `get_time_delay` by default uses the `Response` object so we are good in this regard but if we want to export those detectors we will not take into account the added response!

NuRadioReco/detector/RNO_G/rnog_detector_mod.py
```diff
@@ -0,0 +1,251 @@
...
+
+        component = {
+            'weight': weight,
+            'y-axis_units': ['mag', 'rad'],
+            'mag': [1, 1],
+            'phase': [0, 0],
+            'frequencies': [1e-3, 1e1],
+            'name': name,
+            'time_delay': time_delay
+        }
```
fschlueter, line 215: @shallmann I think this is a bug. The database will return responses where the phase has a phase shift according to `time_delay`. This shift is removed in the constructor of the `Reponse` class. However, if the S21 parameter does not have this shift, it will be introduced and you double count the delay if you apply the response and change the trace start time.

fschlueter: Fix it

fschlueter:
> Hey guys, going through this PR I identified a couple of action items:
> 
> * [ ]  Implement a `is_modified` tag
> * [ ]  Implementing a more detailed status variable which gives information how the detector was modified
> * [ ]  Implementing a function which removes all modifications.
> * [ ]  Deciding whether this should live in its own class or not
> 
> The question I have. Do we need this class sooner or later? Should we merge it soon and work on the open items in the future?

Any opinion about it? If we start shipping files based on this class we should probably merge it? Another question I have in addition to the one above. If we do all of this do we still need a dedicated class or could we do all of this in one class?

philippwindischhofer:
> Hey guys, going through this PR I identified a couple of action items:
> 
> * [ ]  Implement a `is_modified` tag
> * [ ]  Implementing a more detailed status variable which gives information how the detector was modified
> * [ ]  Implementing a function which removes all modifications.
> * [ ]  Deciding whether this should live in its own class or not
> 
> The question I have. Do we need this class sooner or later? Should we merge it soon and work on the open items in the future?

Hey @fschlueter---what about keeping this as a separate class? This would immediately also solve your other points. A modified detector can be easily recognized by its type. A `ModDetector` can be exported as an "official" modification (e.g. through calibration), distributed to users, and then imported again as a plain `Detector`. If analysis operations only work with `Detector` objects, they could be sure to never do anything "illegal".

What do you think?

fschlueter:
> > Hey guys, going through this PR I identified a couple of action items:
> > 
> > * [ ]   Implement a `is_modified` tag
> > * [ ]   Implementing a more detailed status variable which gives information how the detector was modified
> > * [ ]   Implementing a function which removes all modifications.
> > * [ ]   Deciding whether this should live in its own class or not
> > 
> > The question I have. Do we need this class sooner or later? Should we merge it soon and work on the open items in the future?
> 
> Hey @fschlueter what about keeping this as a separate class? This would immediately also solve your other points. A modified detector can be easily recognized by its type. A `ModDetector` can be exported as an "official" modification (e.g. through calibration), distributed to users, and then imported again as a plain `Detector`. If analysis operations only work with `Detector` objects, they could be sure to never do anything "illegal".
> 
> What do you think?

Sure I am open to this question. But how do we differentiate an officially modified and an unofficially (and thus illegal) detector. We can not really right (which I guess is also fine and we really on the sanity of the users/collaborators). We could still implement an option to erase everything and reload it from the DB

philippwindischhofer:
> But how do we differentiate an officially modified and an unofficially (and thus illegal) detector. We can not really right (which I guess is also fine and we really on the sanity of the users/collaborators).

One idea could be to add a "change notes" field to the exported `json`, which would get printed if the thus modified `Detector` is loaded. In this way the user would get notified of the change history of the detector they just loaded.

> We could still implement an option to erase everything and reload it from the DB

What would be the use case of such an option? Couldn't one just create a new `Detector` object with default constructor options, which (if I understand correctly) would automatically fetch everything from the DB?

> I also think we should merge this once these modified files get circulated. @philippwindischhofer do you want to add the script you used to export the calibrated detectors?

Agreed. I have the script generating these files here: https://github.com/RNO-G/antenna-positioning/pull/27 Maybe it's better kept in the calibration code repository? There I also have a quick-n-dirty `diff` script that can compare two `Detectors` and highlight any differences in channel positions and cable delays.

On that note: currently I'm storing the postfit ice model using the `additional_data` field in the `json`. Let me know what your thoughts are on this. This is just a temporary solution, I think, and we should make the ice model an integral part of the detector class.

fschlueter:
> On that note: currently I'm storing the postfit ice model using the `additional_data` field in the `json`. Let me know what your thoughts are on this. This is just a temporary solution, I think, and we should make the ice model an integral part of the detector class.

agreed.

fschlueter:
> I have the script generating these files here: [RNO-G/antenna-positioning#27](https://github.com/RNO-G/antenna-positioning/pull/27) Maybe it's better kept in the calibration code repository? There I also have a quick-n-dirty `diff` script that can compare two `Detectors` and highlight any differences in channel positions and cable delays.

Okay I think that is okay but we should make sure that this is well documented.

fschlueter:
> What would be the use case of such an option? Couldn't one just create a new `Detector` object with default constructor options, which (if I understand correctly) would automatically fetch everything from the DB?

True

fschlueter:
> One idea could be to add a "change notes" field to the exported `json`, which would get printed if the thus modified `Detector` is loaded. In this way the user would get notified of the change history of the detector they just loaded.

I like the idea.

fschlueter:
> > One idea could be to add a "change notes" field to the exported `json`, which would get printed if the thus modified `Detector` is loaded. In this way the user would get notified of the change history of the detector they just loaded.
> 
> I like the idea.

@philippwindischhofer implements this idea with #889. Does that mean we can merge this PR too? From my side we could merge it and add missing features while we work with the class. What do you think @philippwindischhofer

fschlueter:
FFT approve this PR than.

philippwindischhofer (approved):

PR #691, opened 2024-06-13
Add mattak iterator to readRNOGDataMattak

CamphynR:
This PR adds the mattak dataset iterator to readRNOGDataMattak to load data in batches in the run() function. The batch size can be set by the max_in_mem parameter.

### DIFF
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -730,9 +730,9 @@ class readRNOGData:
 
             # read all event infos of the entire dataset (= run)
             for evtinfo, wf in dataset.iterate(start = 0, stop = dataset.N(),
-                                                              calibrated = self._read_calibrated_data,
-                                                              selector = self._selectors,
-                                                              max_entries_in_mem = self._max_in_mem):
+                                               calibrated = self._read_calibrated_data,
+                                               selectors = self._selectors,
+                                               max_entries_in_mem = self._max_in_mem):
                 event_idx += 1
 
                 self.logger.debug(f"Processing (filtered) event number {event_idx} out of (unfiltered) total {self._n_events_total}")
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -149,6 +149,7 @@ def _all_files_in_directory(mattak_dir):
     req_files = ["daqstatus.root", "headers.root", "pedestal.root"]
     for file in req_files:
         if not os.path.exists(os.path.join(mattak_dir, file)):
+            logging.error(f"File {file} could not be found in {mattak_dir}")
             return False
 
     return True
@@ -361,7 +362,7 @@ class readRNOGData:
                                  f"\n\tSelect runs with max. trigger rate of {max_trigger_rate / units.Hz} Hz"
                                  f"\n\tSelect runs which are between {self._time_low} - {self._time_high}")
 
-        self._selectors = []                   
+        self._selectors = []
         self.add_selectors(self._check_for_valid_information_in_event_info)
         self.add_selectors(selectors, select_triggers)
 
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -207,7 +207,6 @@ class readRNOGData:
         # Initialize run table for run selection
         self.__run_table = None
 
-        self.__temporary_dirs = []
         if load_run_table:
             if run_table_path is None:
                 try:
@@ -891,12 +890,6 @@ class readRNOGData:
                 f"\n\tRead {self.__counter} events   (skipped {self.__skipped} events, {self.__invalid} invalid events)"
                 f"\n\tTime to initialize data sets  : {self._time_begin:.2f}s"
                 f"\n\tTime to read all events       : {self._time_run:.2f}s")
-
-        # Clean up links and temporary directories.
-        for d in self.__temporary_dirs:
-            self.logger.debug(f"Remove temporary folder: {d}")
-            os.unlink(os.path.join(d, "combined.root"))
-            os.rmdir(d)
 
 
 ### we create a wrapper for readRNOGData to mirror the interface of the .nur reader
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -360,7 +360,7 @@ class readRNOGData:
                                  f"\n\tSelect runs with max. trigger rate of {max_trigger_rate / units.Hz} Hz"
                                  f"\n\tSelect runs which are between {self._time_low} - {self._time_high}")
 
-        self.add_selectors([self._meta_selector, self._check_for_valid_information_in_event_info, self._select_events])
+        self.add_selectors([self._meta_selector, self._check_for_valid_information_in_event_info])
         self.add_selectors(selectors, select_triggers)
 
         # Read data
@@ -552,7 +552,7 @@ class readRNOGData:
         return dataset
 
 
-    def _select_events(self, evtinfo, event_idx=None):
+    def _select_events(self, evtinfo):
         """ Filter an event base on its EventInfo and the configured selectors.
 
         Parameters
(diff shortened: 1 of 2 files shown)


NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -250,7 +250,8 @@ def begin(self,
             run_time_range=None,
             max_trigger_rate=0 * units.Hz,
             mattak_kwargs={},
-            overwrite_sampling_rate=None):
+            overwrite_sampling_rate=None,
+            max_in_mem = 256):
```
fschlueter, line 254: Add an explanation in the doc-string for this argument please

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -724,29 +728,20 @@ def run(self):
         for dataset in self._datasets:
             dataset.setEntries((0, dataset.N()))
 
-            # read all event infos of the entier dataset (= run)
-            event_infos = dataset.eventInfo()
-            wfs = None
-
-            for idx, evtinfo in enumerate(event_infos):  # returns a list
+            # read all event infos of the entire dataset (= run)
+            for evtinfo, wf in dataset.iterate(start = 0, stop = dataset.N(),
```
fschlueter, line 732: if you use `setEntries` you do not have to specify start and stop.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -724,29 +728,20 @@ def run(self):
...
+                                               max_entries_in_mem = self._max_in_mem):
                 event_idx += 1
 
-                self.logger.debug(f"Processing event number {event_idx} out of total {self._n_events_total}")
-                t0 = time.time()
-
-                if self._filter_event(evtinfo, event_idx):
-                    continue
+                self.logger.debug(f"Processing (filtered) event number {event_idx} out of (unfiltered) total {self._n_events_total}")
+                t0 =time.time()
```
fschlueter, line 739: space is missing. Although I am not sure if that still makes sense. Because the reading of the waveforms which is the thing which takes time is now not registered by this.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -724,29 +728,20 @@ def run(self):
...
 
                 if not self._check_for_valid_information_in_event_info(evtinfo):
-                    continue
-
-                # Just read wfs if necessary
-                if wfs is None:
-                    wfs = dataset.wfs()
-
-                waveforms_of_event = wfs[idx]
+                    continue
```
fschlueter, line 742: Can you try passing this function as selector? The the `_filter_event` function also. That should speed things up

CamphynR: Ok, but I had to invert the logic in the _filter_event function to mimic the selectors logic i.e. I changed it to return True if the event is to be read and False if the event should be skipped.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
CamphynR: Yes I leave it out for now, I guess for a true timing we would need to time in the mattak iterator itself

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -316,6 +316,10 @@ def begin(self,
             (i.e., stored in the mattak files) only when the stored sampling rate is invalid (i.e. 0 or None).
             If None, nothing is overwritten and the sampling rate from the mattak file is used. (Default: None)
             NOTE: This option might be necessary when old mattak files are read which have this not set.
+
+        max_in_mem: int
+            Set the maximum number of events that can be stored in memory. The datareader will divide
+            the data in batches based on this number.
```
fschlueter, line 322: please mention that this is only relevant for the uproot backend

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -361,6 +365,7 @@ def begin(self,
         # Read data
         self._time_begin = 0
         self._time_run = 0
+        self._event_idx = 0
```
fschlueter, line 368: can you add a comment `# only for logging` at the end of the line

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -556,17 +563,18 @@ def _filter_event(self, evtinfo, event_idx=None):
         -------
 
         skip: bool
-            Returns True to skip/reject event, return False to keep/read event
+            Returns False to skip/reject event, return True to keep/read event
```
fschlueter, line 566: Nice catch! Do you think we should slightly rename the function to better describe this change. For example: `_select_events`

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -638,6 +646,9 @@ def _check_for_valid_information_in_event_info(self, event_info):
         is_valid: bool
             Returns True if all information valid, false otherwise
         """
+        # Count events processed here since this function will always be called to check event validity
+        self._event_idx += 1
```
fschlueter, line 650: This I do not like. This is to hidden. Cant we put it in some "main" function.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -724,30 +735,27 @@ def run(self):
...
         """
-        event_idx = -1
+        self._event_idx = 0
         for dataset in self._datasets:
             dataset.setEntries((0, dataset.N()))
 
+            # _filter_event is a wrapper for the selectors defined in begin()
+            selectors = [self._filter_event, self._check_for_valid_information_in_event_info] if len(self._selectors) !=0 \
+                   else [self._check_for_valid_information_in_event_info]
+
```
fschlueter, line 745: should those be merged with the selectors you can pass as argument to this class?

fschlueter:
@CamphynR what do you think if we slightly rewrite the function `set_selectors` to `add_selectors`? Instead of setting `self._selectors` one would add selectors to the internal list and use this one to pass to iterate.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
fschlueter: Having some kind of timing would be nice but we can add this later again.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
CamphynR: What do you mean?

fschlueter: Should they be merged with 
```       
selectors: list of Callables
      List of Callable(eventInfo) -> bool to pass to mattak.Dataset.iterate to select events.
```

CamphynR: Ow yeah that can cause confusion I have renamed it

CamphynR:
I'm not sure I understand the benefit of doing that. Do you want to bypass the _select_events function?

fschlueter:
@CamphynR I took the liberty to push some changes to this PR. I have to apologies, my previous comment was confusing because I forgot what `select_event` actually does...

fschlueter:
Okay, I think I found a slightly simpler solution. Now the new meta_selector is not needed because the logging is done in `select_events`. I actually do not have time to test it. Would you?

CamphynR:
Sure, I moved some logging and initialized the empty selectors list in begin, this should pass the tests

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -552,17 +567,20 @@ def _filter_event(self, evtinfo, event_idx=None):
         -------
 
         skip: bool
-            Returns True to skip/reject event, return False to keep/read event
+            Returns False to skip/reject event, return True to keep/read event
         """
+        self._event_idx += 1
```
fschlueter, line 572: are we not double counting here? I would prefer not having this counter in this function

fschlueter (approved):

PR #692, opened 2024-07-01
Extend rnog reader

fschlueter:
This adds a connivence function to the reader which I assume a lot of people would can make a lot of use of. It also may violates a bit the concept of a reader in NuRadio but we are pirates arent we? @sjoerd-bouma do you agree?

### DIFF
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -730,9 +730,9 @@ class readRNOGData:
 
             # read all event infos of the entire dataset (= run)
             for evtinfo, wf in dataset.iterate(start = 0, stop = dataset.N(),
-                                                              calibrated = self._read_calibrated_data,
-                                                              selector = self._selectors,
-                                                              max_entries_in_mem = self._max_in_mem):
+                                               calibrated = self._read_calibrated_data,
+                                               selectors = self._selectors,
+                                               max_entries_in_mem = self._max_in_mem):
                 event_idx += 1
 
                 self.logger.debug(f"Processing (filtered) event number {event_idx} out of (unfiltered) total {self._n_events_total}")
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -149,6 +149,7 @@ def _all_files_in_directory(mattak_dir):
     req_files = ["daqstatus.root", "headers.root", "pedestal.root"]
     for file in req_files:
         if not os.path.exists(os.path.join(mattak_dir, file)):
+            logging.error(f"File {file} could not be found in {mattak_dir}")
             return False
 
     return True
@@ -361,7 +362,7 @@ class readRNOGData:
                                  f"\n\tSelect runs with max. trigger rate of {max_trigger_rate / units.Hz} Hz"
                                  f"\n\tSelect runs which are between {self._time_low} - {self._time_high}")
 
-        self._selectors = []                   
+        self._selectors = []
         self.add_selectors(self._check_for_valid_information_in_event_info)
         self.add_selectors(selectors, select_triggers)
 
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -207,7 +207,6 @@ class readRNOGData:
         # Initialize run table for run selection
         self.__run_table = None
 
-        self.__temporary_dirs = []
         if load_run_table:
             if run_table_path is None:
                 try:
@@ -891,12 +890,6 @@ class readRNOGData:
                 f"\n\tRead {self.__counter} events   (skipped {self.__skipped} events, {self.__invalid} invalid events)"
                 f"\n\tTime to initialize data sets  : {self._time_begin:.2f}s"
                 f"\n\tTime to read all events       : {self._time_run:.2f}s")
-
-        # Clean up links and temporary directories.
-        for d in self.__temporary_dirs:
-            self.logger.debug(f"Remove temporary folder: {d}")
-            os.unlink(os.path.join(d, "combined.root"))
-            os.rmdir(d)
 
 
 ### we create a wrapper for readRNOGData to mirror the interface of the .nur reader
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -360,7 +360,7 @@ class readRNOGData:
                                  f"\n\tSelect runs with max. trigger rate of {max_trigger_rate / units.Hz} Hz"
                                  f"\n\tSelect runs which are between {self._time_low} - {self._time_high}")
 
-        self.add_selectors([self._meta_selector, self._check_for_valid_information_in_event_info, self._select_events])
+        self.add_selectors([self._meta_selector, self._check_for_valid_information_in_event_info])
         self.add_selectors(selectors, select_triggers)
 
         # Read data
@@ -552,7 +552,7 @@ class readRNOGData:
         return dataset
 
 
-    def _select_events(self, evtinfo, event_idx=None):
+    def _select_events(self, evtinfo):
         """ Filter an event base on its EventInfo and the configured selectors.
 
         Parameters
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -569,9 +569,9 @@ class readRNOGData:
         skip: bool
             Returns False to skip/reject event, return True to keep/read event
         """
-        self._event_idx += 1
         self.logger.debug(f"Processing event number {self._event_idx} out of total {self._n_events_total}")
 
+        self.__counter += 1  # for logging
         if self._selectors is not None:
             for selector in self._selectors:
                 if not selector(evtinfo):
@@ -749,9 +749,8 @@ class readRNOGData:
 
                 evt = self._get_event(evtinfo, wf)
 
-                self.__counter += 1
                 yield evt
-                
+
 
     def get_event_by_index(self, event_index):
         """ Allows to read a specific event identifed by its index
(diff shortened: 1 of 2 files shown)


sjoerd-bouma:
I can definitely see the use case for this. The only reservation I have right now is that I'm not 100% sure what the difference is between the waveforms obtained this way and the waveforms obtained from the `run` function... Right now, am I correct in saying that `get_waveforms` will always return uncalibrated waveforms in ADC and with block offsets? Although I guess this will change once/if calibration gets implemented in the mattak reader?

If that's the case, it might be good to put some kind of caveat in the docstring and/or rename the function to `get_raw_waveforms` so the user doesn't get something they don't expect. And if it is due to change in the future I could also see this potentially causing issues in analysis code down the line... another alternative is to hide this function (by prepending a single `_`) so it's clear this is not a supported method that won't break compatibility, but is still usable by power users that (think they) know what they're doing. Though I think it's useful enough that it would be nice if this is not necessary.

Those are my thoughts, if you think I'm overthinking this I think I'm okay with just pressing approve, but let me know what you think.

fschlueter:
@sjoerd-bouma brilliant comment as always. I haven't thought about this being so in the tunnel of looking at raw waveforms.
Let me see what can be done ...

fschlueter:
What do you say now?

sjoerd-bouma:
Couple more thoughts:
- I am not sure how I feel about having a different block offset removal algorithm for `get_waveforms` - I know you put a warning in the docstring, but this is still going to be a small difference that is easy to overlook when troubleshooting. The block offsetter code has a standalone function; I've tried implementing it in `get_waveforms`, and it works, but it's around a factor 2 slower than the median subtraction (in my recollection the difference was less - maybe I didn't benchmark properly last time? :see_no_evil:). 
If we can live with the performance impact, it might be preferable to stick with this, to avoid this discrepancy. I can push this if you agree.
- A previous iteration of the RNO-G data reader had a similar feature to `get_waveforms`, but rather than return an array of all waveforms, it was an iterator over a user-specified number of waveforms (e.g. it yielded them in sets of 1000). The risk with the current implementation is that the user asks for the waveforms of 100000 events, which will obviously result in memory overflow and a crash. This is not necessarily a problem but does affect the use cases somewhat.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -640,6 +646,52 @@ def get_events_information(self, keys=["station", "run", "eventNumber"]):
...
+                    if remove_block_offsets:
+                        wfs = _baseline_correction(wfs, warn=False)
+
+                    if self._read_calibrated_data:
+                        wfs *= units.mV
+                    else:
+                        # wf stores ADC counts
+                        if self._convert_to_voltage:
+                            # convert adc to voltage
+                            wfs *= (self._adc_ref_voltage_range / (2 ** (self._adc_n_bits) - 1))
```
sjoerd-bouma, line 689: The danger of recycling old code... is reintroducing old bugs ;)
`*=` does not work if no block offsets are removed, because the left hand side is an array of `int`s, so you cannot store floats in it. `wfs = wfs * units.mV` and similar on line 689 will work.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -640,6 +646,52 @@ def get_events_information(self, keys=["station", "run", "eventNumber"]):
...
+        Returns
+        -------
+
+        wfs: np.array
+            Waveforms of all "selected" events. The wavefroms are either calibrated or not
+            based on the class config.
+        """
+
+        if self._events_waveforms is None:
+            self._events_waveforms = []
```
sjoerd-bouma, line 671: I can kind of see the appeal of 'caching' the result in the class? But this does compound the risk of using a lot of memory (see also my earlier comment), and in the current implementation it means any arguments passed to `get_waveforms` are ignored after the initial call (e.g. if I first get the waveforms without block offset removal then the second time I call the function I will get them without block offset removal regardless of what argument I pass to the function).

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
fschlueter: uff good catch!

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
fschlueter: In the workflow I have been using so far it came in handy but you have good arguments.

fschlueter:
> A previous iteration of the RNO-G data reader had a similar feature to get_waveforms, but rather than return an array of all waveforms, it was an iterator over a user-specified number of waveforms (e.g. it yielded them in sets of 1000). The risk with the current implementation is that the user asks for the waveforms of 100000 events, which will obviously result in memory overflow and a crash. This is not necessarily a problem but does affect the use cases somewhat.

I would leave this as a future feature which people can implement when they want.

fschlueter:
> I am not sure how I feel about having a different block offset removal algorithm for get_waveforms - I know you put a warning in the docstring, but this is still going to be a small difference that is easy to overlook when troubleshooting. The block offsetter code has a standalone function; I've tried implementing it in get_waveforms, and it works, but it's around a factor 2 slower than the median subtraction (in my recollection the difference was less - maybe I didn't benchmark properly last time? 🙈).
If we can live with the performance impact, it might be preferable to stick with this, to avoid this discrepancy. I can push this if you agree.

It might be that on a single trace its less but I assume your solution uses a for loop to loop over all traces? Can you commit your implementation? And can you maybe allow my old block_removeal as a 4th option?

sjoerd-bouma:
I've undeprecated (dedeprecated? precated?) your median subtraction code. `get_waveforms` should now do exactly the same as if you went through `get_event` (there is some code duplication, unfortunately...), but has the option to overwrite the `apply_baseline_correction` variable if the user wants more precision or more performance for whatever reason. I also implemented a check on which baseline correction was used so the variable does not get ignored on subsequent calls (if the user added additional `selectors` since the last cache, I guess this will still return the cached (incorrect) waveforms however...).

I've also added a warning if the user requests more than 2000 events, which depending on the selectors corresponds to up to 1 GB of waveforms. The warning doesn't do anything, i.e. I still anticipate this causing a lot of crashes, but at least it should be clear to the user what they did wrong.

sjoerd-bouma (dismissed):
You should probably have a brief look but I think I've addressed most of my initial reservations.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -640,6 +638,78 @@ def get_events_information(self, keys=["station", "run", "eventNumber"]):
 
         return self._events_information
 
+    def get_waveforms(self, apply_baseline_correction=None):
+        """ Return waveforms of events passing the selectors which may have been specified
+
+        Note that
```
fschlueter, line 644: ?

fschlueter:
I thought you wanted to remove the  caching of the waveforms?

fschlueter:
```if (apply_baseline_correction is None) and not self._read_calibrated_data:```

I am not sure that this is correct? Do we understand the cause of the offsets? If the offsets are actual voltage fluctuations it would (maybe) correct to first apply a voltage calibration and than remove the block offsets.

sjoerd-bouma:
> I thought you wanted to remove the caching of the waveforms?

I thought you wanted to keep it :smile: I think it should reset when in all the cases I was worried about by now, but I can also remove it anyway - it would halve the memory requirement. Let me know what you prefer.

sjoerd-bouma:
> `if (apply_baseline_correction is None) and not self._read_calibrated_data:`
> 
> I am not sure that this is correct? Do we understand the cause of the offsets? If the offsets are actual voltage fluctuations it would (maybe) correct to first apply a voltage calibration and than remove the block offsets.

I think you're right, I just kept the order in your code I think but I can swap this around. The difference will probably be negligible compared to our precision anyway.

fschlueter:
> > I thought you wanted to remove the caching of the waveforms?
> 
> I thought you wanted to keep it 😄 I think it should reset when in all the cases I was worried about by now, but I can also remove it anyway - it would halve the memory requirement. Let me know what you prefer.

Do as you prefer. No strong opinions

fschlueter:
> > `if (apply_baseline_correction is None) and not self._read_calibrated_data:`
> > I am not sure that this is correct? Do we understand the cause of the offsets? If the offsets are actual voltage fluctuations it would (maybe) correct to first apply a voltage calibration and than remove the block offsets.
> 
> I think you're right, I just kept the order in your code I think but I can swap this around. The difference will probably be negligible compared to our precision anyway.

What do you mean with odering? I would just remove the `not self._read_calibrated_data:` from the if condition

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
sjoerd-bouma: !

sjoerd-bouma:
> > > `if (apply_baseline_correction is None) and not self._read_calibrated_data:`
> > > I am not sure that this is correct? Do we understand the cause of the offsets? If the offsets are actual voltage fluctuations it would (maybe) correct to first apply a voltage calibration and than remove the block offsets.
> > 
> > 
> > I think you're right, I just kept the order in your code I think but I can swap this around. The difference will probably be negligible compared to our precision anyway.
> 
> What do you mean with odering? I would just remove the `not self._read_calibrated_data:` from the if condition

Actually (as in the docstring), block offsets would only be removed if `read_calibrated_data` was False. The idea, if I recall correctly, was that block offset removal would be handled by mattak in the calibration process. As this is not currently the case, I actually agree that it probably makes more sense for the block offset removal to happen regardless of the value of `read_calibrated_data`, so I changed it. Maybe I ought to also put this change in the changelog, but I don't think many people have been reading in calibrated data anyway until now?

sjoerd-bouma:
Other changes:
- added a `max_events` parameter to `get_waveforms`; this should ensure users have to work a little bit harder to get memory overflow crashes. If you don't care about memory and just always want all event waveforms to be returned you can pass `max_events=None`.
- Switched to using Python's lru_cache to cache the waveforms rather than trying to keep track of everything manually. The cache is cleared every time `add_selectors` is called (including in the `begin` function).

fschlueter (commented):
Looks fine to me (can not approve since I opened the PR)

sjoerd-bouma (approved):

fschlueter:
ups. Sorry forgot about the 24h rule

Issue #331, opened 2021-10-18
RNO-G data reader breaks after uproot version 4.1.2


dansmithphysics:
When using the RNO-G data reader with uproot version greater than 4.1.2 (current version is 4.1.5), the following error is throw:

`raise ValueError(
ValueError: basket 0 in tree/branch /hdr;1:hdr/TObject/fBits has the wrong number of entries (expected 636, obtained 2544) when interpreted as AsDtype('uint8')
    in file ./run476/headers.root`

It looks like this is a bug in uproot itself, fBits should be a UInt_t, a 4 byte integer but uproot is trying to fit the data into a 1 byte integer. Between uproot version 4.1.2 and 4.1.3, there was a change in how branches were iterated over that has known issues for data types. Will follow up with uproot to see if I can get the bug fixed. In the mean time, using version uproot 4.1.2 resolves the issue.

sjoerd-bouma:
This does not seem to happen on newer versions of uproot anymore - not sure which version fixed it exactly, but I think this can be closed.

PR #694, opened 2024-07-04
unpin awkward and uproot

sjoerd-bouma:
It seems #331 has been fixed (if I recall correctly, this only affected the reading of RNO-G files, which seems to work again now according to the tests), and, with the release of numpy 2.0 (which is not compatible with older versions of awkward/uproot), I think we should unpin uproot and awkward again so that everyone can use the latest version of numpy.


fschlueter (approved):
I assume you tested it?

sjoerd-bouma:
> I assume you tested it?

I think I did, and so did the unit tests - seems to work fine,

PR #696, opened 2024-07-09
Feature/rnog update vpol response

blhendricks:
Add current XFdtd Vpol model RVEL

### DIFF
--- a/NuRadioReco/detector/antenna_models_hash.json
+++ b/NuRadioReco/detector/antenna_models_hash.json
@@ -47,6 +47,7 @@
   "RNOG_vpol_4inch_half_n1.73.pkl": "f08d263233503f79e5be10c18bc7587b8b421c1d",
   "RNOG_vpol_4inch_center_n1.73.pkl": "5f429ed9ed08175a7f75fd44422367d2278bf2e1",
   "RNOG_vpol_4inch_wall_n1.73.pkl": "da24017eb80f7a68348a674be2c527457fe19992",
+  "RNOG_vpol_v2_5inch_center_n1.75.pkl":"b589906a6871bb093fdded9dcb4369631031f5d4",
   "RNOG_quadslot_v1_n1.74.pkl": "e9c4946dd2176489249c862f32ef272ab627dc28",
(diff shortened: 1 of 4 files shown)


cg-laser (dismissed):
I copied the antenna model to the DESY server. The PR can be merged now.

sjoerd-bouma:
Thanks @blhendricks and @cg-laser! There were a couple of hyphens (rst wants the number of `-` to be the same length as the title), and `2.3.0-dev` is unreleased, so there was no need to update the version number, though I realize the instructions in the documentation are unclear/wrong on this so that should be updated at some point.

sjoerd-bouma (approved):

fschlueter:
Are these new simulations what we should use in the future? Than we have to add it to the hardware database

anelles:
Christian wrote that they are added to the antenna model server, which should already be enough. 
@blhendricks should tells us whether these are the new defaults. 
Also, Bryan, please merge this whenever you are ready. NuRadioMC policy is that every PR is approved by someone else and then 24 hours later, whoever started the PR should merge.

blhendricks:
Yes, this should be the new default for the Vpols.

fschlueter:
@jakobhenrichs

fschlueter:
@blhendricks can you link some resources or post some plots which show the difference between old and new?

blhendricks:
eff_length_vpol_model_compare

Here is a plot showing the difference between the old and new vpol model. The old model is from the old 4 inch diameter, 50 cm long vpols simulated in a 5.75 inch borehole. The new one is from Dan's sims of the currently deployed 5 inch diameter, 60 cm long vpol model in the 11.2 inch borehole.

When the change to a 11.2 inch diameter borehole was decided, Dan redesigned the vpols to recover the lower frequencies lost from not coupling to the ice as well. Changes to the feed were also apparently made, but I don't know the details of those changes. He had these (current model) simulations before he left but apparently they were never added to nuradiomc, as Brian recently reached out to me about.

There isn't a dramatic difference between them at most frequencies (though differences are more apparent off-boresight and at higher freqs), but I'm assuming since the up-to-date model wasn't included in nuradiomc, all prior analyses have been done with the outdated model?

PR #697, opened 2024-07-10
use fixed noise and pulse windows for SNR calculation

PhilippLaub:
There was a bug in the calculation of the trace SNR, where strong pulses would be assigned a low SNR and vice versa. This is suspected to originate from the pulse and noise windows used for the calculation of the SNR. 

This fix uses fixed windows for the noise, so that the tapered edges of the LOFAR traces are excluded from the noise RMS calculation, preventing the noise RMS to blow up eventually. It is furthermore assumed that the pulse is located somewhat close to the middle of the trace and the noise window is chosen accordingly so that the pulse is not included in the noise RMS calculation. The pulse window for now is the full trace, since only the maximum of the hilbert envelope is used for the pulse calculation.

### DIFF
--- a/NuRadioReco/modules/LOFAR/stationPulseFinder.py
+++ b/NuRadioReco/modules/LOFAR/stationPulseFinder.py
@@ -275,7 +275,7 @@ class stationPulseFinder:
             station_even_list = []
             station_odd_list = []
             for channel in station.iter_channels():
-                if channel.get_id() == int(channel.get_group_id()[1:]):
+                if channel.get_id() == channel.get_group_id():
                     station_even_list.append(channel.get_id())
                 else:
                     station_odd_list.append(channel.get_id())
--- a/NuRadioReco/modules/LOFAR/stationPulseFinder.py
+++ b/NuRadioReco/modules/LOFAR/stationPulseFinder.py
@@ -64,7 +64,7 @@ class stationPulseFinder:
 
         self.direction_cartesian = None  # The zenith and azimuth pointing towards where to beamform.
 
-    def begin(self, window=500, noise_window=10000, cr_snr=3, good_channels=6, logger_level=logging.WARNING):
+    def begin(self, window=800, noise_window=10000, cr_snr=3, good_channels=6, logger_level=logging.WARNING):
         """
         Sets the window size to use for pulse finding, as well as the number of samples away from the pulse
         to use for noise measurements. The function also defines what an acceptable SNR is to consider a
@@ -128,6 +128,7 @@ class stationPulseFinder:
             all_real_traces = np.array([station.get_channel(channel).get_trace() for channel in channel_ids])#debug
             beamed_fft = mini_beamformer(all_spectra, frequencies, channel_positions, self.direction_cartesian)
             beamed_timeseries = fft.freq2time(beamed_fft, sampling_rate, n=station.get_channel(channel_ids[0]).get_trace().shape[0])
+            optimal_amp = np.sum( [np.max(np.abs(trace)) for trace in all_real_traces] )
 
             analytic_signal = hilbert(beamed_timeseries)
             amplitude_envelope = np.abs(analytic_signal)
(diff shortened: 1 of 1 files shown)


NuRadioReco/modules/LOFAR/stationPulseFinder.py
```diff
@@ -275,7 +275,7 @@ def run(self, event, detector):
             station_even_list = []
             station_odd_list = []
             for channel in station.iter_channels():
-                if channel.get_id() == channel.get_group_id():
+                if channel.get_id() == int(channel.get_group_id()[1:]):
```
MijnheerD, line 278: On the lofar/development branch, the channel group ID is already an integer :) So this can be reverted

NuRadioReco/modules/LOFAR/stationPulseFinder.py
```diff
@@ -299,11 +299,21 @@ def run(self, event, detector):
...
+
+            # there was a bug with the windows for the SNR calculation, where strong signals had low SNR and vice versa. 
+            # Since the pulse should be somewhere around the middle of the trace, we use a fixed noise window (without the tapered edges of the trace). 
+            # For the signal window, the whole trace is used since we need the maximum. 
+            # Narrowing this down may increase performance and may be worth looking into in the future.
+            self._check_station_triggered(
+                station, 
+                position_array, 
+                ant_same_orientation, 
+                signal_window=[0, -1],
```
MijnheerD, line 314: Why can't we keep the window calculation as before? I thought we only needed to change the noise windows to not include the tapered edges?

MijnheerD (changes requested):

NuRadioReco/modules/LOFAR/stationPulseFinder.py
PhilippLaub: Agreed, will do that :)

NuRadioReco/modules/LOFAR/stationPulseFinder.py
PhilippLaub: Apparently not because then strong signals would still have high SNRs. But I found that strong signals had low SNRs, so there must be some issue with the peak amplitude. I suspect this comes from the signal window.

MijnheerD: I'm sorry, but it is not clear to me whether you tried to keep the signal window as before, and only change the noise window?

MijnheerD:
Could you have a look at the implementation in `_signal_windows_polarisation` of the same class and see if you can figure out why it is not producing the desired results?

NuRadioReco/modules/LOFAR/stationPulseFinder.py
```diff
@@ -261,7 +262,7 @@ def run(self, event, detector):
             The detector related to the event.
         """
         zenith = event.get_hybrid_information().get_hybrid_shower("LORA").get_parameter(showerParameters.zenith)
-        azimuth = event.get_hybrid_information().get_hybrid_shower("LORA").get_parameter(showerParameters.azimuth)
+        azimuth = np.mod( event.get_hybrid_information().get_hybrid_shower("LORA").get_parameter(showerParameters.azimuth), 2* np.pi )
```
MijnheerD, line 265: Is this really necessary or just a remain of trying different solutions?

NuRadioReco/modules/LOFAR/stationPulseFinder.py
```diff
@@ -275,7 +276,8 @@ def run(self, event, detector):
             station_even_list = []
             station_odd_list = []
             for channel in station.iter_channels():
-                if channel.get_id() == channel.get_group_id():
+                # if channel.get_id() == channel.get_group_id(): #tested it, does not work
+                if channel.get_id() == int(channel.get_group_id()[1:]): #this works
```
MijnheerD, line 280: I am sorry, but this cannot be correct. In `readLOFARData.py` we use this line: 
```python
channel_group: int = detector.get_channel_group_id(station_id, channel_id)
```
So `channel.get_group_id()` should return an integer and not a string. Do you perhaps test your code using old NUR files which were processed with the old reader code?

PhilippLaub: Ah, yes that is possible. Ok, I will revert this again.

NuRadioReco/modules/LOFAR/stationPulseFinder.py
PhilippLaub: I think it is just a remainder. I will remove it.

MijnheerD (approved):

MijnheerD:
Thanks for all the work! You can merge in 24h in case nobody has any other comments :)

PR #698, opened 2024-07-19
Hotfix/read rnog data select events

sjoerd-bouma:
A minor bugfix in `get_event_by_index` which wasn't updated to reflect the new signature of `_select_events`. And unrelatedly a new feature which I always wanted to have, which is the option to read all keys in `EventInfo` without having to remember and type the correct names of all of them.


fschlueter (approved):
Thanks Sjoerd!

PR #703, opened 2024-08-01
exclude readARAData from documentation because of missing .so files

sjoerd-bouma:
Excludes the `.araroot` module from the online documentation, because sphinx tries to import all modules it documents, and this raises an error for the araroot module due to missing `.so` files.

Annoyingly, sphinx _would_ be able to document this module perfectly if we wrapped the .so-loading in a try/except block, but that would get rid of the error when trying to import the module without the .so files, which I think is behaviour we want to keep.


fschlueter (approved):
Thanks Sjoerd!

Issue #699, opened 2024-07-30
Incorrect phasing in the noise-gen script


colemanalan:
https://github.com/nu-radio/NuRadioMC/blob/19e807e04a4aa3ddeda5e06c4008a94f15f3ec37/NuRadioReco/utilities/noise.py#L105-L110

This part of the for loop doesn't actually do what we want. Both if-statements do the same roll so that a delay of e.g. `5` is the same as `-5`.

Example `traces[i] == [0, 1, 2, 3, 4, 5]`

if `r = +2`, we want output to be `[4, 5, 0, 1, 2, 3]`
 * Line 106: `[2, 3, 4, 5, 0, 0]`
 * Line 107: `[2, 3, 4, 5, 0, 1]`

if `r = -2`, we want output to be `[2, 3, 4, 5, 0, 1]`
 * Line 109: `[2, 3, 4, 5, 0, 0]`
 * Line 110: `[2, 3, 4, 5, 0, 1]`

I guess the direction that things get moved for positive values of `r`. But I don't think that we want both positive and negative delays should give the same result.

----
For comparison, if we use numpy's `roll`, one gets

```python
In [1]: np.roll([i for i in range(6)], 2)
Out[1]: array([4, 5, 0, 1, 2, 3])
In [2]: np.roll([i for i in range(6)], -2)
Out[2]: array([2, 3, 4, 5, 0, 1])
```

---
Since we want to use the switch at
https://github.com/nu-radio/NuRadioMC/blob/19e807e04a4aa3ddeda5e06c4008a94f15f3ec37/NuRadioReco/utilities/noise.py#L533-L537
it is probably preferred that we use numpy def (also what is used in the actual PA trigger module for simulations). So we could fix this by removing the `elif` loop and define `r = -rolling[i]` to get the same result.

PR #705, opened 2024-08-01
Incorrect phasing in the noise-gen script

colemanalan:
Issue being addressed:
Solves issue #699

### DIFF
--- a/NuRadioReco/utilities/noise.py
+++ b/NuRadioReco/utilities/noise.py
@@ -98,6 +98,9 @@ def rolled_sum_slicing(traces, rolling):
         1D array of summed traces
     """
 
+    if rolling[0]:
+        raise RuntimeError(f"Cannot have a roll value of {rolling[0]}!=0 for channel 0")
+
     # assume first trace always has no rolling
     sumtr = traces[0].copy()
     for i in range(1, len(traces)):
--- a/NuRadioReco/utilities/noise.py
+++ b/NuRadioReco/utilities/noise.py
@@ -102,10 +102,7 @@ def rolled_sum_slicing(traces, rolling):
     sumtr = traces[0].copy()
     for i in range(1, len(traces)):
         r = rolling[i]
-        if r > 0:
-            sumtr[:-r] += traces[i][r:]
-            sumtr[-r:] += traces[i][:r]
-        elif r < 0:
+        if r != 0:
             sumtr[:r] += traces[i][-r:]
             sumtr[r:] += traces[i][:-r]
         else:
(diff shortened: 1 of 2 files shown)


NuRadioReco/utilities/noise.py
```diff
@@ -102,10 +102,7 @@ def rolled_sum_slicing(traces, rolling):
     sumtr = traces[0].copy()
     for i in range(1, len(traces)):
         r = rolling[i]
-        if r > 0:
-            sumtr[:-r] += traces[i][r:]
-            sumtr[-r:] += traces[i][:r]
-        elif r < 0:
+        if r != 0:
```
shallmann, line 105: Thanks for spotting and correcting. That does the correct thing now... 

A simple test would have avoided this in the first place :/
I verified it works for rolling<trace length...that should not happen, but one could put
`r = rolling[i]%len(sumtr)` for completeness (with that, `r` would never have been negative in the first place)

I used this to test:
```
def test_rolling_functions(n_tests=100, n_traces=5):
    """ tests for equality of output of `rolled_sum_roll` and `rolled_sum_slicing` """
    traces = [np.random.randint(-10,11,100) for i in range(n_traces)]

    for i in range(n_tests):
        rollings = np.random.randint(-100,100,n_traces)
        rollings -= rollings[0]

        assert np.alltrue(rolled_sum_roll(traces, rollings) == rolled_sum_slicing(traces, rollings)), "not equal"
```

NuRadioReco/utilities/noise.py
```diff
@@ -331,7 +328,8 @@ def __init__(self, detector_filename, station_id, triggered_channels,
             delays -= np.max(delays)
 
             roll = np.array(np.round(np.array(delays) * self.sampling_rate * self.upsampling)).astype(int)
-            self.beam_time_delays[iBeam] = roll
+            # subtract off the delay of antenna 0 because `rolled_sum_slicing` assumes this is the case
+            self.beam_time_delays[iBeam] = roll - roll[0]
```
shallmann, line 332: yes perfect

shallmann (approved):

NuRadioReco/utilities/noise.py
colemanalan: Thanks for the double check!

changelog.txt
```diff
@@ -29,6 +29,7 @@ can then iterate over a channel group, which yields all channels for a given gro
 loaded again when reading in the Detector from a nur file
 - Added LOFAR coordinates to Detector site coordinates
 - Implemented mattak dataset iterator in readRNOGDataMattak.run()
+- Fixed bug in phased-array noise generation script
```
cg-laser, line 32: can you specify what exactly was wrong? Something like: "Previous version always generated triggered noise that triggered in half of the beams"

shallmann (approved):
Fine for me.

PR #706, opened 2024-08-02
Update HDF5_structure.rst

fschlueter:


fschlueter:
Given that this is just a minor change in the documentation I will merge it right away

PR #707, opened 2024-08-02
Refactor NuRadioReco/modules/efieldToVoltageConverter.py

fschlueter:
Moving some repetitive code in its own function.


NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -35,11 +34,12 @@ def __init__(self, log_level=logging.WARNING):
...
         self.__max_upsampling_factor = None
         self.antenna_provider = None
-        self.begin()
         self.logger = logging.getLogger('NuRadioReco.efieldToVoltageConverter')
         self.logger.setLevel(log_level)
+        self.begin()
+
 
-    def begin(self, debug=False, uncertainty=None,
+    def begin(self, debug=False, uncertainty={},
```
sjoerd-bouma, line 42: The reason many default arguments in NuRadioMC are `None` rather than the object it defaults to is because of [this](https://stackoverflow.com/questions/41686829/why-does-pycharm-warn-about-mutable-default-arguments-how-can-i-work-around-the). 
While you're probably okay here (I don't think `uncertainty` is modified inside the begin function?) it's probably good practice not to use mutable objects as default arguments in general, so I'd suggest leaving it the way it was.

fschlueter: Can we use or?

fschlueter: Is that okay @sjoerd-bouma

fschlueter:
ping

sjoerd-bouma:
Is it clear to you why the unit tests are failing? It's not clear to me, but they don't seem to fail on other branches as far as I can tell...

fschlueter:
Hm I am puzzled. I did not change anything?

fschlueter:
A miracle, the tests run through now ...

fschlueter:
@sjoerd-bouma could you do me the honour of approving?

fschlueter:
ping

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -74,20 +74,17 @@ def begin(self, debug=False, uncertainty=None,
...
         # some uncertainties are systematic, fix them here
-        if('sys_dx' in self.__uncertainty):
-            self.__uncertainty['sys_dx'] = np.random.normal(0, self.__uncertainty['sys_dx'])
-        if('sys_dy' in self.__uncertainty):
-            self.__uncertainty['sys_dy'] = np.random.normal(0, self.__uncertainty['sys_dy'])
-        if('sys_dz' in self.__uncertainty):
-            self.__uncertainty['sys_dz'] = np.random.normal(0, self.__uncertainty['sys_dz'])
-        if('sys_amp' in self.__uncertainty):
-            for iCh in self.__uncertainty['sys_amp'].keys():
+        self.__uncertainty = uncertainty or {}
```
sjoerd-bouma, line 79: I wonder what percentage of python users finds this more intuitive than the four-line equivalent... I'm not a convert yet.

sjoerd-bouma (approved):

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -267,3 +266,39 @@ def end(self):
         dt = timedelta(seconds=self.__t)
         self.logger.info("total time used by this module is {}".format(dt))
         return dt
+
+
+def calculate_time_shift_for_cosmic_ray(det, station, efield, channel_id):
```
sjoerd-bouma, line 271: Also not sure if it's worth making this function private, I guess we don´t expect it to be used on its own. But this is anyway not something we're very consistent about throughout NuRadio.

NuRadioReco/modules/efieldToVoltageConverter.py
fschlueter: if that is the case: ciao.

NuRadioReco/modules/efieldToVoltageConverter.py
fschlueter: next PR? Is this function useful for the other efield2Voltage converter modules?

sjoerd-bouma: Yeah I'm happy for this to be merged as is.

Issue #702, opened 2024-07-31
Definition of Fresnel coefficients


philippwindischhofer:
Hi all,

_(First of all, I have only recently started using NuRadioMC and my question below might simply be a consequence of me misunderstanding the code or its coordinate system. Apologies if that is the case.)_

My question is: does anybody happen to have a reference for where the Fresnel coefficients `r_p` and `r_s` (defined [here](https://github.com/nu-radio/NuRadioMC/blob/a56e0b0cc9106a0bad32069f8e5c60bb543491fa/NuRadioReco/utilities/geometryUtilities.py#L144-L171)) have been taken from? I am wondering particularly about the overall sign of `r_p` (where e.g. Jackson differs from NuRadioMC), and its phase shift in the case of total internal reflection. I attach more information below.

1. Sign of `r_p`: Looking at Eq. 7.60 in Jackson (page 220 [here](http://www.fisica.ugto.mx/~ggutj/CV/Classical_Electrodynamics_Jackson_1a_Edicion.pdf)) and converting to the same notation used in NuRadioMC, there seems to be an overall sign difference relative to NuRadioMC. The same is true for the definition given on [Wikipedia](https://en.wikipedia.org/wiki/Fresnel_equations#Complex_amplitude_reflection_and_transmission_coefficients). If anybody remembers the original reference that was used in the code, I'd love to take a look at it to understand this difference.

[...]

PR #708, opened 2024-08-04
Fresnel coefficients

philippwindischhofer:
These are the changes that are being discussed in #702


cg-laser:
I looked at some of the test printouts and I'm surprised by how much they fail, e.g, 

```
attribute max_amp_shower_and_ray not almost equal
[[1.15973026e-08 3.59956175e-01]
 [6.05853951e-08 4.99576163e-01]
 [1.71753687e-08 3.11614636e-01]
 [6.42167179e-08 4.93779904e-01]
 [8.38073202e-09 8.36464761e-08]
 [1.17274847e-08 7.49235866e-08]
 [6.83285122e-09 7.54794339e-08]
 [3.25724437e-09 7.58291203e-08]]
[[2.90221718e-05 2.77085818e-05]
 [3.15053776e-05 1.01170651e-05]
 [3.00007825e-05 2.80823000e-05]
 [3.14096781e-05 1.03940245e-05]
 [4.07112342e-05 3.36148530e-05]
 [3.98448441e-05 3.28416734e-05]
 [3.81546119e-05 3.15717069e-05]
 [3.89905849e-05 3.23170932e-05]]
[[2.90221721e-05 2.03746e-05]
 [3.15053795e-05 2.02169928e-05]
 [3.00007830e-05 2.14104808e-05]
 [3.96801e-05 2.05326193e-05]
 [4.07112345e-05 3.36148502e-05]
 [3.98448446e-05 3.28416710e-05]
 [3.81546121e-05 3.15717045e-05]
 [3.89905850e-05 3.23170907e-05]]
```
The first array shows the deviation, the second and the third the absolute numbers. first column is the first RT solution, second column the second RT solution. 
top 4 rows are 4 LPDAs at -1m, bottom 4 rows are 4 dipoles a -5m. 
No deviation for the direct trajectories and the second RT to the dipoles (likely because they are refracted rays). Only change is the reflected RT solutions to the LPDAs, so this is good. But why does the maximum amplitude change by 50%? Doesn't the change only result in a polarity flip? Then the maximum (of the magnitude) of the trace should remain the same.

cg-laser:
I double checked, the "maximum" is the maximum of the Hilbert envelope of the trace `channel_rt_data['max_amp_shower_and_ray'][iCh, iS] = sim_channel[chp.maximum_amplitude_envelope]`

philippwindischhofer:
The changes here result in a sign change only for `r_p` for non-TIR geometries (where the sign was inconsistent before). `r_s` does not pick up a sign flip (but only a complex conjugation). To the extent that the `E_phi` is important, this can cause a change in the signal amplitude.

I don't know the test scenario well enough to be able to tell if this is why the pipeline is failing or not ...

sjoerd-bouma:
I had a brief look at the SingleEvent test, and indeed the only difference I can see is that for the reflected signal the relative polarity of the theta and phi components has flipped; for some events/surface channels theta and phi are of similar magnitude, which would explain the large relative difference found by the test for some events (as was also concluded on the call today).
image


@philippwindischhofer / @cg-laser - I briefly lost connection to the call while we were discussing this, but are there any more tests you want to run, or shall I update the reference files? (I guess I would make a PR to `philippwindischhofer:phwindis_fresnel`?)

philippwindischhofer:
Thanks a lot for having a look @sjoerd-bouma !

From my perspective it's fine to update the reference files. @cg-laser said on the call that he also wants to have a last look at the output of the tests before we go ahead.

sjoerd-bouma:
Okay, then I'll update the reference files once #763 has been merged - I don't think it's worth Philipp spending the time to generate the new reference files which is not very transparent (though he is of course free to disagree :).

philippwindischhofer:
Hey @sjoerd-bouma! Sorry, this fell off my radar. I just checked and it looks like #763 is now merged. Is there anything I can / should do to get the reference files updated, or are you going to take care of it?

sjoerd-bouma:
I have made a PR to philippwindischhofer:phwindis_fresnel to update the test files. I think everything should be okay once that is merged. I did notice that for the trigger time tests, one trigger time was shifted by .4 ns (for a direct triggering pulse), which is slightly unusual, and for another event that triggered on a reflected pulse, the trigger time was shifted by exactly 1 ns. 

This is probably an artifact of how our triggering is implemented, but I haven't done the necessary digging through the code to make it make sense in my head - maybe someone who is more familiar with the details (@fschlueter  / @cg-laser ?) can confirm that this makes sense.

cg-laser:
> I have made a PR to philippwindischhofer:phwindis_fresnel to update the test files. I think everything should be okay once that is merged. I did notice that for the trigger time tests, one trigger time was shifted by .4 ns (for a direct triggering pulse), which is slightly unusual, and for another event that triggered on a reflected pulse, the trigger time was shifted by exactly 1 ns.
> 
> This is probably an artifact of how our triggering is implemented, but I haven't done the necessary digging through the code to make it make sense in my head - maybe someone who is more familiar with the details (@fschlueter / @cg-laser ?) can confirm that this makes sense.

I think we can always have small interference between both pulses which can explain slight shifts in trigger times. I would need to look at which channel (antenna depth) this actually happened. 
Are the test results available somewhere?

philippwindischhofer:
Looks like all tests have passed now! Thanks a lot @sjoerd-bouma !

philippwindischhofer:
Hey @sjoerd-bouma thanks for approving! Should I merge or should we wait for @cg-laser?

cg-laser (approved):

philippwindischhofer:
Thanks @cg-laser, merging now!

PR #715, opened 2024-08-07
Cleanup of functions in LOFAR beamforming utilities

MijnheerD:
I removed a bunch of old code in`[beamforming_utilities.py](https://github.com/nu-radio/NuRadioMC/compare/lofar/development...lofar/beamforming_utilities_cleanup?expand=1#diff-77e866206f274f79671989260ae2167b24b68e18b9e7225f6f2929ef4cab0566) and added the docstrings to go with the remaining functions. 

I also simplified and optimised the function that are still there, by using numpy array magic.


acorstanje:
Why was the 'build' test for the Detector not successful?

sjoerd-bouma:
> Why was the 'build' test for the Detector not successful?

Most likely an issue that has already been solved on the latest `develop` branch, you can make it go away by merging that in.

PR #716, opened 2024-08-09
Implement trigger_channel into the channel class. This allows to simulate trigger and readout channels with different hardware responses for the same physical channel/antenna.

fschlueter:
@cg-laser and @sjoerd-bouma I went ahead and started an implementation which will allow us to treat "trigger channels" better.

Originally I wanted to restructure the channel class to be basically an collection of base_traces. But I could not think of a solution which would keep the same interface as now for the first channel. 

Any thoughts?


fschlueter:
I think we will merge these changes with #743. If not I will reactivate this PR.

PR #717, opened 2024-08-20
Remove first if condition, the argparser will throw an error anyway. …

fschlueter:
…Improve doc-string

This mostly improves the doc-string. 

I have to say the code if `if len(args.files) == 1:` is hard to read. I imagine it was used for some specific library (Gen2?). Is it necessary to keep it where it is? Maybe we can store it in the analysis repo of the Gen2 analysis and simplify this code?

My intention: When I first looked at this script (and the previous doc-string) I thought the script is not suited for my purpose and would need quite some changes to be. I only later learned (when looking at a script by Steffen) that I indeed can use it simply by passing several inputs.

### DIFF
--- a/NuRadioMC/utilities/merge_hdf5.py
+++ b/NuRadioMC/utilities/merge_hdf5.py
@@ -242,20 +242,13 @@ if __name__ == "__main__":
         logger.setLevel(log_val)
 
     if len(args.files) == 1:
-        filenames = glob.glob("{}/*/*.hdf5.part????".format(args.files[0]))
-        filenames = np.append(filenames, glob.glob("{}/*/*.hdf5.part??????".format(args.files[0])))
-        filenames = sorted(filenames)
-        filenames2 = []
-        for i, filename in enumerate(filenames):
-            filename, ext = os.path.splitext(filename)
-            if(ext != '.hdf5'):
-                if(filename not in filenames2):
-                    d = os.path.split(filename)
-                    a, b = os.path.split(d[0])
-                    filenames2.append(filename)
-
+        root_directory = args.files[0]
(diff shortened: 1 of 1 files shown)


cg-laser (commented):
Thanks for improving the doc string. 
Why did you remove `f(len(args.files) < 1):` etc? Doesn't this script now crashes if you don't specify any file? Or is this validated by the argparser?

fschlueter:
> Thanks for improving the doc string. Why did you remove `f(len(args.files) < 1):` etc? Doesn't this script now crashes if you don't specify any file? Or is this validated by the argparser?

Exactly the argparser will `complain`.  `files` is a positional argument. So the argparser will exit the program with an appropriate error message

cg-laser:
Please keep the `==1` option. For most productions, NuRadioMC generates files in subfolders. So this command automatically merged parallel simulations runs into a single hdf5 file. The event generator has the option to automatically generate many input files based on the maximum number of events per file. If you use this feature in the event generation, you can use the `==1` feature of the hdf5_merge tool.

fschlueter:
@cg-laser okay, I kept it. I tried to simplify the code a bit. The only functional change is that I switched from `glob("...hdf5.part????")` and `glob("...hdf5.part??????")` to only `glob("...hdf5.part*")`. Is there a specific reason to allow only 4 and 6 symbols?

fschlueter:
ping

fschlueter:
Maybe @sjoerd-bouma can have a look.

fschlueter:
Can we merge that soon or close it?

anelles (approved):
I think it is fine to allow "*" instead of 4 or 6 symbols. Random coincidences seem unlikely.

PR #718, opened 2024-08-30
Fix bugs in get_cable_delay() get_time_delay() of the rnog_detector class ...

fschlueter:
… Some more modifications. Some fixes changes in the response class as well.

These changes are already included in the `rnog_trigger` and `rnog_trigger_3.0` branches. But it is crucial that the fixes get merged sooner to the develop branch. 

This PR also adds a few new features and modification to both classes. However, the main things are:

RNO-G detector:
- The function get_cable_delay returns now the time delay of the entire response chain, not just the cables.
- If one selects only cables, now also responses which contain "fiber" and "cable" are used.
- Take golden components correctly into account to get the total time delay

Response class:
- How to treat cases where one divides through very low amplifications. Avoid exploding responses when dividing through golden components.

### DIFF
--- a/NuRadioReco/detector/RNO_G/rnog_detector.py
+++ b/NuRadioReco/detector/RNO_G/rnog_detector.py
@@ -37,6 +37,9 @@ def _keys_not_in_dict(d, keys):
 
     Returns False if d["key1"]["key2"] exsits, True otherwise.
     """
+    if isinstance(keys, str):
+        keys = [keys]
+
     d_tmp = d
     for key in keys:
         try:
@@ -59,7 +62,7 @@ def _check_detector_time(method):
 
 
 class Detector():
-    def __init__(self, database_connection='RNOG_test_public', log_level=logging.INFO, over_write_handset_values={},
+    def __init__(self, database_connection='RNOG_test_public', log_level=logging.INFO, over_write_handset_values=None,
                  database_time=None, always_query_entire_description=True, detector_file=None,
                  select_stations=None, create_new=False):
         """
@@ -75,8 +78,10 @@ class Detector():
             Other options are: `logging.WARNING`, `logging.DEBUG`, ...
 
         over_write_handset_values : dict (Default: {})
-            Overwrite the default values for the manually set parameter which are not (yet) implemented in the database.
-            (Default: {}, the acutally default values for the parameters in question are defined below)
+            Overwrite the default values for (channel) parameters which are not (yet) implemented in the database.
+            You can not specify keys which already exist in the database. If the value is a dict it should be contain
+            a value for each channel_id (key). (Default: None, the acutally default values for the parameters in question
+            are defined below)
 
         database_time : `datetime.datetime` or `astropy.time.Time`
             Set database time which is used to select the primary measurement. By default (= None) the database time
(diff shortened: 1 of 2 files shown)


NuRadioReco/detector/RNO_G/rnog_detector.py
philippwindischhofer, line 1: Hi @fschlueter just starting to take a look, so please bear with me and my potentially dumb comments.

It seems that you might want to call `response._calculate_time_delay()` [here](https://github.com/nu-radio/NuRadioMC/pull/718/files#diff-710718ba5a76e872ff86449c09d02e860766ca8f684902a20085e8f57dbe3e45R1226) instead of `response._get_time_delay()`?

(I cannot put an in-line comment there because it wasn't changed in your PR.)

philippwindischhofer: (I'm also not sure if the `weight` is propagated correctly into the local `Response` that is being used if `use_stored == False`?)

fschlueter: yes another good point!

fschlueter: How about now?

NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -1179,28 +1202,29 @@ def get_time_delay(self, station_id, channel_id, cable_only=False, use_stored=Tr
...
+                    time_delay += weight * value["cable_delay"]
 
             else:
                 ydata = [value["mag"], value["phase"]]
                 response = Response(value["frequencies"], ydata, value["y-axis_units"],
                                     name=key, station_id=station_id, channel_id=channel_id,
                                     log_level=self.__log_level)
 
-                time_delay += response._get_time_delay()
+                time_delay += weight * response._calculate_time_delay()
```
philippwindischhofer, line 1227: Nit-picky comment: It might be slightly nicer to pass `weight` to `response` directly and let the object take care of it in the calculation of the time delay. But I don't think there is a way in which the current manual application can result in an incorrect answer, so either way is fine.

fschlueter: We can discuss that yes. But lets merge this first.

fschlueter:
Can that be merged?

NuRadioReco/detector/response.py
```diff
@@ -304,24 +328,32 @@ def plot(self, ax1=None, show=False, in_dB=True, plt_kwargs={}):
...
             _gain = gain(freqs)
 
             name = name.replace("_", " ")
             ls = "-" if weight == 1 else "--"
 
+            if name.startswith("golden"):
+                name = name.replace("golden downhole components", "ref. comp.")
+
+            if name.endswith(" "):
+                name = name[:-1]
```
sjoerd-bouma, line 341: `name.strip()` removes leading and trailing whitespace and saves one line over your implementation ;)

sjoerd-bouma (approved):
I found one nitpick! But thanks Felix and Philipp, I agree this should get merged.

NuRadioReco/detector/response.py
fschlueter: Well if I fix it now you have to approve again

sjoerd-bouma: To the extent that it is within my powers I give you permission to treat the first approval as the start of the 24h window. But I'm also perfectly happy with this staying like this.

fschlueter: It does not remove whitespaces within the string?

sjoerd-bouma: It removes leading and trailing whitespace only; alternatively there's `lstrip`/`rstrip` to do only one of those.

fschlueter: crazy, not what I would have expected. I will fix it next time I touch that class

Issue #695, opened 2024-07-09
Numpy 2.0 is going to break things


clark2668:
**Describe the issue**
Numpy 2.0 is going to deprecate some things, like `np.infty`. This means that lines like:
`from NuRadioMC.EvtGen.generator import generate_eventlist_cylinder` start throwing errors:
```
Traceback (most recent call last):
  File "/data/condor_builds/users/baclark/RNOG/simulation/plot_sim_event/generate_event/step1_generate_event_list.py", line 3, in <module>
    from NuRadioMC.EvtGen.generator import generate_eventlist_cylinder
  File "/data/condor_builds/users/baclark/RNOG/simulation/NuRadioMC/NuRadioMC/EvtGen/generator.py", line 14, in <module>
    from NuRadioMC.simulation.simulation import pretty_time_delta
  File "/data/condor_builds/users/baclark/RNOG/simulation/NuRadioMC/NuRadioMC/simulation/simulation.py", line 3, in <module>
    from radiotools import helper as hp
  File "/data/condor_builds/users/baclark/RNOG/simulation/software/miniconda/lib/python3.11/site-packages/radiotools/helper.py", line 3, in <module>
    from radiotools.atmosphere import models as atm
  File "/data/condor_builds/users/baclark/RNOG/simulation/software/miniconda/lib/python3.11/site-packages/radiotools/atmosphere/models.py", line 395, in <module>
    class Atmosphere():
  File "/data/condor_builds/users/baclark/RNOG/simulation/software/miniconda/lib/python3.11/site-packages/radiotools/atmosphere/models.py", line 631, in Atmosphere
    def get_atmosphere(self, zenith, h_low=0., h_up=np.infty, observation_level=0):
                                                    ^^^^^^^^
  File "/data/condor_builds/users/baclark/RNOG/simulation/software/miniconda/lib/python3.11/site-packages/numpy/__init__.py", line 397, in __getattr__
    raise AttributeError(
AttributeError: `np.infty` was removed in the NumPy 2.0 release. Use `np.inf` instead.. Did you mean: 'info'?
```

I guess this is formally an issue with radiotools, but I have to imagine things like this are sprinkled into NuRadioMC too.

MijnheerD:
I think this is purely an issue with radiotools. Installing the version from Github instead of PyPi actually solves this issue (ie. doing `pip install git+https://github.com/nu-radio/radiotools` ). I have not encountered this issue when using any NuRadio modules so far.

[...]

PR #720, opened 2024-09-04
[Hotfix] pin numpy<2

sjoerd-bouma:
See #695 for discussion

Currently, `.nur` files generated with numpy 2.* cannot be read in using earlier versions of numpy. This means we cannot guarantee compatibility even between two instances of the same version of NuRadioMC. This PR avoids this issue by pinning `numpy<2`; in order to switch to numpy 2.*, we will probably have to switch to `numpy>=2` in some future release of NuRadioMC in order to maintain cross-compatibility for a fixed NuRadioMC versions.

(This is a proposed hotfix release v2.2.2; if accepted, this should also be merged into `develop` afterwards)

### DIFF
--- a/NuRadioReco/detector/antennapattern.py
+++ b/NuRadioReco/detector/antennapattern.py
@@ -1384,7 +1384,7 @@ class AntennaPatternAnalytic(AntennaPatternBase):
 
             index = np.argmax(freq > self._cutoff_freq)
             Gain = np.ones_like(freq)
-            from scipy.signal import hann
+            from scipy.signal.windows import hann
             gain_filter = hann(2 * index)
             Gain[:index] = gain_filter[:index]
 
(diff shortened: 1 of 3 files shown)


fschlueter:
Can we start converting numpy arrays to lists ?

fschlueter:
@sjoerd-bouma the notes from Uppsala say that you wanted to have a look if it is feasable to convert numpy arrays before serialization. Do you think you get around this any time soon?

sjoerd-bouma:
> @sjoerd-bouma the notes from Uppsala say that you wanted to have a look if it is feasable to convert numpy arrays before serialization. Do you think you get around this any time soon?

Probably not before November - if you want to have a look yourself before then feel free.

anelles:
Looks like it is past November now @sjoerd-bouma ;)

sjoerd-bouma:
Closing this in favour of #782 which we will hopefully get merged soon (still want to backport some changes to how astropy objects are stored).

Issue #541, opened 2023-05-24
C++ get_attenuation occasionally fails and returns nans


sjoerd-bouma:
It seems that for some geometries, the C++ analytic raytracer returns `np.nan` for the attenuation (at least when using `SP1` as the attenuation model), which then results in traces that are all `nans`. Using the Python version works. I'm a bit surprised as I hadn't run into this before, but also can't seem to find an earlier version of NuRadioMC on which this does not happen.

MWE below:

```
from NuRadioMC.SignalProp import analyticraytracing 
from NuRadioMC.utilities import medium
import numpy as np

use_cpp = {'C++':True, 'Python':False}
x_channel = [17.3, -10.0, -142.0]
x_nu = [210.75915126907, 267.081824987715, -86.5823449330406]

for key, value in use_cpp.items():
    print(f"Using {key} raytracer")
    analyticraytracing.cpp_available = value
    rt = analyticraytracing.ray_tracing(medium.get_ice_model('southpole_2015'), attenuation_model='SP1',)
    rt.set_start_and_end_point(x_nu, x_channel)
    rt.find_solutions()

    for iS in range(rt.get_number_of_solutions()):
        attenuation = rt.get_attenuation(iS, np.fft.rfftfreq(2048, .2), 5)
        if np.any(np.isnan(attenuation)):
            print('attenuation contains nans!', attenuation)
        else:
            print('no nans! Hooray!')
```

Not sure if this is similar to the issue in #521?

colemanalan:
Actually, we may have some numerical instabilities here. I was looking at how the `result` for the first 3 freq bins are affected by changing z. If you use

`-85.5823449330406`

- 0.0377103
- 0.0875237
- 0.10958

For `-86.5823449330406` (the one that gives nans)

- 0.0320964
- 0.0744749
- 0.0932363

And for `-87.5823449330406` 

- 0.0378141
- 0.0877644
- 0.109881

So clearly the bad values aren't bounded properly. Probably a fix of this will involve finding the weirdness of the integrand that cuases this to fail or to slightly perturb the start/end points (which hopefully fixes the issue) and return some kind of interpolation

fschlueter:
I am currently testing a solution which was suggested by @EnriqueHuesca. If you swap the integration function from `gsl_integration_qags` to `gsl_integration_qag` you do not get nans and the results agree wekk when shifting the z-coordinate by 1 meter up or down. 
```
z=-85....
result: 0.0377103
[...]

PR #722, opened 2024-09-16
Raytracer does not use scipy.integrate.quad for attenuation

christophwelling:
Using scipy.integrate.quad to calculate the attenuation can occasionally cause problems.
With this PR, the integral will by default be calculated using a simple Rieman sum over a fixed number of point. While this is a bit less precise, the difference is far smaller than the uncertainty on the attenuation length itself.


NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -127,8 +125,8 @@ def __init__(self, medium, attenuation_model="SP1",
             The signal attenuation is calculated using a numerical integration
             along the ray path. This calculation can be computational inefficient depending on the details of 
             the ice model. An optimization is implemented approximating the integral with a discrete sum with the loss
-            of some accuracy, See PR #507. This optimization is used for all ice models listed in 
-            speedup_attenuation_models (i.e., "GL3"). With this argument you can explicitly activate or deactivate
+            of some accuracy, See PR #507. This optimization is used for all ice models.
+            With this argument you can explicitly activate or deactivate
             (True or False) if you want to use the optimization. (Default: None, i.e., use optimization if ice model is
             listed in speedup_attenuation_models)
```
fschlueter, line 131: The doc string needs to be adjusted

fschlueter:
@christophwelling can you finish that?

sjoerd-bouma:
I don't want to get too involved with this - probably simplifying and unifying the way the attenuation length is calculated in Python is a good idea, and I think this should get merged once the docstring is updated, but I'm not aware of any cases where the current implementation causes errors. 

The 'bigger' problem is in the C++ backend (#541), where the integration fails silently because of **much** stricter convergence limits than in the Python case. I believe @EnriqueHuesca has a solution already (other than just relaxing the convergence requirement), which I hope he will contribute soon : )

cg-laser:
Hi everyone, I justed merged #660  which makes the python version (using numba) as fast as the C++ implementation (except for the attenuation length calculation). 
My suggestion is that we make python/numba + the numerical attenuation length calculation (this PR) the new default. 
We could even retire the C++ implementation which always had more problems in finding solutions than the Python implementation. 

However, there is one important thing that we need to check for! @christophwelling argued that an accuracy of 10% would be acceptable because the uncertainty of the attenuation length is not better. This is true, BUT we need much, much smaller relative error. When we calculate the attenuation for two rays to two nearby antennas, the result should be identical. In general, for one event, the relative error between the paths to different antennas needs to be on the <1% level, because we use the amplitude differences in a reconstruction to determine the viewing angle and vertex distance. 

@christophwelling can you merge the recent develop into your branch and implement such a test?

christophwelling:
attenuation_integration
I already checked it when I implemented this. The difference between the integration methods is much smaller than 1%. The MB model looks a bit weird, but that's just because the attenuation length parametrization approaches 0 when you get too deep. But since this is much deeper than the MB ice, this should never happen anyway.

christophwelling:
I commited a fix for the error in the raytracing test. The test still fails, but I think that's just because it now defaults to a different way of calculating the integral.
If I remove the part where the raytracer integrates the attenuation around the path inflection point using quad, the fraction of attenuation lenghts that are deviating from the reference actually drops to 0.5%, so I think this is just due to the fact that the C++ raytracer does not use quad to calculate the attenuation integral.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -901,12 +899,9 @@ def dt(t, C_0, frequency):
...
                         
-                        integrand = integrate.quad(self.ds, segments[idx], segments[idx + 1], args=(C_0), epsrel=1e-2, points=[z_turn])[0]
-                        attenuation = (attenuation_util.get_attenuation_length(z_turn, f, self.attenuation_model) for f in freqs)
-                        att_int = np.array(
-                            [integrand, attenuation])
-
-                        attenuation_exp_tmp[:, idx] = att_int
+                        
+                        att = np.array([attenuation_util.get_attenuation_length(z_turn, f, self.attenuation_model) for f in freqs])
+                        attenuation_exp_tmp[:, idx] = 1. / att
```
fschlueter, line 904: I am a bit confused. I have a PR #908 which fixes a bug which this PR apparently also fixes but the results looks different. @christophwelling why don't you do `integrand / att`? Is `integrand` by design 1?

fschlueter: Okay I convinced myself with the example in NuRadio that this is wrong and the correct equation is `integrand / att`. Will change it.

fschlueter:
I am currently running some comparison between the optimized and non optimized calclation for the attenuation factor with the python raytracer. This is for 900 different sources on a rectangular grid with x=-3000..-10, z=-3000..-100, the receiver at 0, 0, -100 and the GL1 model. The histogram marginalises over all positions, ray types, and frequencies. The maximum deviation is typically very low. It is mostly positive (i.e., optimized > not optimized). Weirdly in this test the optimized version took longer...

image

fschlueter:
This is for GL3
image

fschlueter:
In my tests for GL1 the calcuation via quad is often faster ... So to be honest I am not seeing the point of this PR.

fschlueter:
After talking to @christophwelling, @EnriqueHuesca, @sjoerd-bouma it is not entirely clear which problem this PR tries to solve: I.e., is there an issue with the use of `scipy.integrate.quad` for ice models other than GL3 for which this function is computational inefficient. Because for the smooth attenuation length profiles described by the models other than GL3 the use of `quad` is actually faster in most cases than the "optimized calculation" for the GL3 model. I tested this with real ray-traced paths but also with this MWE. 

```
from NuRadioMC.utilities import attenuation

import scipy.integrate
import numpy as np
import time


freq = 400 * units.MHz
depths = np.linspace(-10, -2990, 1000)

n = 100

for model_name, model_int in attenuation.model_to_int.items():
    if model_name == "MB1":
        continue  # skip this model for now because it does not go to 3000 m

    t0 = time.time()
    for _ in range(n):
        scipy.integrate.quad(
            attenuation.get_attenuation_length,
            depths[0], depths[-1], args=(freq, model_name),
            epsrel=1e-2
        )
    print(f"Integration over attenuation length profile for {model_name} took {time.time() - t0:.3f} seconds for {n} iterations")

    t0 = time.time()
    for _ in range(n):
        np.sum([
            attenuation.get_attenuation_length(depth, freq, model_name) * (depths[1] - depths[0])
            for depth in depths
        ])
    print(f"Summing over attenuation length profile with {len(depths)} steps for {model_name} took {time.time() - t0:.3f} seconds for {n} iterations")
```
Hence I will close this PR. We can get back to it if problems with the attenuation calculation for the other ice models pop up.

PR #725, opened 2024-09-19
Update fft.py

fschlueter:
Adding missing interface to rely sole on utilities (otherwise the user has to use np.rfft themself.

### DIFF
--- a/NuRadioReco/utilities/fft.py
+++ b/NuRadioReco/utilities/fft.py
@@ -30,9 +30,27 @@ which are defined using the latter convention.
 
 import numpy as np
 
+def freqs(num_samples, sampling_rate):
+    """
+    Returns frequency bins for FFT.
+  
+    Parameters
+    ----------
+    num_samples: int
+        The number of sample in the time domain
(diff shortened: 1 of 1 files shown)


NuRadioReco/utilities/fft.py
```diff
@@ -30,23 +30,46 @@
 
 import numpy as np
 
+def freqs(num_samples, sampling_rate):
+    """
```
sjoerd-bouma, line 34: I'm trying to think if there's a clean way to extend this such that a user can call this with the number of frequencies instead of `num_samples`, by adding an optional switch or having both `num_samples` and `num_freqs` or something similar. But I can't think of one that I like sufficiently, and I don't know that we _need_ this.

sjoerd-bouma (approved):
This is a nice addition - one comment which if you have a good solution for might be worth adding, otherwise I think this can be merged.

Maybe if you could remerge the latest develop (or just re-run the tests) because I'm not sure why they failed - this change shouldn't have affected anything.

PR #726, opened 2024-10-02
Extend hardware rnog detector class to read addtional attenuator from…

fschlueter:
… hardware dB and add it to class

There might be still some sign error - will fix it once we have the data in the database. Also this assumes the attenuation/gain in dB. 

@philippwindischhofer this is the idea of the implementation. We can also rename things, no strong opinion here.

### DIFF
--- a/NuRadioReco/detector/RNO_G/rnog_detector.py
+++ b/NuRadioReco/detector/RNO_G/rnog_detector.py
@@ -845,14 +845,16 @@ class Detector():
             is_equal = False
             if "drab_board" in components and "iglu_board" in components:
 
-                is_equal = np.allclose(measurement_components_dic["drab_board"]["mag"],
-                                       measurement_components_dic["iglu_board"]["mag"])
+                is_equal = np.allclose(
+                    measurement_components_dic["drab_board"]["mag"],
+                    measurement_components_dic["iglu_board"]["mag"])
 
                 if is_equal:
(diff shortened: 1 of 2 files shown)


philippwindischhofer:
Hi @fschlueter thanks for getting started on this!

By way of documentation, let me briefly record here which components are affected by this. Looking at the attached plots, only SURFACE boards measured in 2022 with the measurement protocol `surface_board_10dBattenuation_erlangen2022 ` seem to be affected.

These are `C0018`, `C0009`, `C0010`, `C0003`, `C0012`, `C0011`, `C0019`, `C0013`, `C0004`, `C0006`, `C0001`, `C0008`, `C0007`, `C0002`, `C0014`, `C0015`, `C0005`, `C0016`, `C0017`.

IGLUs or DRABs are not affected. As we did not store S11 information for optical fiber measurements, it's impossible to verify at the single-measurement level for this class of components.

drab_S11
iglu_S11
surface_S11

philippwindischhofer:
Updated the attenuator settings in the DB for the surface boards listed above, cf. https://github.com/RNO-G/hardware-webinterface/pull/14

philippwindischhofer:
Thanks for correcting the sign @fschlueter! I can't approve nor merge this, unfortunately, so let's wait for @sjoerd-bouma!

fschlueter:
why not? you should be able to approve this!

philippwindischhofer (approved):

PR #728, opened 2024-10-03
Fix in period-of-validity logic for RNO-G HW DB

philippwindischhofer:
Fixes a bug in removing channels and devices that are not commissioned at the requested time.

Simple script to validate this fix:

```
from datetime import datetime
from NuRadioReco.detector.RNO_G.db_mongo_write import Database

db = Database(database_connection = 'RNOG_public')
db.set_detector_time(datetime.utcnow())
station_info = db.get_general_station_information(23)

print(station_info)
```

Crashes before this PR, prints station information after.


fschlueter:
Oh damn, I not only broke the webinterface I broke the whole damn thing ? :D

fschlueter (approved):

fschlueter:
Since this bug causes the webinterface to return an error I am taking the executive decision to merge it right away and rebuild the webserver.

PR #731, opened 2024-10-10
Implement add_to_trace method (refactor)

MartinRavn:
I implemented a method in the BaseTrace class that adds the trace of a different channel to the current trace without changing the binning, i.e., the same readout window.

### DIFF
--- a/.github/workflows/run_tests.yaml
+++ b/.github/workflows/run_tests.yaml
@@ -112,6 +112,12 @@ jobs:
         export GSLDIR=$(gsl-config --prefix)
         export PYTHONPATH=$(pwd):$PYTHONPATH
         NuRadioMC/test/SignalGen/test_build.sh
+    - name: "Validate separate trigger channels"
+      if: always()
+      run: |
+        export GSLDIR=$(gsl-config --prefix)
+        export PYTHONPATH=$(pwd):$PYTHONPATH
+        NuRadioMC/test/SingleEvents/validate_separate_trigger_channels.sh
     - name: "Signal propagation tests"
       if: always()
       run: |
--- a/NuRadioReco/framework/base_trace.py
+++ b/NuRadioReco/framework/base_trace.py
@@ -246,6 +246,75 @@ class BaseTrace:
         if 'trace_start_time' in data.keys():
             self.set_trace_start_time(data['trace_start_time'])
 
+    def add_to_trace(self, channel):
+        """
+        Adds the trace of another channel (with potentially diffrent length and _trace_start_time) to
+        the trace of this channel without changing the sampling. If used on an empty trace with a defined
+        self._sampling_rate and self.self._trace_start_time, and self._time_trace containing zeros, this
+        function can be seen as recording the channel in a specific readout window.
+
+        Parameters
+        ----------
+        channel: BaseTrace
+            Channel to add to current channel
+        """
+
+        assert self.get_number_of_samples() is not None, "No trace is set for this channel"
+        assert self.get_sampling_rate() == channel.get_sampling_rate(), "Sampling rates of the two channels do not match"
+
+        tt_readout = self.get_times()
+        t0_readout = self.get_trace_start_time()
+        t1_readout = t0_readout +  self.get_number_of_samples() / self.get_sampling_rate()
+        sampling_rate_readout = self.get_sampling_rate()
+        n_samples_readout = self.get_number_of_samples()
+
+        tt_channel = channel.get_times()
+        t0_channel = channel.get_trace_start_time()
+        t1_channel = t0_channel + channel.get_number_of_samples() / channel.get_sampling_rate()
+        sampling_rate_channel = channel.get_sampling_rate()
+        n_samples_channel = channel.get_number_of_samples()
+
+        # We handle 1+2x2 cases:
+        # 1. Channel is completely outside readout window:
+        if t1_channel < t0_readout or t1_readout < t0_channel:
+            return
+        # 2. Channel starts before readout window:
+        if t0_channel < t0_readout:
+            i_start_readout = 0
+            t_start_readout = t0_readout
+            i_start_channel = int( (t0_readout-t0_channel) * sampling_rate_channel ) + 1 # The first bin of channel inside readout
+            t_start_channel = tt_channel[i_start_channel]
+        # 3. Channel starts after readout window:
+        elif t0_channel >= t0_readout:
+            i_start_readout = int( (t0_channel-t0_readout) * sampling_rate_readout ) # The bin of readout right before channel starts
+            t_start_readout = tt_readout[i_start_readout]
+            i_start_channel = 0
+            t_start_channel = t0_channel
+        # 4. Channel ends after readout window:
+        if t1_channel >= t1_readout:
+            i_end_readout = n_samples_readout
+            t_end_readout = t1_readout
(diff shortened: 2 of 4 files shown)


NuRadioReco/framework/base_trace.py
```diff
@@ -246,6 +246,75 @@ def deserialize(self, data_pkl):
...
+        Adds the trace of another channel (with potentially diffrent length and _trace_start_time) to
+        the trace of this channel without changing the sampling. If used on an empty trace with a defined
+        self._sampling_rate and self.self._trace_start_time, and self._time_trace containing zeros, this
+        function can be seen as recording the channel in a specific readout window.
+
+        Parameters
+        ----------
+        channel: BaseTrace
+            Channel to add to current channel
+        """
```
fschlueter, line 260: In our call last week @sjoerd-bouma was not quite happy about the naming as it can be confused with the `__add__` operator. I am not sure I can come up with a better name... However, I think the doc-string can more clearly reflect this difference. A suggestion from my side (please extend with the information which are missing):
```
        """
        Adds the trace of another channel to the trace of this channel. The traces are only added within the time window of the "this" channel.

        Parameters
        ----------

        channel : BaseTrace
            The channel whose trace is to be added to the trace of this channel.
        """
```

fschlueter:
A general comment. I was wondering if it makes sense to use `__add__` within this method to make the code easier to understand. Would something like this work?
```
    def add_to_trace(self, channel):
        """
        Adds the trace of another channel to the trace of this channel. The traces are only added within the time window of the "this" channel.

        Parameters
        ----------

        channel : BaseTrace
            The channel whose trace is to be added to the trace of this channel.
        """

        assert self.get_number_of_samples() is not None, "No trace is set for this channel"
        assert self.get_sampling_rate() == channel.get_sampling_rate(), "Sampling rates of the two channels do not match"

        t_start = self.get_trace_start_time()
        nsamples = self.get_number_of_samples()

        # Use __add__ to combine both traces. We made sure that the sampling rates are the same
        # We will "cut out" the readout window in the following...
        self = self + channel

        # We only have two cases to consider: Either "this" trace or "channel" is first. 
        if self.get_trace_start_time() == t_start:
            # self was first trace, so we only need to adjust the number of samples
            self.set_trace(self.get_trace()[:nsamples], self.get_sampling_rate())
        else:
            # channel was first trace
            bin_shift = int((t_start - self.get_trace_start_time()) * self.get_sampling_rate())  # is positive
            residual_time_shift = (t_start - self.get_trace_start_time()) - bin_shift / self.get_sampling_rate()
            self.apply_time_shift(-residual_time_shift)
            self.set_trace(self.get_trace()[bin_shift:bin_shift + nsamples], self.get_sampling_rate())
            self.set_trace_start_time(t_start)
```
This solution might be not the most efficient one and for example we have to apply the time shift twice. Is that an issue?

.github/workflows/run_tests.yaml
```diff
@@ -112,6 +112,12 @@ jobs:
         export GSLDIR=$(gsl-config --prefix)
         export PYTHONPATH=$(pwd):$PYTHONPATH
         NuRadioMC/test/SignalGen/test_build.sh
+    - name: "Validate separate trigger channels"
+      if: always()
+      run: |
+        export GSLDIR=$(gsl-config --prefix)
```
fschlueter, line 118: is that line actually needed or is that a copy past relict?

fschlueter:
@cg-laser, the review was not concluded. Do you want to continue the review on the refactor branch?

NuRadioReco/framework/base_trace.py
MartinRavn: Thanks. I think it is clearer now.
One other important difference is that the inputs to __add__ commute whereas they don't for add_to_trace.

fschlueter: @MartinRavn I did not commit my proposed doc-string. It was a proposal to you. I agree that can be added to the doc string as well

MartinRavn:
> A general comment. I was wondering if it makes sense to use `__add__` within this method to make the code easier to understand. Would something like this work?
> 
> ```
>     def add_to_trace(self, channel):
>         """
>         Adds the trace of another channel to the trace of this channel. The traces are only added within the time window of the "this" channel.
> 
>         Parameters
>         ----------
> 
>         channel : BaseTrace
>             The channel whose trace is to be added to the trace of this channel.
>         """
> 
>         assert self.get_number_of_samples() is not None, "No trace is set for this channel"
>         assert self.get_sampling_rate() == channel.get_sampling_rate(), "Sampling rates of the two channels do not match"
> 
>         t_start = self.get_trace_start_time()
>         nsamples = self.get_number_of_samples()
> 
>         # Use __add__ to combine both traces. We made sure that the sampling rates are the same
>         # We will "cut out" the readout window in the following...
>         self = self + channel
> 
>         # We only have two cases to consider: Either "this" trace or "channel" is first. 
>         if self.get_trace_start_time() == t_start:
>             # self was first trace, so we only need to adjust the number of samples
>             self.set_trace(self.get_trace()[:nsamples], self.get_sampling_rate())
>         else:
>             # channel was first trace
>             bin_shift = int((t_start - self.get_trace_start_time()) * self.get_sampling_rate())  # is positive
>             residual_time_shift = (t_start - self.get_trace_start_time()) - bin_shift / self.get_sampling_rate()
>             self.apply_time_shift(-residual_time_shift)
>             self.set_trace(self.get_trace()[bin_shift:bin_shift + nsamples], self.get_sampling_rate())
>             self.set_trace_start_time(t_start)
> ```
> 
> This solution might be not the most efficient one and for example we have to apply the time shift twice. Is that an issue?

I actually did this initially, but Christian suggested the more efficient implementation. We risk dealing with very long traces when using add. I save up to a second when adding the pulses for 16 antennas.

fschlueter:
Why? That seems surprising to me. The only reason I can think of is to use `apply_time_shift` on a very long trace. But in principle we can "cut" on the very long traces before we shift.

.github/workflows/run_tests.yaml
MartinRavn: Probably not. I don't fully understand how the workflows work, so I just copied and pasted. Which lines specifically are not needed?

fschlueter: the export

cg-laser:
> @cg-laser, the review was not concluded. Do you want to continue the review on the refactor branch?

ups, that was unintentional. I needed the function to continue work in the refactor branch. Github must have detected that I had merged it locally and then pushed...

NuRadioReco/framework/base_trace.py
MartinRavn: Yes. I have a mix of yours and mine that I will push tomorrow.

MartinRavn:
> Why? That seems surprising to me. The only reason I can think of is to use `apply_time_shift` on a very long trace. But in principle we can "cut" on the very long traces before we shift.

In general we can not shift the trace after we have cut it. We risk that the signal rolls over the edge, or that we throw some of the signal away that would be shifted "into view". Maybe there is a clever way to do it.

fschlueter:
I agree that why I shift in my solution before trimming as well. But we could have some intermediate trimming, we never shift more than one signal so we could first trim to `start -1:end+1`, shift and than trim a second time `start:end` that should be very fast

PR #732, opened 2024-10-10
refactor get_time_delay function in rno-g detector

fschlueter:
This reduces code for the most common way of using the function `cable_only=False` `use_store=True`


NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -1203,31 +1203,33 @@ def get_time_delay(self, station_id, channel_id, cable_only=False, use_stored=Tr
...
-            if use_stored:
-                if "time_delay" not in value and "cable_delay" not in value:
-                    self.logger.warning(
-                        f"The signal chain component \"{key}\" of station.channel "
-                        f"{station_id}.{channel_id} has no cable/time delay stored... (hence return time delay without it)")
+        if use_stored and not cable_only:
+            resp = self.get_signal_chain_response(station_id, channel_id)
+            return resp.get_time_delay()
+        elif use_stored and cable_only:
+            time_delays = resp.get_time_delays()
```
shallmann, line 1210: `resp` not defined inside the function

NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -1203,31 +1203,33 @@ def get_time_delay(self, station_id, channel_id, cable_only=False, use_stored=Tr
...
+            names = resp.get_names()
+            time_delay = 0
+            for name, dt in zip(names, time_delays):
+                if re.search("cable", name) is None and re.search("fiber", name):
```
shallmann, line 1214: shouldn't it be `if re.search("cable", name) is None and re.search("fiber", name) is None`?

NuRadioReco/detector/response.py
```diff
@@ -380,10 +380,14 @@ def plot(self, ax1=None, show=False, in_dB=True, plt_kwargs={}):
         else:
             return fig, ax
 
-    def get_time_delay(self):
+    def get_time_delay(self, ):
```
shallmann, line 383: what does the comma do here?

shallmann (commented):
did you test that it gives consistent results? I added some comments.

fschlueter:
> did you test that it gives consistent results? I added some comments.

Nope :(

fschlueter:
Maybe a more general question also to @cg-laser, @sjoerd-bouma and @philippwindischhofer : I do not like this implementation with the `cable_only` (using strings in the component name ...). Is there a use case where one is interested in only the cable/fiber delay and not the delay of the entire response chain? If that is not the case I will delete the entire option `cable_only`

fschlueter:
okay I removed the function `cable_only` entirely. Now the function `get_cable_delay` calls `get_time_delay` and this returns the time delay of the entire chain possibly including amplifier and filter.

philippwindischhofer:
Hey @fschlueter sorry to be slow, just catching up.

I think for most (non-expert) use-cases, we don't need to look separately at the cable delay. The (relative) delay between two separate signal chains is what is most important, I think.

fschlueter:
@shallmann do you want to approve?

shallmann:
on it... just looking at the changes once more now

NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -1280,20 +1260,24 @@ def get_site_coordinates(self, station_id=None):
...
-                            database_connection='RNOG_public', select_stations=24)
+                            database_connection='RNOG_public', select_stations=13)
 
-    det.update(datetime.datetime(2023, 8, 2, 0, 0))
+    det.update(datetime.datetime(2023, 7, 2, 0, 0))
 
+    print(det.get_time_delay(13, 0))
+    print(det.get_time_delay(13, 0, cable_only=True))
+    # print(det.get_time_delay(13, 0, cable_only=False, use_stored=False))
+    print(det.get_time_delay(13, 0, cable_only=True, use_stored=False))
```
shallmann, line 1270: The optional argument now no longer exists, no?

NuRadioReco/detector/response.py
shallmann: I trust you on resolving this, but maybe I can learn: Can you tell me what is the benefit of introducing the comma is here?

shallmann (approved):
Please check if the print in the __main__ is intended. Otherwise good to go.

shallmann (approved):
Thanks for cleaning up!

PR #734, opened 2024-10-18
Reduce date + time stamp in logging

fschlueter:
I love how the now logging looks! But I feel that the "date and timestamp" is a bit large which decreases readability.

Currently I am for example getting: 
image

This PR changes it to: 
image

Or maybe a good compromise is:
image

### DIFF
--- a/NuRadioReco/utilities/logging.py
+++ b/NuRadioReco/utilities/logging.py
@@ -84,7 +84,7 @@ def _setup_logger(name="NuRadioReco", level=None):
 
     Notes
     -----
-    This function is only meant to be called once, on import, as part of the `__init__.py` scripts of the base packages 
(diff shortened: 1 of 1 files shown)


sjoerd-bouma (approved):
This is the format that [I use myself](https://github.com/nu-radio/NuRadioMC/pull/664#issuecomment-2027075237) so of course it gets a quick approval from me - you should merge it quickly before someone else disagrees ;)

(I'm half-joking, of course I'm happy for other people to weigh in. But I much prefer the shorter format, I think the information on either side is rarely relevant and just pushes the actual logger message further away.)

MijnheerD (approved):

fschlueter:
we could read the format from file (if exists) `.logging_format.txt` which we add to the gitignore. This way everyone can overwrite their personal format if they want

PR #735, opened 2024-10-18
Add caching to BaseTrace.get_frequencies()

fschlueter:
Not that this was a bottle neck but why not saving some computing time. A small test with a 2048 sample waveforms improved the speed by a factor of 10 from 1.9us to 180ns.

PR #736, opened 2024-10-21
Allow to get frequency spectrum from windowed trace.

fschlueter:
This PR is based of #735 and the diff will simplify once this is merged.


shallmann (commented):
Implementation looks ok, requires that the user knows what he is doing to got get weird stuff...
Two potential issues:
1/ `window = [1,0,1,0,1,0,1,0,0,1,1,1]` will be totally valid, but probably not meaningful, masking a window `[1,1,1,...,1,1,0,0,0,0,0,1,1,...,1,1,1]` might also introduce weird features
2/ the other problem might need more attention: how do we get the corresponding frequencies? The get_frequencies() should also be able to take the same window then, no?

shallmann:
@fschlueter yes that's fine with me if you add it to the get_frequencies()

fschlueter:
@shallmann have a look :)

shallmann:
Hmm.. why do you think the window can be `int`/`float` only? for a single value, an fft does not really make sense?

The other thing that came to my mind and might cause problems - although I did not check: The implementation works for 1D, an `electric_field` can have 3D traces, your argument is probably that for electric fields one would never set the window and hence the calculation is not required to work in > 1D?

fschlueter:
true, what shape has a 3D efield, `nsamples, pol` or `pol, nsamles`? In the former case the code should work. The idea behind window being an int/float is that one can pass the number of samples directly. Although, I agree the name of the variable does not reflect that.

fschlueter:
@shallmann I only test the code in my ipython shell with pseudo data but I think it should work like this. Do you like this "solution"?

NuRadioReco/framework/base_trace.py
```diff
@@ -177,8 +195,29 @@ def add_trace_start_time(self, start_time):
...
+        """
+        if window_or_nsamples is None:
+            nsamples = self.get_number_of_samples()
+        else:
+            if isinstance(window_or_nsamples, (int, float)):
+                nsamples = int(window_or_nsamples)
+            else:
+                nsamples = int(np.sum(window_or_nsamples))
+
+        return get_frequencies(nsamples, self._sampling_rate)
```
sjoerd-bouma, line 220: I would probably have a slight preference for having the same `window` argument as in `get_frequency_spectrum`, for simplicity... but I can live with this.

NuRadioReco/framework/base_trace.py
```diff
@@ -77,13 +77,31 @@ def get_filtered_trace(self, passband, filter_type='butter', order=10, rp=None):
...
-            self.__time_domain_up_to_date = False
-        return np.copy(self._frequency_spectrum)
+    def get_frequency_spectrum(self, window=None):
+        """
+        Returns the frequency spectrum.
+
+        Parameters
+        ----------
+        window: array of bools (default: None)
+            If not None, specifies the time window to be used for the FFT. Has to have the same length as the trace.
```
sjoerd-bouma, line 87: I'm not so sure about the term `window`, which to me implies e.g. Tukey/Hamming/etc. windows, none of which are supported here.  I suppose that's clear from the `array of bools` type specification, but maybe it's worth mentioning the word `mask` somewhere or even changing the name of the argument.

fschlueter:
@sjoerd-bouma

sjoerd-bouma (approved):
I would have stuck with the simpler argument name (I'm not sure to what extent `window_mask` makes it clearer that only masks/rect windows are supported), and just added some clarification to the doc string, but I can live with this. 

Must be unrelated, but any idea why the birefringence test fails? Maybe it's not seeded?

PR #656, opened 2024-03-13
Further LOFAR development

MijnheerD:
This is a draft PR to follow up on the newest additions from the LOFAR side. These should mainly concern LOFAR specific modules, except for some framework changes that might happen do accommodate LOFAR specific requirements.

PR #740, opened 2024-11-06
galactic noise adder upgrades

karenterveer:
more sky models, speed up module.

It should run exactly like before, just with the added option of being able to choose from more sky models during the begin() call. For some sky models, additional modules are necessary, but if they are not installed the module will default to the standard galaxy model.
The generation of sky powers was moved to begin() as otherwise the generation happens in every single run, slowing down the model considerably.


NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -50,32 +72,101 @@ def begin(
...
             calculated by interpolation the log10 of the temperature
             The interpolation_frequencies have to cover the entire passband
             specified in the run method.
         """
         self.__debug = debug
         self.__n_side = n_side
-        self.__interpolaiton_frequencies = interpolation_frequencies
+
+        if freq_range is None:
+            freq_range = np.array([30, 80]) * units.MHz
```
sjoerd-bouma, line 95: Relatedly - changing from a default of [10,1100] to [30,80] MHz is quite a big change, and e.g. for RNO-G the new default is definitely not what we want. I would prefer to keep the old default and just raise an error (or a warning) if the chosen model is not compatible with the frequency range.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -91,132 +182,135 @@ def run(
...
+                delta_phases = (
+                        2 * np.pi * freqs[passband_filter] / (scipy.constants.c * units.m / units.s) * n_air *
+                        (
+                                sin_zenith *
+                                (np.cos(azimuth) * channel_pos_x + channel_pos_y * np.sin(azimuth)) +
+                                channel_depth *
+                                ((n_ice / n_air) ** 2 + sin_zenith ** 2) /
+                                np.sqrt((n_ice / n_air) ** 2 - sin_zenith ** 2)
+                        )
+                )
```
sjoerd-bouma, line 295: I think this is Moritz' change to account for the same signal in different antennas. This is definitely an upgrade we want to have, but as mentioned in #591 I don't think this is correct for deep in-ice antennas (I am okay with not fixing that for now, although maybe issuing a warning for channel depths over 10 m would be good?), and it also doesn't look right to me for above-surface antennas right now (that probably should be fixed and should be relatively straightforward).

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -91,132 +182,135 @@ def run(
...
+
+                # add random polarizations and phase to electric field
+                polarizations = np.random.uniform(0, 2. * np.pi, len(S))
```
sjoerd-bouma, line 298: ...also an open question whether random polarizations are right (technically, this implies that the signal from each signal is fully linearly polarized in a random direction for each antenna. I don't think this is right, we should think about how to do this correctly, even if the impact in the end is probably small).

sjoerd-bouma (commented):
Thanks Karen! It would be great to finally have Moritz' changes from #591 implemented (even though his commit credit seems to have disappeared into the ether :/). Unfortunately my questions from #591 haven't gone away yet ;)

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -50,32 +72,101 @@ def begin(
...
             12 * n_side ** 2, so a larger value for n_side will result better accuracy
             but also greatly increase computing time.
-        interpolation_frequencies: array of float
+        freq_range: array of len=2, default: [30, 80] * units.MHZ
             The sky brightness temperature will be evaluated for the frequencies
-            in this list. Brightness temperature for frequencies in between are
+            within this limit. Brightness temperature for frequencies in between are
             calculated by interpolation the log10 of the temperature
             The interpolation_frequencies have to cover the entire passband
             specified in the run method.
```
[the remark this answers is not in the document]
sjoerd-bouma: ...alternatively, we can keep the argument `interpolation_frequencies`. Your call.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -91,132 +182,135 @@ def run(
...
+
+        # check that or all channels channel.get_frequencies() is identical
+        last_freqs = None
+        for channel in station.iter_channels():
+            if (not last_freqs is None) and (
+                    not np.allclose(last_freqs, channel.get_frequencies(), rtol=0, atol=0.1 * units.MHz)):
+                logger.error("The frequencies of each channel must be the same, but they are not!")
+                return
+
+            last_freqs = channel.get_frequencies()
```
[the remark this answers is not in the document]
karenterveer: I think these were also changes that Moritz implemented - I am also not sure in which case they would be needed, but he clearly deemed this important. I also think we can keep it since it does not seem to cause any problems

NuRadioReco/modules/channelGalacticNoiseAdder.py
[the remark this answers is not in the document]
karenterveer: Good suggestion, I kept both and print a warning when interpolation_frequencies is used. 
I much prefer freq_range since it does not require the user to set up the frequencies themselves, especially since log interpolation seems like a more ideal choice for Galactic Noise.

NuRadioReco/modules/channelGalacticNoiseAdder.py
karenterveer: (oops, I changed the default for LOFAR related reasons but meant to change it back for the commit)

karenterveer:
Have implemented the changes, except the points raised in #591 - But I agree that this should be addressed or (for now) removed before merge.
As for the random polarizations it might be good to check how big the impact really is, but as this is not a new change in the GalacticNoiseAdder I think it can be kept for now.

fschlueter:
> Unfortunately my questions from https://github.com/nu-radio/NuRadioMC/pull/591 haven't gone away yet ;)

What question?

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -1,13 +1,32 @@
...
+
+try:
+    from radiocalibrationtoolkit import *  # Documentation: https://github.com/F-Tomas/radiocalibrationtoolkit/tree/main contains SSM, GMOSS, ULSA
+except:
+    logger.warning("radiocalibrationtoolkit import failed. Consider installing it to use more sky models.")
+
+try:
+    from pylfmap import LFmap  # Documentation: https://github.com/F-Tomas/pylfmap needs cfitsio installation
+except:
+    logger.warning("LFmap import failed. Consider installing it to use LFmap as sky model.")
```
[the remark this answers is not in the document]
fschlueter: @karenterveer I think the idea is to except the import error and not raise an import error for any exception thrown:
```
try:
    from pylfmap import LFmap  # Documentation: https://github.com/F-Tomas/pylfmap needs cfitsio installation
except ImportError:
    logger.warning("LFmap import failed. Consider installing it to use LFmap as sky model.")
```

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -50,32 +73,107 @@ def begin(
...
+            # define interpolation frequencies. Set in logarithmic range from freq_range[0] to freq_range[1],
+            # rounded to 0 decimal places to avoid import errors from LFmap abd tabulated models.
+            self.__interpolation_frequencies = np.around(
+                np.logspace(
+                    *np.log10(freq_range), num=15
+                ), 0
+            ) * units.MHz
+        else:
+            self.__interpolation_frequencies = interpolation_frequencies * units.MHz
+            logger.warning("DeprecationWarning: Optional argument 'interpolation_frequencies' was replaced by 'freq_range'.")
```
fschlueter, line 110: Why would one deprecate it? However you can drop the unit in l107 and add it in l99.

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: spppppppace :)

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -1,13 +1,32 @@
...
 import numpy as np
-import matplotlib.pyplot as plt
 import scipy.constants
 import scipy.interpolate
-import pygdsm
+
+logger = logging.getLogger('NuRadioReco.channelGalacticNoiseAdder')
+
+try:
+    from radiocalibrationtoolkit import *  # Documentation: https://github.com/F-Tomas/radiocalibrationtoolkit/tree/main contains SSM, GMOSS, ULSA
```
fschlueter, line 14: Can we add this as an optional dependency in the installation script? Also can we use explicit imports here?

karenterveer: The problem is that some of the modules included in the radiocalibrationtoolkit are not straightforward to install in my experience.
https://github.com/F-Tomas/radiocalibrationtoolkit/tree/main the documentation states to import the packages in the toolkit like this...

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: > However you can drop the unit in l107 and add it in l99.

actually this is important for the same reason as the comment above.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -91,132 +189,135 @@ def run(
...
+                                                                     lonlat=True)
         pixel_longitudes *= units.deg
         pixel_latitudes *= units.deg
-        galactic_coordinates = astropy.coordinates.Galactic(l=pixel_longitudes * astropy.units.rad, b=pixel_latitudes * astropy.units.rad)
+
+        galactic_coordinates = astropy.coordinates.Galactic(l=pixel_longitudes * astropy.units.rad,
+                                                            b=pixel_latitudes * astropy.units.rad)
         local_coordinates = galactic_coordinates.transform_to(local_cs)
+
         n_ice = ice.get_refractive_index(-0.01, detector.get_site(station.get_id()))
```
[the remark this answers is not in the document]
fschlueter: And the time! It might be more a problem for my current "processing style" where every station is a dual polarised antenna but in principle if you have several station you want to process at the same time you might be able to save something here!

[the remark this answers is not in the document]
fschlueter: So I tried it with this function:
```
#@functools.lru_cache(maxsize=128)
def get_local_coordinates(coordinates, time, n_side):
    print(coordinates, time, n_side)
    site_latitude, site_longitude = coordinates
    site_location = astropy.coordinates.EarthLocation(lat=site_latitude * astropy.units.deg,
                                                        lon=site_longitude * astropy.units.deg)

    local_cs = astropy.coordinates.AltAz(location=site_location, obstime=time)

    pixel_longitudes, pixel_latitudes = healpy.pixelfunc.pix2ang(n_side,
                                                                    range(healpy.pixelfunc.nside2npix(n_side)),
                                                                    lonlat=True)
    pixel_longitudes *= units.deg
    pixel_latitudes *= units.deg

    galactic_coordinates = astropy.coordinates.Galactic(l=pixel_longitudes * astropy.units.rad,
                                                        b=pixel_latitudes * astropy.units.rad)
    local_coordinates = galactic_coordinates.transform_to(local_cs)

    return local_coordinates
```
Turns out it does not really safe any time because apparently this code is quite fast. However, you might want to add it anyway because it makes the code a little bit more readable?

fschlueter:
Is that correct? Why does one multiply two different units into a variable back to back
```
    pixel_longitudes *= units.deg
    pixel_latitudes *= units.deg

    galactic_coordinates = astropy.coordinates.Galactic(l=pixel_longitudes * astropy.units.rad,
                                                        b=pixel_latitudes * astropy.units.rad)
```

fschlueter:
@karenterveer and @sjoerd-bouma I took the liberty to address a few of the things I commented. Please let me know if you are happy with that.

NuRadioReco/modules/channelGalacticNoiseAdder.py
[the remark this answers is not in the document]
fschlueter: But it also does not allow the user to set the frequency spacing themself since the number of bins is fixed. I would keep both - why deprecating one?

NuRadioReco/modules/channelGalacticNoiseAdder.py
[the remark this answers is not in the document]
fschlueter: But only returning an error and than not adding the noise? Wouldn't it be better to raise an error?
This might make it a bit more readable? 
```
        # check that or all channels channel.get_frequencies() is identical
        freqs = station.get_channel(station.get_channel_ids()[0]).get_frequencies()
            if not np.allclose(freqs, channel.get_frequencies(), rtol=0, atol=0.1 * units.MHz):
                logger.error("The frequencies of each channel must be the same, but they are not!")
```

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -27,55 +47,132 @@ class channelGalacticNoiseAdder:
...
-        self,
-        event,
-        station,
-        detector,
-        passband=None
+            self,
+            event,
+            station,
+            detector,
+            passband=None
```
fschlueter, line 173: isnt that typical called `bandpass`?

sjoerd-bouma: https://en.wikipedia.org/wiki/Passband

The bandpass window of a bandpass filter is called the passband, I think ; )

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: Hm but this is because we use the same phase in the projection of the two polarisations right?

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: + maybe a global sign flip (i.e., `-2 pi ...`)

fschlueter: For the additional in-ice part I am getting `z * cos(theta) (n_ice/ n_air - sin^2(theta))`. But this could very well be wrong. Well it is wrong because it does not take into account refraction at the air-ice boundary but even if we ignore that it is probably wrong.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -91,132 +188,167 @@ def run(
...
+            t_phi = geometryUtilities.get_fresnel_t_s(zenith, n_ice, 1)
+            fresnel_zenith = geometryUtilities.get_fresnel_angle(zenith, n_ice, 1.)
+
+            if fresnel_zenith is None:
+                continue
+
+            temperature_interpolator = scipy.interpolate.interp1d(self.__interpolation_frequencies,
+                                                                  np.log10(self.__noise_temperatures[:, i_pixel]),
+                                                                  kind='quadratic')
+            noise_temperature = np.power(10, temperature_interpolator(freqs[passband_filter]))
```
fschlueter, line 242: I benchmarked a couple of interpolation methods. We could probably speed up this by a factor of ~ 2 if we would cache the interpolation functions. However, why dont we cache the `e_field_amplitude` for a given vector of frequencies (after interpolation), this could also only be optional as it would increase the time spend in `begin()` but it should make running it faster (you could also actually cache it for a given pixel and frequency vector). And it might makes the the frequency interpolation unnecessary!

fschlueter: here is the benchmarking:
```
In [13]: %timeit np.interp(x, x0, y0)
4.52 µs ± 17.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [14]: %timeit scipy.interpolate.interp1d(x0, y0)(x)
16.1 µs ± 53.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [15]: %timeit scipy.interpolate.interp1d(x0, y0, kind="quadratic")(x)
55 µs ± 77.1 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [16]: f = scipy.interpolate.interp1d(x0, y0, kind="quadratic")

In [17]: %timeit f(x)
21.4 µs ± 61.9 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [18]: f_lin = scipy.interpolate.interp1d(x0, y0)

In [19]: %timeit f_lin(x)
8.9 µs ± 22.3 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

fschlueter:
Another thing @sjoerd-bouma when implementing coherent noise, it would make sense if you can provide a list of arrival directions which are used to calculate the phase delay with a plane wave for the pixel in which the direction is (otherwise one uses the pixel center) - Or maybe not because this amplifies the noise from this particular direction to much since it is not coming from the entire pixel?

fschlueter:
Not sure If you can make sense of my thoughts.

sjoerd-bouma:
> Just as a note: The module pygdsm added in their newest version (1.5.4) match-case statements to their code. These statements are only supported for python 3.10+. To make pygdsm work with older python versions one needs to install version 1.5.2. I think we should add the information somewhere (maybe it is already added, but I was not able to find it).

I have an [open PR on pygdsm](https://github.com/telegraphic/pygdsm/pull/29) that fixes this, but it seems development there is not very active. We could consider temporarily switching to my fork to fix this.

sjoerd-bouma:
> Another thing @sjoerd-bouma when implementing coherent noise, it would make sense if you can provide a list of arrival directions which are used to calculate the phase delay with a plane wave for the pixel in which the direction is (otherwise one uses the pixel center) - Or maybe not because this amplifies the noise from this particular direction to much since it is not coming from the entire pixel?

I'm not sure if I understand this point correctly, but in case I did, this is actually a very good point that I had overlooked earlier - we go from uncorrelated noise in each antenna (not good) to coherent noise from 192 pixels (n_side=4, the default)... which probably strongly overestimates the coherence and for something like SKA will just result in 192 point sources.

I'm not sure if I understand your suggestion for how to fix this, though. Should we pick a random direction 'inside' each pixel for each antenna?

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -27,55 +47,132 @@ class channelGalacticNoiseAdder:
...
-        self.__interpolaiton_frequencies = interpolation_frequencies
+        self.solid_angle = healpy.pixelfunc.nside2pixarea(self.__n_side, degrees=False)
+
+        if interpolation_frequencies is None:
+            if freq_range is None:
+                freq_range = np.array([10, 1100]) * units.MHz
+
+            # define interpolation frequencies. Set in logarithmic range from freq_range[0] to freq_range[1],
+            # rounded to 0 decimal places to avoid import errors from LFmap and tabulated models.
+            self.__interpolation_frequencies = np.around(np.logspace(*np.log10(freq_range), num=15), 0)
```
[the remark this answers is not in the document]
jakobhenrichs: I guess how it was done before should solve the issue (see https://github.com/nu-radio/NuRadioMC/pull/740#discussion_r1831047070).

NuRadioReco/modules/channelGalacticNoiseAdder.py
sjoerd-bouma: I think you've got your signs wrong, `x cos \phi` should have the same sign as `y sin \phi`. And a positive phase implies a negative time delay in the NuRadio convention, so that is currently accurate as well. I get a result for the depth term that's different from both what's in there and your suggestion - maybe we should compare notes at some point!

NuRadioReco/modules/channelGalacticNoiseAdder.py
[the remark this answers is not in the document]
sjoerd-bouma: Good catch! Will fix

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: I get the sign from the rotation matrix around the z axis 
image

fschlueter:
> > Another thing @sjoerd-bouma when implementing coherent noise, it would make sense if you can provide a list of arrival directions which are used to calculate the phase delay with a plane wave for the pixel in which the direction is (otherwise one uses the pixel center) - Or maybe not because this amplifies the noise from this particular direction to much since it is not coming from the entire pixel?
> 
> I'm not sure if I understand this point correctly, but in case I did, this is actually a very good point that I had overlooked earlier - we go from uncorrelated noise in each antenna (not good) to coherent noise from 192 pixels (n_side=4, the default)... which probably strongly overestimates the coherence and for something like SKA will just result in 192 point sources.
> 
> I'm not sure if I understand your suggestion for how to fix this, though. Should we pick a random direction 'inside' each pixel for each antenna?

You expanded my point actually. I meant that if we are beamforming into a specific direction we have to take into account coherent noise from this very direction. Right now we will have noise which is somewhat coherent as it originated form the pixel center. However, I agree that will (and this is true for every pixel direction) overestimate the acutal coherence from this very location... Not quite sure how we can handle this maybe we add a time jitter according to the pixel width?

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -279,20 +283,10 @@ def run(
...
-                                (np.cos(azimuth) * channel_pos_x + channel_pos_y * np.sin(azimuth)) +
-                                channel_depth *
-                                ((n_ice / n_air) ** 2 + sin_zenith ** 2) /
-                                np.sqrt((n_ice / n_air) ** 2 - sin_zenith ** 2)
-                        )
-                )
+                dt = geometryUtilities.get_time_delay_from_direction(
+                    curr_fresnel_zenith, azimuth, channel_pos, n=curr_n)
+
+                delta_phases = - 2*np.pi * freqs[passband_filter] * dt
```
fschlueter, line 289: space

NuRadioReco/modules/channelGalacticNoiseAdder.py
sjoerd-bouma: I think (hope) I got my maths right ;) but I've resorted to just using `geometryUtilities.get_time_delay_for_direction` instead anyway. If you want to double check your maths I think the easiest is to just take the dot product of arrival direction and antenna position. And positive x at phi=0 should definitely give the same sign as positive y and phi=90 deg ;)

sjoerd-bouma:
> > > Another thing @sjoerd-bouma when implementing coherent noise, it would make sense if you can provide a list of arrival directions which are used to calculate the phase delay with a plane wave for the pixel in which the direction is (otherwise one uses the pixel center) - Or maybe not because this amplifies the noise from this particular direction to much since it is not coming from the entire pixel?
> > 
> > 
> > I'm not sure if I understand this point correctly, but in case I did, this is actually a very good point that I had overlooked earlier - we go from uncorrelated noise in each antenna (not good) to coherent noise from 192 pixels (n_side=4, the default)... which probably strongly overestimates the coherence and for something like SKA will just result in 192 point sources.
> > I'm not sure if I understand your suggestion for how to fix this, though. Should we pick a random direction 'inside' each pixel for each antenna?
> 
> You expanded my point actually. I meant that if we are beamforming into a specific direction we have to take into account coherent noise from this very direction. Right now we will have noise which is somewhat coherent as it originated form the pixel center. However, I agree that will (and this is true for every pixel direction) overestimate the acutal coherence from this very location... Not quite sure how we can handle this maybe we add a time jitter according to the pixel width?

Let's see if we can come up with something clever here... I don't think this is a very straightforward problem, as the coherence depends not only on the pixel size (in fact, for a single antenna we of course don't need to worry about this) but also on the antenna-to-antenna distance, and getting this right for >2 antennas will be difficult, as one needs to consider all possible pairings. I would be open to deferring this to a future PR.

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: You are right, I am just wondering where I am doing something wrong in my calculation? Do I have to rotated by `theta = - azimuth` because then since `cos(x) = cos(-x)` and `sin(x) = -sin(-x)` I would get a `+` in between the x and y term

fschlueter: But I agree your solution with the utility is better (did you check for efficiency?).

fschlueter: probably this is irrelevant

sjoerd-bouma: > probably this is irrelevant

Like I mentioned somewhere below, everything is `get_antenna_response_vectorized` right now - if we can think of a way to speed this up by an order of magnitude or more we can look at other parts of this code for performance, but until then this is a waste of time, I think.

NuRadioReco/modules/channelGalacticNoiseAdder.py
sjoerd-bouma: @karenterveer I also tried to install the radiocalibrationtoolkit, but it seems it's a bit picky (e.g. it requires an older version of pygdsm, there's also an install-time dependency on `wheel` that's not documented because it still relies on an outdated `setup.py` to install)... if you still remember how to install these modules it would be nice if you could document the instructions!

Another reason why we should get this right is that right now this module is incompatible with pygdsm and radiocalibrationtoolkit at the same time - we need pygdsm 1.5 to avoid an ImportError for `GlobalSkyModel16`, but radiocalibrationtoolkit still expects `GlobalSkyModel2016` which was apparently deprecated at some point in pygdsm.

sjoerd-bouma:
I think we need to figure out our dependencies before merging this. (see also https://github.com/nu-radio/NuRadioMC/pull/740#discussion_r1891609083) - is anyone actually using `radiocalibrationtoolkit` (@karenterveer ?)? 
Right now the galacticnoiseadder assumes both pygdsm > 1.5 (it will raise an ImportError if not), which is not actually compatible with the radiocalibrationtoolkit. In addition, even answering the question 'is anyone using it' might be hard to answer right now, as radiocalibrationtoolkit attempts to silently overwrite some pygdsm modules but is wrapped in a try/except here.

fschlueter:
<3

MijnheerD:
Just interjecting here because I am looking into the merge of the lofar branch. Are the changes from the `feature/galacticNoiseAdderUpgrade` branch included in this PR?

karenterveer:
> I think we need to figure out our dependencies before merging this. (see also [#740 (comment)](https://github.com/nu-radio/NuRadioMC/pull/740#discussion_r1891609083)) - is anyone actually using `radiocalibrationtoolkit` (@karenterveer ?)? Right now the galacticnoiseadder assumes both pygdsm > 1.5 (it will raise an ImportError if not), which is not actually compatible with the radiocalibrationtoolkit. In addition, even answering the question 'is anyone using it' might be hard to answer right now, as radiocalibrationtoolkit attempts to silently overwrite some pygdsm modules but is wrapped in a try/except here.

I do not know of anyone who uses the modules in radiocalibrationtoolkit. However, for LOFAR we use pylfmap (this one should not make any issues I hope), so it would be nice to keep that.

karenterveer:
> Just interjecting here because I am looking into the merge of the lofar branch. Are the changes from the `feature/galacticNoiseAdderUpgrade` branch included in this PR?

Yes, the changes are included

sjoerd-bouma:
> > I think we need to figure out our dependencies before merging this. (see also [#740 (comment)](https://github.com/nu-radio/NuRadioMC/pull/740#discussion_r1891609083)) - is anyone actually using `radiocalibrationtoolkit` (@karenterveer ?)? Right now the galacticnoiseadder assumes both pygdsm > 1.5 (it will raise an ImportError if not), which is not actually compatible with the radiocalibrationtoolkit. In addition, even answering the question 'is anyone using it' might be hard to answer right now, as radiocalibrationtoolkit attempts to silently overwrite some pygdsm modules but is wrapped in a try/except here.
> 
> I do not know of anyone who uses the modules in radiocalibrationtoolkit. However, for LOFAR we use pylfmap (this one should not make any issues I hope), so it would be nice to keep that.

In that case I think I would propose removing the radiocalibrationtoolkit (at least for now) - its dependencies conflict, it seems tricky to install, and I'm not even sure that it would do what we expect (e.g. I think it adds a CMB contribution to the pygdsm models, which may or may not have been added to them in a more recent update anyway if I understand its rather brief changelog correctly).

I'll see if I can add pylfmap to the installer.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -27,55 +48,132 @@ class channelGalacticNoiseAdder:
     would be too slow, the model is evaluated for a few frequencies and the log10
     of the brightness temperature is interpolated in between.
     """
+
     def __init__(self):
-        self.__debug = None
         self.__zenith_sample = None
         self.__azimuth_sample = None
```
fschlueter, line 54: Remove unused variables

sjoerd-bouma:
Okay, I'm not sure who wants to review this (maybe @fschlueter  or @anelles?) but I think I would merge this as is, particularly as the initial commit is already merged into develop via #656 . The channelGalacticNoiseAdder should now work, is also tested in the CI, and has progressed somewhat towards adding coherent noise. The status of the optional dependencies is that
- pygdsm (github version) can be pip installed on python >=3.6 again!
- radiocalibrationtoolkit has been removed due to incompatibilities with the (also used) most recent version of pygdsm
- pylfmap is installed - to use it, another couple of not-so-straightforward dependencies have to be installed (see [pylfmap/issues/2](https://github.com/F-Tomas/pylfmap/issues/2) ). This can probably be solved by converting the LFmap data to healpy format once and providing it as a download. This shouldn't be too much work, but I would be happy to defer this to a new PR.

The documentation test failing is unrelated and will be fixed by #800

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -197,33 +204,26 @@ def run(
...
-        pixel_latitudes *= units.deg
-
-        galactic_coordinates = astropy.coordinates.Galactic(l=pixel_longitudes * astropy.units.rad,
-                                                            b=pixel_latitudes * astropy.units.rad)
-        local_coordinates = galactic_coordinates.transform_to(local_cs)
+        local_coordinates = get_local_coordinates((site_latitude, site_longitude), station_time, self.__n_side)
 
         n_ice = ice.get_refractive_index(-0.01, detector.get_site(station.get_id()))
-        n_air = 1.000292  # TODO: This value applies under standard conditions. Does it hold for Greenland?
+        n_air = ice.get_refractive_index(depth=1, site=detector.get_site(station.get_id()))
```
fschlueter, line 217: Isn't `ice` a deprecated class? If it also just return a hardcoded float I would prefer to define it here openly in the code.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -242,35 +242,39 @@ def run(
             noise_temperature = np.power(10, temperature_interpolator(freqs[passband_filter]))
```
fschlueter, line 242: This and the lines before: Why not caching the `scipy.interpolate.interp1d` functions

fschlueter (approved):

NuRadioReco/modules/channelGalacticNoiseAdder.py
sjoerd-bouma: Better to use the (to be deprecated) class, IMO - we will catch all instances of this when we deprecate it.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -279,23 +283,13 @@ def run(
...
-                        (
-                                sin_zenith *
-                                (np.cos(azimuth) * channel_pos_x + channel_pos_y * np.sin(azimuth)) +
-                                channel_depth *
-                                ((n_ice / n_air) ** 2 + sin_zenith ** 2) /
-                                np.sqrt((n_ice / n_air) ** 2 - sin_zenith ** 2)
-                        )
-                )
+                dt = geometryUtilities.get_time_delay_from_direction(
+                    curr_fresnel_zenith, azimuth, channel_pos, n=curr_n)
```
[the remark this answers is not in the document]
sjoerd-bouma: For shallow in-ice antennas, it probably does alright. For deep in-ice antennas, this will mess up the coherence of the noise. We're kind of doing something bad with this PR, which is 'semi-implementing' coherent noise; to a sufficiently sensitive detector we simulate the galaxy as a small number of point sources. 

We could still raise a warning for channels with a depth < 5 m (for example), but seeing as there are so many caveats with the 'coherence' that we simulate I think we don't need to single them out... we could make an issue and consider it as a todo for a future PR to get coherence right?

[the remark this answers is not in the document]
fschlueter: I still think such a warning is appropriate

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: Yes we discussed this.

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: You do not mind to ask `ice` for `n_air`?

fschlueter: also the argument `depth=1` is very confusing. I would assume that depth is positively defined and that depth = 1 is equal to z=-1. I hope this is not the case.

sjoerd-bouma: What I'm worried about is having multiple hardcoded values for `n_air` in different places in the code base, which potentially differ in the last digit(s), which will probably somehow lead to some very hard to debug behaviour in some bizarre edge case far into the future. I am not going to defend the `ice` class which indeed should be deprecated.

sjoerd-bouma:
ping @fschlueter - fixed a merge conflict and added a warning when simulating noise for deep in-ice channels (where indeed not just the coherence but also the arrival angle and therefore antenna response will be simulated incorrectly).

fschlueter (approved):

PR #742, opened 2024-11-15
Add different colors to the different logging levels

fschlueter:
This PR adds some different colors to the log level in the logging message. E.g.: 
image
image

What is missing is something for the loglevel STATUS

### DIFF
--- a/NuRadioReco/utilities/logging.py
+++ b/NuRadioReco/utilities/logging.py
@@ -128,10 +128,37 @@ def get_fancy_formatter():
     -------
     formatter : logging.Formatter
     """
-    formatter = logging.Formatter(
-        '\033[33;20m%(levelname)s - \033[93m%(asctime)s - \033[32m%(name)s - \033[0m%(message)s',
+
+    class CustomFormatter(logging.Formatter):
+
+        def __init__(self, format, datefmt):
+            super().__init__(datefmt=datefmt)
+            grey = "\033[38;1m"
+            yellow = "\033[33;1m"
+            purple = "\033[35;1m"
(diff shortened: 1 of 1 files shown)


fschlueter:
> can we just add in the dict `logging.STATUS: yellow + "%(levelname)s - " + reset + format,` ,i.e., does the interpreter understand `logging.STATUS` maybe @MijnheerD

hmm that seems not to work - is that STATUS correctly implemented?

MijnheerD:
So logging.STATUS is not implemented currently, because the idea was to any potential clashed with other packages which implement the STATUS logging level. To implement your colorscheme, we can simply using the `LOGGING_STATUS` variable (the `FORMATS` dictionary is just mapping level `int` to a format) as I did in the last commit [6db7dce](https://github.com/nu-radio/NuRadioMC/pull/742/commits/6db7dce7c45a8b76eaba5208f56e84b25ea8ff16) .

If you wish to implement the `logging.STATUS` attribute, this is also possible by simply adding 
```python
setattr(logging, "STATUS", LOGGING_STATUS)
```
on line 17 (right after adding the level name). But then we open the possibility for clashes to occur, so I guess we need to decide whether it is worth it.

MijnheerD (approved):

fschlueter:
I am confused by what the `LOGGING_STATUS` does? Which numerical value gets it assigned?

MijnheerD:
It's just our internal variable for the logging value`logging.py - line 3: LOGGING_STATUS = 25`

fschlueter:
Ah great, I overlooked that!

PR #741, opened 2024-11-13
Adding a module to simulate the RNO-G trigger board (FLOWER) response

fschlueter:

PR #743, opened 2024-11-15
Simulate trigger channels for RNO-G

fschlueter:
This PR is based on #716 and #741. 

Currently the "everything regarding the trigger channels happens in the triggerBoardResponse module and the trigger module. I will also add a script which shows how to simulate the trigger.


fschlueter:
@cg-laser that highlights the idea I have - feel free to comment. Currently everything happens in the triggerBoardResponse.py but in principle we could also do it in the HardwareResponseIncoperator. The one downside with how it is implemented right now is that I first fold in the radiant response and a cable and later fold them out which is somewhat awkward.

sjoerd-bouma:
I'll try to have a look at this later this week but just so we don't forget - if we go with this implementation the trigger channel should probably be serialized as well, right? Right now I don't think it would be stored if you write to a .nur file.

fschlueter:
Hm I am getting this error sometimes (propabably only appears if more than one event triggered):
```
  File "/Volumes/software/nuRadioMC/NuRadioMC/examples/08_RNO_G_trigger_simulation/simulate.py", line 277, in <module>
    sim.run()
  File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1652, in run
    self._output_writer_hdf5.add_event_group(output_buffer)
  File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/output_writer_hdf5.py", line 127, in add_event_group
    assert all(np.atleast_1d(self._mout_groups_attributes[sid][key_hdf5] == np.array(channel_values))), f"station {sid} key {key_hdf5} is {self._mout_groups_attributes[sid][key_hdf5]}, but current channel is {np.array(channel_values)}"
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: operands could not be broadcast together with shapes (13,) (24,)
```
any idea @cg-laser ?

fschlueter:
> Hm I am getting this error sometimes (propabably only appears if more than one event triggered):
> 
> ```
> Traceback (most recent call last):
>   File "/Volumes/software/nuRadioMC/NuRadioMC/examples/08_RNO_G_trigger_simulation/simulate.py", line 277, in <module>
>     sim.run()
>   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/simulation.py", line 1652, in run
>     self._output_writer_hdf5.add_event_group(output_buffer)
>   File "/Users/felix/software/nuRadioMC/NuRadioMC/simulation/output_writer_hdf5.py", line 127, in add_event_group
>     assert all(np.atleast_1d(self._mout_groups_attributes[sid][key_hdf5] == np.array(channel_values))), f"station {sid} key {key_hdf5} is {self._mout_groups_attributes[sid][key_hdf5]}, but current channel is {np.array(channel_values)}"
>                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ValueError: operands could not be broadcast together with shapes (13,) (24,)
> ```
> 
> any idea @cg-laser ?

so for some reason, the second triggered event has only 15 channels?

fschlueter:
Okay, I think in principle I am finished. I will validate the simulation results soon and I will add the phased array trigger to the example. What is missing is to use `get_trigger_channel()` in all trigger modules.

NuRadioReco/framework/station.py
```diff
@@ -92,9 +92,35 @@ def iter_channels(self, use_channels=None, sorted=False):
...
+            if use_channels is None:
+                yield channel.get_trigger_channel()
+            else:
+                if channel_id in use_channels:
+                    yield channel.get_trigger_channel()
+
     def get_channel(self, channel_id):
         return self.__channels[channel_id]
 
+    def get_trigger_channel(self, channel_id):
```
cg-laser, line 106: we could also not implement this function to make clear that `trigger_channels` are not saved at station level, but are a sub-element of a channel. Users would need to write `station.get_channel(id).get_trigger_channel()` instead of `station.get_trigger_channel(id)`

fschlueter:
> > @cg-laser that highlights the idea I have - feel free to comment. Currently everything happens in the triggerBoardResponse.py but in principle we could also do it in the HardwareResponseIncoperator. The one downside with how it is implemented right now is that I first fold in the radiant response and a cable and later fold them out which is somewhat awkward.
> 
> My main critique here is that this is not only awkward but could make things impossible as bandpass filters can't be inverted. As the bandpass for the FLOWER is narrower than for the RADIANT would work for the current RNO-G,but for other systems it might break.

Fair point! Let me see if I can easily change this to the hardware response incorperator.

cg-laser (commented):
is there a way to automatically determine the bandwidth of the trigger channel signal chain? Currently, we detect all modules that have `get_filter` (or similar) implemented and multiply all filters together on-the-fly to calculate the bandwidth. This is used in NuRadioMC to calculate the Vrms from the noise temperature. 

It seems that this is no longer possible now. I guess, the triggerBoardRespone module could just implement the `get_filter` function which provided the "diff" to get from the DAQ signal chain to the trigger signal chain. 

Or we do it all in the `hardwareResponseIncorporator`. However, for pure simulation studies, it would be nice to have a generic "triggerBoardResonse" module that would create additional trigger channels. Right now, we do this within the custom user detector simulation code and never save the trigger channels anywhere. I'm thinking of a `triggerBandpassFilter` module. 
Or, this can be done in the usercode (https://github.com/nu-radio/analysis-scripts/blob/main/gen2-tdr-2021/detsim/D01detector_sim.py#L124) Instead of making a copy of the station, we would directly copy the trigger channels, apply an additional bandpass, and save them as trigger channels. But having a `triggerBandPass` module would simplify this code substantially.

fschlueter:
@cg-laser, have a look at `NuRadioMC/examples/08_RNO_G_trigger_simulation/simulate.py` where I calculate the thermal noise rms for the trigger channels. We can add such a function to response class for example.

NuRadioReco/framework/station.py
fschlueter: I added this interface for connivence because it makes the diff necessary to the trigger module minimal (I have not applied it to any but the hi-lo one yet). We could delete it for the reason you mention..

fschlueter:
> > @cg-laser that highlights the idea I have - feel free to comment. Currently everything happens in the triggerBoardResponse.py but in principle we could also do it in the HardwareResponseIncoperator. The one downside with how it is implemented right now is that I first fold in the radiant response and a cable and later fold them out which is somewhat awkward.
> 
> My main critique here is that this is not only awkward but could make things impossible as bandpass filters can't be inverted. As the bandpass for the FLOWER is narrower than for the RADIANT would work for the current RNO-G,but for other systems it might break.

This is now resolved. I am now creating and applying trigger responses in the hardwareResponseIncorperator.

fschlueter:
Besides the issue #770 this PR is almost done. We can modify more trigger modules to use the trigger channel getters. So feel free to comment.

fschlueter:
@cg-laser, Sjoerd and I just had a zoom call to discuss this issue. A short summary:
- The unit tests are failing because now we add the cable delay after we added noise instead of before. That means that even if after adding the cable delay the signals are at the exact same sample/position in time as before, they are different relative to the noise we added. This explains the failing unit tests.
- We agreed that we still want this change although it might be difficult to verify that everything is okay (because of the failing tests) for a couple of reasons. As already discuss is it easier when adding trigger channels, it is conceptually more clean, and if we start adding coherent noise (for example galactic noise) adding the cable delay before we add noise is simply wrong.
- The way forward would be to double check my code and verify that the failing tests/differences in the results are within the tolerance we expect when changing noise. 

Any thoughts comments?

fschlueter:
Just thinking out loud here. Do we need to keep, in simulations, all `trace_start_time`s the same? This clearly simplifies the trigger simulation. Of course we could account for different start times in the trigger simulations but it might be annoying to do that (in particular if we have added noise). I do not see a reason why this would be impossible, but is there one I do not see?

fschlueter:
I will close this PR for the moment. The main functionality will be merged with PR #788. I will add a issue for the missing feature which was attempted in this PR.

PR #744, opened 2024-11-17
remove * units.mV to yield everything in V (to get everything in mV should use: / units.mv)

CamphynR:
When using the rnog reader to loop over waveforms it was noticed that the full calibration (using the voltage calibration coefficients) was of by several orders of magnitude compared to expected values and linear conversion (V = cst * ADC). Using * units.mV seems to be the problem.
This PR changes the reader to always yield results in V. If results in mV are desired,
the  " * units.mV " should be replaced by a " / units.mV " and also added to the linear calibration logic.

### DIFF
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -688,7 +688,7 @@ class readRNOGData:
                 selectors=self._select_events)):
 
                 if self._read_calibrated_data:
-                    wfs = wfs * units.mV
+                    wfs = wfs
                 else:
                     # wf stores ADC counts
                     if self._convert_to_voltage:
@@ -788,7 +788,7 @@ class readRNOGData:
         for channel_id, wf in enumerate(waveforms):
             channel = NuRadioReco.framework.channel.Channel(channel_id)
             if self._read_calibrated_data:
-                channel.set_trace(wf * units.mV, sampling_rate * units.GHz)
+                channel.set_trace(wf, sampling_rate * units.GHz)
             else:
                 # wf stores ADC counts
                 if self._convert_to_voltage:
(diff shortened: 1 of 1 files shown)


fschlueter (approved):
Good catch!

cg-laser:
I didn't have a complete look at the module, but at some point the unit should be defined, i.e., we should not assume that the base unit remains Volts. I.e., at the first interaction with an external library, all variable should be multiplied with the unit have in the external library.

fschlueter:
> I didn't have a complete look at the module, but at some point the unit should be defined, i.e., we should not assume that the base unit remains Volts. I.e., at the first interaction with an external library, all variable should be multiplied with the unit have in the external library.

okay, that would mean multiplying by `units.V` @CamphynR can you do that?

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -688,12 +688,12 @@ def get_waveforms(self, apply_baseline_correction=None, max_events=1000):
...
 
                 if self._read_calibrated_data:
-                    wfs = wfs
+                    wfs = wfs * units.V
                 else:
                     # wf stores ADC counts
                     if self._convert_to_voltage:
                         # convert adc to voltage
-                        wfs = wfs * (self._adc_ref_voltage_range / (2 ** (self._adc_n_bits) - 1))
+                        wfs = wfs * (self._adc_ref_voltage_range / (2 ** (self._adc_n_bits) - 1)) * units.V
```
fschlueter, line 696: hm this is not necessary actually because `self._adc_ref_voltage_range` is already defined in volt

fschlueter (approved):

fschlueter:
@CamphynR thanks!

fschlueter:
Go ahead and merge this I would say!

PR #745, opened 2024-11-19
Remove \d to get rid of SyntaxWarning

MijnheerD:
This is a nitpicky PR, but everytime I import `NuRadioReco.detector.detector` I got a `SyntaxWarning: invalid escape sequence '\d' `. There were two places were this came from:
1. detector.py: A quote and backslash were misplaced, resulting in \d appearing in the string on line 151 (this was just a mistake)
2. db_mongo_read.py: In this case the \d was not a mistake, but rather part of a regex expression. But the \d is equivalent to [0-9], so I replaced this to stop getting the `SyntaxWarning` .


fschlueter (approved):
Appreciate your nit-picking! Especially if you solve it!

PR #754, opened 2024-11-20
Hotfix: Add empty channels also if config[noise] is false and the cha…

fschlueter:
…nnel is noiseless.

In certain situations missing channels (because of no raytracing solution) were not added to the station at the end. This PR fixes that. I also added a helper function `_add_empty_channel` to avoid code duplication. 

Edit: I also added a commit to store the station-level attributes in the hdf5 files. I think this got maybe lost in the refactoring?


cg-laser (approved):
thanks for catching this!

PR #756, opened 2024-11-20
add core and energy estimate from LORA!

karenterveer:
In addition to reading in the zenith and azimuth from LORA, the reader now also reads the core and energy estimation from LORA and adds them to the hybrid shower showerParameters.


MijnheerD (approved):
Excellent work!!

MijnheerD:
image

PR #757, opened 2024-11-21
Do not use refractive index of ice at LOFAR/Auger

sjoerd-bouma:
The old `NuRadioReco.utilities.ice` ice 'model' is still used internally by some modules, including the voltageToEfieldConverter which is used also by LOFAR. This leads to unwanted behaviour for antennas with z-coordinates < 0.

In future, this module should probably be deprecated because it also doesn't do the right thing for deep in-ice antennas (or any ice where the assumption that n_ice=1.3 is wrong). But this would be slightly more work, depending on how we want to implement it (e.g. should air-to-ice raytracing be included).

### DIFF
--- a/NuRadioMC/utilities/medium.py
+++ b/NuRadioMC/utilities/medium.py
@@ -1,3 +1,11 @@
+"""
+Module providing the ice models in NuRadioMC.
+
+For more details on the implementation, and the available modules,
+see the documentation :doc:`here </NuRadioMC/pages/Manuals/icemodels>`.
+
+"""
+
 from NuRadioMC.utilities import medium_base
 import itertools
 import numpy as np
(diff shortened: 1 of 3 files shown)


cg-laser (approved):
I agree, we should probably deprecate this module. The fix for now looks good. 

And yes, the `efieldToVoltageConverter` only works for local antennas. Two assumptions need to be fulfilled
- straight line propagation. i.e. constant index-of-refraction
- same efield in all antennas

so this only works for very compact antenna clusters or the dual-polarizided antennas used in air-shower arrays.

fschlueter:
The value of `1.000293` for the refractivity of air is only a reasonable value for sea level (and not an altitude of ~3km) and probably not for arctic conditions with very dry air neither.

sjoerd-bouma:
> The value of `1.000293` for the refractivity of air is only a reasonable value for sea level (and not an altitude of ~3km) and probably not for arctic conditions with very dry air neither.

I think generally even the difference between this value and 1 is irrelevant for most of the contexts in  which it is used... but of course long-term, something better should be implemented.

PR #761, opened 2024-11-21
Refactor rnog noise importer module. WS only

fschlueter:


NuRadioReco/modules/measured_noise/RNO_G/noiseImporter.py
```diff
@@ -207,14 +205,15 @@ def run(self, evt, station, det):
...
 
     def end(self):
         self._noise_reader.end()
         n_use = np.array(list(self._n_use_event.values()))
         sort = np.flip(np.argsort(n_use))
-        self.logger.info("\n\tThe five most used noise events have been used: {}"
-                         .format(", ".join([str(ele) for ele in n_use[sort][:5]])))
+        self.logger.info(
+            "\n\tThe five most used noise events have been used: {}"
+            .format(", ".join([str(ele) for ele in n_use[sort][:5]])))
```
MijnheerD, line 218: Wouldn't this be cleaner as an f-string?

MijnheerD (approved):
Proud of you! :)

NuRadioReco/modules/measured_noise/RNO_G/noiseImporter.py
fschlueter: I think there is one other benefit (especially when you use `.debug()`) that this way the string just gets formatted when it is actually printed and not all the time, but not 100 sure

PR #763, opened 2024-11-21
Trigger Readout Window

cg-laser:
the trace_start_time will include the (channel-specific) trigger time delays at the end of a NuRadioMC simulation. 
For a clean implementation and to avoid bugs in old code, a new module channelReadoutWindowCutter was written. The previous `triggerTimeAdjuster` module will be deprecated in a future release.


fschlueter:
Do we want to move the deprecated modules into a folder `deprecated` (still allowing to import them from `NuRadioReco.modules` ofc?

cg-laser:
> Do we want to move the deprecated modules into a folder `deprecated` (still allowing to import them from `NuRadioReco.modules` ofc?

no, because then the usercode will fail and they won't get the deprecation message ;-) 
for the next release, I would just delete them.

NuRadioReco/modules/triggerTimeAdjuster.py
```diff
@@ -18,6 +18,9 @@ def __init__(self, log_level=logging.NOTSET):
         logger.setLevel(log_level)
         self.__sampling_rate_warning_issued = False
         self.begin()
+        logger.warning(f"triggerTimeAdjuster is deprecated and will be removed soon. In most cased you can safely delete the application\
+                       of this module as it is automatically applied in NuRadioMC simulations. If you really need to use this module, \
+                       please use the channelReadoutWindowCutter module instead.")
```
sjoerd-bouma, line 23: 1. using `\` this way does not lead to the right formatting - all the whitespace from the indentation is also included. To write the string over multiple lines you can just close and open the string (see e.g. the sampling rate warning).
2. This doesn't really matter for now, but can you change this to
```
import warnings

warnings.warn("Warning message", DeprecationWarning)
```
The reason for this is that this will allow us to raise errors from deprecation warnings in the tests (not yet implemented), which will make it easier to find and fix deprecated code used within NuRadioMC.

NuRadioReco/modules/triggerTimeAdjuster.py
```diff
@@ -76,6 +79,9 @@ def run(self, event, station, detector, mode='sim_to_data'):
             If the ``trigger_name`` was specified in the ``begin`` function,
             only this trigger is considered.
         """
+        logger.warning(f"triggerTimeAdjuster is deprecated and will be removed soon. In most cased you can safely delete the application\
+                       of this module as it is automatically applied in NuRadioMC simulations. If you really need to use this module, \
+                       please use the channelReadoutWindowCutter module instead.")
```
sjoerd-bouma, line 84: See previous comment

NuRadioReco/modules/channelLengthAdjuster.py
```diff
@@ -11,6 +11,8 @@ class channelLengthAdjuster:
     def __init__(self):
         self.number_of_samples = None
         self.offset = None
+        logger.warning("In most cases it is advisable to run a trigger module and use the channelReadoutWindowCutter module to cut the traces to the readout window \
+                       instead of this simple module.")
```
sjoerd-bouma, line 15: Formatting (don't use `\`, just close and open the string, otherwise you get a bunch of whitespace)

sjoerd-bouma (commented):
This looks fine to me - as far as I can tell it really only changes the trace start times back to how they were pre-2.2.0. 

I just have some nitpicky comments about multiline strings and a request to use the `warnings` module for DeprecationWarnings as this will make it easier to catch this stuff in the tests in the future.

sjoerd-bouma:
> > Do we want to move the deprecated modules into a folder `deprecated` (still allowing to import them from `NuRadioReco.modules` ofc?
> 
> no, because then the usercode will fail and they won't get the deprecation message ;-) for the next release, I would just delete them.

I think Felix's suggestion was to move the modules to a `deprecated` folder, and change the original module to `from deprecated import *`. You keep the functionality but move deprecated code away. I don't have strong feelings about this either way (though maybe we would want to name the module `_deprecated` to avoid it showing up as a suggestion for users).

NuRadioReco/modules/triggerTimeAdjuster.py
cg-laser: fixed

cg-laser:
thanks for the comments!  I fixed them. We still need to update the tests once we agree on the implementation

fschlueter:
What is with the `channelLengthAdjuster` do we still need this module?

fschlueter:
> What is with the `channelLengthAdjuster` do we still need this module?

Probably we can keep it because it is a simple module which one can use in reconstruction pipelines?

cg-laser:
yes, this module might be useful for some studies where running a trigger module would be overkill

fschlueter (dismissed):

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -0,0 +1,137 @@
...
+    def __check_sampling_rates(self, detector_sampling_rate, channel_sampling_rate):
+        if not self.__sampling_rate_warning_issued: # we only issue this warning once
+            if not np.isclose(detector_sampling_rate, channel_sampling_rate):
+                logger.warning(
+                    'triggerTimeAdjuster was called, but the channel sampling rate '
+                    f'({channel_sampling_rate/units.GHz:.3f} GHz) is not equal to '
+                    f'the target detector sampling rate ({detector_sampling_rate/units.GHz:.3f} GHz). '
+                    'Traces may not have the correct trace length after resampling.'
+                )
+                self.__sampling_rate_warning_issued = True
```
fschlueter, line 137: Do we need this warning at all? We ensure that we have the correct number of samples with:
```

                # this should ensure that 1) the number of samples is even and
                # 2) resampling to the detector sampling rate results in the correct number of samples
                # (note that 2) can only be guaranteed if the detector sampling rate is lower than the
                # current sampling rate)
                number_of_samples = int(
                    2 * np.ceil(
                        detector.get_number_of_samples(station.get_id(), channel.get_id()) / 2
                        * sampling_rate / detector_sampling_rate
                    ))
```
Maybe we can only check that `detector_sampling_rate <= channel_sampling_rate`?

sjoerd-bouma: I think when I implemented this I thought about it sufficiently to realize that if `sampling_rate < detector_sampling_rate` we cannot guarantee the correct number of samples (with the way resampling is currently implemented); I am not actually sure if the correct number of samples is guaranteed in all cases if `sampling_rate > detector_sampling_rate`; my feeling was that this is only true in most cases which is why I left the warning as it is. If you can work out the maths and figure out that this is wrong and we don't need the warning we can change this : )

sjoerd-bouma:
I'm a bit confused, and think potentially more changes are needed. The triggerTimeAdjuster (now channelReadoutWindowCutter) is called in line 1522; later non-trigger channels are added still, with (in the new convention) wrong trace start times; I am also not 100% sure that the slightly hacky looking adjustments to pre_trigger_time by making a deep copy of the sim channels was correct, but it probably wouldn't be correct anymore.

Is there any reason why we do not just call the triggerTimeAdjuster (now channelReadoutWindowCutter) at the end of the simulation, after downsampling to the detector sampling rate has already occurred? I feel like this would be much less error-prone. I guess there will be a small performance impact because we have to generate more noise, but probably this only makes a small difference in the grand scheme of things?

fschlueter:
I also think this is not yet correct. Let me elaborate a bit. The `SimChannels` from the non-trigger channels do not all have the same `trace_start_time` since they were simulated with `apply_det_response_sim` (maybe an obvious fact but I wanted to put it here). For those non-trigger channels, the `Channel` object is created with `self._add_empty_channel(station, channel_id)` which does:
```
        # we need to use any other channel to get the correct trace_start_time. All channels have the same start time at the end
        # of the simulation.
        channel.set_trace_start_time(station.get_channel(station.get_channel_ids()[0]).get_trace_start_time())
 ``` 
 So it creates a channel with the same `trace_start_time` as the first channel (which at this point has to be a trigger channel  I think/hope). For this trigger channel the `trace_start_time` is correct and already takes the `pre_trigger_time` into account. However, a new channel can have for the same trigger a different `pre_trigger_time` (we want to simulate different readout delays with that right?). So if we add the traces of all the `SimChannel` to this `Channel` we have to correct the `trace_start_time` for this channel and not for the `SimChannels` as currently done:
 ```
    if not station.has_channel(sim_channel.get_id()):
        # add empty channel with the correct length and time if it doesn't exist yet.
        self._add_empty_channel(station, channel_id)

    # Add the sim_channel to the station channel:
    channel = station.get_channel(sim_channel.get_id())
    # we need to account for the pre trigger time of the trigger that was used to determine the readout window
    pre_trigger_time = station.get_primary_trigger().get_pre_trigger_time_channel(channel_id)
    sim_channel_copy = copy.deepcopy(sim_channel)
    sim_channel_copy.set_trace_start_time(sim_channel.get_trace_start_time() + pre_trigger_time)
    channel.add_to_trace(sim_channel_copy)
```
Hence I am suggesting to change the function `_add_empty_channel` to: 
```
    def _add_empty_channel(self, station, channel_id, primary_trigger):
        """ Adds a channel with an empty trace (all zeros) to the station with the correct length and trace_start_time """
        channel = NuRadioReco.framework.channel.Channel(channel_id)
        n_samples = int(round(self._det.get_number_of_samples(station.get_id(), channel_id))
                        * self._config['sampling_rate'] / self._det.get_sampling_frequency(station.get_id(), channel_id))
        channel.set_trace(np.zeros(n_samples), self._config['sampling_rate'])
        
        trigger_channel_id = station.get_channel_ids()[0]
        trigger_channel_trace_start_time = station.get_channel(trigger_channel_id).get_trace_start_time()
        channel_trace_start_time = trigger_channel_trace_start_time + (
            primary_trigger.get_pre_trigger_time_channel(channel_id) - primary_trigger.get_pre_trigger_time_channel(trigger_channel_id))
        
        channel.set_trace_start_time(channel_trace_start_time)
        station.add_channel(channel)
```
And the code block in the run method to:
```
for evt in output_buffer[station_id].values():
        if not station.has_channel(sim_channel.get_id()):
            # add empty channel with the correct length and time if it doesn't exist yet.
            self._add_empty_channel(station, channel_id)

        # Add the sim_channel to the station channel:
        channel = station.get_channel(sim_channel.get_id())
        channel.add_to_trace(sim_channel_copy)
```

fschlueter:
FYI, I noticed that the function `_add_empty_channel` just exists in `develop` yet ... My code snippet should still work I hope.

fschlueter:
I started implementing my ideas into a new branch: `https://github.com/nu-radio/NuRadioMC/tree/refactor-triggertimeadjuster-fs`. Alongside the changes described above, I removed everything from the deprecated module + raising an Error now, and drastically reduced the code in the new module which now also uses the `add_to_trace` function.

fschlueter:
Running the test which fails on #820 on this branch gives us a lot of:
```
WARNING - 2025-02-04 14:16:36,095 - NuRadioReco.channelReadoutWindowCutter - trigger time is before 'trigger offset window' (requested samples before trigger = 275,trigger time sample = 166), the trace needs to be rolled by 109 samples first = 21.80ns
``` 
This could be the reason why the unit test fails? @cg-laser you updated at one point the unit tests with values which should match this simulations?

fschlueter:
What is also confusing me. Why can we roll the traces without adjusting `trace_start_time`? This module should not effect the signal time, correct? So when you roll a trace you would need to adjust for that by correcting the `trace_start_time`? However, the module and also the processor seemed to have had this feature since always. So is that a bug? It does of course not matter to much if you always roll all channels by the same amount but this is or was never guaranteed.

fschlueter:
> What is also confusing me. Why can we roll the traces without adjusting `trace_start_time`? This module should not effect the signal time, correct? So when you roll a trace you would need to adjust for that by correcting the `trace_start_time`? However, the module and also the processor seemed to have had this feature since always. So is that a bug? It does of course not matter to much if you always roll all channels by the same amount but this is or was never guaranteed.

Okay, apologies. I confused myself again. The rolling is just applied to bring the pulse to the correct position in the trace which matches the trace_start_time.

fschlueter:
@sjoerd-bouma I updated the documentation as you wished. However, I think now we are getting errors because we can not reference `NuRadioReco.modules._deprecated.triggerTimeAdjuster`. Any suggestion?

cg-laser (commented):
Thanks a lot Felix, I had some discussion with Martin this morning and it would be good to further improve the `add_to_trace` function. The function should get a `precision` parameter which determines if an additional residual time correction is performed. Now this is done inexplicitely by rounding to 5 digits. We can also allow more tolerance by default, e.g. 1ps should be sufficiently precise.

fschlueter:
Fine, but that can be done in a different PR?

fschlueter:
I just added a flag to `add_to_trace(channel)` that by default an error is raised if the trace of `channel` does not cover the entire readout window. The reason behind this is simple: If we have simulations with noise and this happens the resulting trace of the readout window will contain a bunches of zeros. Now the error actually is raised for `non-trigger` channels in one test. I actually implemented the flag exactly for this scenario. Background is that for non-trigger channels we do not ensure that traces are long enough. However, what I forgot is that for those channels the noise is not yet added. Which means we should be good. If everyone is okay with that I will just set the flag to False for this particular case of non-trigger channels.

fschlueter:
Okay, the `trace_start_time` failed for one test for 55 ns which sounds like it is the default `pre_trigger_time`. Shall we just update the reference values than?

MartinRavn:
I have two remaining concerns about the `channelReadoutWindowCuttter`:

1. It may not be obvious to a user that the function should not be used when the readout window is not almost completely inside the simulated trace. Now, the function will fail if the window is completely outside the trace, but maybe we should raise an error when the overlap is less than e.g. half the `number_of_samples`.
2. In `channelReadoutWindowCuttter` we round both the `trigger_time_sample` and `samples_before_trigger` to integers, but in in `simulation.py` when we treat the non-trigger channels, continuous time shifts are allowed. This means that the relative timing between antennas can be up to half a sample wrong if the trigger time and/or pre-trigger times are not multiples of `1/sampling_rate`. I believe it can be solved by calculating the residual time offset in `channelReadoutWindowCuttter` and adding it to the final trace start time.

Are these issues big enough that we should solve them, or can we leave it as it is?

fschlueter:
> I have two remaining concerns about the `channelReadoutWindowCuttter`:
> 
> 1. It may not be obvious to a user that the function should not be used when the readout window is not almost completely inside the simulated trace. Now, the function will fail if the window is completely outside the trace, but maybe we should raise an error when the overlap is less than e.g. half the `number_of_samples`.
> 2. In `channelReadoutWindowCuttter` we round both the `trigger_time_sample` and `samples_before_trigger` to integers, but in in `simulation.py` when we treat the non-trigger channels, continuous time shifts are allowed. This means that the relative timing between antennas can be up to half a sample wrong if the trigger time and/or pre-trigger times are not multiples of `1/sampling_rate`. I believe it can be solved by calculating the residual time offset in `channelReadoutWindowCuttter` and adding it to the final trace start time.
> 
> Are these issues big enough that we should solve them, or can we leave it as it is?

@MartinRavn, to you first point: I think we discussed such a check and I am agree we can implement it. And to your second point, will instead of using `trigger_time - pre_trigger_time` using `trigger_time - int(pre_trigger_time * sampling_rate) / sampling_rate` for the trace_start_time solve the problem?

MartinRavn:
We also need to take into account that trigger_time_channel has been rounded, so we could set the trace_start_time to:

`int(np.round((trigger_time - channel.get_trace_start_time()) * sampling_rate)) / sampling_rate + int(pre_trigger_time * sampling_rate) / sampling_rate`

What I came up with was to calculate the residual time offset:

`channel.get_times()[trigger_time_sample - samples_before_trigger] - (trigger_time - pre_trigger_time)`

and add it to the trace_start_time at the end.

fschlueter:
Are you sure this is necessary? It is only important that `t_trigger_in_channel_trace = trig.get_trigger_time() - channel.get_trace_start_time()` gives us the correct "location", it is not important that this location coincides with a sample.

MartinRavn:
I would say it is necessary. We set:
`channel.set_trace_start_time(trigger_time - pre_trigger_time)`
Which will not be the true trace start time since we took the bin nearest to `trigger_time - pre_trigger_time`

MartinRavn:
The `validate_separate_trigger_channels` test fails if I run it with a pre-trigger time which is not a multiple of `1/sampling_rate`. I did it by adding `pre_trigger_times = 55 * units.ns + 1.17341 * units.ns` to [this](https://github.com/nu-radio/NuRadioMC/blob/2a5acc46bebe95958a28d38acfbad9b7f1387b41/NuRadioReco/modules/trigger/highLowThreshold.py#L238) line. However, none of the solutions I suggested solves the problem, so the test maybe fails for unrelated reasons.

MartinRavn:
I get this error:
image

cg-laser:
Re throwing an error if there is no complete overlap in `add_to_trace`: Many use cases that will use this function only deal with (noiseless) signal. In this use case, the default behavior is that the two traces are not overlapping. So, I'm not sure how good it would be to have this as a default setting. At the minimum, we should tell the user in the error message, that this behaviour can be deactivated by a function argument. 

I agree with Martin, we need to set the trace start time correctly, otherwise we introduce arbitrary time shifts between antennas due to integer rounding. Adding the residual time shift to the trace start time (as Martin suggested) seems sufficient to me. 
@MartinRavn can you provide more content of the failed test? the test should also show which quantify is compared that leads to the difference.

fschlueter:
> Re throwing an error if there is no complete overlap in `add_to_trace`: Many use cases that will use this function only deal with (noiseless) signal. In this use case, the default behavior is that the two traces are not overlapping. So, I'm not sure how good it would be to have this as a default setting. At the minimum, we should tell the user in the error message, that this behaviour can be deactivated by a function argument.
> 
> I agree with Martin, we need to set the trace start time correctly, otherwise we introduce arbitrary time shifts between antennas due to integer rounding. Adding the residual time shift to the trace start time (as Martin suggested) seems sufficient to me. @MartinRavn can you provide more content of the failed test? the test should also show which quantify is compared that leads to the difference.

Yes that is a good suggestion (adding the suggestion to flip the flag in the error message). I am not sure I understand the situation you are describing but I am happy if we change the default to false if we set it to true in situations where we want it.

MartinRavn:
I believe it is the voltage traces that do not match. This is some of the information which came before what I posted a few days ago: 
image

fschlueter:
@MartinRavn any updates? Can you update the reference values for the failing test?

cg-laser:
the problem is that the test is not supposed to fail. It runs NuRadioMC twice, once with separate trigger channels, once without. The result should be exactly the same. If the test fails, we do something wrong.

NuRadioReco/framework/base_trace.py
```diff
@@ -299,20 +299,26 @@ def deserialize(self, data_pkl):
         if 'trace_start_time' in data.keys():
             self.set_trace_start_time(data['trace_start_time'])
 
-    def add_to_trace(self, channel):
+    def add_to_trace(self, channel, min_residual_time_offset=1e-5, raise_error=True):
```
cg-laser, line 302: unit missing. It should be `min_residual_time_offset=1e-5 * units.ns`

MartinRavn:
It was also my understanding that the `validate_separate_trigger_channels` has no reference values and just compares two NuRadioMC simulations that should give the same result. I will think about a way to solve the problem.

fschlueter:
> the problem is that the test is not supposed to fail. It runs NuRadioMC twice, once with separate trigger channels, once without. The result should be exactly the same. If the test fails, we do something wrong.

Right now only the "Single Event Test" is failing. And this is because the trace_start_times in the reference fail is different (because we never checked them). What Martin posted is a test failing for him if he changes the pre_trigger_time to not be a multiple of the sampling rate. 

So the code is still buggy and we have to fix it but the test failing in the CI is not this, the CI is not sensitive to the bug. Correct Martin?

fschlueter:
Fine! We still need to update the reference file for the Single Event Test

MartinRavn:
> > the problem is that the test is not supposed to fail. It runs NuRadioMC twice, once with separate trigger channels, once without. The result should be exactly the same. If the test fails, we do something wrong.
> 
> Right now only the "Single Event Test" is failing. And this is because the trace_start_times in the reference fail is different (because we never checked them). What Martin posted is a test failing for him if he changes the pre_trigger_time to not be a multiple of the sampling rate.
> 
> So the code is still buggy and we have to fix it but the test failing in the CI is not this, the CI is not sensitive to the bug. Correct Martin?

Sorry, I missed this comment. Yes, the CI is not sensitive to the bug since all tests have pre-trigger times which are multiples of 1/sampling_rate.

MartinRavn:
I have now implemented the most stringent errors that will still pass if there is a trigger on noise on the edge of the trace. I have tested that it only gives a warning if the trigger is on the very first or last bin of the trace.

The reference file for the Single Event Test still needs to be updated. I am not sure I can do this since I can't run the test locally.

MartinRavn:
Now a Test all examples fails because of the errors I implemented, however, I think something wrong is being done in one of the examples. It has given a lot of warnings for a long time. Here is the log of the test from a commit (2dbae80) to develop from two weeks ago: 
image

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -95,39 +95,50 @@ def run(self, event, station, detector):
...
+            elif pre_trigger_time_channel + number_of_samples / sampling_rate > trace_length / sampling_rate:
+                msg = ("End of the readout window is outside the end of the trace (trigger time = {}ns, "
+                       "pre-trigger time = {}ns, length of readout window {}ns, end of trace {}, requested time after trace = {}ns), "
+                       "the trace will be rolled over the edge to fit in the readout window").format(
+                            trigger_time, pre_trigger_time, number_of_samples/sampling_rate, channel.get_trace_start_time() + trace_length/sampling_rate,
+                            pre_trigger_time_channel + number_of_samples / sampling_rate - trace_length / sampling_rate)
+                logger.warning(msg)
+
+            # "roll" the start of the readout window to the start of the trace
+            channel.apply_time_shift(-pre_trigger_time_channel, silent=True)
```
fschlueter, line 136: I understood christian such that we should only "fall back" to fourier shifting if the pre_trigger_time is not a multiple of the sampling rate?

MartinRavn: I implemented that in the `apply_time_shift` method. So it should only use `np.roll` if the time shift is a multiple of `1/sampling_rate`.

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -95,39 +95,50 @@ def run(self, event, station, detector):
...
-
-            # shift trace to be in the correct location for cutting
-            trace = trace[cut_samples_beginning:(number_of_samples + cut_samples_beginning)]
+            if pre_trigger_time_channel < -pre_trigger_time:
+                msg = ("Start of the readout window is more than pre_trigger_time before the start of the "
+                       "trace (trigger time = {}ns, pre-trigger time = {}ns, start of trace {}ns, requested "
+                       "time before trace = {}ns), this would result in rolling over the edge of the trace "
+                       "and is not the intended use of this function").format(
+                            trigger_time, pre_trigger_time, channel.get_trace_start_time(), pre_trigger_time_channel)
+                logger.error(msg)
```
fschlueter, line 111: one general comment. I prefer f-strings, i.e., `f"trigger time = {trigger_time}"` because they are much easier to read. There is one very good argument to use `.format`: If you for example use `logger.debug` you should use `.format` because the string gets only "produced" if debug is enabled. But in this case where a if condition has to be fulfilled first it is okay to use f-strings

fschlueter: just for the future!

NuRadioReco/framework/base_trace.py
```diff
@@ -257,12 +258,23 @@ def apply_time_shift(self, delta_t, silent=False):
...
         """
-        if delta_t > .1 * self.get_number_of_samples() / self.get_sampling_rate() and not silent:
+        if np.abs(delta_t) > .1 * self.get_number_of_samples() / self.get_sampling_rate() and not silent:
             logger.warning('Trace is shifted by more than 10% of its length')
-        spec = self.get_frequency_spectrum()
-        spec *= np.exp(-2.j * np.pi * delta_t * self.get_frequencies())
-        self.set_frequency_spectrum(spec, self._sampling_rate)
+        
+        if np.abs(np.round(delta_t * self.get_sampling_rate()) - delta_t * self.get_sampling_rate()) < fourier_shift_threshold:
+            roll_by = int(np.round(delta_t * self.get_sampling_rate()))
```
fschlueter, line 270: can you use `round` instead of `np.round` it seems to be much faster (factor 30) (not that this matters much):
```
In [1]: %timeit round(4.9)
22.1 ns ± 0.081 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [2]: %timeit np.round(4.9)
650 ns ± 6.27 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

NuRadioReco/framework/base_trace.py
```diff
@@ -257,12 +258,23 @@ def apply_time_shift(self, delta_t, silent=False):
             Turn off warnings if time shift is larger than 10% of trace length
             Only use this option if you are sure that your trace is long enough
             to acommodate the time shift
+        fourier_shift_threshold: float (default: 1e-5 * units.ns)
+            Threshold for the Fourier shift. If the shift is closer to a multiple of
+            1 / sampling_rate than this, the trace is rolled instead of using the Fourier 
+            shift theorem to save time and avoid numerical errors in the Fourier transforms.
         """
-        if delta_t > .1 * self.get_number_of_samples() / self.get_sampling_rate() and not silent:
+        if np.abs(delta_t) > .1 * self.get_number_of_samples() / self.get_sampling_rate() and not silent:
```
fschlueter, line 266: same comment as below no need to use numpy

NuRadioReco/framework/base_trace.py
fschlueter: you could also simplify this by `abs(delta_t % sampling_rate) < thres`

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -0,0 +1,171 @@
...
+                logger.warning(msg)
+            elif trigger_time > channel.get_trace_start_time() + trace_length / sampling_rate:
+                msg = ("Trigger time is after the end of the trace (trigger time = {}ns, end of trace {}ns), "
+                       "this would result in rolling over the edge of the trace and is not the intended use "
+                       "of this function").format(
+                            trigger_time, channel.get_trace_start_time() + trace_length / sampling_rate,
+                            pre_trigger_time_channel + number_of_samples / sampling_rate - trace_length / sampling_rate)
+                logger.error(msg)
+                raise AttributeError(msg)
+            elif pre_trigger_time_channel + number_of_samples / sampling_rate > trace_length / sampling_rate:
```
fschlueter, line 127: I think you have to use  `detector_sampling_rate` here `number_of_samples / sampling_rate`

fschlueter:
FYI, I think @sjoerd-bouma modified the same reference nur file in #649. I do not now how correctly resolve merge requests in binary files. I set the reference file to pass this PR. After we merge it, the test will probably fail on develop and we have to open a new PR to fix it.

fschlueter:
Urg, why is that? Different floating point arithmetic on my machine and the CI machine?
image

NuRadioReco/framework/base_trace.py
MartinRavn: I thought about simplifying it like that, but it doesn't work when `delta_t` is just below a multiple of `1/sampling_rate`, e.g., `0.9 % 1 = 0.9`.

fschlueter: you are right that does not work. Sorry

MartinRavn: my suggestion also doesn't work. So I will just keep it as it was but without numpy.

fschlueter:
@sjoerd-bouma the test in question is *not* failing for me locally. Can you double check this? Or we just merge since we will have to fix the CI anyway due to a merge conflict with the focusfactor branch.

fschlueter (dismissed):

sjoerd-bouma (changes requested):
With both @MartinRavn and @fschlueter having looked at this in so much detail, I trust that fundamentally this PR will do what it's supposed to do : ) Still, some minor comments attached.

Also, the reference file does not pass the tests on my local machine.

sjoerd-bouma:
> @sjoerd-bouma the test in question is _not_ failing for me locally. Can you double check this? Or we just merge since we will have to fix the CI anyway due to a merge conflict with the focusfactor branch.

It does fail for me. Do you want me to update it?

MartinRavn:
> It does fail for me. Do you want me to update it? There shouldn't be any remaining merge conflicts, I hope, otherwise you should just merge develop back in (but I thought you had already done this?).

The test also passes for me, so it would be great if you could update it.

NuRadioReco/framework/base_trace.py
```diff
@@ -332,37 +349,59 @@ def add_to_trace(self, channel):
...
+            if raise_error:
+                logger.error("The channel is completely outside the readout window")
+                raise ValueError('The channel is completely outside the readout window')
             return
+
+        def floor(x):
+            return int(np.floor(round(x, 5)))
+
+        def ceil(x):
+            return int(np.ceil(round(x, 5)))
```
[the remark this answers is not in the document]
MartinRavn: If we have two channels that are supposed to have the same time grid but with different `trace_start_times`, the times we get from `BaseTrace.get_times` will not line up perfectly due to numerical errors. This caused an error and is a common way a user would use the function. Rounding doesn't solve the problem completely, since it will still fail if the grids are offset by half of the fifth decimal, but this is an unlikely situation.

NuRadioReco/framework/base_trace.py
```diff
@@ -300,20 +311,26 @@ def deserialize(self, data_pkl):
...
-        time window of "this" channel.
-        If this channel is an empty trace with a defined _sampling_rate and _trace_start_time, and a
-        _time_trace containing zeros, this function can be seen as recording a channel in the specified
-        readout window.
+        Adds the trace of another channel to the trace of this channel.
+
+        The trace of "channel" is only added within the time window of "this". If "this" is an empty
+        trace (i.e., a trace containing zeros) with a defined trace_start_time, this function can be
+        seen as recording a channel in the specified readout window. Hence, "this" is referred
+        to as the "readout" in the comments of this function.
```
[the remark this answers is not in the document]
MartinRavn: I improved the entire docstring now.

NuRadioReco/framework/base_trace.py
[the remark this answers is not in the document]
cg-laser: can we change it to something physical? I.e. `digits = int(math.log10(1/(0.01*units.ps)))`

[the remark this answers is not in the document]
fschlueter: should we use `min_residual_time_offset` here?

fschlueter:
> > @sjoerd-bouma the test in question is _not_ failing for me locally. Can you double check this? Or we just merge since we will have to fix the CI anyway due to a merge conflict with the focusfactor branch.
> 
> It does fail for me. Do you want me to update it? There shouldn't be any remaining merge conflicts, I hope, otherwise you should just merge develop back in (but I thought you had already done this?).

yes please

NuRadioReco/framework/base_trace.py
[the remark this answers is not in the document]
fschlueter: e.g. `int(np.ceil(round(x, int(np.log10(1 / min_residual_time_offset))))`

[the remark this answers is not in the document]
fschlueter: ah lol - what christian said :D

NuRadioReco/framework/base_trace.py
```diff
@@ -300,20 +311,27 @@ def deserialize(self, data_pkl):
...
         ----------
         channel: BaseTrace
-            The channel whose trace is to be added to the trace of this channel.
+            The channel whose trace is to be added to the trace of the current channel.
+        min_residual_time_offset: float (default: 1e-5)
+            Minimum residual time between the target bin of this channel and the target bin of the channel
+            to be added. Below this threshold the residual time shift is not applied to increase performance
+            and minimize numerical artifacts from Fourier transforms.
+        raise_error: bool (default: True)
+            If True, an error is raised if the incoming channel is not fully contained in the current channel.
```
sjoerd-bouma, line 333: I think the error is raised in the opposite case, no?

NuRadioReco/framework/base_trace.py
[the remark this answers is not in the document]
sjoerd-bouma: @MartinRavn what was the error that would be caused by this? That you would get the 'wrong' number of samples due to different rounding at the first and last sample? (i.e. the error that's now in line 403?).

My OCD also wonders if this should be a function of the grid spacing (i.e. `1/sampling_rate`) rather than an absolute number. But probably 0.01 ps will always be fine. (Anyway, there are probably larger timing errors introduced by the way we currently handle resampling... different discussion).

NuRadioReco/framework/base_trace.py
sjoerd-bouma: ... also in the docstring

sjoerd-bouma (commented):
Some more minor comments/questions.

Also, just for future reference (thanks Cosmin for teaching me!) - using `git mv` rather than `git rm` & `git add` preserves the file history, which e.g. for the triggerTimeAdjuster -> channelReadoutWindowCutter would have been useful.

NuRadioReco/framework/base_trace.py
MartinRavn: No. What is a bit unclear is that an error is raised both if the incoming channel is completely outside the readout window and if it is only partially inside the readout window. These two cases give different error messages. I can clarify this in the docstring.

NuRadioReco/framework/base_trace.py
```diff
@@ -332,37 +350,59 @@ def add_to_trace(self, channel):
...
+            i_end_channel = n_samples_channel
 
         # Determine the remaining time between the binning of the two traces and use time shift as interpolation:
         residual_time_offset = t_start_channel - t_start_readout
-        tmp_channel = copy.deepcopy(channel)
-        tmp_channel.apply_time_shift(residual_time_offset)
-        trace_to_add = tmp_channel.get_trace()[i_start_channel:i_end_channel]
+        if np.abs(residual_time_offset) >= min_residual_time_offset:
+            tmp_channel = copy.deepcopy(channel)
+            tmp_channel.apply_time_shift(residual_time_offset)
```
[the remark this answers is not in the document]
MartinRavn: I have also been confused about this a couple of times, but the current implementation is correct.

The target bins that we cast the channel into are a little to the left of the selected bins of the channel. This means that we should shift the channel a little to the right before we cast it for the physical signal to be at the correct point in time.

NuRadioReco/framework/base_trace.py
```diff
@@ -354,40 +355,38 @@ def add_to_trace(self, channel, min_residual_time_offset=1e-5, raise_error=True)
                 logger.error("The channel is completely outside the readout window")
                 raise ValueError('The channel is completely outside the readout window')
             return
-
+        
         def floor(x):
-            return int(np.floor(round(x, 5)))
+            return int(np.floor(round(x, int(np.log10(1/(0.01*units.ps))))))
```
fschlueter, line 360: why not using `min_residual_time_offset` here?

NuRadioReco/framework/base_trace.py
```diff
@@ -354,40 +355,38 @@ def add_to_trace(self, channel, min_residual_time_offset=1e-5, raise_error=True)
                 logger.error("The channel is completely outside the readout window")
                 raise ValueError('The channel is completely outside the readout window')
             return
-
+
```
fschlueter, line 358: please do not add whitespace here

NuRadioReco/modules/_deprecated/triggerTimeAdjuster.py
```diff
@@ -0,0 +1,9 @@
+import warnings
+
+class triggerTimeAdjuster:
+
+    def __init__(self, *args, **kwargs):
+        warnings.error("The module `triggerTimeAdjuster `is deprecated. In most cased you can safely delete the application "
+                       "of this module as it is automatically applied in NuRadioMC simulations. If you really need to use this module, "
+                       "please use the channelReadoutWindowCutter module instead.", DeprecationWarning)
+        raise NotImplementedError
```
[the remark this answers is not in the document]
fschlueter: yes - not sure how to fix that.

NuRadioReco/framework/base_trace.py
```diff
@@ -354,40 +355,38 @@ def add_to_trace(self, channel, min_residual_time_offset=1e-5, raise_error=True)
...
             i_start_channel = ceil((t0_readout - t0_channel) * sampling_rate_channel) # The first bin of channel inside readout
             t_start_channel = tt_channel[i_start_channel]
         # 3. Channel starts after readout window:
         elif t0_channel >= t0_readout:
-            if t0_channel > t0_readout and raise_error:
-                logger.error("The channel starts after the readout window")
-                raise ValueError('The channel starts after the readout window')
-
             i_start_readout = floor((t0_channel - t0_readout) * sampling_rate_readout) # The bin of readout right before channel starts
             t_start_readout = tt_readout[i_start_readout]
```
fschlueter, line 377: I am confused. Changing the error message to the other place does not make much sense to me...

fschlueter:
weird I manually rerun the CI this morning after it failed but now it passed..

NuRadioReco/framework/base_trace.py
sjoerd-bouma: I just meant that the docstring said the opposite of what actually happened in code - I do believe the code was previously correct, so I have taken the liberty of reverting your change there, and just adjusting the docstring. (It was my understanding that we want the readout window to be completely inside the incoming channel - if this is wrong please let me know!)

NuRadioReco/framework/base_trace.py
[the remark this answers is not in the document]
sjoerd-bouma: Rather than thinking about this, I just tested it, and you are of course correct!

NuRadioReco/modules/_deprecated/triggerTimeAdjuster.py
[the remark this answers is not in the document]
sjoerd-bouma: Well, the straightforward alternative is to raise a deprecation warning but not remove the entire class yet, so it is still available. But I don't know if many simulations were produced with the old triggerTimeAdjuster - they would all have been with `develop` anyway so I'm not sure to what extent we want to insist on stability.

NuRadioReco/framework/base_trace.py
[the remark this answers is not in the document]
sjoerd-bouma: As far as I'm concerned this is a slightly hacky workaround for the problem (the sampling rates are already ensured to be the same, so one can always calculate a matching integer number of samples, and just set t_last = t_start + n_bins * dt).
If in 8 years from now this leads to a weird machine-precision issue with some future unforeseen application of NuRadioReco, I will not be held responsible. But I can't think of an actual practical example where this doesn't work, so if you don't want to invest the time into thinking of a cleaner solution I won't insist.

NuRadioReco/framework/base_trace.py
MartinRavn: Sorry. I must have been confused yesterday. It is correct now.

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -0,0 +1,219 @@
...
+        detector: `NuRadioReco.detector.detector.Detector`
+        """
+        counter = 0
+        for i, (name, instance, kwargs) in enumerate(event.iter_modules(station.get_id())):
+            if name == 'channelReadoutWindowCutter':
+                counter += 1
+        if counter > 1:
+            logger.warning('channelReadoutWindowCutter was called twice. '
+                           'This is likely a mistake. The module will not be applied again.')
+            return 0
```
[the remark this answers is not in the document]
fschlueter: But for this we have already a different module right? The `channelLengthAdjuster`. That being said I am happy with that being removed. I think @cg-laser implemented it at one point for a reason but I think the reason is obsolet now?!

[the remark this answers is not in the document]
sjoerd-bouma: Okay, probably the use case I mentioned needs further modification of either this module or the channelLengthAdjuster anyway. Another tricky use case I can think of is if we add resimulation from .nur files as a feature, when this might also cause issues (not sure, I guess what would happen is this warning would be raised but it might not cause any problems if we keep the current structure where the additional channels are already initialized with the correct window). There may be other cases. But they should be debuggable if they happen because you raise a warning :)

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -17,10 +17,10 @@ class channelReadoutWindowCutter:
 
     def __init__(self, log_level=logging.NOTSET):
         logger.setLevel(log_level)
-        self.__sampling_rate_warning_issued = False
         self.begin()
 
     def begin(self):
+        self.__sampling_rate_warning_issued = False
```
fschlueter, line 23: why

sjoerd-bouma: We can raise the warning for every channel, for every event, or just once. Currently it's raised just once, though I decided to reset it in `begin` instead of the `__init__`. Doesn't make much of a difference.

sjoerd-bouma (approved):

NuRadioReco/framework/base_trace.py
fschlueter: okay fine!

fschlueter:
@MartinRavn I think the honour of merging this PR is yours. But dont wait to long I might ca not control myself.

fschlueter:
🥳

PR #765, opened 2024-11-21
Lofar/add pipeline plots

karenterveer:
The module modules/LOFAR/pipelineVisualization_LOFAR.py creates some nice plots at the end of the pipeline, i.e. the polarisation of antennas, the reconstructed arrival directions for all stations and also the fluence/timing/direction plot. Unless otherwise specified, it saves the plots in the current working directory. Furthermore it raises warnings if the polarisations in single antennas are very off or if the event time was before the outer station clocks were synced with the core station clocks.

### DIFF
--- a/NuRadioReco/modules/LOFAR/pipelineVisualizer_LOFAR.py
+++ b/NuRadioReco/modules/LOFAR/pipelineVisualizer_LOFAR.py
@@ -85,13 +85,15 @@ class pipelineVisualizer:
             self.logger.warning("No radio core found, using LORA core instead")
             core = lora_core
 
-        for i, station in enumerate(event.get_stations()):
-        #for station in event.get_stations():
+        azimuth = event.get_hybrid_information().get_hybrid_shower("LORA").get_parameter(showerParameters.azimuth)
+        zenith = event.get_hybrid_information().get_hybrid_shower("LORA").get_parameter(showerParameters.zenith)
+        
+        i=0
+        #for i, station in enumerate(event.get_stations()):
+        for station in event.get_stations():
 
             if station.get_parameter(stationParameters.triggered):
 
-                zenith = station.get_parameter(stationParameters.cr_zenith)
-                azimuth = station.get_parameter(stationParameters.cr_azimuth)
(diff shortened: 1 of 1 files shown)


NuRadioReco/modules/LOFAR/pipelineVisualizer_LOFAR.py
```diff
@@ -0,0 +1,397 @@
...
+
+from NuRadioReco.utilities import units
+from NuRadioReco.framework.parameters import stationParameters, channelParameters, showerParameters
+from NuRadioReco.modules.base.module import register_run
+
+
+class pipelineVisualizer:
+    """
+    Creates debug plots from the LOFAR pipeline - 
+    This is the pipelineVisualizerTM for LOFAR.
```
PhilippLaub, line 26: *glorious pipelineVisualizerTM :)

NuRadioReco/modules/LOFAR/pipelineVisualizer_LOFAR.py
```diff
@@ -0,0 +1,397 @@
...
+                good_antennas = good_antennas_dict[station.get_id()]
+                if len(good_antennas) >= min_number_good_antennas:
+                    for channel_id in good_antennas:
+                        timelags.append(station.get_channel(channel_id).get_parameter(channelParameters.signal_time))
+
+        for i, station in enumerate(event.get_stations()):
+            # plot absolute station positions
+            if station.get_parameter(stationParameters.triggered):
+                station_pos = detector.get_absolute_position(station.get_id())
+                ax.scatter(station_pos[0], station_pos[1], color=cmap(norm(i)), s=20, label=f'Station CS{station.get_id():03d}')
```
PhilippLaub: maybe a different marker would be good to differentiate between station positions and antenna positions later (square or star?)

karenterveer: good point! (pun intended)

PhilippLaub: don't use points!

Issue #500, opened 2023-02-27
Simulation of ADC expects symmetric voltage to ADC relation


fschlueter:
The title of the issue might be confusing, sorry for that.

I have the suspicion that the following line is dangerous: 

**NuRadioReco/modules/analogToDigitalConverter.py:40**
```
lsb_voltage = adc_ref_voltage / (2 ** (adc_n_bits - 1) - 1)
```
with `ads_ref_voltage` being:
```
    adc_ref_voltage: float
        Voltage corresponding to the maximum number of counts given by the
        ADC: 2**adc_n_bits - 1
```
However, this is only true if voltage_max = - voltage_min, correct? I.e. if that holds:
```
v_ref_peak_to_peak = (voltage_max - voltage_min) = 2 voltage_max 
lsb_voltage = v_ref_peak_to_peak / (2 ** (adc_n_bits) - 1) = voltage_max / (2 ** (adc_n_bits - 1) - 1)
```
(compare the exponents).

cg-laser:
I think in general you are correct but this is so far a "simulation only" module. If we want to model, e.g., RNO-G in the future, we would plug in the measured voltage to ADC conversion. So in the future, one could add the option to read the conversion from some calibration table/database etc. 
I don't think it would be useful to just add an `offset` parameter to this module.

fschlueter:
Agree, but why not changing the code to use:
```
v_ref_peak_to_peak / (2 ** (adc_n_bits) - 1)
```
This is also correct with an offset (?). After all an offset, i.e., a pedestal is typically removed.

fschlueter:
I stumbled across this last week again. I am feeling that this will cause a bug sooner or later. For example If I understand the code correct,  to simulate the RADIANT response correctly we want to simulate the effect of an 12-bit digitiser with a range of 2.5V. However, with the current code we would need to set the reference voltage to 1.25V correct? This is awkward.

ryankrebs016:
Bringing this convo from the above PR to here. 

[...]

PR #768, opened 2024-11-22
Allow negative delays in trace_utilities.delay

fschlueter:
This is not finished yet and will break the tests. We have to take a few decisions here. Currently I am always cropping the traces after delaying to remove the now unphysical signals at the beginning or end. However, in case of a positive delay this will change the trace start time (as it did before). This is now calculated in function. If we want to keep that we have to change the code where this function is used. I also changed the calculation of the number of samples to crop. We might need to check if that calculation agrees now with calculations outside of this function. And finally, why do we have the argument `delayed_samples` I would remove it.


fschlueter:
Okay one unit test is still failing which I will investigate. However, apparently the new warning I added is firing a lot during our simulation examples. Are we time-shifting traces with noise? Shouldn't we corp those traces than by default?

fschlueter:
```
CPP version of ray tracer is available
fraction of triggered events = 114/10000 = 0.011 -> 8.07561 km^3 sr
effective volume deviates 2119.7 sigma (3%) from the mean
deviation is more than 3 sigma -> this should only happen in less than 1\% of the tests. Rerun the test and see if the error persists.
```
what is going on here?

fschlueter:
The entire module `analogToDigitalConverter.py` and its long history is an enigma for me - which makes me very scared. @cg-laser. What is this module suppose to be and what not?

fschlueter:
Shall it simulate clock glitches in an ADC? Because this functionality exist, was changed several times throughout the modules history, and is - as far as I can tell - buggy.

anelles:
@fschlueter  if you don't like what analogToDigitalConverter.py does, can we take it to a different issue? I am having a hard time keeping track of whether we have a problem now or whether we don't.

fschlueter:
Sure, there is also already an issue #500 (not really related to my criticism here though) which I would like to see discussed. Regarding this PR: It does change what the modules does -> it slightly changed the result of one unit test. I double checked what both codes (old and new) did and came to the conclusion that the old code was buggy.

fschlueter:
> I think this looks good.
> 
> I'm also in favor of rewriting and improving the digitalToAnalog converter. I think this goes back to Daniels work(?). I never worked on it myself so I can't provide more insights.
> 
> Changing the meaning of detector descriptions would be error prone. However, we can add new fields with slightly different (but more meaningful) names. Old settings can be detected and throw a deprecation error. This way, the user knows how to fix the code.
> 
> Can we discuss this in the next inicemc call? I'm still not fully sure about the best approach.

That sounds good!

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -304,49 +303,84 @@ def apply_butterworth(spectrum, frequencies, passband, order=8):
...
     """
     Delays a trace by transforming it to frequency and multiplying by phases.
-    Since this method is cyclic, the trace has to be cropped. It only accepts
-    positive delays, so some samples from the beginning are thrown away and then
-    some samples from the end so that the total number of samples is equal to
+
+    Since this method is cyclic, the delayed trace has to be cropped. A positive
+    delay means that the trace is shifted to the right, i.e., its delayed. A
+    negative delay would mean that the trace is shifted to the left. Samples
+    from the beginning or end are thrown away. Optionally one can crop the
```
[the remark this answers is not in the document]
fschlueter: Done!

NuRadioReco/modules/analogToDigitalConverter.py
```diff
@@ -306,20 +305,21 @@ def get_digital_trace(self, station, det, channel,
...
-        trace = delay_trace(trace, MC_sampling_frequency, adc_time_delay, delayed_samples)
-
-        times = times + 1.0 / adc_sampling_frequency
-        times = times[:len(trace)]
+        if adc_time_delay:
+            # Random clock offset
+            trace, dt_tstart = delay_trace(trace, sampling_rate, adc_time_delay)
+            if dt_tstart > 0:
+                # by design dt_tstart is a multiple of the sampling rate
+                times = times[int(dt_tstart / sampling_rate):]
```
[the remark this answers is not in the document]
fschlueter: done

fschlueter:
@sjoerd-bouma or @cg-laser something else or can we merge this?

sjoerd-bouma:
Should we be worried about the fact that we might now get an odd-length trace from `delay_trace`? I seem to recall that for some reason we insist on always having even-length traces. But maybe I'm misremembering.

fschlueter:
> Should we be worried about the fact that we might now get an odd-length trace from `delay_trace`? I seem to recall that for some reason we insist on always having even-length traces. But maybe I'm misremembering.

I think this might be handled by the module currently. But I can add a safeguard:
```
if corped_samples % 2:
    corped_samples += 1
```

fschlueter:
@sjoerd-bouma fine?

cozzyd:
Doesn't it make sense to zeropad by ceil(delay) in the correct direction prior to doing the shift? Then you should never get cyclic behavior? (Then you can crop again to retain the length). 

i.e. something like
```
pad = ceil(abs(delay))
trace = np.insert(trace, 0 if delay < 0 else len(trace) -1, np.zeroes(pad)))
... 
trace = trace[pad:] if delay > 0 else trace[:-pad]] 
```

fschlueter:
@cozzyd if I understand you correctly you first want to add zeros, such that only the zeros get cycled and than cut the cycled zeros out again? But this way you change nothing, ie you did not change the position of the signal in the trace. Or you do but now have a patch of zeros in the trace?

sjoerd-bouma (dismissed):

NuRadioMC/test/Veff/1e18eV/T03check_output.py
```diff
@@ -41,12 +39,9 @@ def calculate_veff(fin):
 
 print('fraction of triggered events = {:.0f}/{:.0f} = {:.3f} -> {:.6g} km^3 sr'.format(n_triggered, n_events, n_triggered / n_events, Veff / units.km ** 3))
 
-delta = (Veff / units.km ** 3 - Veff_mean) / Veff_sigma
 rdelta = (Veff / units.km ** 3 - Veff_mean) / Veff_mean
```
cg-laser, line 42: Why did you change that? This makes the test much less stringent.

fschlueter: No, it was actually not used to decide the outcome of the test. It was only used to report the deviation in sigmas which since `Veff_sigma = 0.0001` it reported crazy high numbers which did not help in understanding what happens.

fschlueter: @cg-laser ?

cozzyd:
> @cozzyd if I understand you correctly you first want to add zeros, such that only the zeros get cycled and than cut the cycled zeros out again? But this way you change nothing, ie you did not change the position of the signal in the trace. Or you do but now have a patch of zeros in the trace?

Ah ok, I misunderstood the point of this. I thought the point of this was to do a sub-sample delay (because otherwise, you would just use np.roll or something). I don't think even with noise you can trust the rolled values though, because they will no longer obey the system bandwidth? (Except in the ideal case of full-band).

NuRadioMC/test/Veff/1e18eV/T03check_output.py
fschlueter: I really would like to get this merged asap. So please @cg-laser comment. My changes do not change what the test does. They only change what numbers are printed into the terminal. But I think these numbers were anyway very misleading.

fschlueter: @cg-laser can you please resolve the request for change or comment?

fschlueter: @cg-laser apologies for the change and my pushy behaviour. Indeed I changed the test condition. I changed that back to 3 sigma (sigma is a very small number). However, we are not reporting the deviation in sigma anymore but in %. This was, imo, very confusing and the reason I made the change in the first place. 

This being said, I noticed that the to `T03_check*py` files are basically the same besides 1 hard-coded float and one string (trigger name). I removed one file and allowed both tests to use one file.

cg-laser (approved):
thanks Felix

PR #771, opened 2024-11-27
remove per-station RadioShowers and use one RadioShower per event instead

PhilippLaub:
So far, for an event, the readLOFARData module adds one RadioShower for each station. Mitja suggested to use one RadioShower for the entire event instead, where one could store general event/shower information such as the direction of the incoming cosmic particle.


MijnheerD (approved):
Looks good!

PR #772, opened 2024-11-27
read propagation settings from config file if given

cg-laser:
I noticed that some setings for the propagation module were not read from the config file and needed to be explicitly passed to the class as arguments. 
I propose to read these settings instead from the config if given


fschlueter:
The change make sense to me. Can you add to the doc-string that the default values are taken from the config file and setting the argument explicitly will overwrite those values?

PR #774, opened 2024-12-02
Allow to set a sim_station to a station with `add_sim_station`

fschlueter:
Let the function do what the doc-string suggests.

### DIFF
--- a/NuRadioReco/framework/station.py
+++ b/NuRadioReco/framework/station.py
@@ -19,8 +19,7 @@ class Station(NuRadioReco.framework.base_station.BaseStation):
 
     def set_sim_station(self, sim_station):
         """
-        Sets the SimStation of the Station.
-        If a SimStation is already present, it is overwritten.
+        Sets the SimStation of the Station. If a SimStation is already present, it is overwritten.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


fschlueter:
For some reason this commit smuggled itself into two of my other PR, can we merge it?

NuRadioReco/framework/station.py
```diff
@@ -39,7 +38,10 @@ def add_sim_station(self, sim_station):
         sim_station : NuRadioReco.framework.sim_station.SimStation
             The SimStation to add to the Station.
         """
-        self.__sim_station = self.__sim_station + sim_station
+        if self.__sim_station is None:
+            self.__sim_station = sim_station
+        else:
+            self.__sim_station = self.__sim_station + sim_station
```
sjoerd-bouma, line 44: To me this (which I realize is not your change) still isn't what the docstring says - the addition operator somehow combines two stations rather than overwriting the old sim station. I'm guessing this is implemented this way because it is used in some way in the new simulation.py, but can you please update the docstring to actually reflect this? 

Then I won't argue about how it's unintuitive that this is what `set_sim_station` does...

fschlueter: the doc-string says
```
Adds a SimStation to the Station. If a SimStation is already present, the new SimStation is merged to the existing one.
``` 
which is okay I think

sjoerd-bouma: Ah my bad, I just saw the docstring for set_sim_station had changed and missed that this was the add_sim_station function.

sjoerd-bouma (approved):

PR #775, opened 2024-12-03
Docu of the Antenna model orientation description

fschlueter:
This is my shot at improving the documentation. The key point for me is that I always understood the rotation angles as an rotation around the orientation vector/axis and not that the angles define a second unit vector within the same coordinate system as the first. I tried to make this even more explicit as before (knowing what I know now the documentation was understandable but I thought others might have the same initial problems I had).

Anyway, I actually think sketch is also quite misleading. I would draw the two vectors into the same LPDA, add a little coordinate system and explain the angles for the sketch as one of the examples.

@christophwelling did you do the original sketch, could you make a new version?


cg-laser (approved):
looks good to me

fschlueter:
I would replace the sketch with this?!
orientation 001

christophwelling:
We used to have a drawing like yours in the documentation. The problem with it was that it was hard to tell if the rotation vector was pointing along the tines or perpendicular to it.That is still a little bit of an issue with your drawing. Maybe you can figure out a way to make this clearer?

fschlueter:
like this? 
image

fschlueter:
Sorry @cg-laser I dismissed your approval, can you do it again?

sjoerd-bouma (approved):

PR #776, opened 2024-12-04
Hotfix detector.py

fschlueter:
I did not run the code but just from looking at it I thought this was wrong? @sjoerd-bouma would you mind double reading this?

### DIFF
--- a/NuRadioReco/detector/detector.py
+++ b/NuRadioReco/detector/detector.py
@@ -57,7 +57,7 @@ def Detector(*args, **kwargs):
             - kwargs["source"] == "rnog_mongo" -> `NuRadioReco.detector.RNO_G.rnog_detector`
             - kwargs["source"] == "sql" -> `NuRadioReco.detector.detector_base`
             - kwargs["source"] == "json" or "dictionary" -> `NuRadioReco.detector.detector_base` or
-              `NuRadioReco.detector.generic_detector`
+            `NuRadioReco.detector.generic_detector`
 
         For 'kwargs["source"] == "json"', whether to use "detector_base" or "generic_detector"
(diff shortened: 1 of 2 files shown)


fschlueter:
For the first commit in this PR I have an extra PR #774

NuRadioReco/detector/detector.py
```diff
@@ -57,7 +57,7 @@ def Detector(*args, **kwargs):
             - kwargs["source"] == "rnog_mongo" -> `NuRadioReco.detector.RNO_G.rnog_detector`
             - kwargs["source"] == "sql" -> `NuRadioReco.detector.detector_base`
             - kwargs["source"] == "json" or "dictionary" -> `NuRadioReco.detector.detector_base` or
-              `NuRadioReco.detector.generic_detector`
+            `NuRadioReco.detector.generic_detector`
```
sjoerd-bouma, line 60: bad rst formatting (previous was correct)

sjoerd-bouma (commented):
As far as I can tell, this PR wouldn't actually change anything? If `source=='dictionary'`, `station_dict` and `dictionary` are the same, so nothing changes. And if `source=='json'`, the `dictionary` argument is ignored (as it should be), so the fact that you now pass a dictionary instead of `None`, while arguably not desirable, doesn't change anything.

I would probably recommend closing this. `station_dict` was originally (and still is) used to check whether the detector is generic or not.

NuRadioReco/detector/detector.py
fschlueter: okay, I changed it because my editor complained ...

fschlueter:
I missed the part that the source=="json" is correctly handled within the classes

PR #773, opened 2024-11-29
Documentation / poetry improvements and fixes

sjoerd-bouma:
I suppose technically this could have been two PRs but no functionality should be affected.
1. Optional dependencies weren't really implemented correctly in pyproject.toml (I accept full responsibility). This made very little difference, but the advantage is they can now be installed also using the conventional pip syntax `pip install nuradiomc[documentation, proposal]` to install the correct optional dependencies. Previously, optional dependencies were implemented as developer dependencies, and would never be installed during `pip` installation.
2. The rest is minor fixes to the documentation: some missing docstrings from the recent refactoring, and an attempt to somewhat tidy up the new 'overview of times' page, though I find the content somewhat unclear still. The table of contents on code documentation pages now also only shows top-level submodule and subpackage names instead of every method of every class inside them, which I think improves clarity,

PR #778, opened 2024-12-09
Clean up imports in  voltageToAnalyticEfieldConverterNeutrinos

fschlueter:
Also fixes broken channelResampler.begin() call.


sjoerd-bouma (approved):
Thanks for the merge conflict with #773 ;) (just joking, the cleanup is nice)

I should actually see if this module still works... I think there is an example for it, but I've never run it, and it's not part of the tests.

PR #779, opened 2024-12-12
Add argument to skip/not do consitency check in antenna pattern init.…

fschlueter:
… Some clean up

For a (to be honest to) large antenna pattern (pkl file size of 2GB) the consistency checks to 15 seconds! Increasing the init time from 5 to 20 seconds. 

What might be missing is a smart logic to automatically switch of the check (for example if the hash-check was successful.

### DIFF
--- a/NuRadioReco/detector/antennapattern.py
+++ b/NuRadioReco/detector/antennapattern.py
@@ -613,8 +613,8 @@ def get_pickle_antenna_response(path):
 
         download_from_dataserver(remote_path, path)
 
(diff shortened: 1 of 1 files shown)


anelles (approved):
Looks fine. Maybe just add a "desired feature issue" to keep track of logic that you suggest?

sjoerd-bouma (approved):
I'm rewriting this module somewhat at the moment but we can merge this for now - it's a useful feature and who knows how long it'll take before I've implemented the changes I want ;)

PR #780, opened 2024-12-12
add missing import statement modules/base/module.py 

jakobhenrichs:
There was a missing import statement in modules/base/module.py, which sometimes caused an error in line 77 of the file. With this PR the missing import statement is added.

### DIFF
--- a/NuRadioReco/modules/base/module.py
+++ b/NuRadioReco/modules/base/module.py
@@ -3,6 +3,7 @@ from timeit import default_timer as timer
 import NuRadioReco.framework.event
 import NuRadioReco.framework.base_station
 import NuRadioReco.detector.detector_base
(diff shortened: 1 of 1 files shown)


fschlueter:
Hi Jakob thanks for doing this. I found this in a couple of branches I have open but the fix never made it into the develop

fschlueter (approved):

fschlueter:
I approved it, however, what I think would be a litte bit of a better solution would be to import `import NuRadioReco.detector.detector as detectors` and than test for `detectors.rnog_detector.Detector` or `detectors.detector_base.Detector`. But his works as well

jakobhenrichs:
Hi Felix, I changed the import statement and if condition according to your suggestions.

fschlueter (approved):

PR #782, opened 2024-12-19
Store numpy objects to be compatible across numpy 2.0

sjoerd-bouma:
This fixes numpy 2.0 cross-compatibility by using `numpy.save` and `numpy.load` internally. As pickle stores also the function used to unpickle stored data this should be fully backwards compatible (i.e. reading old .nur files is unchanged).

There is some performance impact to reading (writing is essentially unchanged), but at least on my laptop (SSD) the file IO still dominates. So possibly on superfast PCIE SSDs the performance impact is noticeable, but the `.nur` format is probably not optimal for very-high throughput in general.

This replaces #720. My proposal is to merge this to master as a hotfix first, and then to develop after.

Medium to long term, other upgrades we can consider are safe(r) pickling (see https://docs.python.org/3/library/pickle.html#restricting-globals), and before I thought of this I got halfway to using `h5py` to store NuRadio events. I think it may still be worth finishing that up - there would be benefits to security, readability and potentially performance. But I think both of these can be discussed and implemented in future PRs.

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -5,7 +5,7 @@ on: [push, workflow_dispatch]
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/deploy_documentation.yaml
+++ b/.github/workflows/deploy_documentation.yaml
@@ -3,7 +3,7 @@ name: Deploy Documentation
 on: [push, pull_request, workflow_dispatch]
 jobs:
   prepare:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     outputs:
       runtests: ${{ steps.check.outputs.runtests }}
     steps:
@@ -22,7 +22,7 @@ jobs:
   build:
     needs: [prepare]
     if: needs.prepare.outputs.runtests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
       - uses: actions/checkout@v3 # get the develop branch
--- a/.github/workflows/run_tests.yaml
+++ b/.github/workflows/run_tests.yaml
@@ -4,7 +4,7 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   prepare:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     outputs:
       runtests: ${{ steps.check.outputs.runtests }}
     steps:
@@ -23,7 +23,7 @@ jobs:
   build:
     needs: [prepare]
     if: needs.prepare.outputs.runtests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
     - uses: actions/checkout@v3
--- a/NuRadioMC/EvtGen/generate_unforced.py
+++ b/NuRadioMC/EvtGen/generate_unforced.py
@@ -10,7 +10,7 @@ from scipy.interpolate import interp1d
 from NuRadioMC.simulation.simulation import pretty_time_delta
 from NuRadioMC.EvtGen.generator import write_events_to_hdf5
 from NuRadioMC.utilities import inelasticities
-import pickle
+from NuRadioReco.utilities.io_utilities import pickle
 import os
 import time
 import logging
--- a/NuRadioMC/EvtGen/generate_unforced.py
+++ b/NuRadioMC/EvtGen/generate_unforced.py
@@ -10,7 +10,7 @@ from scipy.interpolate import interp1d
 from NuRadioMC.simulation.simulation import pretty_time_delta
 from NuRadioMC.EvtGen.generator import write_events_to_hdf5
 from NuRadioMC.utilities import inelasticities
-from NuRadioReco.utilities.io_utilities import pickle
+import pickle
 import os
 import time
 import logging
--- a/NuRadioMC/SignalGen/ARZ/scripts/A01preprocess_shower_library.py
+++ b/NuRadioMC/SignalGen/ARZ/scripts/A01preprocess_shower_library.py
@@ -3,7 +3,7 @@ from NuRadioReco.utilities import units
 import os
 from scipy import interpolate as intp
 import glob
-from NuRadioReco.utilities.io_utilities import pickle
+import pickle
 import sys
 
 rho = 0.924 * units.g / units.cm**3  # density g cm^-3
--- a/NuRadioMC/SignalGen/ARZ/scripts/A01preprocess_shower_library.py
+++ b/NuRadioMC/SignalGen/ARZ/scripts/A01preprocess_shower_library.py
@@ -3,7 +3,7 @@ from NuRadioReco.utilities import units
 import os
 from scipy import interpolate as intp
 import glob
-import pickle
+from NuRadioReco.utilities.io_utilities import pickle
 import sys
 
 rho = 0.924 * units.g / units.cm**3  # density g cm^-3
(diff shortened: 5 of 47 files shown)


sjoerd-bouma:
So I probably didn't pick the ideal test file originally - it had no numpy ints or strings, and very long traces. Although I've now (I think) fixed the missing numpy scalars, I've also done some more extensive benchmarking... and it turns out that for smaller trace lengths the overhead _is_ significant.

Benchmark result on `master`:
```
WARNING:test-io:Write speed:   4.08 ms / event (17500 events total).
WARNING:test-io:Read speed :   0.57 ms / event (17500 events total).
```
...and on the current branch
```
WARNING:test-io:Write speed:   3.68 ms / event (17500 events total).
WARNING:test-io:Read speed :   2.35 ms / event (17500 events total).
```

The reason seems to be that `numpy.load` has a relatively high constant performance penalty, i.e. it becomes competitive with `pickle.loads` only for very large arrays. I've tried to benchmark that as well:

image

(I certainly don't trust some of the features here - who knows what's going on in the background with different parts of memory and CPU cache etc - but hopefully the overall trends can be believed. See also https://stackoverflow.com/a/58942584 for an independent, more exhaustive benchmark)

The short version is that `pickle` is much, much faster for loading small arrays, and as we currently store traces individually (each class has its own `serialize` function), we will probably not be able to do better unless we decide to store multiple (potentially many/all) traces 'together' (although in principle traces can have arbitrary shapes, so this probably comes with its own difficulties and performance penalties). So we probably have to come back to this after Christmas.

sjoerd-bouma:
So following the previous InIceMC call (slides [here](https://uppsala.app.box.com/s/oyldlu31cpd9f9sxwpph5v4yrqj62ro9/file/1745583344709)), I've decided for now to use the [fastnumpyio](https://github.com/divideconcept/fastnumpyio) library to speed things up. This essentially stores numpy arrays in a simpler way, resulting in performance similar to that of pickle

```
WARNING:test-io:Write speed:   3.44 ms / event (17500 events total).
WARNING:test-io:Read speed :   0.61 ms / event (17500 events total).
```

Potential downsides are that this code (it's a grand total of 10 lines of code) is probably not very widely used or tested. If anyone can think of some clever corner cases that we should test I'm happy to have a look at them.

We also discussed moving to a more widely-adopted standard format such as HDF5 or ASDF, but decided this is a goal for the medium or longer-term future.

sjoerd-bouma:
Also some open-source lawyer should tell me if including the MIT licence of the original code in the module docstring is sufficient or not : )

anelles:
> Also some open-source lawyer should tell me if including the MIT licence of the original code in the module docstring is sufficient or not : )

I think copying the license seems to be the right choice. (Not a lawyer either, but ...)

anelles:
Overall, I think this is a good short-term solution. Thanks @sjoerd-bouma!
I would be happy to approve it, but since I was not part of the call (and hence missed the discussions), I would give others the floor. Ping me, if no one else feels like it.

fschlueter:
Sorry for having not replied earlier. One reason is that the solution is a bit unsettling for me. The code from `fastnumpyio` looks like a lot magic happens... I have no concrete argument against it and I trust you if you think that is the best and practical solution... 
So if others (@cg-laser) have nothing to object I think we can merge this.

fschlueter (dismissed):

cg-laser (approved):
Hi Sjoerd, thanks! Looks good to me. 

(Although I would have been ok to break forward compatibility ;-) But this is a neat solution. Thanks)

sjoerd-bouma:
Apologies @cg-laser and @fschlueter - I also wanted to backport the `astropy.time` fixes in #652 and #520. Hopefully this allows 2.2.2 to produce stable files (stable reading is probably more of a challenge, but then I suppose most people will use some flavour of `develop` by now).

More frustratingly I also thought of an alternative solution involving `numpy.memmap` (instead of fastnumpyio) that may or may not be fast enough; however implementing and testing this would take the investment of another undetermined amount of time.

cozzyd:
I'm far from an expert on this, but I thought the "correct" way to set up a dispatch_table was with the `copyreg` module? https://docs.python.org/3/library/copyreg.html

sjoerd-bouma:
@cozzyd you are technically correct (the best kind of correct!), so I've changed it. For what it's worth, the [source code](https://github.com/python/cpython/blob/b5eed8c80e4e8b626d2eb13ce58df2864278ca03/Lib/copyreg.py#L12) of `copyreg.pickle()` just does exactly what I had implemented myself, but seeing as there's no difference I see no reason not to use the implementation that is suggested by the Python documentation.

@fschlueter / @cg-laser I think this PR can be reviewed/approved once the tests have passed.

cozzyd:
One difference is you may be able to load the normal pickle module now?

sjoerd-bouma:
> One difference is you may be able to load the normal pickle module now?

Internally, nothing is different - the global dispatch table is modified, so also previously there was no reason to import `pickle` from `io_utilities`. However, we still need the code in `io_utilities` to run, so it has to be imported somewhere. I've done this now by always importing pickle from io_utilities, I suppose a less error-prone method would be to load this module in both the `NuRadioReco` and `NuRadioMC` __init__.py, if you think that would be cleaner/safer?

fschlueter:
I think the failing test is not related to the changes here. I have seen this failing elsewhere today. Is that something which just happens quite often (due to randomness) or is that something new?

cozzyd:
> > One difference is you may be able to load the normal pickle module now?
> 
> Internally, nothing is different - the global dispatch table is modified, so also previously there was no reason to import `pickle` from `io_utilities`. However, we still need the code in `io_utilities` to run, so it has to be imported somewhere. I've done this now by always importing pickle from io_utilities, I suppose a less error-prone method would be to load this module in both the `NuRadioReco` and `NuRadioMC` `__init__.py`, if you think that would be cleaner/safer?

ah ok, I didn't look closely enough and thought somehow you were exposing a special pickle module object that had been modified correctly.  Is there any downside to always loading it in __init__.py? (will it e.g. cause surprises to 3rd party code?). If so, to avoid forgetting, we could do a with a check similar to decorator check script.

sjoerd-bouma:
> ah ok, I didn't look closely enough and thought somehow you were exposing a special pickle module object that had been modified correctly. Is there any downside to always loading it in `__init__.py`? (will it e.g. cause surprises to 3rd party code?). If so, to avoid forgetting, we could do a with a check similar to decorator check script.

I was initially reluctant to put it in the `__init__` because we have two of them, one for NuRadioReco and one for NuRadioMC. But seeing as all IO stuff is handled by NuRadioReco anyway, it's probably sufficient to put it in that one.  I'll sleep on it but I'll probably move it over to the NuRadioReco `__init__.py` tomorrow.

Re third party code: I would be surprised if there's a lot of third party code that relies on pickling numpy arrays (which, as we have also found out, is generally a bad idea : ) and it would probably be affected equally regardless of where we implement the pickling override.

fschlueter:
Dont we want to also include it in the init for NuRadioMC? I assume you thought about this and decided against it?

sjoerd-bouma:
> Dont we want to also include it in the init for NuRadioMC? I assume you thought about this and decided against it?

Yes, I spent some time thinking about this... in the end, I figured the main thing is we ensure that `.nur` files use this pickling mechanism, which means NuRadioReco is sufficient; if someone decides to pickle e.g. a couple of arrays from NuRadioMC-specific examples I think we're probably less concerned with the long-term readability of the data. For the HDF5 output of NuRadioMC this shouldn't affect anything anyway.

The alternative would be to put `import NuRadioReco` in the NuRadioMC `__init__.py`, which while possible I didn't like aesthetically. We could still do this if you prefer?

fschlueter (approved):

fschlueter:
So we merge this?

PR #784, opened 2025-01-13
[CI] Change default Python version to 3.10

sjoerd-bouma:
`ubuntu-latest` has transitioned to 24.04, which no longer supports our default Python version of 3.7. This PR changes the default CI Python version to 3.10. Older Python versions are still tested on pushes to `develop` as before (by manually specifying older versions of Ubuntu).

Note that this change will need to make it into the next release on `master` as well, as the release to PyPI will fail otherwise.

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -19,10 +19,10 @@ jobs:
         ref: 'refs/heads/master'
     - uses: actions/checkout@v4
 
-    - name: Set up Python 3.7
+    - name: Set up Python 3.10
       uses: actions/setup-python@v5
       with:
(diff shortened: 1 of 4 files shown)


fschlueter (approved):

fschlueter:
I am a bit confused. I have a branch (without an associated PR yet) which fails because `pygdsm` is missing. I though it is related to this fix but I merged this branch in mine...

fschlueter:
@sjoerd-bouma I think we should go ahead and merge this as it blocks further development (because the CI on other PRs fail)

sjoerd-bouma:
Okay, I'll break the 24-hour rule; no code is affected anyway, just the CI.

Issue #785, opened 2025-01-13
Upgrade CI tests to Ubuntu 24.04


sjoerd-bouma:
Our actions run on `ubuntu-latest`, which is currently transitioning to `24.04`. This version no longer support our default python version of 3.7, so all the tests will fail for a bit.

This is fixed by #784 upgrading our default Python version to 3.10. If we want to make another release to PyPI, this change will have to be included in the `build_and_release` action prior.

PR #786, opened 2025-01-13
Update hardwareResponseIncorporator.py

avijai2000:
Correcting the "phase_only" mode amp_response in the hardwareResponseIncorporater.py file to ensure that the magnitude of the phase is equal to 1 as expected.


sjoerd-bouma (approved):
The tests are failing because of #785, but this fix looks good (I also made sure that it actually passes the tests https://github.com/nu-radio/NuRadioMC/actions/runs/12767510283/job/35586021705). Thanks @avijai2000 for fixing this!

PR #787, opened 2025-01-14
Allow to overwrite sim_channel objects within a sim_station. Unify in…

fschlueter:
…terface for add_channel of sim_station and station but keep default behaviour at it currently is

### DIFF
--- a/NuRadioReco/framework/station.py
+++ b/NuRadioReco/framework/station.py
@@ -135,7 +135,7 @@ class Station(NuRadioReco.framework.base_station.BaseStation):
 
         return list(channel_ids)
 
-    def add_channel(self, channel, overwrite=False):
+    def add_channel(self, channel, overwrite=True):
         """
         Adds a channel to the station. If a channel with the same id is already present, it is overwritten.
 
@@ -143,7 +143,7 @@ class Station(NuRadioReco.framework.base_station.BaseStation):
         ----------
         channel : `NuRadioReco.framework.channel.Channel`
             The channel to add to the station.
-        overwrite : bool, (Default: False)
+        overwrite : bool, (Default: True)
             If True, allow to overwrite an existing channel (i.e., a channel with the same id). 
             If False, raise AttributeError if a channel with the same id is being added.
         """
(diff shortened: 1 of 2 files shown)


NuRadioReco/framework/station.py
```diff
@@ -135,7 +135,24 @@ def get_channel_ids(self, return_group_ids=False):
 
         return list(channel_ids)
 
-    def add_channel(self, channel):
+    def add_channel(self, channel, overwrite=True):
```
anelles, line 138: I am not sure I like overwrite = True as default ... that seems dangerous.

NuRadioReco/framework/station.py
```diff
@@ -135,7 +135,24 @@ def get_channel_ids(self, return_group_ids=False):
...
+        """
+        Adds a channel to the station. If a channel with the same id is already present, it is overwritten.
+
+        Parameters
+        ----------
+        channel : `NuRadioReco.framework.channel.Channel`
+            The channel to add to the station.
+        overwrite : bool, (Default: True)
+            If True, allow to overwrite an existing channel (i.e., a channel with the same id). If False, raise AttributeError
```
anelles, line 147: Want to add a linebreak -- while we are at it?

anelles (commented):
Overwriting functionality seems good, but I would prefer to default to 'False"

NuRadioReco/framework/station.py
fschlueter: Well I thought I keep the original behaviour as default behaviour to not break anything...

fschlueter: Okay lets see what happens. I will divert all blame towards you :D

fschlueter:
@anelles I changed the default again to allow overwriting of `Station.Channel`s to maintain the original behaviour. I would suggest to merge it like it is and discuss the change of the default in a new PR/issue.

anelles:
Interesting that we had no safety checks before overwriting .... as I said, seems dangerous. 
But yeah, start the new PR and then go ahead with this one.

PR #789, opened 2025-01-16
Implementing a new importing scheme

fschlueter:
This is a guilty pleasure thing. I dislike the current scheme of importing modules and framework classes and they require to much code. 

This PR adds imports into the `__init__.py` files of the packages `NuRadioReco.modules` and `NuRadioReco.framework`. Here are two snippets: 
```
from .base_shower import BaseShower
from .base_station import BaseStation
from .base_trace import BaseTrace
...
```
and 
```
from .analogToDigitalConverter import analogToDigitalConverter as AnalogToDigitalConverter
from .beamFormingDirectionFitter import beamFormingDirectionFitter as BeamFormingDirectionFitter
```
Notice that in the latter the name of the class objects is changed to have a capital letter as first letter. This does two things: It allows backwards compatibility (old import statements are still working) and it implements a typical syntax where classes use capital letters (similar as we do in the framework). However, it might introduce some confusion that the same classes are now accessible with two names (one with and one without the capital letter). 

This new import scheme allows the following code (pseudocode):
```
from NuRadioReco import framework as fwk, modules

event = fwk.Event(0, 0)
station = fwk.Station(0, 0)

channelNoiseAdder = modules.ChannelNoiseAdder()

....
```

One major disadvantage is, that now all modules are always loaded which: I) might take more time (although maybe only the first time, afterwards with caching it should be relatively fast) II) makes it a bit more cumbersome to implement modules which use optional dependencies.


fschlueter:
@cg-laser any thoughts? Do you like this idea or not so much? It does hide a little bit what things are - for me that is okay and the benefits in code writing out way this and over disadvantages (maybe there are more I am not aware of yet). @sjoerd-bouma do you have any thoughts as well?

anelles:
I really looks to me that we have bigger fish to fry than working on this ... 
My problem with this is indeed that it hides and makes it less transparent to new users.

cg-laser:
I see the appeal of these changes. It cleans up the code (a bit) and avoids lengthy lines. The main disadvantage is that it requires "magic" __init__ functions that provide somewhat magic functionality. It gets less obvious where a certain module is implemented. 
And more important. Now we need the extra step of registering all modules in the init function and as you already said, all modules are always loaded. So far it was no problem to just add more and more modules (in subdirectories). And if a certain user or collaboration needed another dependency, no problem to the other users. This changes now. 
The other thing is that user modules are difficult to integrate. Or one still needs to use the "old" syntax to load in a custom module that is not part of the NuRadioReco repository.

fschlueter:
> As I found with the module checker from #813, it is non-trivial to import all modules. At least one is broken on import (due to internal API change), and several require more complex dependencies (e.g. AraRoot or snowshovel). As far as I can tell, snowshovel is not even publicly available so there is no good solution there.
> 
> Of course, you can add try around the imports and just ignore modules that don't work or have unmet dependencies, but you have to do that for each one, which is annoying. Another alternative is to use the import replacement from fuckit.py https://github.com/ajalt/fuckitpy?tab=readme-ov-file#as-a-replacement-for-import, though introducing a dependency on fuckit.py may encourage usage elsewhere...

This is what I have been doing.

fschlueter:
I agree with all what was said. However, I also think this could actually newcomers because it simplifies the import process. Anyway, @sjoerd-bouma when we talked a couple of weeks ago you said you might want to come up with a different solution?

sjoerd-bouma:
As a comment on the technical implementation - I would envision modules like `event.py`, with an accompanying folder `_event` in which the code could be stored as one class per `.py` file as before. `event.py` would then look something like
```
from ._event.channel import Channel
from ._event.station import Station
(...)

__all__ = ['Channel', 'Station', ...]
```

This would result in the structure suggested above; the usage of `__all__` to list public classes/functions etc. I believe is relatively standard and used also by e.g. Python and numpy. As long as we do not directly clobber previously used namespaces, it should be straightforward to maintain backwards compatibility initially by adding stubs at the current class/function locations (and probably raising a deprecation warning if the old strucutre is used).

The advantage, as I see it, would be a shorter, easier to navigate module structure, especially in interactive development environments and the online documentation (I would envision each top-level module gets a dedicated description, instead of the current status where modules are generally undocumented, and one has to look at the classes or the 'manuals' section of the documentation to figure out what's where). The disadvantages are change, and possibly a slightly more opaque module structure for developers, though I would hope that most active developers are capable of following import statements.

I would estimate that implementing this, if agreed upon, and once we have a module structure everyone is happy with, would take @fschlueter or myself 1-2 days.

MijnheerD:
I just came across this PR, so my apologies if I am coming in after the facts, but I wanted to add something else that could be accomplished with this change. This is related to a discussion we had before New Year's, about the suggestion of adding type hints to the entire code base. I am not sure if this was ever followed up on, but I had a look and I found that my editor at least can infer type hints from docstrings as well. The real problem is that the current import scheme does not seem to play along with this. Let me explain with an example:

```python
from NuRadioReco.framework.event import Event
import NuRadioReco.detector.detector


def foo(evt):
    """
    Parameters
    ----------
    evt: Event
    """
    return None

def bar(det):
    """
    Parameters
    ----------
    det: NuRadioReco.detector.detector.Detector
    """
    return None
```
Here, when using the function `foo()`, the argument type will be correctly inferred. So when I would try to pass it a variable which is not of type `Event`, my editor would warn me. However, this is not the case with the function `bar()`. I am not sure exactly what goes wrong, but the type `NuRadioReco.detector.detector.Detector` is not resolved correctly (I am guessing it has something to do with it only being evaluated at runtime). Things would be much easier if type hinting did work, because then when scripting the editor would know that `Event.get_station(0)` returns a `Station` object and can auto-suggest functions. 

There are two reasons why I suggest it here. First one is that am wondering if using the `__all__` construction like @sjoerd-bouma suggested could help with this. From my understanding, it defines the symbols to be exported from the module, so maybe it can help resolve these imports. Secondly, the current naming scheme of things makes it difficult to simply adjust scripts to use the `from ... import` scheme, which would be the solution. Constructions like
```python
readCoREASDetector = NuRadioReco.modules.io.coreas.readCoREASDetector.readCoREASDetector()
```
are very common, especially in the example scripts. Because we tend to name the objects the same as the classes, a `from ... import` is not possible (because then the variable would overwrite the imported namespace). If we would go ahead with changing the import scheme, I think it could be worthwhile to try and solve this at the same time.

sjoerd-bouma:
> I don't think this new structure is a good idea, but it is difficult to have this discussion in writing. I suggest to have a phone call to discuss this in more detail and to go through the pro and cons. I can also explain better the reasoning behind the current structure.

I would be happy to have such a discussion! Maybe we can/should use a dedicated NuRadioMC call for this (where we announce this topic in advance, so those who are not interested do not need to show up?)

sjoerd-bouma:
> I just came across this PR, so my apologies if I am coming in after the facts, but I wanted to add something else that could be accomplished with this change. This is related to a discussion we had before New Year's, about the suggestion of adding type hints to the entire code base. I am not sure if this was ever followed up on, but I had a look and I found that my editor at least can infer type hints from docstrings as well. The real problem is that the current import scheme does not seem to play along with this. Let me explain with an example:
> 
> ```python
> from NuRadioReco.framework.event import Event
> import NuRadioReco.detector.detector
> 
> 
> def foo(evt):
>     """
>     Parameters
>     ----------
>     evt: Event
>     """
>     return None
> 
> def bar(det):
>     """
>     Parameters
>     ----------
>     det: NuRadioReco.detector.detector.Detector
>     """
>     return None
> ```
> 
> Here, when using the function `foo()`, the argument type will be correctly inferred. So when I would try to pass it a variable which is not of type `Event`, my editor would warn me. However, this is not the case with the function `bar()`. I am not sure exactly what goes wrong, but the type `NuRadioReco.detector.detector.Detector` is not resolved correctly (I am guessing it has something to do with it only being evaluated at runtime). Things would be much easier if type hinting did work, because then when scripting the editor would know that `Event.get_station(0)` returns a `Station` object and can auto-suggest functions.
> 
> There are two reasons why I suggest it here. First one is that am wondering if using the `__all__` construction like @sjoerd-bouma suggested could help with this. From my understanding, it defines the symbols to be exported from the module, so maybe it can help resolve these imports. Secondly, the current naming scheme of things makes it difficult to simply adjust scripts to use the `from ... import` scheme, which would be the solution. Constructions like
> 
> ```python
> readCoREASDetector = NuRadioReco.modules.io.coreas.readCoREASDetector.readCoREASDetector()
> ```
> 
> are very common, especially in the example scripts. Because we tend to name the objects the same as the classes, a `from ... import` is not possible (because then the variable would overwrite the imported namespace). If we would go ahead with changing the import scheme, I think it could be worthwhile to try and solve this at the same time.

You've identified two separate issues here :upside_down_face: `NuRadioReco.detector.detector.Detector` is not understood, because it's incorrect - `Detector` is a (factory) function, not a class, this should instead be the base class that all detectors inherit from (which has also become problematic, because the RNO-G detector does **not** inherit from DetectorBase... as I said, separate discussion).

The second issue is due to NuRadioMC ignoring about [3 or 4 PEP8 conventions](https://peps.python.org/pep-0008/#prescriptive-naming-conventions) in its module and class naming schemes: classes should be capitalized, class instances should be lowercase. This is what causes the second problem you outline, and **if** we restructure NuRadioMC I would strongly suggest we fix this by following PEP8.

MijnheerD:
> You've identified two separate issues here 🙃 `NuRadioReco.detector.detector.Detector` is not understood, because it's incorrect - `Detector` is a (factory) function, not a class, this should instead be the base class that all detectors inherit from (which has also become problematic, because the RNO-G detector does **not** inherit from DetectorBase... as I said, separate discussion).

The problem still occurs when replacing  `NuRadioReco.detector.detector.Detector`  by a real class like `NuRadioReco.framework.station.Station` (maybe I am doing something wrong though). So I believe there is more at play than just the understanding of the object, I really think the imports are resolved differently.

MijnheerD:
> > You've identified two separate issues here 🙃 `NuRadioReco.detector.detector.Detector` is not understood, because it's incorrect - `Detector` is a (factory) function, not a class, this should instead be the base class that all detectors inherit from (which has also become problematic, because the RNO-G detector does **not** inherit from DetectorBase... as I said, separate discussion).
> 
> The problem still occurs when replacing `NuRadioReco.detector.detector.Detector` by a real class like `NuRadioReco.framework.station.Station` (maybe I am doing something wrong though). So I believe there is more at play than just the understanding of the object, I really think the imports are resolved differently.

I did a bit more testing and it looks like the problem I encountered is that types in docstrings are not evaluated for type hints, in the sense that the path is not followed. So this does not work:
```python
import NuRadioReco.framework.station

def bar(var):
    """
    Parameters
    ----------
    var: NuRadioReco.framework.station.Station
    """
    return None

bar(0)  # no warning that type is wrong
```
However, what does work is adding an explicit type hint in the function definition:
```python
import NuRadioReco.framework.station

def bar(var: NuRadioReco.framework.station.Station):
    """
    Parameters
    ----------
    var: NuRadioReco.framework.station.Station
    """
    return None

bar(0)  # now my IDE lets me know that type is wrong
```
So it looks like going to a `from ... import` scheme would still be the easiest way to go. FYI, I am using Pycharm which has its own type checker tool, it might be different when using VSCode.

PR #790, opened 2025-01-16
Adding a script for the standard data processing of RNO-G data

fschlueter:
This PR implements standard procedures to read and process RNO-G data. I includes the following things:

- A new module `dataProviderRNOG` which wraps around the most essential modules including the data reader module which every one should use.
- a module `processing.py` which adds a function `process_event` which adds a couple of standard modules to the processing sequence you typically want to use.
- a example script `data_analysis_example.py` which uses the `dataProviderRNOG` and `process_event`.
- Two new RNO-G related modules the `channelGlitchDetector` and the `statuionDeepCrVariable.py`.
- Many more minor changes.


NuRadioReco/framework/event.py
```diff
@@ -562,14 +565,57 @@ def get_hybrid_information(self):
...
+        for station in self.get_stations():
+            for channel in station.iter_channels():
+                if (channel.has_parameter(chp.RPR)):
+                    RPR_ch = channel.get_parameter(chp.RPR)
+                    if (RPR_ch == np.inf):
+                        return np.inf
+                    else:
+                        RPR_all.append(RPR_ch)
+
+        return np.mean(RPR_all)
```
fschlueter, line 609: @avijai2000, great idea to add those functions! I have two small comments.
- can you change the name of the variables `SNR_all` and `RPR_all` to use small letters? In the pep8 python style (which we try to mostly follow) variables typically do not start with capital letters. 
- can you implement the feature to specify channel_ids similar to calculate the mean? This is, I think, very useful to e.g. calculate the mean over the PA.

And one additional thought, maybe we should implement those functions on station level and than add a wrapper for those at event level with a check that the event only has one station?

avijai2000: Sure yeah that sounds good, I'll make these changes

NuRadioReco/framework/station.py
```diff
@@ -261,6 +261,36 @@ def get_magnetic_field_vector(self, time=None):
...
+        Returns the average SNR across all channels
+        """
+        snr_all = []
+        for channel in station.iter_channels():
+            if (channel.has_parameter(chp.SNR)):
+                snr_ch = channel.get_parameter(chp.SNR)
+                if (snr_ch == np.inf):
+                    return np.inf
+                else:
+                    snr_all.append(snr_ch)
```
fschlueter, line 275: @avijai2000, I think currently `channels_to_include` is not used. You can pass it to `station.iter_channels(channels_to_include)`. Can you also rename the function to something like `get_average_signal_to_noise_ratio` (at least I am a fan or being quite explicit). Furthermore, I would probably raise and error (or warning) if the channel parameter are missing). Also I think the enums are not wrong (at least in case of the RPR). All comments also apply to the RPR function

sjoerd-bouma:
I'm not sure if this is already ready for review; I just noticed you are adding a number of new methods to the `Event` class. I'm not set in stone against this but typically properties like SNR, glitches etc. are obtained by dedicated modules and stored in `parameters`. They can then be obtained in subsequent analysis by `get_parameter` or dictionary access (`station[parameter]`) where needed.

What I am worried about is two things: splitting information across multiple places, and cluttering the basic `Event` class with many methods that may or may not be relevant to different users. But if you or @fschlueter feel that there is a good reason to implement things this way I am willing to listen to arguments.

NuRadioReco/framework/station.py
avijai2000: Just made the changes! Let me know what you think

fschlueter:
> I'm not sure if this is already ready for review; I just noticed you are adding a number of new methods to the `Event` class. I'm not set in stone against this but typically properties like SNR, glitches etc. are obtained by dedicated modules and stored in `parameters`. They can then be obtained in subsequent analysis by `get_parameter` or dictionary access (`station[parameter]`) where needed.
> 
> What I am worried about is two things: splitting information across multiple places, and cluttering the basic `Event` class with many methods that may or may not be relevant to different users. But if you or @fschlueter feel that there is a good reason to implement things this way I am willing to listen to arguments.

@sjoerd-bouma, I understand your concern! Indeed I think we should make sure that we are not duplicating data in the various data-structure "levels". What we have started implementing are getter functions to simplify the access to certain data. Smart getter functions like `get_average_snr` which allow you to on-the-fly calculate the avg SNR for, e.g., the PA channels, also add useful functionality and allow us easy access to such data without the need to extra store them. 

One problem I see with the current functions is that they access channel information for an event while "by-passing" the station level. One solution could be to implement those functions on the station-level and implement wrapper for those on the event level. Regarding cluttering the basic `event` class. Yes I am also not super happy with this. Do you think it is time to implement a subclass of `Event` a `RNOGEvent` for example?

NuRadioReco/framework/station.py
```diff
@@ -261,33 +261,37 @@ def get_magnetic_field_vector(self, time=None):
             logger.error(f"Reference reconstruction not set / unknown: {self.__reference_reconstruction}")
             raise ValueError(f"Reference reconstruction not set / unknown: {self.__reference_reconstruction}")
 
-    def avg_SNR(self, channels_to_include = [0, 1, 2, 3, 5, 6, 7, 22, 23]):
+    def get_average_signal_to_noise_ratio(self, channels_to_include = [0, 1, 2, 3, 5, 6, 7, 22, 23]):
```
fschlueter, line 264: Can you set the default value for `channels_to_include` to `None`? This class should be a rather generic class and the channels id only make sense for RNO-G.

NuRadioReco/framework/station.py
```diff
@@ -261,33 +261,37 @@ def get_magnetic_field_vector(self, time=None):
...
         """
         Returns the average SNR across all channels
         """
         snr_all = []
-        for channel in station.iter_channels():
+        for channel in station.iter_channels(channels_to_include):
             if (channel.has_parameter(chp.SNR)):
```
fschlueter, line 270: can you remove the parentheses from the if condition?

NuRadioReco/framework/station.py
```diff
@@ -261,33 +261,37 @@ def get_magnetic_field_vector(self, time=None):
...
                 snr_ch = channel.get_parameter(chp.SNR)
                 if (snr_ch == np.inf):
                     return np.inf
                 else:
                     snr_all.append(snr_ch)
+            else:
+                raise LookupError(f"Channel {channel.get_id()} has no parameter SNR")
```
fschlueter, line 277: ValueError or KeyError?

cg-laser:
I second Sjoerd that I'm afraid that event and station classes get too cluttered with analysis/RNO-G specific functions. On the other hand, we want to offer this flexibility.

Wouldn't the better way to implement utility functions such as 
`get_average_SNR(station, include_channels=[xxx])`
that we provide in `NuRadioReco/utlities/RNOG_utilities.py` 
Or a more general name, depending on how general the functions are. 

So instead of writing `station.get_average_SNR` one writes `get_average_SNR(station)`

cg-laser:
Hi everyone, I provided an example data analysis script with extensive documentation.

I have one technical question. Why is the hardwareResponeIncorporator configured to only remove the phases but not correct for amplitudes? I suggest to do the full correction. A bandpass filter is applied before that module so that we won't have any "division by zero" problems.

NuRadioReco/framework/parameters.py
```diff
@@ -63,6 +63,26 @@ class channelParameters(Enum):
...
+    max_corr_coords = 4 #: azimuth, zenith corresponding to max correlation
+    max_corr = 5 #: maximum correlation value obtained through reconstruction
+    csw_snr = 6 #: signal to noise ratio of coherently summed waveform
+    csw_rpr = 7 #: root power ratio of coherently summed waveform
+    csw_hilbert_snr = 8 #: signal to noise ratio of hilberted coherently summed waveform
+    csw_impulsivity = 9 #: coherently summed waveform's impulsivity and other statistical measures (dict)
+    surf_corr_ratio = 10 #: ratio comparing max surface correlation to max overall correlation
+    max_surf_corr = 11 #: maximum surface correlation
+    avg_snr = 12 #: average signal to noise ratio across all channels ("peak to peak")
+    avg_rpr = 13 #: average root power ratio across all channels
```
fschlueter, line 85: With the new getter function we do not need these right?

NuRadioReco/framework/parameters.py
```diff
@@ -63,6 +63,26 @@ class channelParameters(Enum):
...
+
+class eventParametersRNOG(Enum):
+    max_corr_coords = 4 #: azimuth, zenith corresponding to max correlation
+    max_corr = 5 #: maximum correlation value obtained through reconstruction
+    csw_snr = 6 #: signal to noise ratio of coherently summed waveform
+    csw_rpr = 7 #: root power ratio of coherently summed waveform
+    csw_hilbert_snr = 8 #: signal to noise ratio of hilberted coherently summed waveform
+    csw_impulsivity = 9 #: coherently summed waveform's impulsivity and other statistical measures (dict)
+    surf_corr_ratio = 10 #: ratio comparing max surface correlation to max overall correlation
+    max_surf_corr = 11 #: maximum surface correlation
```
fschlueter, line 83: Important parameters! I was just wondering if we should move them to `stationParametersRNOG` and maybe rename them to have less acronyms?

NuRadioReco/framework/event.py
```diff
@@ -562,14 +565,53 @@ def get_hybrid_information(self):
...
+    def has_glitch(self):
+        """
+        Returns true if any channel in any station has a glitch
+        """
+        for station in self.get_stations():
+            for channel in station.iter_channels():
+                if channel.has_parameter(chp.glitch) and channel.get_parameter(chp.glitch):
+                    return True
+
+        return False
```
fschlueter, line 577: I think we should implement this function also in utilities. Maybe in an RNO-G specific file, i.e., `rnog_station_utilities.py`?

NuRadioReco/framework/event.py
```diff
@@ -562,14 +565,53 @@ def get_hybrid_information(self):
...
+        """
+        if (len(list(self.get_stations())) == 1):
+            for station in self.get_stations():
+                return station.get_average_root_power_ratio()
+        else:
+            avg_rprs = {}
+            for station in self.get_stations():
+                avg_rprs[station.get_id()] = station.get_average_root_power_ratio()
+                
+        return avg_rprs
```
fschlueter, line 605: those can be removed now?

NuRadioReco/examples/RNOG/rnog_config.yaml
```diff
@@ -0,0 +1,61 @@
+# Config file for the standard RNO-G data processing
```
fschlueter, line 1: I guess this file is not needed anymore and can be removed?

fschlueter:
@all I decided to remove the `channelBeamFormingDirectionFitter` module from this branch and opened a new one for it (#833). I think we should merge this PR asap and continue further development on new branches/PR.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -427,7 +427,8 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
         The mean rms of the lowest few segment rms values
     """
     split_array = np.array_split(trace, segments)
-    rms_of_splits = np.std(split_array, axis=1)
+    split_array = np.array(split_array, dtype="object") #Objectify dtype to allow timetraces indivisible by amount of segments 
+    rms_of_splits = [np.std(split) for split in split_array]
```
fschlueter, line 431: This is an elegant fix. But I was wondering if it is not better to raise an error and in cases where the traces are not dividable by 4 we have to change the number?

cg-laser:
I improved the signal reconstructor. The window size should have a physical unit (i.e. ns) and not bins which would implicitly depend on the sampling rate. 

I also added another "advanced" example where all modules are defined in file, as a basis for reco pipelines that deviate from the default processing. 

In this file, I also simplified the cmd arguments. I suggest to to the same in the other script if no one disagrees. The user always needs to specify an output file. 

Shouldn't we remove the run_reconstruction.py example? As a minimum it should be rewritten to use the standard processing.

NuRadioReco/utilities/trace_utilities.py
JethroStoffels: Having some sort of print/log for the indivisibility does seem like a good idea indeed
I'd be more in favor of raising a warning then as I'd like to leave the possibility of splitting the timetrace in as many segments as one needs. In case of enough samples per event trace, having some segments have one more sample shouldn't make a big difference anyhow I believe

NuRadioReco/modules/channelSignalReconstructor.py
```diff
@@ -151,7 +151,8 @@ def get_SNR_and_RPR(self, station_id, channel, det, stored_noise = False, rms_st
             snr['peak_amplitude'] = np.max(np.abs(trace[signal_window_mask])) / noise_rms
 
         #Calculate peak to peak voltage SNR using the RMS of the split waveform
-        snr['peak_2_peak_amplitude_split_noise_rms'] = np.amax(trace_utilities.maximum_peak_to_peak_amplitude(channel.get_trace(), self.__coincidence_window_size)) 
+        coincidence_window_size_bins = int(round(self.__coincidence_window_size * channel.get_sampling_rate()))
+        snr['peak_2_peak_amplitude_split_noise_rms'] = np.amax(trace_utilities.maximum_peak_to_peak_amplitude(channel.get_trace(), coincidence_window_size_bins))
```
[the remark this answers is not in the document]
JethroStoffels: This change to bins would for sure change the resulting values if not accounted for (from what I see there's mainly an additional factor of sampling rate). I'd have to take a look at what values make sense to use for the coincidence window and then adjust accordingly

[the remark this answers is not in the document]
cg-laser: I think this should make it much easier to properly define it. Now, you can specify the physical time of the coincidence, e.g. 6 nano seconds. 
The previous code was quite buggy, because, depending on the upsampling done in a previous module, the results would change.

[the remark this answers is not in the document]
fschlueter: Can we add a error if `coincidence_window_size_bins`
```
coincidence_window_size_bins = int(round(self.__coincidence_window_size * channel.get_sampling_rate()))
```
is 0 or 1 (both values do not really make sense right?). 
@JethroStoffels, since the RNOG-sampling rate is not 1 GHz, the 6 ns are not 6 bins. (where is the 6 coming from?). Can you change this in the rnog example scripts (not in the class).

fschlueter:
Okay, all changes discussed yesterday are implemented. Shall we first merge #849 into this branch before merging this into develop? Anyway someone else has to approve this PR.

fschlueter:
I just noticed that the `window_size` in `stationDeepCRVariables` is still defined in bins while the `window_size` in `channelSignalReconstructor` is now defined in ns. Shall we fix that? Someone volunteering?

martin01021207:
> Okay, all changes discussed yesterday are implemented. Shall we first merge #849 into this branch before merging this into develop? Anyway someone else has to approve this PR.

@fschlueter  PR #849 has been merged to this branch

martin01021207:
> I just noticed that the `window_size` in `stationDeepCRVariables` is still defined in bins while the `window_size` in `channelSignalReconstructor` is now defined in ns. Shall we fix that? Someone volunteering?

@fschlueter I just made some changes in `stationDeepCRVariables`, please take a look and see if that's okay. But my question remains to be that if we should rename this script/class before we merge, because it's all about CSW, and it's not that specific for just CR analysis. What do you think?

fschlueter:
> > I just noticed that the `window_size` in `stationDeepCRVariables` is still defined in bins while the `window_size` in `channelSignalReconstructor` is now defined in ns. Shall we fix that? Someone volunteering?
> 
> @fschlueter I just made some changes in `stationDeepCRVariables`, please take a look and see if that's okay. But my question remains to be that if we should rename this script/class before we merge, because it's all about CSW, and it's not that specific for just CR analysis. What do you think?

@martin01021207 yes I think that is a good idea. Please adapt the doc strings to reflect the new name and general purpose of the module. Can you also add a brief explanation what the CSW is?

martin01021207:
> > > I just noticed that the `window_size` in `stationDeepCRVariables` is still defined in bins while the `window_size` in `channelSignalReconstructor` is now defined in ns. Shall we fix that? Someone volunteering?
> > 
> > 
> > @fschlueter I just made some changes in `stationDeepCRVariables`, please take a look and see if that's okay. But my question remains to be that if we should rename this script/class before we merge, because it's all about CSW, and it's not that specific for just CR analysis. What do you think?
> 
> @martin01021207 yes I think that is a good idea. Please adapt the doc strings to reflect the new name and general purpose of the module. Can you also add a brief explanation what the CSW is?

@fschlueter Okay I just renamed the script and added some doc strings and descriptions, please take a look. If everything is good now this branch can also be merged to the main (develop), right?

fschlueter:
@martin01021207 excellent! I just slightly modified the doc-string. Feel free to approve this PR (I can not). After a period of 24h we can merge it.

martin01021207 (dismissed):

sjoerd-bouma (changes requested):
Thanks for all the work to those involved! I would like to request that we add docstrings to the `dataProviderRNOG` module before merging, though.

NuRadioReco/utilities/rnog_utilities.py
```diff
@@ -0,0 +1,19 @@
+from NuRadioReco.framework.parameters import channelParametersRNOG as chpRNOG
+from NuRadioReco.framework.event import Event
+
+
+def has_glitch(event_or_station):
+    """
+    Returns true if any channel in any station has a glitch
```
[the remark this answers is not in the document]
fschlueter: Can do. I was just thinking we could also add the function to the module script (but not to the class)?

fschlueter:
@sjoerd-bouma please have a look!

NuRadioReco/modules/RNO_G/dataProviderRNOG.py
```diff
@@ -21,6 +50,17 @@ def __init__(self):
 
 
     def begin(self, files, reader_kwargs={}, det=None):
```
sjoerd-bouma, line 52: if det is required, why not use `begin(self, files, det, **reader_kwargs)`? (or `reader_kwargs={}` as now)? Not that it makes a big difference...

NuRadioReco/utilities/rnog_utilities.py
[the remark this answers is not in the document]
sjoerd-bouma: That is, adding it to `dataProviderRNOG`? Alternatively, it might in any case be worth to put `rnog_utilities.py` under `modules/RNO_G` instead of under `utilities` to minimize clutter, what do you think?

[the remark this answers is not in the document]
sjoerd-bouma: Also, as `dataProviderRNOG` is an io module, I originally looked for it under `modules/io/RNO_G` instead of `modules/RNO_G`. But here I don't have a strong preference.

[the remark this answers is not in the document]
fschlueter: No I would suggest to put it into the channelGlitchDetector module. I do not like to mix utility modules under modules/ but LOFAR does that already right?

NuRadioReco/modules/RNO_G/dataProviderRNOG.py
fschlueter: done

NuRadioReco/utilities/rnog_utilities.py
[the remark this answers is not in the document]
sjoerd-bouma: For me 'utilities' is not so clearly demarcated from other module functions, and I would generally prioritize keeping experiment-specific code in as few locations as possible (currently 3, would be 4 if we introduce rnog_utilities under utilities). I'm happy to have `has_glitch` in the `channelGlitchDetector` module.

sjoerd-bouma (approved):
Looks good, thanks everyone!

martin01021207:
Shall we merge this PR?

fschlueter:
Can someone reapprove this PR (I dismissed the prev. approval because I solved a merge conflict which appeared in the meantime)

sjoerd-bouma (approved):

PR #791, opened 2025-01-17
Change default to disallow overwritting channel objects in Station

fschlueter:
See PR #787 for an explanation.


fschlueter:
@christophwelling, I had to modify your unit test because it would overwrite channel objects. I believe removing the first solves the issue but please have a look at tell me if with that the test still makes sense.

christophwelling (approved):
Makes sense to me. The efieldToVoltageConverter should replace the channels you removed, so you have the same result as if you just overwrote them.

PR #793, opened 2025-01-17
[bugfix] actually adjust trace_start_time in channelStopFilter

sjoerd-bouma:
While going through the definitions of times in #773, I noticed that the channelStopFilter doesn't actually adjust the `trace_start_time`. This is a very minor bug, as all traces are affected the same way, but it's probably worth doing it correctly.


fschlueter (approved):

fschlueter (changes requested):
I think you have to change the `trace_start_time` by `int(np.round(prepend * sampling_rate))) / sampling_rate`. And you can use add_trace_start_time(...)` instead

fschlueter:
> I think you have to change the `trace_start_time` by `int(np.round(prepend * sampling_rate))) / sampling_rate`. And you can use add_trace_start_time(...)` instead

or simply use `prepend` which should be already in units of time but not `prepend * sampling_rate`

fschlueter:
And I guess you have to update the reference values for the failed reconstruction?

fschlueter (approved):

PR #794, opened 2025-01-17
Adding a module for sine wave subtraction

alisanozdrina:
the sine sub module for CW filtering. currently failing for a few channels. needs a review to understand the issue.


NuRadioReco/modules/channelSineSubtraction.py
```diff
@@ -0,0 +1,249 @@
+
+import numpy as np
+import sys
+import logging
+# Add the directory where your library is located
+#sys.path.append('/Users/nozdrina.1/Documents/phys/software/NuRadioMC')
```
fschlueter, line 6: Can these lines be removed?

NuRadioReco/modules/channelSineSubtraction.py
```diff
@@ -0,0 +1,249 @@
...
+        Estimate the amplitude of a specific harmonic in the waveform.
+
+        Args:
+            wf (np.ndarray): Input waveform (1D array).
+            target_freq (float): Target frequency for which to estimate amplitude.
+            sampling_rate (float): Sampling rate of the waveform (GHz).
+
+        Returns:
+            float: Estimated amplitude of the target frequency.
+        """
```
fschlueter, line 29: Can we correct the indentation here? Aline the lines 20-29 with 19. Also in the rest of NuRadio we use this kind of syntax for function doc-strings:
```
 Paramters
 ----------
 wf: np.ndarray
     Input waveform (1D array).
target_freq:  float
     Target frequency for which to estimate amplitude.
sampling_rate: float
     Sampling rate of the waveform (GHz).

Returns
--------
ampl: float
     Estimated amplitude of the target frequency.
```
Do you think you could change this?

NuRadioReco/modules/channelSineSubtraction.py
```diff
@@ -0,0 +1,249 @@
...
+    Perform sine subtraction on a waveform to remove CW noise.
+
+    Args:
+        wf (np.ndarray): Input waveform (1D array).
+        sampling_rate (float): Sampling rate of the waveform (Hz).
+        peak_prominence (float): Threshold for identifying prominent peaks in the FFT spectrum.
+        saved_noise_freqs (list, optional): A list to store identified noise frequencies.
+
+    Returns:
+        np.ndarray: Corrected waveform with CW noise removed.
```
fschlueter, line 59: Here as well

NuRadioReco/modules/channelSineSubtraction.py
```diff
@@ -0,0 +1,249 @@
...
+    ----------
+    channel : NuRadio channel class
+        channel from which to get trace
+    ax : matplotlib.axes
+        ax on which to plot
+    label : string
+        plotlabel
+    plot_kwargs : dict
+        options for plotting
+    """
```
fschlueter, line 142: Like this :)

NuRadioReco/modules/channelSineSubtraction.py
```diff
@@ -0,0 +1,249 @@
...
+
+    bin_index = np.argmin(np.abs(frequencies - target_freq))
+    amplitude = np.abs(fft_spectrum[bin_index]) / len(wf)  # Normalize
+    if bin_index > 0:
+        amplitude *= 2  # Adjust for double-sided spectrum
+    #print(f"Amplitude of {target_freq} Hz harmonic: {amplitude}")
+    return amplitude
+
+
+def sine_sub(wf: np.ndarray, sampling_rate: float = 3.2, peak_prominance: float = 6.0, saved_noise_freqs : list = None):
```
fschlueter, line 48: can we change the name of the function to be a bit more self explanatory: `sinewave_subraction` for example

NuRadioReco/modules/channelSineSubtraction.py
```diff
@@ -0,0 +1,249 @@
...
+    Returns:
+        np.ndarray: Corrected waveform with CW noise removed.
+    """
+    dt = 1 / sampling_rate # in ns
+    t = np.arange(0, len(wf) * dt, dt) # in ns
+    # zero meaning, just in case
+    wf = wf - np.mean(wf)
+
+    def sinusoid(t, amplitude, noise_frequency, phase):
+        return amplitude * np.cos(2 * np.pi * noise_frequency * t + phase)
```
fschlueter, line 67: I mean it does not really matter but why are you using a cos here instead of a sin?

NuRadioReco/modules/channelSineSubtraction.py
```diff
@@ -0,0 +1,249 @@
...
+                group.append(peak_idxs[i])
+            else:
+                # Calculate the mean frequency for the current group of neighbors
+                noise_freqs.append(np.mean(freqs[group]))
+                # Start a new group with the current peak
+                group = [peak_idxs[i]]
+
+        # Don't forget to append the last group
+        if group:
+            noise_freqs.append(np.mean(freqs[group]))
```
fschlueter, line 96: I am actually not sure if that is a good idea. Why have you implemented it like this? Maybe becuase the ARA tools does it? Can we ask someone from ARA why it is done this way? I would rather (by default at least) remove the sine wave bin by bin.

NuRadioReco/modules/channelSineSubtraction.py
```diff
@@ -0,0 +1,249 @@
...
+
+            # Fit the sinusoidal model to the waveform
+            try:
+                params, _ = curve_fit(sinusoid, t, wf, p0=initial_guess)
+                estimated_amplitude, estimated_freq, estimated_phase = params
+
+                # Generate the estimated CW noise
+                estimated_cw_noise = estimated_amplitude * np.cos(
+                    2 * np.pi * estimated_freq * t + estimated_phase
+                )
```
fschlueter, line 114: here you could use the function defined above!

NuRadioReco/modules/channelSineSubtraction.py
```diff
@@ -0,0 +1,249 @@
...
+            float: Estimated amplitude of the target frequency.
+        """
+    if wf.size == 0:
+        raise ValueError("Input waveform is empty.")
+    if target_freq < 0 or target_freq > sampling_rate / 2:
+        raise ValueError("Target frequency is out of range (0 to Nyquist frequency).")
+
+    fft_spectrum = np.fft.fft(wf)
+    frequencies = np.fft.fftfreq(len(wf), d=1 / sampling_rate)
+
```
fschlueter, line 37: We should use the functions defined in `NuRadioReco.utilities.fft` as also done elsewhere in this module already. If we do that we also do not have to fix the amplitude below (i.e., the normalisation should be fine with no extra effort).

NuRadioReco/modules/channelSineSubtraction.py
```diff
@@ -0,0 +1,249 @@
...
+                                 peak_prominance=self.peak_prominance, saved_noise_freqs=self.save_filtred_freqs)
+            channel.set_trace(trace_fil, fs)
+
+
+if __name__ == "__main__":
+    import argparse
+    import os
+    from NuRadioReco.modules.io.RNO_G.readRNOGDataMattak import readRNOGData
+    import requests
+    import aiohttp
```
fschlueter, line 207: Remove those dependencies.

NuRadioReco/modules/channelSineSubtraction.py
alisanozdrina: It was initially bin by bin, but it didn't perform well while testing this on a big dataset.  c++ sinesub does it bin by bin and some CW lines remains untouched. I changed it to the mean value of the peak and it worked without fail on one of the station CR burn sample.

NuRadioReco/modules/channelSineSubtraction.py
alisanozdrina: i just don't like the fact that sine function has a zero value at x=0. It probably doesn't matter, but in my experience with fitting functions (any type of minimization problem) is that sometimes they get stuck at zero. If you feel strongly about it, I can change it and test.

fschlueter: Okay, I actually thought of this and I understand your reasoning but you could also accomplish that by starting the phase at `pi/2`.

However, thinking of the fitting procedure I have a couple of other questions / comments: Would it make sense to first only fit the phase and keep the amplitude and frequency fixed (after all we have strong priors for those). And in a second fit I would fit all three parameter together or only the amplitude and phase and fix the frequency the entire time? Do you think this is worth trying?

NuRadioReco/modules/channelSineSubtraction.py
fschlueter: Exactly, and I would imagine that this frequency is typically contained with one bin and not split across bins. In any case I would replace the mean by a weighted mean using the amplitudes in each bin (but I Imagine that the one bin is always dominating the amplitude spectrum?

PR #795, opened 2025-01-20
Clean up of station. Do not return arianna magnetic field if ref…

fschlueter:
…erence reconstruction is RD

This PR mostly does cosmetic changes to `station.py` (wrong in commit message). The only functional change is in the function `get_magnetic_field_vector`. Now an error is raised if the reference reconstruction is set to "RD" and not the arianna magnetic field is silently returned.

### DIFF
--- a/NuRadioReco/framework/station.py
+++ b/NuRadioReco/framework/station.py
@@ -30,7 +30,7 @@ class Station(NuRadioReco.framework.base_station.BaseStation):
 
     def add_sim_station(self, sim_station):
         """
-        Adds a SimStation to the Station. If a SimStation is already present, 
+        Adds a SimStation to the Station. If a SimStation is already present,
         the new SimStation is merged to the existing one.
 
         Parameters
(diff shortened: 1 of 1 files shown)


fschlueter:
@sjoerd-bouma I hope this could be a quick review :)

NuRadioReco/framework/station.py
```diff
@@ -188,51 +188,62 @@ def get_reference_reconstruction(self):
...
+            logger.error(
+                "Magnetic field for `self.__reference_reconstruction == 'RD'` not implemented yet. "
+                "Please use `radiotools.helpers.get_magnetic_field_vector(site)` for the site you are interested in.")
+            raise NotImplementedError(
+                "Magnetic field for `self.__reference_reconstruction == 'RD'` not implemented yet. "
+                "Please use `radiotools.helpers.get_magnetic_field_vector(site)` for the site you are interested in."
+            )
+        else:
+            logger.error(f"Reference reconstruction not set / unknown: {self.__reference_reconstruction}")
+            raise ValueError(f"Reference reconstruction not set / unknown: {self.__reference_reconstruction}")
```
sjoerd-bouma, line 214: Technically breaks backwards compatibility but I agree this should be changed. Only I think the module is called `radiotools.helper` and not `helpers`?

sjoerd-bouma (commented):
One typo, other than that looks good. We should probably deprecate all the `get_reference_reconstruction` etc. methods though (different PR, I think).

sjoerd-bouma (approved):

PR #796, opened 2025-01-20
Fix trigger time and readout delays set by the readRNOGDataMattak module

fschlueter:
to the NuRadio convention: The trigger time is not not absolute anymore but relative to the station/event time. It also now read outs and applies the readoutDelays which are stored along side the data.

This PR corrects for the convention correctly documented with PR #773.

What is yet missing is a verification whether the "intrinsic" readout delays are correctly represented with function `get_time_offset`. Furthermore, we have to fix the `Event` class to not store Astropy objects in nur files.

### DIFF
--- a/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
+++ b/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
@@ -775,7 +775,6 @@ class readRNOGData:
         evt = NuRadioReco.framework.event.Event(event_info.run, event_info.eventNumber)
         station = NuRadioReco.framework.station.Station(event_info.station)
         station.set_station_time(astropy.time.Time(trigger_time, format='unix'))
-        evt.set_event_time(astropy.time.Time(trigger_time, format='unix'))  # Typically event time = station time
 
         trigger = NuRadioReco.framework.trigger.Trigger(event_info.triggerType)
         trigger.set_triggered()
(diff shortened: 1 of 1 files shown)


NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
```diff
@@ -775,16 +775,18 @@ def _get_event(self, event_info, waveforms):
         evt = NuRadioReco.framework.event.Event(event_info.run, event_info.eventNumber)
         station = NuRadioReco.framework.station.Station(event_info.station)
         station.set_station_time(astropy.time.Time(trigger_time, format='unix'))
+        evt.set_event_time(astropy.time.Time(trigger_time, format='unix'))  # Typically event time = station time
```
sjoerd-bouma, line 778: Converting to an astropy.time.Time object isn't quite ´free'; it's fast enough that it doesn't matter here but seeing as you sometimes like superfluous optimizations, it could be made quicker here by creating it only once.

sjoerd-bouma (dismissed):
Thanks Felix! We should get this merged soon. Just one comment but up to you if you want to address it or not.

NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py
fschlueter: I removed that line. I think we agree that for data the event time does not need to be set.

sjoerd-bouma: That's also fine. What I thought was actually `evt.set_event_time(station.get_station_time())` (or an equivalent) so you don't call the 'expensive' `astropy.time.Time` initialization twice. But I have no objections to not setting the event time in data.

sjoerd-bouma (approved):

PR #797, opened 2025-01-20
Move the module noise from NuRadioReco/utilities to NuRadioMC/examples

fschlueter:
I had a hard time finding this code even know I existed. It is completely unfindable if you do not know that it exists.


sjoerd-bouma:
How is this more findable in examples (intended for scripts, excluded from the documentation) than in utilities? :confused:

fschlueter:
> How is this more findable in examples (intended for scripts, excluded from the documentation) than in utilities? 😕

Well when I want to simulate noise I go into the example folder of NuRadio**MC** and not in a utilities folder in NuRadioReco. For me a utilities folder never contains standalone scripts but only "tool" functions. Maybe this is just me but I looked several times for it and could not find it. I only found it now by searching on slack because I remembered that Christian posted it one (that is not a joke!).

anelles:
Fair enough on the not finding, but I also tend to disagree with this  moving into an example. I would have moved it to NuRadioMC/utilities

fschlueter:
I talked to sjoerd earlier. My proposal now is to split the file, add one script which can be used to simulate noise to `NuRadioMC/simulation/noise.py` and keep the utilities module with some functions (could also be in NuRadioMC). I will bing you when I have implemented this.

Issue #524, opened 2023-03-31
Defining trigger time in NuRadioReco 


fschlueter:
Should it be defined absolut or relative (to station time). In what format do we want to store it (astropy, float tuple, ...). Do we need ns precision?

sjoerd-bouma:
Some more comments:
- we currently store astropy.time.Time objects by calling `.value` (see https://github.com/nu-radio/NuRadioMC/blob/9b588ee8ba3cea753b30086652c6ec22f78dcffb/NuRadioReco/framework/base_station.py#L241-L247). I'm not sure if this ensures the precision we're interested in. E.g. trying to set an astropy.time.Time object with a `np.float128`:
```
from astropy.time import Time
import numpy as np
time_128 = np.float128(int(Time.now().unix)) + np.float128(1e-9)
print(time_128, type(time_128))
aptime = Time(time_128, format='unix')
print(aptime.value, type(aptime.value))
```
results in the output
```
1680768370.000000001 <class 'numpy.float128'>
1680768370.0 <class 'numpy.float64'>
```
If we do store our times as unix doubles, the precision is limited to [52 bits](https://numpy.org/doc/stable/user/basics.types.html#array-types-and-conversions-between-types), which means ~ 400 ns. This will limit what we can do with multiple-station coincidences.

fschlueter:
Just to put the information here and to restart the discussion. For the (just merged) mattak reader we set the trigger time absolute and the traces start time is relative to that.

fschlueter:
> Some more comments:
> 
> * we currently store astropy.time.Time objects by calling `.value` (see https://github.com/nu-radio/NuRadioMC/blob/9b588ee8ba3cea753b30086652c6ec22f78dcffb/NuRadioReco/framework/base_station.py#L241-L247
>   ). I'm not sure if this ensures the precision we're interested in. E.g. trying to set an astropy.time.Time object with a `np.float128`:
> 
> ```
> from astropy.time import Time
> import numpy as np
> time_128 = np.float128(int(Time.now().unix)) + np.float128(1e-9)
> print(time_128, type(time_128))
> aptime = Time(time_128, format='unix')
> print(aptime.value, type(aptime.value))
> ```
> 
> results in the output
> 
> ```
> 1680768370.000000001 <class 'numpy.float128'>
> 1680768370.0 <class 'numpy.float64'>
> ```
> 
[...]

PR #798, opened 2025-01-20
Implement sub nanosecond precision for station and event time in nur files

fschlueter:
This PR does two things. 

- It implements sub nanosecond precision for station and event time in nur files.
- Properly implements the event time with an internal representation using astropy. This also changes its default value from 0 to None.

This PR addresses the issue #524.


fschlueter:
ping

sjoerd-bouma (approved):

PR #799, opened 2025-01-20
Update phased trigger and digitizer

ryankrebs016:
Changes:

- Refactor phased array module so it acts as a base to the phased array triggers to reduce identical code between modules.
- Added digital envelope trigger that is an optional trigger for RNO-G.
- Remove extra -1 when calculating V/ADC during digitization.
- Small things to accommodate the runner when simulating events.


fschlueter:
@ryankrebs016 I merged the branch `simulate_trigger_channels_rnog_simplified` into this one and also changed the "traget" of the PR to this branch (at least for the moment). This shows better the changes you made.

NuRadioMC/examples/08_RNO_G_trigger_simulation/generate_data_set.py
```diff
@@ -0,0 +1,349 @@
+#!/bin/env python3
```
fschlueter, line 1: what is the purpose of this script? Compared to `simulate.py` does it only add the two triggers and allows to use the NuRadioMCRunner? Can we not merge the two scripts into one (maybe we can have an second script which takes the simulation as defined in the first script but uses the runners?)

NuRadioMC/utilities/Veff.py
```diff
@@ -276,7 +276,10 @@ def get_Veff_Aeff_single(
     else:
         raise AttributeError(f"attributes do neither contain volume nor area")
 
-    Vrms = fin.attrs['Vrms']
+    try:
+        Vrms = fin.attrs['Vrms']
+    except:
+        Vrms = 1
```
fschlueter, line 282: hm, why is `Vrms` not available? It should? Is `1` a good default value? I think you might be able to make this code a one-liner `Vrms = fin.attrs.get('Vrms', 1)` if hdf5 attributes are compatible with python dictionaries.

NuRadioReco/detector/RNO_G/RNO_single_station_only_PA.json
```diff
@@ -72,7 +72,7 @@
       "ant_type": "RNOG_vpol_4inch_center_n1.73",
       "trigger_adc_sampling_frequency" : 0.472,
       "trigger_adc_nbits": 8,
-      "trigger_adc_noise_nbits": 3.321,
+      "trigger_adc_noise_nbits": 2.585,
```
fschlueter, line 75: has there been a change in the firmware? Is that available in the data?

NuRadioReco/modules/analogToDigitalConverter.py
```diff
@@ -36,9 +36,7 @@ def perfect_comparator(trace, adc_n_bits, adc_ref_voltage, mode='floor', output=
     digital_trace: array of floats
         Digitised voltage trace in volts or ADC counts
     """
-
-    lsb_voltage = adc_ref_voltage / (2 ** (adc_n_bits - 1) - 1)
-
+    lsb_voltage = adc_ref_voltage / (2 ** (adc_n_bits) - 1)
```
fschlueter, line 39: See discussion in #500, if we change this (which I in principle would be very happy with) we would need to update all existing json detector files and would break the code for all users with their own detector files in a very subtle and dangerous way.

NuRadioReco/modules/analogToDigitalConverter.py
```diff
@@ -290,7 +288,7 @@ def get_digital_trace(self, station, det, channel,
 
             adc_ref_voltage = det_channel[adc_ref_voltage_label] * units.V
         else:
-            adc_ref_voltage = Vrms * (2 ** (adc_n_bits - 1) - 1) / (2 ** (adc_noise_n_bits - 1) - 1)
+            adc_ref_voltage = Vrms * (2 ** (adc_n_bits) - 1) / (2 ** (adc_noise_n_bits ) - 1)
```
fschlueter, line 291: See comment above

NuRadioReco/modules/analogToDigitalConverter.py
ryankrebs016: That code implies that the voltage/adc is twice what it really is if you assume adc_n_bits is what's given in the det files. It's not. adc_n_bits here is -1 from what's in the det file as it goes through the ref voltage in the comment below. It works out in the end but it's confusing

NuRadioReco/detector/RNO_G/RNO_single_station_only_PA.json
ryankrebs016: No changes to firmware. See the next two comment threads for why this number changed

NuRadioReco/modules/analogToDigitalConverter.py
fschlueter: I absolutely agree with you, I stumbled across this twice (more than a year apart from each other so I had forgotten). The issue is if we change it now we will effect all other users using this module. Maybe we should have this discussion in the issue I linked above?

NuRadioReco/detector/RNO_G/RNO_single_station_only_PA.json
fschlueter: Okay but you removed the `-1` there, why would you change this number here? Or do you only have to change the nbits if the `-1` is in the voltage/adc conversion? Sorry something does not yet add up for me.

NuRadioReco/modules/RNO_G/triggerBoardResponse.py
```diff
@@ -126,58 +99,71 @@ def apply_adc_gain(self, station, det, trigger_channels, avg_vrms=None):
...
-            self.logger.debug(msg)
-
-            # find the ADC gain from the possible values that makes the realized
-            # vrms closest-to-but-greater-than the ideal value
-            amplified_vrms_values = vrms * self._triggerBoardAmplifications
-            mask = amplified_vrms_values > ideal_vrms
-            if np.any(mask):
-                gain_to_use = self._triggerBoardAmplifications[mask][0]
-                vrms_after_gain.append(amplified_vrms_values[mask][0])
+            volts_per_adc = self._adc_input_range / (2 ** total_bits)
```
fschlueter, line 131: there is a `-1` missing (not in the exponent). If you have a 8bit system you want to divide by 2 ** 8 -1 = 255

NuRadioReco/modules/RNO_G/triggerBoardResponse.py
```diff
@@ -228,22 +214,26 @@ def run(self, evt, station, det, trigger_channels, vrms=None, apply_adc_gain=Tru
...
+            trigger_board_vrms, ideal_vrms, ret_gain_values = self.apply_adc_gain(station, det, trigger_channels, vrms, gain_values)
         else:
             trigger_board_vrms = vrms
-            ideal_vrms = vrms
+            ideal_vrms = np.mean(vrms)
+            ret_gain_values = None
 
         if digitize_trace:
             self.digitize_trace(station, det, trigger_channels, ideal_vrms)
+            trigger_board_vrms = self.get_vrms(station, trigger_channels)
```
fschlueter, line 234: I think this line is a problem for two reasons. Lets start with the important one. You recalculate the vrms here which you use to define the trigger thresholds. However, if the traces are in ADC counts so should be the thresholds (I think). Hence if you `self._adc_output == "counts"` you have to pass the `trigger_board_vrms` through `np.floor(trigger_board_vrms).dtype(int)`. The second reason, before if you passed `vrms` as an argument to this module, what was returned was the amplified vrms. Now you always return the `std(amplified_trace)`. I am not sure what we acutally want to do here but point out that you change what the module does.

fschlueter (changes requested):

NuRadioReco/modules/RNO_G/triggerBoardResponse.py
ryankrebs016: The first reason isn't so much an issue as the traces are gain normalized and digitized so the output is in adc or voltage. Though I agree this can be removed. It was leftover when I was calculating trace rms awkwardly and was skeptical that the gain and digitization was affecting the rms and then thresholds. The assumption would be that if thresholds are stable (on an event by event basis) then the thresholds should be based on theoretical rms values. When calculating the thresholds in simulation, the code does the gain normalization and digitization so the resulting threshold is already informed that things get digitized etc

ryankrebs016: locally I already removed this

ryankrebs016:
Do we want to include the change from adc_noise_nbits to something like adc_noise_counts to remove the ambiguous noise calculation?

NuRadioReco/modules/RNO_G/triggerBoardResponse.py
fschlueter: No - when you calculate the STD from a trace of ints you do not get and int back.

fschlueter: When you removed this line, how do you convert the VRMS to have the "unit" ADC? I use the following code currently:
```
        if vrms is None:
            vrms = self.get_noise_vrms_per_trigger_channel(station, trigger_channels)

        if apply_adc_gain:
            equalized_vrms, ideal_vrms = self.apply_adc_gain(station, det, trigger_channels, vrms)
        else:
            equalized_vrms = vrms
            ideal_vrms = vrms

        if digitize_trace:
            self.digitize_trace(station, det, trigger_channels, ideal_vrms)
            if self._adc_output == "counts":
                lsb_voltage = self._adc_input_range / (2 ** self._n_bits - 1)
                equalized_vrms = np.floor(equalized_vrms / lsb_voltage).astype(int)
                logger.debug("obs. Vrms {} ADC".format(equalized_vrms))

```

ryankrebs016: I'm confused, doesn't something like np.std(array_of_ints) return a float? Also vrms should not be forced to an int. The conversion should happen on the threshold if trying to match firmware

fschlueter: `np.std(array_of_ints)` returns a float. And yes you are right, we have to floor/round just when defining the threshold.

fschlueter:
> Do we want to include the change from adc_noise_nbits to something like adc_noise_counts to remove the ambiguous noise calculation?

Not quite sure what you mean.

fschlueter:
@ryankrebs016 merging #826 caused massive merge conflicts (sorry for that). I can offer to fix those conflicts?  Other option would be to checkout a fresh branch and only port over the changes you want to apply. What do you think?

fschlueter:
> @ryankrebs016 merging #826 caused massive merge conflicts (sorry for that). I can offer to fix those conflicts? Other option would be to checkout a fresh branch and only port over the changes you want to apply. What do you think?

?

fschlueter:
Is that ready to be reviewed?

ryankrebs016:
yes. Though for whatever reason it’s been running crazy slow after merging
________________________________

NuRadioMC/simulation/simulation.py
```diff
@@ -1607,7 +1607,7 @@ def run(self):
                 # we loop through all non-trigger channels and simulate the electric fields for all showers.
                 # then we apply the detector response to the electric fields and find the event in which they will be visible in the readout window
                 non_trigger_channels = list(set(self._det.get_channel_ids(station_id)) - set(channel_ids))
-                if len(non_trigger_channels):
+                if len(non_trigger_channels)>0 and self._outputfilenameNuRadioReco is not None:
```
fschlueter, line 1610: why adding this condition - we store also information of non-trigger channels in hdf5 files.

NuRadioMC/simulation/simulation.py
```diff
@@ -1757,7 +1759,7 @@ def add_filtered_noise_to_channels(self, evt, station, channel_ids):
 
     def _add_empty_channel(self, station, channel_id):
         """ Adds a channel with an empty trace (all zeros) to the station with the correct length and trace_start_time """
-        trigger = station.get_primary_trigger()
+        trigger = station.get_first_trigger()
```
fschlueter, line 1762: I understand that in principle the primary trigger matters. But I believe the idea was that we can simulate "identical" triggers with different thresholds and we keep the freedom to specify the trigger we are most interested in. Correct @cg-laser? Otherwise I agree that in principle first is correct here.

NuRadioMC/utilities/Veff.py
```diff
@@ -276,7 +276,7 @@ def get_Veff_Aeff_single(
     else:
         raise AttributeError(f"attributes do neither contain volume nor area")
 
-    Vrms = fin.attrs['Vrms']
+    Vrms = fin.attrs.get('Vrms', 1)
```
fschlueter, line 279: WIthout checking the context that looks fishy

NuRadioReco/modules/RNO_G/hardwareResponseIncorporator.py
```diff
@@ -218,12 +221,15 @@ def run(self, evt, station, det, temp=293.15, sim_to_data=False, phase_only=Fals
                 # zero first bins to avoid DC offset
                 trig_trace_fft[0] = 0
 
-                # Add trigger channel
-                trig_channel = copy.deepcopy(channel)
-                trig_channel.set_frequency_spectrum(
-                    trig_trace_fft, channel.get_sampling_rate())
+                # Check if already has trigger channel, otherwise make a new one.
+                if channel.has_extra_trigger_channel():
+                    channel.get_trigger_channel().set_frequency_spectrum(trig_trace_fft, channel.get_sampling_rate())
```
fschlueter, line 226: what is the use-case for this. I did not allow the extra trigger channels to exist before intentionally.

NuRadioReco/modules/RNO_G/triggerBoardResponse.py
```diff
@@ -1,11 +1,54 @@
...
 import logging
 logger = logging.getLogger("NuRadioReco.triggerBoardResponse")
 
+main_low_angle = np.deg2rad(-60)
+main_high_angle = np.deg2rad(60)
+phasing_angles = np.arcsin(np.linspace(np.sin(main_low_angle), np.sin(main_high_angle), 12))
+sampling_rate = 472 * units.MHz
+
+#define the keyword arguments for the trigger modules here. these settings emulate the trigger firmware
+#the keys should only appear here and not in as named keyword arguments!
```
fschlueter, line 15: Why?

fschlueter (commented):
I will have to continue reviewing this in batches - sorry.

NuRadioMC/simulation/simulation.py
ryankrebs016: Mostly impatience with how long simulations take. When comparing triggers I only need if the event triggered or not. I can remove this check once I'm done

NuRadioMC/utilities/Veff.py
ryankrebs016: This might be leftovers from a messed up simulation script that wasn't saving stuff. I will revert

NuRadioReco/modules/RNO_G/hardwareResponseIncorporator.py
ryankrebs016: For the trigger efficiency scripts, it takes the neutrino signal and rescales it over various SNRs inside of a loop embedded in the trigger part of the simulation. When I was tracking down why the efficiency scripts were taking 10x longer, I thought it was generating a bunch of nested channels. It wasn't... I'll remove this

NuRadioReco/modules/RNO_G/triggerBoardResponse.py
ryankrebs016: I added the default RNO-G trigger settings that can be passed as a dict of kwargs to the triggers (wasn't sure where to keep it). It's a way to keep track of what settings the trigger modules need to act like RNO-G. There's a bunch of them for the beamformed ones

ryankrebs016:
Let me know if you want me to preemptively go through my code to revert the smaller dumb changes (but maybe some useful ones? wishful thinking I guess) before you have to check everything

NuRadioMC/simulation/simulation.py
ryankrebs016: This might be a useful speed up though especially for trigger simulations only. If this is something you think I should add to a config file, let me know

fschlueter: Absolutely, wanted to say that!

NuRadioMC/examples/08_RNO_G_trigger_simulation/generate_data_set.py
fschlueter: @ryankrebs016 I still think we should merge the two scripts.

NuRadioReco/modules/phasedarray/phasedArray.py
```diff
@@ -0,0 +1,215 @@
+from NuRadioReco.modules.base.module import register_run
+from NuRadioReco.utilities import units
+import logging
+import scipy
+import numpy as np
+from scipy import constants
+
+logger = logging.getLogger('phasedTriggerBase')
```
fschlueter, line 8: Please change logger name to inherit log level from top module, i.e., `NuRadioReco.modules.trigger....`

NuRadioReco/modules/phasedarray/phasedArray.py
```diff
@@ -0,0 +1,215 @@
...
+
+cspeed = constants.c * units.m / units.s
+
+class phasedArray():
```
fschlueter, line 12: Is it worth to change the class name to something like `phasedArrayTrigger` or `phasedArrayTriggerBase`?

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -116,7 +116,8 @@ def run(self, evt, station, det,
             Vrms=None,
             trigger_adc=True,
             clock_offset=0,
-            adc_output='voltage'):
+            adc_output='voltage',
+            **kwargs):
```
fschlueter, line 120: why `**kwargs` is that used somewhere?

fschlueter:
@ryankrebs016 some of the tests are failing because the import of the old module fails `No module named 'NuRadioReco.modules.phasedarray.triggerSimulator'` what module replaces this one?

fschlueter:
I am also wondering if it would be easier to split this PR into 2 or 3 PRs to make the review process easier: 1. Changing the trigger modules, 2. adding the new example scripts. 3. for all the small changes to the other scripts.

NuRadioReco/modules/RNO_G/triggerBoardResponse.py
```diff
@@ -1,11 +1,54 @@
...
+                "coeff_gain": 256 }
+
+envelopeTriggerKwargs={
+                "phasing_angles": phasing_angles,
+                "ref_index": 1.75,
+                "trigger_adc": True,
+                "trigger_filter": None,
+                "apply_digitization": False,
+                "adc_output": "counts",
+                "upsampling_factor": 1,
```
fschlueter, line 39: Is that meant to be one?

ryankrebs016: 1 or 2 on the envelope trigger are doable in firmware. I think a bad 2x upsampling did worse than no upsampling at some point

PR #801, opened 2025-01-22
Implement a new class ParameterStorage to let classes like BaseStation inherit from

fschlueter:
Okay, this PR does more than I initially intended. It implements a new base class `ParameterStorage` to reduce a lot of Boilerplate code. However, this new class brings also an additional big benefit: If we want to add additional parameters to a class (for example `ARIANNAParameters`) we can do so now for many experiments (SKA, RNO-G, ...) without exploding code!

This PR is not yet finished, I first want to see if you like the idea (@sjoerd-bouma, @cg-laser) before investing more time. Things to do:

- [x]  Apply it to all classes which use parameter storages
- [x] Adjust the `serialize` and `deserialize` functions
- [x] Create deprecation warnings for the `get_ARIANNA_parameter()` interface


NuRadioReco/framework/parameter_storage.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __setitem__(self, key, value):
+        self.set_parameter(key, value)
+
+    def __getitem__(self, key):
+        return self.get_parameter(key)
+
+    @check_key
+    def get_parameter(self, key):
+        return self._parameters[key]
```
sjoerd-bouma, line 34: why not instead of the decorator just have a method
```
def __check_key(self, key):
    if not isinstance(key, self._parameter_type):
    ...
```
and then in e.g. get_parameter
```
def get_parameter(self, key):
    self.__check_key(key)
    return self._parameters[key]
```
?
The advantage being you don't have to mess around with `args` and `kwargs` in the decorator.

fschlueter:
Or, with the new class it would also be easy to allow keys from two (or more) different parameter types to be added to one parameter storage. I made a quick test that this might work (although we have to check if that still works serializing/deserializing
```
from NuRadioReco.framework.parameters import stationParameters as stp, channelParameters as chp

test_dict = {stp.nu_zenith: 1, chp.zenith: 2}. # both enums have the numerical value 1
print(test_dict)
```

NuRadioReco/framework/parameter_storage.py
```diff
@@ -0,0 +1,57 @@
...
+
+def check_key(func):
+
+    def wrapper(*args):
+        key = args[1]
+        parameter_class = args[0]._parameter_type
+        if not isinstance(key, parameter_class):
+            logger.error(f"Parameter {key} needs to be of type {parameter_class}")
+            raise ValueError(f"Parameter {key} needs to be of type {parameter_class}")
+        return func(*args)
```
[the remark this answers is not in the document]
fschlueter: Indeed that would not work

fschlueter:
Sjoerd made the name of this branch really meaningless :D

sjoerd-bouma:
Other comment - I'm wondering if the module `parameter_storage` should be private as we probably don't want users to initialize this class. (I'm not sure if that would mess up the documentation as classes generally also link to their parent class). The thinking being that with the large amount of modules/classes already in NuRadioMC we should try to avoid exposing potentially confusing 'internal' classes.

NuRadioReco/framework/parameter_storage.py
```diff
@@ -0,0 +1,49 @@
...
+    def get_parameter_error(self, key):
+        self._check_key(key)
+        return self._parameter_covariances[(key, key)] ** 0.5
+
+    def remove_parameter(self, key):
+        self._check_key(key)
+        self._parameters.pop(key, None)
+
+    def get_parameters(self):
+        return self._parameters
```
sjoerd-bouma, line 49: Looks like this is indeed how it was implemented previously, but this looks dangerous to me - shouldn't we return a deepcopy? I _think_ it makes more sense for this to be read-only, but let me know if you disagree.

NuRadioReco/framework/base_station.py
```diff
@@ -17,64 +19,18 @@
...
 
-class BaseStation():
+class BaseStation(NuRadioReco.framework.parameter_storage.ParameterStorage):
 
     def __init__(self, station_id):
-        self._parameters = {}
-        self._ARIANNA_parameters = {}
-        self._parameter_covariances = {}
+        super().__init__(parameters.stationParameters)
+        self.ARIANNA_parameters = NuRadioReco.framework.parameter_storage.ParameterStorage(parameters.ARIANNAParameters)
```
sjoerd-bouma, line 26: missing `_` (see tests)

fschlueter:
Okay, lets see if the CI is happy with me. I have implemented the new parameter storage into all necessary classes. And I am now allowing parameters from several parameter types to be stored in one parameter storage. This will allow to add more experiment specific data into these parameter storages without duplicating more and more code. This means also that the parameter type is now also stored in the files. An example for this is the `Event` class which stores now `eventParameters` and `generatorAttributes` in the same parameter storage.

fschlueter:
@cg-laser can you approve this? This would in come handy to continue with the RNO-G data processing branch

NuRadioReco/framework/parameter_storage.py
```diff
@@ -0,0 +1,101 @@
...
+    def serialize(self):
+        parameters = {str(key): self._parameters[key] for key in self._parameters}
+        parameter_covariances = {
+            (str(key[0]), str(key[1])): self._parameter_covariances[key]
+            for key in self._parameter_covariances}
+
+        data = {
+            "_parameters": parameters,
+            "_parameter_covariances": parameter_covariances,
+            "_parameter_types": [parameter_type.__name__ for parameter_type in self._parameter_types]
```
sjoerd-bouma, line 71: Why are you storing which types of parameters are accepted by a class? This is already defined in the __init__ (I actually think it should be defined before the __init__, but this is nitpicky) of each class, and I think this should have precedence. But maybe I'm just missing something.

NuRadioReco/framework/parameter_storage.py
```diff
@@ -0,0 +1,101 @@
...
+        return data
+
+
+    def deserialize(self, data):
+        # for backward compatibility
+        if 'parameters' in data:
+            data['_parameters'] = data['parameters']
+
+        if 'parameter_covariances' in data:
+            data['_parameter_covariances'] = data['parameter_covariances']
```
sjoerd-bouma, line 83: Is there a reason why you are now using `_parameters` to store the parameters, instead of `parameters` as previously? Would just changing it to `parameters` mean you could get rid of this backwards-compatibility check?

NuRadioReco/framework/particle.py
```diff
@@ -1,26 +1,20 @@
 from __future__ import absolute_import, division, print_function
-import NuRadioReco.framework.parameters as parameters
-import NuRadioReco.framework.parameter_serialization
+from NuRadioReco.framework.parameters import particleParameters as paP
```
sjoerd-bouma, line 2: I'm not a huge fan of obscure abbreviations of classes anyway, and paP looks pretty ugly to me... but I'm aware this is not the only place in the code where this occurs, and `particleParameters.vertex_times` is indeed very long. Won't reject the PR over this but I don't like it ;)

NuRadioReco/framework/parameter_storage.py
```diff
@@ -0,0 +1,101 @@
+import NuRadioReco.framework.parameters
+
+import itertools
+import logging
+logger = logging.getLogger('NuRadioReco.framework.parameter_storage')
+
+
+class ParameterStorage:
```
sjoerd-bouma, line 8: I'm still curious if it's worth making this module private, because this module should only be used internally, so this somewhat 'clutters' the documentation. I can test this on Monday.

NuRadioReco/framework/parameter_storage.py
sjoerd-bouma: So the conclusion is to leave it as is for now?

sjoerd-bouma (commented):
Just left some minor comments/questions, but I won't be able to approve before Monday, and anyway I don't think I'm impartial on this (I'm a big fan!), so maybe it's worth someone else (@cg-laser  or @anelles) having a brief look at this as well.

NuRadioReco/framework/parameter_storage.py
fschlueter: I think it is a good idea but maybe for another PR?

NuRadioReco/framework/parameter_storage.py
fschlueter: The first class I touched used `_paramters` I thought all did but realised only later that several (maybe more) used `parameter`. So we need this anyway...

NuRadioReco/framework/parameter_storage.py
fschlueter: Well with the PR, we are adding the possibility to store parameters of different types in the same storage (dict). For one class this is currently done in the init function (event with eventParameters and generatorAttributes). However, we allow to add more possible types via a function which should be set in modules/user code. The idea is that for example in the mattak/RNO-G data reader we would do something like `station.add_paramter_type(stationParameterRNOG)`. For those additional parameter types defined at run time we need to store them.

anelles (commented):
I agree, very nice change and it used to be cumbersome indeed.

fschlueter:
> This is great Felix.
> 
> The only downside I can see is that you will break usercode that used `get_ARIANNA_parameter` or `get_generator_attributes`. However this will be an easy fix. We could think about just adding a deprecation warning instead of fully removing the functions, but as the userbase is still smallish, I'm ok with just deleting the functions. But we should make this clear in the changelog.

@cg-laser, I have added deprecation warnings in both cases. In case of `get_generator_attributes` it will issue a warning an than return the `return self.get_parameter(key)`. In case of `get_ARIANNA_parameter` just an error is raised saying one should use `get_parameter`

NuRadioReco/framework/parameter_storage.py
fschlueter: I have already merge this branch into other branches so I would really like doing this in another PR.

NuRadioReco/framework/parameter_storage.py
```diff
@@ -0,0 +1,102 @@
...
+    def has_parameter_error(self, key):
+        self._check_key(key)
+        return (key, key) in self._parameter_covariances
+
+    def remove_parameter(self, key):
+        self._check_key(key)
+        self._parameters.pop(key, None)
+
+    def get_parameters(self):
+        return copy.copy(self._parameters)
```
sjoerd-bouma, line 61: I would have thought you need a `deepcopy` (e.g. if the parameter value of one of the parameters is a numpy array, or a dict, or...)

NuRadioReco/framework/parameter_storage.py
sjoerd-bouma: I'm not a fan of merging public API and then making it private after, but I guess we can live with a lower standard of backwards-compatibility on `develop`.

NuRadioReco/framework/parameter_storage.py
fschlueter: yes I agree a deepcopy is probably what you want. I will open a new PR

PR #802, opened 2025-01-22
reduced LOFAR.json size

karenterveer:
Removed empty or "null" station and channel attributes to reduce the file size of "LOFAR.json". These were irrelevant to the LOFAR detector description.


fschlueter:
Great, more than 100 000 lines scrapped! A few thoughs, all channels seem to have the same `amp_type`, `cab_type`, and `cab_time_delay`. Is that correct? (I think in particular the latter can not be. However, this I think opens the room to really reduce the size further by using the generic detector class. And one other thing. For the channel positions you are using a lot of digits after the dot. I think in a json file this translate to larger files. You could corp these number to mm precision? After all I doubt that you would know the precision with such high accuracy.

fschlueter:
The commission and decommission time are also the same for all channels.

karenterveer:
> Great, more than 100 000 lines scrapped! A few thoughs, all channels seem to have the same `amp_type`, `cab_type`, and `cab_time_delay`. Is that correct? 

the channels all have the same amplifiers, however there are three different cable lengths (50, 80 and 115m). But it would be possible to create generic channels for these three cable types

MijnheerD:
From my memory, the GenericDetector made it such that loading the LOFAR description was very slow (in the order of minutes). Switching to a complete detector description made it much faster. It might be that more recent developments have solved this issue, so you can always try to test with a GenericDetector

karenterveer:
Alright, thanks Mitja! I'll test it and let you know.

fschlueter:
@karenterveer, I was told that the data you are getting have the cable delay already removed. I would suggest to set the cab_delay in the json file to 0 than. And if further I would remove the cable type. Is this information used in any of your reconstruction / analysis? If not I would suggest to remove it.

fschlueter:
The cable lengths too? Although they are different, if they are never used it is unnecessary information (unless you can see a future where they are indeed used)

karenterveer:
no, the cable lengths are used as depending on the cable length a different cable attenuation file applies in the hardware response (see detector/LOFAR/analog_components).

fschlueter:
> no, the cable lengths are used as depending on the cable length a different cable attenuation file applies in the hardware response (see detector/LOFAR/analog_components).

Ah okay, my bad. Than please dismiss my comment

karenterveer:
Okay, I have now tested the Generic detector, loaded both the current version and the generic version and printed the runtimes, and I guess it's safe to say that a generic detector is not an option for us:
image

that being said, I have cropped the positions to mm precision and removed the cable time delays. I think this is the best we can do (for now)

fschlueter:
This is wild, @christophwelling do you understand how this can be? @karenterveer thank you!

fschlueter (approved):

PR #803, opened 2025-01-23
[RNO-G] make channelBlockOffsetFitter automatically pick appropriate block offset removal mode

sjoerd-bouma:
- channelBlockOffsetFitter can now take the mode 'auto'; in this case the approximate block offset removal (faster) is used unless the estimated block offsets are larger than 50% of the Vrms, in which case it will automatically perform the full fit. This should be a considerable speed-up compared to using `fit` all the time with a minimal impact to accuracy.
- added a `run` method which is essentially just an alias of `remove_offsets`, with the only difference an additional `det` parameter (which is ignored) to be consistent with the `run` method of other NuRadio classes.
- minor bugfix: previously the block offsets for all channels were stored in each channel; now only the block offsets corresponding to the channel are stored in `channelParameters.block_offsets`.

### DIFF
--- a/NuRadioReco/modules/RNO_G/channelBlockOffsetFitter.py
+++ b/NuRadioReco/modules/RNO_G/channelBlockOffsetFitter.py
@@ -53,8 +53,8 @@ class channelBlockOffsets:
 
         Parameters
         ----------
-        event: Event object | None
-        station: Station
+        event : `NuRadioReco.framework.event.Event` | None
+        station : `NuRadioReco.framework.station.Station`
             The station to add block offsets to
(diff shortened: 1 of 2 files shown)


fschlueter (dismissed):

fschlueter (approved):

anelles:
Thanks for this.  I had wondered already why there was a block-offset discussion on-going while we already have this. Let's get to merging this quickly.

fschlueter:
Since @sjoerd-bouma told me he might not be available and this is also a bugfix I am taking the liberty to merge it myself.

PR #805, opened 2025-01-24
Implement quick interface to get waveforms from an event

fschlueter:
I thought this function would be helpful in making NuRadio more accessible to newcomers. Thoughts? @anelles @sjoerd-bouma

### DIFF
--- a/NuRadioReco/framework/event.py
+++ b/NuRadioReco/framework/event.py
@@ -156,10 +156,13 @@ class Event:
 
         Returns
         -------
-        times: np.ndarray
+        times: np.ndarray(nr_stations, nr_channels, nr_samples)
             A numpy array containing the times of the waveforms.
-        waveforms: np.ndarray
+            The returned array is squeezed:
+            (1, 10, 2048) -> (10, 2048) or (2, 1, 2048) -> (2, 2048).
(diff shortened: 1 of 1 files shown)


NuRadioReco/framework/event.py
```diff
@@ -137,6 +138,44 @@ def set_id(self, evt_id):
...
+        channel_id: int or list of ints (Default: None)
+            The channel id(s) of the channel(s) for which the waveforms should be returned.
+            If `None`, the waveforms of all channels are returned.
+
+        Returns
+        -------
+        times: np.ndarray
+            A numpy array containing the times of the waveforms.
+        waveforms: np.ndarray
+            A numpy array containing the waveforms.
```
anelles, line 162: Can you add here what it actually returns? I assume it is nChannel x amplitudes ?

anelles (commented):
Seems useful yes. But if it is for newcomers a bit more documentation would be great.

NuRadioReco/framework/event.py
fschlueter: you mean the shape?

fschlueter: Like this or is this now to confusing?

anelles (approved):

PR #806, opened 2025-01-27
Channel glitch detector

philippwindischhofer:
@fschlueter I had to fix a few strange-looking things in b7687e6a73e0a211840b942028b631ebd73cba86 to make it run. Please have a look, maybe I missed something?

@avijai2000 this replaces your glitch detection logic with the one in https://github.com/RNO-G/analysis-tools/blob/main/rnog_analysis_tools/glitch_unscrambler/glitch_detection_per_event.py


NuRadioReco/examples/RNOG/rnog_config.yaml
```diff
@@ -11,7 +11,7 @@ readRNOGDataMattak:
     # Use mattak to calibrate data on the fly. Requires calibration files to be available
     read_calibrated_data: False
     # Do not apply a baseline correction in the read module, do it afterwards
-    apply_baseline_correction: null
+    apply_baseline_correction: none
```
fschlueter, line 14: so `null` converts to `None` and I thought I had added `None` as possilbe value for this module. Maybe it did not yet make it into your branch, sorry for this.

NuRadioReco/examples/RNOG/rnog_standard_data_processing.py
```diff
@@ -71,11 +71,11 @@ def process_data(config):
     hardwareResponseIncorporator = NuRadioReco.modules.RNO_G.hardwareResponseIncorporator.hardwareResponseIncorporator()
     hardwareResponseIncorporator.begin()
 
-    channelBlockOffsetFitter = NuRadioReco.modules.RNO_G.channelBlockOffsetFitter.channelBlockOffsetFitter()
+    channelBlockOffsetFitter = NuRadioReco.modules.RNO_G.channelBlockOffsetFitter.channelBlockOffsets()
```
fschlueter, line 74: autsch, again sorry that you had to fix the code, it is always a problem if you do not run it before. That being said I think we need to add this to the CI asap

NuRadioReco/framework/event.py
```diff
@@ -26,7 +26,7 @@
 class Event(NuRadioReco.framework.parameter_storage.ParameterStorage):
 
     def __init__(self, run_number, event_id):
-        super().__init__([parameters.eventParameters, parameters.generatorAttributes])
+        super().__init__([evp, gta])
```
fschlueter, line 29: here again, sorry

fschlueter:
@philippwindischhofer two things:
- You have a lot of hardcoded numbers in your code `128`, `2048`. Although they are likely to not change I would feel better if we initalize them as variables in `init` or `begin` method. I do not see the need to make them arguments but at least variables (maybe even with a helpful comment or descriptive name) would be nice.
- so you replaced entirely Aishwaryas glitch detection with the test statistics you are calculating. However, I would keep a binary parameter `glitch` and rather add your test statistic as `glitch_test_statistc`. Also see the helper function `has_glitch` in event.py which expects glitch to be a bool. So unless you can come up with a discriminant and convert your statistics to a bool I would suggest to keep Aishwarya's code

philippwindischhofer:
Hi @fschlueter I implemented your comments as we discussed yesterday. Attached is an example plot for Station 13 / Run 1083.

I also noticed another "nuisance" that I commented on above. Perhaps you (or @sjoerd-bouma) know an elegant way out?

rnog_glitch_test

philippwindischhofer:
Here is another example of a run that @avijai2000 pointed me to (Station 11, run 683) where the glitching happens intermittently throughout a run.

rnog_glitch_station11_run683

fschlueter:
@philippwindischhofer could you post example waveforms for the plots you posted here. The runs you are showing were known to be effected by glitching right?

fschlueter:
> Hi @fschlueter I implemented your comments as we discussed yesterday. Attached is an example plot for Station 13 / Run 1083.
> 
> I also noticed another "nuisance" that I commented on above. Perhaps you (or @sjoerd-bouma) know an elegant way out?
> 
> rnog_glitch_test

Would the best conclusion be that all waveforms from channels 0, 2, 3 are good and all from ch 1 are bad? If so the best solution for the future would be to run over all data and create bad-periods?

NuRadioReco/examples/RNOG/rnog_standard_data_processing.py
```diff
@@ -71,11 +71,11 @@ def process_data(config):
...
     channelBlockOffsetFitter.begin()
 
     channelGlitchDetector = NuRadioReco.modules.RNO_G.channelGlitchDetector.channelGlitchDetector()
-    channelGlitchDetector.begin(config["channelGlitchDetector"]["begin"]["kwargs"]["max_deviation"])
+    channelGlitchDetector.begin()
```
fschlueter, line 78: dont you want to add the cut value as argument?

NuRadioReco/examples/RNOG/make_glitch_summary_plot.py
```diff
@@ -1,6 +1,7 @@
 import argparse, numpy as np
 from NuRadioReco.modules.io import eventReader
 from NuRadioReco.framework.parameters import channelParametersRNOG as chp
+import NuRadioReco.framework.parameters as parameters
```
fschlueter, line 4: Das ist aber doppelt gemoppelt

philippwindischhofer:
Hi @fschlueter,

Here are some waveforms from Station 13 / Run 1083 (the first glitch summary plot from above).

It's a bit difficult to visualize the glitches by looking at "typical" RF-triggered waveforms. In some cases, the presence of a glitch is clearly apparent, while in others it's more difficult to see by eye. That's exactly the point behind the construction of the test statistic ... it combines information from across the event.

station13_run1083_ch3
station13_run1083_ch2
station13_run1083_ch1
station13_run1083_ch0

philippwindischhofer:
> Would the best conclusion be that all waveforms from channels 0, 2, 3 are good and all from ch 1 are bad? If so the best solution for the future would be to run over all data and create bad-periods?

To answer your question here: for Station 13 / Run 1083, my conclusion is that CH 0 and 1 are glitching, while CH 2 and 3 are fine.

philippwindischhofer:
> The runs you are showing were known to be effected by glitching right?

And, yes, these runs were previously flagged as being affected by glitching (discovered by Christoph, Jakob, or Aishwarya).

fschlueter:
> > Would the best conclusion be that all waveforms from channels 0, 2, 3 are good and all from ch 1 are bad? If so the best solution for the future would be to run over all data and create bad-periods?
> 
> To answer your question here: for Station 13 / Run 1083, my conclusion is that CH 0 and 1 are glitching, while CH 2 and 3 are fine.

So you would not trust the few waveforms with a ts < 0 in channels 0 and 1?

philippwindischhofer:
> > > Would the best conclusion be that all waveforms from channels 0, 2, 3 are good and all from ch 1 are bad? If so the best solution for the future would be to run over all data and create bad-periods?
> > 
> > 
> > To answer your question here: for Station 13 / Run 1083, my conclusion is that CH 0 and 1 are glitching, while CH 2 and 3 are fine.
> 
> So you would not trust the few waveforms with a ts < 0 in channels 0 and 1?

No, I wouldn't feel comfortable ... 

To me there is no clear indication that the ts < 0 events come from a different population than those with ts > 0 in the same channel and run. And it's not clear what the mechanism would be that causes glitches in one event, but not the next.

philippwindischhofer:
Yes, it would make sense as a rough way to e.g. flag the entire run for closer inspection. I need to read more `NuRadio` code to see if there is a way of keeping per-run metadata in the `nur` file somehow.

philippwindischhofer:
@fschlueter added now!

NuRadioReco/modules/RNO_G/channelGlitchDetector.py
```diff
@@ -2,12 +2,7 @@
 from NuRadioReco.utilities import units
 from NuRadioReco.framework.parameters import channelParametersRNOG as chp
 
-
-
-import numpy as np
-import logging
-
-logger = logging.getLogger('NuRadioReco.RNO_G.channelGlitchDetector')
+import numpy as np, logging
```
fschlueter, line 5: uh, I actually like this syntax but we use it nowhere and it is not pep8 conform I think

NuRadioReco/modules/RNO_G/channelGlitchDetector.py
```diff
@@ -114,3 +135,8 @@ def unscramble(trace):
             
             ch.set_parameter(chp.glitch, glitch_disc)
             ch.set_parameter(chp.glitch_test_statistic, glitch_ts)
+
+            # update glitching statistics
+            if not ch_id in self.events_glitching_per_channel:
+                self.events_glitching_per_channel[ch_id] = 0
+            self.events_glitching_per_channel[ch_id] += glitch_disc
```
fschlueter, line 142: You can use `defaultdict(int)` to simplify the code to simply one line: `self.events_glitching_per_channel[ch_id] += glitch_disc`

NuRadioReco/modules/RNO_G/dataProviderRNOG.py
```diff
@@ -23,29 +23,28 @@ def begin(self, files, reader_kwargs={}, det=None):
 
         self.channelGlitchDetector.begin()
         self.channelBlockOffsetFitter.begin()
-        self.reader.begin(self.files, apply_baseline_correction=None, **reader_kwargs)
+        self.reader.begin(self.files, **reader_kwargs)
```
fschlueter, line 26: Wow, this `apply_baseline_correction=None` should be there

fschlueter:
gosh. I am not sure if I was ever happier

fschlueter (approved):

PR #808, opened 2025-01-27
Implement module for sinewave subtraction

fschlueter:
This PR continues the closed PR #794 (please also see comments from this PR)


fschlueter:
Any updates on this? @alisanozdrina

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,271 @@
...
+"""
+
+
+class channelSinewaveSubtraction:
+    """ Continuous wave (CW) filter module. Uses sine subtraction based on scipy curve_fit. """
+    def __init__(self):
+        pass
+
+    def begin(self,  save_filtred_freqs=False):
+        #self.peak_prominance = peak_prominance
```
cozzyd, line 24: prominence

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,271 @@
...
+        self.save_filtred_freqs = [] if save_filtred_freqs else None
```
cozzyd, line 25: filtered

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,271 @@
...
+
+            trace = channel.get_trace()
+            trace_fil = sinewave_subtraction(
+                trace, peak_prominance, sampling_rate=sampling_rate, 
+                saved_noise_freqs=self.save_filtred_freqs)
+
+            channel.set_trace(trace_fil, sampling_rate)
+
+
+def guess_amplitude(wf: np.ndarray, target_freq: float, sampling_rate: float = 3.2):
```
cozzyd, line 39: note that if you only want the magnitude of one frequency, there are much more efficient methods than to take the full FFT. I believe the Goertzel algorithm is one example.  Alternatively, you may want to guess the amplitude of all the frequencies at the same time, though it's possible subsequent ones may change.

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,271 @@
...
+    wf = wf - np.mean(wf)
+
+    def sinusoid(t, amplitude, noise_frequency, phase):
+        return amplitude * np.sin(2 * np.pi * noise_frequency * t + phase + np.pi/2)
+
+    spec = abs(fft.time2freq(wf, sampling_rate))
+    freqs = fft.freqs(len(wf), sampling_rate)
+
+    # find noise frequencies:
+    rms = np.sqrt(np.mean(np.abs(spec) ** 2))
```
cozzyd, line 110: this is a little funny because the the spectrum has the shape of the filters in it, so the RMS will largely be driven by te difference between the in-band and out-of-band parts of the signal, no? Since you run into the noise floor on the low-pass side, the RMS is also strongly dependent on the gain and dynamic range of the system. If you are going to use the RMS of the spectrum as some sort of baseline I would only consider within the band.

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,271 @@
...
+    peak_idxs = np.where(np.abs(spec) > peak_prominance * rms)[0]
```
cozzyd, line 111: how are you handling non-CW but still relatively narrow-band signals? For example, especially for quickly-falling impulses at the edge of the low-pass, you would get a number of bins significantly larger than the rest, but this is not CW.

cozzyd: ok, I see below that you're grouping them and taking the mean. I guess this is fine since most likely the iteration will be unsuccessful.

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,271 @@
...
+
+        # Initialize a list to hold groups of neighboring peak indices
+        group = [peak_idxs[0]]
+
+        # Loop through the remaining peak indices to group neighboring peaks
+        for i in range(1, len(peak_idxs)):
+            if peak_idxs[i] - peak_idxs[i - 1] == 1:  # If the peak is neighboring the previous one
+                group.append(peak_idxs[i])
+            else:
+                # Calculate the mean frequency for the current group of neighbors
```
cozzyd, line 126: Is it better to take the mean or the maximum value?

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,271 @@
...
+                estimated_amplitude, estimated_freq, estimated_phase = params
+
+                # Generate the estimated CW noise
+                estimated_cw_noise = sinusoid(t, estimated_amplitude, estimated_freq,estimated_phase) 
+
+                logger.info(f"Subtract sinewave with a frequency: {estimated_freq / units.MHz:.1f} MHz, "
+                            f"an amplitude: {estimated_amplitude:.1e} V/GHz and a phase: {estimated_phase / units.deg:.1f} deg")
+
+                # Subtract the estimated CW noise
+                corrected_waveform -= estimated_cw_noise
```
cozzyd, line 159: you should check that the power actually decreases! It may not, for example, if the CW only covers part of the trace.

cozzyd: (or if it wasn't CW to start with, or if it's highly modulated, etc.)

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,271 @@
...
+            saved_noise_freqs.append(noise_freqs)
+
+        for noise_freq in noise_freqs:
+
+            ampl_guess = guess_amplitude(wf, noise_freq, sampling_rate)
+            initial_guess = [ampl_guess, noise_freq, 0.01]
+
+            # Fit the sinusoidal model to the waveform
+            try:
+                params, _ = curve_fit(sinusoid, t, wf, p0=initial_guess)
```
cozzyd, line 149: check if fit converged? Or does curve_fit throw exceptions on lack of convergence? (that would be weird, but I suppose it's possible... )

cozzyd (changes requested):
At the very least, it's important to check that each iteration actually reduces power. I also have some concerns about identifying candidate frequencies, since I don't think the algorithm here is likely to pick up anything other than very large CW (which conversely, means that subtraction may always work!). 

If you'd like, at some point I can try to talk you through some of the design decisions in https://github.com/nichol77/libRootFftwWrapper/blob/master/include/SineSubtract.h (if I can remember any of them).

NuRadioReco/modules/channelSinewaveSubtraction.py
alisanozdrina: I just copied peak searching algo from channelCWNotchFilter. I've changed it to RMS calculation within the band 100-700 MHz

NuRadioReco/modules/channelSinewaveSubtraction.py
alisanozdrina: it should be the same, if the CW peak is "symmetrical", which I thought is always the case?

NuRadioReco/modules/channelSinewaveSubtraction.py
alisanozdrina: curve_fit should raise a RuntimeError if minimization fails. I added checks to see if the fit parameters or the covariance matrix are invalid.

Do you think I should remove the exception and just let the script crash if the fit fails? I thought it be better to just keep going and log the info.

NuRadioReco/modules/channelSinewaveSubtraction.py
alisanozdrina: hmm, I didn't think about it. 

I added check to see if power increases after subtraction, it will skip this peak then. I'll think how to work around partially CW contaminated/modulated events.

NuRadioReco/modules/channelSinewaveSubtraction.py
alisanozdrina: I'll try out Goertzel.

NuRadioReco/modules/channelSinewaveSubtraction.py
cozzyd: It may be a good idea to make the frequency range configurable (or perhaps even to allow multiple disjoint frequency ranges). That could be helpful for other experiments with different bands or bands that have notches in them. The same range(s) could be used for computing the RMS and limiting where peaks are found.

NuRadioReco/modules/channelSinewaveSubtraction.py
cozzyd: For very narrow band CW, you essentially expect the rectangular window function, which will be  indeed symmetric (https://upload.wikimedia.org/wikipedia/commons/6/6a/Window_function_and_frequency_response_-_Rectangular.svg). More generally you won't necessary get a symmetric response (e.g. single-sideband modulation). Also, if the gain of the system has a slope on it, that can skew how many bins you get on each side of the maximum anyway. But you're right, this is probably almost always the same for the types of signals most being considered, and in the case of two nearly equal bins, which will happen if the frequency is halfway between the bins, the mean may be even better than the mode.

NuRadioReco/modules/channelSinewaveSubtraction.py
cozzyd: Ah, ok, yes that would be surprising behavior if the script crashes because the minimization fails (though, from looking at the code, only some failures result in exceptions as far as I can tell)). I agree it's better to let it fail and log, though you may want to add some sort of flag to th event as well (if you don't already, I don't remember).

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -70,8 +72,49 @@ def guess_amplitude(wf: np.ndarray, target_freq: float, sampling_rate: float = 3
...
+    k = int(0.5 + (N * target_freq / sampling_rate))  # Index corresponding to target frequency
+    omega = (2.0 * np.pi * k) / N  # Angular frequency
+    coeff = 2 * np.cos(omega)  # Coefficient for recursion
+
+    # Initialize the previous two values
+    s_prev2 = 0.0
+    s_prev = 0.0
+
+    # Iterate through the waveform to compute the Goertzel algorithm
+    for n in range(N):
```
cozzyd, line 103: Since the  Goertzel algorithm is a type of filter, using one of the scipy signal filter functions may be a more efficient way to implement in Python.

alisanozdrina: you mean writing a spectral amplitude guesser that uses a scipy butter bandpass to isolate the CW frequency and measure its amplitude?

I checked how much time it takes for fft, Goertzel and scipy signal filter to estimate the spectral amplitude of a simulated trace. Strangely, FFT is faster than Goertzel, but I don’t understand why.

exec_time

cozzyd: I suspect it's because FFT is implemented in C or Fortran, while your Goertzel is in Python?  It may be that the FFT time is subdominant anyway so you don't care too much.

cozzyd: and no, what I mean is to write the Goertzel filter as a (complex?) IIR filter and use one of the IIR filter regimes.

alisanozdrina: ok, python Goertzel with IIR filter now is slightly faster than fft. I'll use it as the default function to guess CW amplitude. 
exec_time2

cozzyd:
If you have access to MATLAB it might be a good idea to validate your Goertzel implementation against the MATLAB one (at the very least, see if you get a comparable answer to the FFT method... the normalization may differ).

NuRadioReco/modules/channelSinewaveSubtraction.py
alisanozdrina: I've checked that it gives same results at the fft implemetation, yes.

cozzyd:
I think this looks good to me now?

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,361 @@
...
+class channelSinewaveSubtraction:
+    """ Continuous wave (CW) filter module. Uses sine subtraction based on scipy curve_fit. """
+    def __init__(self):
+        pass
+
+    def begin(self,  save_filtered_freqs=False, freq_band=(0.1, 0.7)):
+        self.save_filtered_freqs = [] if save_filtered_freqs else None
+        self.freq_band = freq_band
+
+    def run(self, event, station, det=None, peak_prominence=4.0):
```
fschlueter, line 27: missing run decorator

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,361 @@
...
+In contrast to the module channelCWNOtchFilter, which uses a notch filter to remove CW noise.
+"""
+
+
+class channelSinewaveSubtraction:
+    """ Continuous wave (CW) filter module. Uses sine subtraction based on scipy curve_fit. """
+    def __init__(self):
+        pass
+
+    def begin(self,  save_filtered_freqs=False, freq_band=(0.1, 0.7)):
```
fschlueter, line 23: missing doc-string

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,361 @@
...
+            sampling_rate = channel.get_sampling_rate()
+
+            trace = channel.get_trace()
+            trace_fil = sinewave_subtraction(
+                trace, peak_prominence, sampling_rate=sampling_rate, 
+                saved_noise_freqs=self.save_filtered_freqs, freq_band=self.freq_band)
+
+            channel.set_trace(trace_fil, sampling_rate)
+    def get_filtered_frequencies(self):
+        return self.save_filtered_freqs
```
fschlueter, line 38: some space for readability, doc-string?

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+def guess_amplitude(wf: np.ndarray, target_freq: float, sampling_rate: float = 3.2):
```
fschlueter, line 41: I like the typing but than please do it in all functions of this file.

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,361 @@
...
+    ampl: float
+        Estimated amplitude of the target frequency.
+    """
+    if wf.size == 0:
+        raise ValueError("Input waveform is empty.")
+
+    if target_freq < 0 or target_freq > sampling_rate / 2:
+        raise ValueError("Target frequency is out of range (0 to Nyquist frequency).")
+
+    fft_spectrum = fft.time2freq(wf, sampling_rate) * sampling_rate / np.sqrt(2) # take into account nuradio normalization
```
fschlueter, line 65: if you use the functions in the `fft` module you do not have to add additional normalization terms.

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,361 @@
...
+    frequencies = fft.freqs(len(wf), sampling_rate)
+
+    # Find amplitude of the 50 Hz harmonic
+
+    bin_index = np.argmin(np.abs(frequencies - target_freq))
+    amplitude = np.abs(fft_spectrum[bin_index]) * 2 / len(wf) # Extract the amplitude (real + imaginary parts)
+
+    return amplitude
+
+def guess_amplitude_iir(wf: np.ndarray, target_freq: float, sampling_rate: float = 3.2):
```
fschlueter, line 75: can you remove all the commented code in this function?

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,361 @@
...
+    imag = s_prev2 * np.sin(omega)
+
+    # Compute magnitude (amplitude)
+    amplitude = np.sqrt(real**2 + imag**2) / scaling_factor
+    #print(amplitude)
+    return amplitude
+
+
+
+def sinewave_subtraction(wf: np.ndarray, peak_prominence: float = 4.0, sampling_rate: float = 3.2,  saved_noise_freqs: list = None, freq_band: tuple = (0.1, 0.7)):
```
fschlueter, line 123: Can you also remove the comments in this function

fschlueter:
Mostly cosmetic comments. One comment about the fft normalisation and a missing decorator. But I think we can merge this soon!

NuRadioReco/modules/channelSinewaveSubtraction.py
alisanozdrina: fixed now

NuRadioReco/modules/channelSinewaveSubtraction.py
alisanozdrina: hmm. I actually tested that on simulated waveforms: wf = np.random.randn(N) + A * np.sin(2 * np.pi * target_freq * t)
I receive correct amplitude with this normalization. it also gives the same results as Goertzel algo function.

fschlueter: Can you elaborate with an example?

fschlueter: The fft module applies a normalisation which it seems you are undoing: 
```
return np.fft.rfft(trace, axis=-1) / sampling_rate * 2 ** 0.5  # an additional sqrt(2) is added because negative frequencies are omitted.
```
I am not sure I understand your motivation for that yet. The NuRadio convention has the goal to conserve energy between the two domains:
```
np.sum(trace**2) * dt = np.sum(spectrum**2) * df
```
Does that mean that the amplitude is not conserved (which is what you care about here)?

alisanozdrina: I checked the amplitude estimates from these two functions (guess_amplitude() and guess_amplitude_iir()), which I use for the sine wave fitter, against simulated waveforms with parameters similar to those of RNO-G

ampl_guess

fft
wf

alisanozdrina: I think this has answer to energy conservation and how it affect the amplitude - https://physics.stackexchange.com/questions/71891/parsevals-theorem-on-a-random-signal

fschlueter: @alisanozdrina I missed that you added an additional `2 / len(wf)` factor a few line below. I moved that up such that all factors are in one line. I also added a comment why we are not using the standard fft normalization here.

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -20,21 +21,52 @@ class channelSinewaveSubtraction:
...
+            Frequency band to calculate baseline RMS of fft spectrum. Used to identify noise peaks.
+            0.1 to 0.7 GHz is the default for RNO-G, based on bandpass. 
+
+        """
         self.save_filtered_freqs = [] if save_filtered_freqs else None
         self.freq_band = freq_band
 
-    def run(self, event, station, det=None, peak_prominence=4.0):
+    @register_run()
+    def run(self, event: 'NuRadioReco.framework.event.Event', station: 'NuRadioReco.framework.station.Station', det: 'NuRadioReco.detector.Detector' = None, peak_prominence: float = 4.0) -> None:
```
fschlueter, line 41: @sjoerd-bouma can we use the backticks `` for typing?

fschlueter:
I think this PR is almost good to go with one outstanding item: the typing using the NuRadio classes. It seems that the CI complains about this. Maybe Sjoerd can help there?

NuRadioReco/modules/channelSinewaveSubtraction.py
sjoerd-bouma: [This](https://peps.python.org/pep-0484/#forward-references) suggests that using a string instead of the actual class is acceptable only if the reference is resolvable once the module is loaded completely, which is not currently the case. I guess I would suggest explicitly importing the classes and removing the `'` altogether.

(backticks don't apply here, as that would be rst syntax that is resolved by sphinx when compiling the documentation, which is unfortunately completely separate from Python typing)

NuRadioReco/modules/channelSinewaveSubtraction.py
```diff
@@ -0,0 +1,404 @@
...
+        """
+        Run the CW filter module on a given event and station. Removes all the CW peaks > peak_prominence * RMS.
+
+        Parameters
+        ----------
+        event: `NuRadioReco.framework.event.Event`
+            Event object to process.
+        station: `NuRadioReco.framework.station.Station`
+            Station object to process.
+        det: `NuRadioReco.detector.detector.Detector` (default: None)
```
fschlueter, line 51: I think the documentation currently fails because of this. `NuRadioReco.detector.detector.Detector` is a factory function and probably not suited to specify a type.

NuRadioReco/modules/channelSinewaveSubtraction.py
fschlueter: But than the typing is correct and not the cause of the failure in the CI? Is that checked anyway?

fschlueter: How do we fix that? `NuRadioReco.detector.detector.Detector` is not a class.

fschlueter:
Okay, I could not fix the typing for the detector object due to the known reason that `NuRadioReco.detector.detector.Detector` is a factory function and not a class. Hence I completely omitted the type for the variable (which is anyway not used in the module). @sjoerd-bouma, can you live with that?
@cozzyd can you approve? (I can't because technically I created this PR).

fschlueter:
Ping @cozzyd @alisanozdrina @sjoerd-bouma

cozzyd (approved):

fschlueter:
Thank you very much @alisanozdrina!

PR #811, opened 2025-01-28
Export rno-g detector without complete response data to reduce file size

fschlueter:
As discussed in the monday call, this PR implements a striped down version of the detector files and adds a `get_time_delay2` function to be used in mattak. I also added a function which should produce a detector file for every possible state of the detector. This accumulates to quite a number of files since we get a file every time we newly commissioned or decommissioned a station/channel.

### DIFF
--- a/NuRadioReco/detector/RNO_G/rnog_detector.py
+++ b/NuRadioReco/detector/RNO_G/rnog_detector.py
@@ -1195,7 +1195,7 @@ class Detector():
         return self.get_time_delay(station_id, channel_id, use_stored=use_stored, trigger=trigger)
 
 
-    def get_time_delay2(self, station_id, channel_id, trigger=False):
+    def _get_time_delay(self, station_id, channel_id, trigger=False):
         """ Returns the sum of the time delay of all components in the signal chain calculated from the phase.
 
         The function returns the values which were precalculated by the hardware database interface. They
@@ -1221,6 +1221,10 @@ class Detector():
         -------
         time_delay: float
             Sum of the time delays of all components in the signal chain for one channel.
+
+        Also see
+        --------
+        get_time_delay
         """
         signal_chain_dict = self.get_channel_signal_chain(
         station_id, channel_id)
@@ -1288,15 +1292,17 @@ class Detector():
 
         See Also
         --------
-        get_time_delay2
         get_cable_delay
         """
         signal_chain_dict = self.get_channel_signal_chain(
             station_id, channel_id)
 
         if use_stored:
-            resp = self.get_signal_chain_response(station_id, channel_id, trigger=trigger)
-            return resp.get_time_delay()
+            try:
+                resp = self.get_signal_chain_response(station_id, channel_id, trigger=trigger)
+                return resp.get_time_delay()
+            except KeyError: # in case the full S21 parameters are not stored
+                return self._get_time_delay(station_id, channel_id, trigger=trigger)
         else:
             time_delay = 0
             if trigger and "trigger_response_chain" not in signal_chain_dict:
--- a/NuRadioReco/detector/RNO_G/rnog_detector.py
+++ b/NuRadioReco/detector/RNO_G/rnog_detector.py
@@ -169,6 +169,10 @@ class Detector():
             self._det_imported_from_file = True
             self._import_from_file(detector_file)
 
+        # This should be set with Detector.update(..) and corresponds to the time of a measurement.
+        # It will be use to decide which components are commissioned at the time of the measurement
+        self.__detector_time = None
+
         # Allow overwriting the hard-coded values
         over_write_handset_values = over_write_handset_values or {}
         self.__default_values.update(over_write_handset_values)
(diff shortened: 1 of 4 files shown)


fschlueter:
@sjoerd-bouma or @jakobhenrichs can someone review this PR?

NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -220,6 +222,16 @@ def export(self, filename, json_kwargs=None, additional_data=None):
                     self._time_periods_per_station[station_id]["modification_timestamps"][idx]]
             }
 
+        if drop_response_data:
+            keys_to_drop = ['frequencies', 'mag', 'phase']
+            for chain in ["response_chain", "trigger_response_chain"]:
+                for station_id in self.__buffered_stations:
```
jakobhenrichs, line 228: Maybe its better to create a deep copy of `self.__buffered_stations` first, likes its done in `export_as_string()`. Otherwise you are altering the buffered detector, right?

NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -1205,6 +1188,75 @@ def get_cable_delay(self, station_id, channel_id, use_stored=True, trigger=False
         # However, having the cable delay without amplifiers is anyway weird.
         return self.get_time_delay(station_id, channel_id, use_stored=use_stored, trigger=trigger)
 
+
+    def get_time_delay2(self, station_id, channel_id, trigger=False):
```
jakobhenrichs, line 1195: I think its very confusing for the user to have two functions that basically do the same thing. Would it not be possible to replace the `get_time_delay(use_stored=True)` with this function? The code will also work if the buffered detector has response data stored, right?

NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -1337,17 +1410,46 @@ def get_component(self, collection="coax_cable", component="daq_drab_flower_2024
         return resp
 
 
-if __name__ == "__main__":
+def produce_detector_files_for_all_time_periods(drop_response_data=False):
+    """
+    This function produces a detector file for each time period necessary
+    """
+    db = det.get_database()
```
jakobhenrichs, line 1413: Should be `db = self.get_database()`

NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -1337,17 +1410,46 @@ def get_component(self, collection="coax_cable", component="daq_drab_flower_2024
...
+    station_ids = self.db[self.__station_collection].distinct("id")
```
jakobhenrichs, line 1414: Should be without `self` or `self.__db`

fschlueter: This function is no member function. self does not exist... I will fix it

NuRadioReco/detector/RNO_G/rnog_detector.py
fschlueter: I like the fact that we use the `total_reponse` object. But I agree that this is awkward.

NuRadioReco/detector/RNO_G/rnog_detector.py
fschlueter: fine

NuRadioReco/detector/RNO_G/rnog_detector.py
fschlueter: one could add a 3rd mode to get_time_delay...

fschlueter: Okay, I think I found a good solution. `get_time_delay2` is not `_get_time_delay()` and this is called in `get_time_delay` if an keyerror occurs.

jakobhenrichs: fine with me :+1:

NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -1337,15 +1415,44 @@ def get_component(self, collection="coax_cable", component="daq_drab_flower_2024
...
+
+        ts = np.unique(np.hstack([station_ts['modification_timestamps'] for station_ts in ts_dict.values()]))
+        for tdx in range(len(ts) - 1):
+            t0 = ts[tdx]
+            t1 = ts[tdx + 1]
+            time = t0 + (t1 - t0) / 2
+
+            det.update(time)
+            if det.has_station(station_id):
+                det.export(
```
jakobhenrichs, line 1443: Is there a reason why we want to create a single file per station and not a single file for all stations together?

fschlueter: I guess to have each file as small as possible. And in practice we probably anyway only look at data of one station at a time. But this is also only a silly helper function which we can modify in the future.

jakobhenrichs: I think its not so nice that the station ids are hardcoded in the function. But you are right, it can be modified in the future.

jakobhenrichs (dismissed):
Looks good to me. Thanks for implementing it so fast!

NuRadioReco/detector/RNO_G/rnog_detector.py
fschlueter: station 14 is not yet in the DB

cozzyd: (would be better to gather unique station ids from the database)

PR #814, opened 2025-01-28
initialize trigger_channels to None in RNO-G hardwareResponseIncorporator

sjoerd-bouma:
One-line change; initialize the `self.trigger_channels` to `None`. The introduction of this attribute meant that the `begin` method of the hardwareResponseIncorporator was no longer optional.

@fschlueter if this is intentional, feel free to close this PR; I can see the value in 'forcing' users to call `begin` so they are (hopefully) more aware of the `trigger_channels` option. But your change broke one of my scripts, and no behaviour is affected by this addition, so :shrug:

### DIFF
--- a/NuRadioReco/modules/RNO_G/hardwareResponseIncorporator.py
+++ b/NuRadioReco/modules/RNO_G/hardwareResponseIncorporator.py
@@ -23,6 +23,7 @@ class hardwareResponseIncorporator:
         self.__time_delays = {}
         self.__t = 0
         self.__mingainlin = None
+        self.trigger_channels = None
 
     def begin(self, trigger_channels=None):
         """


fschlueter (approved):
Fine, thanks for fixing that. Did we clearly define somewhere what the init and begin method should be used for? I think currently every module does it own thing which is very annoying. I would love to clearly define the purpose of both methods and consolidate what we have so far in NuRadio?

sjoerd-bouma:
https://nu-radio.github.io/NuRadioMC/NuRadioReco/pages/nur_modules.html

There is of course some ambiguity between settings that could or couldn't change from event to event, which in the end is up to the user. My feeling is that currently most modules don't use the begin method, but there are definitely some that require it. Implementing inheritance as suggested by Cosmin might be a nice way to make the 'default' begin method optional, and overwrite it explicitly with methods that are non-optional.

fschlueter:
@sjoerd-bouma can you point to example where we actually need the begin method and could not do everything via the constructor (`__init__`)? It would also be nice if we have a module/method which automatically calls the `end` method of all modules which ran.

sjoerd-bouma:
> @sjoerd-bouma can you point to example where we actually need the begin method and could not do everything via the constructor (`__init__`)? It would also be nice if we have a module/method which automatically calls the `end` method of all modules which ran.

readRNOGData?

fschlueter:
> > @sjoerd-bouma can you point to example where we actually need the begin method and could not do everything via the constructor (`__init__`)? It would also be nice if we have a module/method which automatically calls the `end` method of all modules which ran.
> 
> readRNOGData?

But do we really? I agree this module is one of the cases where it could make sense but I could see world where we do everything in the constructor.

PR #816, opened 2025-01-29
Adds the SKALA_v4 antenna model and hashes

sjoerd-bouma:
Adds a function to convert the FEKO simulations for the SKALA_v4 antennas to NuRadio pkl files, and adds their hashes to the antenna_models_hash.json file. They are already on the DESY server.


fschlueter:
Did you hear back from Clancy? Does this code contain a proper conversion from FEKO simulationen to VEL, or is that commiting the "is" situation?

fschlueter (approved):

fschlueter:
To summarised my understanding. In the FEKO file the `EEP` is stored per polarisation. We define the VEL as `U = VEL * E` where `U` is the voltage as the antenna footpoint and `E` the incoming electric field vector. Hence we can substitute eq. 6 in ref [1] and get `VEL = -2j lambda/eta_0 * EEP`. And we since our base unit is `1 Volt` we do not need any additional normalisation? Is that correct? If so I am happy with the PR.

sjoerd-bouma:
> To summarised my understanding. In the FEKO file the `EEP` is stored per polarisation. We define the VEL as `U = VEL * E` where `U` is the voltage as the antenna footpoint and `E` the incoming electric field vector. Hence we can substitute eq. 6 in ref [1] and get `VEL = -2j lambda/eta_0 * EEP`. And we since our base unit is `1 Volt` we do not need any additional normalisation? Is that correct? If so I am happy with the PR.

That's as good as my understanding, though you're missing the impedance of the amplifier (Z_L) there. And while the EEP is technically in volts, it has already been divided by the input voltage (1 V), so it should be unitless here, I think.

fschlueter:
> > To summarised my understanding. In the FEKO file the `EEP` is stored per polarisation. We define the VEL as `U = VEL * E` where `U` is the voltage as the antenna footpoint and `E` the incoming electric field vector. Hence we can substitute eq. 6 in ref [1] and get `VEL = -2j lambda/eta_0 * EEP`. And we since our base unit is `1 Volt` we do not need any additional normalisation? Is that correct? If so I am happy with the PR.
> 
> That's as good as my understanding, though you're missing the impedance of the amplifier (Z_L) there. And while the EEP is technically in volts, it has already been divided by the input voltage (1 V), so it should be unitless here, I think.

Great!

fschlueter:
I assume the files on the desy server are not downscaled. Can you add downscaled versions? Loading ~2Gb files takes quite some time.

sjoerd-bouma:
> I assume the files on the desy server are not downscaled. Can you add downscaled versions? Loading ~2Gb files takes quite some time.

You know what happens when you assume... they're 140 MB each, produced with the default settings in the preprocess function.

fschlueter:
> > I assume the files on the desy server are not downscaled. Can you add downscaled versions? Loading ~2Gb files takes quite some time.
> 
> You know what happens when you assume... they're 140 MB each, produced with the default settings in the preprocess function.

damn, me culpa!

fschlueter:
image

fschlueter:
@sjoerd-bouma is this PR good to go?

fschlueter:
@sjoerd-bouma just to be sure (and have it documented for once here). With the new patterns, the "Xpol" antenna which should describe an LPDA those dipols are aligned east-west (along the x-axis) and is most sensitive to signals coming from north south should have an orientation vector of `[ori_theta, ori_phi, rot_theta, rot_phi] = [0, 0, pi / 2, pi / 2]` and a orthogonal "Ypol" `[0, 0, pi / 2, 0]`?

sjoerd-bouma:
This is probably a bad place to document these things. The 'native' orientation is also stored inside the .pkl files, but I guess I could add this to https://nu-radio.github.io/NuRadioMC/NuRadioReco/pages/detector/antennamodels.html#antenna-models.

Anyway, to actually answer your question, I decided the orientations of the Xpol and Ypol are East-West (0,0,90,90) and North-South (0,0,90,180), respectively; the difference with what you suggested amounts to a polarity flip in the Ypol antenna. Whether this is the correct interpretation of Daniel Ung's simulations is up for debate, but this results in X and Y-pol patterns that are very similar if they have the same orientation.

PR #817, opened 2025-01-29
Explicitly defining volt = 1 as base unit (althouht it can be derived…

fschlueter:
… from the base units for energy and charge). Adding sympol for microvolt = muV. Some cleaning

### DIFF
--- a/NuRadioReco/utilities/units.py
+++ b/NuRadioReco/utilities/units.py
@@ -221,7 +221,7 @@ microampere = micro * ampere
 nanoampere = nano * ampere
 
 # Electric potential [E][Q^-1]
-volt = 1  # == electronvolt / eplus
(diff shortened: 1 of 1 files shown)


sjoerd-bouma:
This is technically bad (the best kind of bad or the worst kind?) - derived units shouldn't be defined explicitly, because this can lead to an inconsistent system of units. Changing it to volt = electronvolt / eplus would be fine by me; if you want to add the explicit comment that this is equal to one that is also fine.

fschlueter:
Okay, I do not disagree.

sjoerd-bouma (approved):
I'm going to foolishly assume no one will think `units.h` refers to one Planck action... at least until the theorists take over NuRadioMC.

PR #818, opened 2025-01-31
adds 2024 RNO-G detector description

sanyukta13:
Added RNO_season_2024.json


fschlueter:
@sanyukta13, I checked a few channels in station 11 and their time_delay changed w.r.t. the 2023 json file. Why is that? Also, we changed the sampling rate for our data in the 2024 season to 2.4 GHz. This is not reflected (but this is also not easy to do with these json files. I maybe we can discuss this at the Monday call later.

fschlueter:
@sanyukta13, we changed the sampling rate of all stations to 2.4 GHz not just the new station 14.

fschlueter:
Furthermore, I found more channels being different between the season 23 and 24 file: Examples are station 12 channels 0 and 23.

sanyukta13:
> Furthermore, I found more channels being different between the season 23 and 24 file: Examples are station 12 channels 0 and 23.

@fschlueter, there have been changes to station 12 positions and cable delays so the discrepancy is a result of the updated values. Please refer to _calib.json_ attached. 
[calib.json](https://github.com/user-attachments/files/18663421/calib.json)

philippwindischhofer:
Hi @sanyukta13 my understanding is that for now we want to keep using only the surveyed locations in these kinds of detector description files. What you mention in terms of post-calibration positions (and cable delays) will receive its proper place in the hardware DB. 

The bug in the cable delays for Station 12 that you mention has already been corrected (I think?) in the hardware DB, and will be factored into upcoming calibration releases.

fschlueter:
Oh I was not aware that this file should already include calibrations, it is also not mentioned anywhere. Where are these calibration results coming from? I would also suggest to at least add two files, one describing the calibrated and one describing the uncalibrated detector. If you do that please provide enough information about the origin of the calibration and so on.

fschlueter:
I do not think it is a good idea to have one file where one station is "calibrated" while the others are not.

sanyukta13:
My bad, I'll revert station 12 description to what it was in 2023.

fschlueter (commented):
Great, thank you very much!

fschlueter (approved):

PR #820, opened 2025-02-03
Fix refactoring and renaming of triggerTimeAdjuster module.

fschlueter:
This PR does:

- Fixes a bug in the method `base_trace.add_to_trace` which only occurred in a situation where the trace and the trace to add had timing which perfectly aligned. @MartinRavn for the fix!
- Remove everything from the deprecated module `triggerTimeAdjuster` and only throws an error now. 
- Fixes the timing of the non-trigger channels.
- Rewrites the new module `channelReadoutWindowCutter` to use the new method `add_to_trace`.


fschlueter:
hm I do not understand why the diff is so large ..

fschlueter:
Maybe we merge this branch directly in develop @cg-laser

NuRadioMC/simulation/simulation.py
```diff
@@ -1721,15 +1715,24 @@ def run(self):
...
-        # we need to use any other channel to get the correct trace_start_time. All channels have the same start time at the end
-        # of the simulation.
-        channel.set_trace_start_time(station.get_channel(station.get_channel_ids()[0]).get_trace_start_time())
+
+        # Set the correct trace start time taking into account different `pre_trigger_times`
+        trigger_channel_id = station.get_channel_ids()[0]
+        trigger_channel_trace_start_time = station.get_channel(trigger_channel_id).get_trace_start_time()
+        channel_trace_start_time = trigger_channel_trace_start_time + (
+            primary_trigger.get_pre_trigger_time_channel(channel_id) -
+            primary_trigger.get_pre_trigger_time_channel(trigger_channel_id))
```
fschlueter, line 1733: hm if I am already having the primary trigger object I do not need the `trigger_channel_trace_start_time = station.get_channel(trigger_channel_id).get_trace_start_time()`

fschlueter:
@cg-laser, I talked to @MartinRavn and he is happy with the changes we made to `base_trace.add_to_trace`. I think we fixed a small additional bug that traces were added with one sample short (missing at the end). I am not sure that has a big effect/explains that the one unit test fails. The unit tests runs the simulation twice, once without and once with noise. Only the comparison with noise fails. Maybe it is due to the changes in `simulation.py`?

PR #821, opened 2025-02-04
Cache antenna vector effective length in channelGalacticNoiseAdder

fschlueter:
…nnelGalacticNoiseAdder

This hides the fact that getting the vector effective length from the `antenna_pattern` class is pretty slow. Since arrays/lists are not hashable we can not directly pass the frequency vector to the cached function but have to pass `self` and use `self.__freqs`. This means we have to make sure ourselves that `self.__freqs` does not change or we have to clear the cache.

I have run some very limited tests (I run it and checked that the resulting nur files are the same (fixing the seed of the module). However, it would be good if someone actually looked at the noise/vel this caching produces. Maybe @PhilippLaub, @JethroStoffels or @karenterveer can do that?

@sjoerd-bouma any comments about the implementation?

### DIFF
--- a/NuRadioReco/modules/channelGalacticNoiseAdder.py
+++ b/NuRadioReco/modules/channelGalacticNoiseAdder.py
@@ -114,9 +114,9 @@ class channelGalacticNoiseAdder:
 
         self.__caching = caching
         self.__freqs = None
-        if self.__caching and self.__n_side > 10:
+        if self.__caching and self.__n_side >= 10:
             logger.warning(
-                "Caching for the vector effective length is enabled (with `maxsize=1024`). `n_side` is larger 10 and thus "
+                "Caching for the vector effective length is enabled (with `maxsize=1024`) and `n_side >= 10`, and thus "
                 "it produces to many different caching entries for two antenna models to be stored of one `station_time`. "
                 "Either decrease `n_side` or increase `maxsize` (has to be done in the source code).")
 
--- a/NuRadioReco/modules/channelGalacticNoiseAdder.py
+++ b/NuRadioReco/modules/channelGalacticNoiseAdder.py
@@ -71,7 +71,8 @@ class channelGalacticNoiseAdder:
             n_side=4,
             freq_range=None,
             interpolation_frequencies=None,
-            seed=None
+            seed=None,
+            caching=True
     ):
         """
         Set up important parameters for the module
@@ -100,6 +101,9 @@ class channelGalacticNoiseAdder:
         seed : {None, int, array_like[ints], SeedSequence}, optional
             The seed that is passed on to the `numpy.random.Philox` bitgenerator used for random
             number generation.
+        caching: bool, default: True
+            If True, the antenna response is cached for each channel. This can speed up this module
+            by a lot. If the frequencies of the channels change, the cache is cleared.
         """
         if debug:
             warnings.warn("This argument is deprecated and will be removed in future versions.", DeprecationWarning)
@@ -107,6 +111,9 @@ class channelGalacticNoiseAdder:
         self.__random_generator = Generator(Philox(seed))
         self.__n_side = n_side
         self.solid_angle = healpy.pixelfunc.nside2pixarea(self.__n_side, degrees=False)
+
+        self.__caching = caching
+        self.__freqs = None
 
         if interpolation_frequencies is None:
             if freq_range is None:
@@ -160,6 +167,12 @@ class channelGalacticNoiseAdder:
             self.__radio_sky = healpy.pixelfunc.ud_grade(self.__radio_sky, self.__n_side)
             self.__noise_temperatures[i_freq] = self.__radio_sky
 
+
+    @functools.lru_cache(maxsize=1024 * 32)
+    def get_cached_antenna_response(self, ant_pattern, zen, azi, *ant_orient):
+        return ant_pattern.get_antenna_response_vectorized(self.__freqs, zen, azi, *ant_orient)
+
+
     @register_run()
     def run(
             self,
@@ -203,6 +216,20 @@ class channelGalacticNoiseAdder:
         freqs = last_freqs
         d_f = freqs[2] - freqs[1]
 
+        # If we cache the antenna pattern, we need to make sure that the frequencies have not changed
+        # between stations. If they have, we need to clear the cache.
+        if self.__caching:
+            if self.__freqs is None:
+                self.__freqs = freqs
+            else:
+                if not np.allclose(self.__freqs, freqs, rtol=0, atol=0.01 * units.MHz):
+                    self.__freqs = freqs
+                    self.get_cached_antenna_response.cache_clear()
+                    logger.warning(
+                        "Frequencies have changed. Clearing antenna response cache. "
+                        "(If this happens often, something might be wrong...")
+
+
         if passband is None:
             passband = [10 * units.MHz, 1000 * units.MHz]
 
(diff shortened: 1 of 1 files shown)


NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -160,6 +167,12 @@ def begin(
             self.__radio_sky = healpy.pixelfunc.ud_grade(self.__radio_sky, self.__n_side)
             self.__noise_temperatures[i_freq] = self.__radio_sky
 
+
+    @functools.lru_cache(maxsize=1024 * 32)
```
sjoerd-bouma, line 171: I guess for the galacticnoiseadder you need a large cache... so what is this, 32000 * len(frequencies) * 2 (polarizations) 128-bit complex numbers? So you would typically increase the memory usage by O(1--2 GB), is that right?

fschlueter: > Also, I'm pretty sure self is not hashed, so you may want to clear the cache on class initialization as well.

I can not follow.

fschlueter: `maxsize` refers to the number of entries in the internal dict which is used for caching not the actual size of the cache. I still read somewhere that it should be a power of two... Right now we allow for ~32k entries which should be by far sufficient.

sjoerd-bouma: My bad for making two points in a single thread. 32000 = 32 * 1024 (close enough). Each entry contains len(frequencies) * len(polarizations) complex numbers. My rough estimate was this is an additional 1-2 GB of memory, assuming 1000-2000 frequency bins, and this could of course get worse for longer traces or large (upsampled) sampling rates. I wondered if you had checked this.

fschlueter: Based on how many entries is that?

sjoerd-bouma: > > Also, I'm pretty sure self is not hashed, so you may want to clear the cache on class initialization as well.
> 
> I can not follow.

The cache does not reset if you initialize a second instance of the channelGalacticNoiseAdder. It probably should? Although it seems you are also checking if the antenna_pattern has changed... so I guess in that case it might not matter. (As an aside, I didn't know you could hash custom classes inside lru_cache, but I guess you've tested this and it does not complain).

fschlueter: Yes I did. And `self.__hash__()` does work so self is hashed. A new instance gets a new hash. Of course the hash does not change when you modify the member variables. But this we keep track of our selfs.

fschlueter: So a vector with 2000 entries and dtype "complex128` gives me a size of 32kb. Taking into account 2 polarisations and 80 different directions from the sky give us 5 MB. That should be fine right? It is hard to see for me that this would increase by a factor of ~100

sjoerd-bouma: If you want 80 directions then why do you make the cache size 32000 instead of 80? I really don't understand where our communication is breaking down at the moment...

fschlueter: I do not think this cache size is allocated when not used. It is only the limit of what it is allowed to allocate. A maximum of 2 GB (which is probably never reached) seems absolutely acceptable for me. I set this number while not over thinking it. Feel free to change it if that is what bothers you.

sjoerd-bouma: I would change it to e.g. 1024. If the intention is for it to use 2 GB then that's a feature that we can discuss. If that's not the intention I'd rather not accidentally use 2 GB of memory (e.g. in a 'realistic' simulation of a detector at different times with different antenna orientations, I see no reason why this wouldn't fill the cache).

fschlueter: Well if you save it is realistic to fill the cache up to 2GB I would leave it as it. I think it is more problematic that people do not realise that they fill up the cache and thus suffer from reduced performance as that 2 GB are becoming a problem. (Changing the station time should hardly effect the cache).

sjoerd-bouma: Changing station time affects the cache - you will ask for different local zeniths and azimuths because the local coordinate system changes as a function of time. In this case you will be asking for new vector effective lengths constantly while storing (unused) vector effective lengths in the cache. This is just a waste of memory for - as far as I can tell - no reason.

You could say "well, just don't use the caching in that case". Or you could just reduce the cache size so there's no downside to using it. I would advocate the latter.

fschlueter: No, I would say if this is actually the case we should change what and keep the zenith and azimuth angle in the local  coordinate system fix and only change the sky temperature for the new position in the sky. This will be much more efficient.

fschlueter: Or would that just move the problem back to the interpolation of the sky map?

fschlueter: Okay sjoerd. YOU WIN.

fschlueter:
so?

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -160,6 +172,12 @@ def begin(
             self.__radio_sky = healpy.pixelfunc.ud_grade(self.__radio_sky, self.__n_side)
             self.__noise_temperatures[i_freq] = self.__radio_sky
 
+
+    @functools.lru_cache(maxsize=1024)
+    def get_cached_antenna_response(self, ant_pattern, zen, azi, *ant_orient):
```
sjoerd-bouma, line 177: Please make this a private function (or add a docstring).

sjoerd-bouma (commented):
One more comment. Other than that this looks like it'll come in handy : )

sjoerd-bouma:
Actually one other comment that is unrelated to this PR - I changed the random number generation to use the same numpy BitGenerator interface as elsewhere in our code in #740.  However, even though the galactic noise adder test is seeded, it reports different resultant Vrms every time... maybe you have some idea why this is happening?

Not super important (at least, while we're not interested in the combination of coherent noise and simulating non-trigger channels separately), but puzzling.

fschlueter:
> Actually one other comment that is unrelated to this PR - I changed the random number generation to use the same numpy BitGenerator interface as elsewhere in our code in #740. However, even though the galactic noise adder test is seeded, it reports different resultant Vrms every time... maybe you have some idea why this is happening?
> 
> Not super important (at least, while we're not interested in the combination of coherent noise and simulating non-trigger channels separately), but puzzling.

To test that the cache is doing what it is suppose to do, I created two nur files with and without the caching (but fixed seed) and run a `diff` in the terminal on them. I would have expected to get the promt `Differences in binary files` if the noise changed?

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: Done (both actually)

sjoerd-bouma (approved):

sjoerd-bouma:
> > Actually one other comment that is unrelated to this PR - I changed the random number generation to use the same numpy BitGenerator interface as elsewhere in our code in #740. However, even though the galactic noise adder test is seeded, it reports different resultant Vrms every time... maybe you have some idea why this is happening?
> > Not super important (at least, while we're not interested in the combination of coherent noise and simulating non-trigger channels separately), but puzzling.
> 
> To test that the cache is doing what it is suppose to do, I created two nur files with and without the caching (but fixed seed) and run a `diff` in the terminal on them. I would have expected to get the promt `Differences in binary files` if the noise changed?

This is even more mysterious... I'm referring to the `NuRadioReco/examples/StandAloneScripts/B01galacticnoiseadder.py`. You can check that its output is different from run to run in the tests, or run it twice yourself, and see that the output changes despite the fixed seed. So how you managed to get the same result twice in a .nur file I don't know!

PR #825, opened 2025-02-06
Refactor ADC module

fschlueter:
I thought before I actually start refactoring the functionality of this module I make a PR for cosmetic refactoring to keep the diff lower in the following PR.


fschlueter:
We can also merge these changes with PR #826 ... Maybe someone new to NuRadio can review this since it is only cosmetic changes? @avijai2000 maybe?

NuRadioReco/modules/analogToDigitalConverter.py
```diff
@@ -237,65 +222,51 @@ def get_digital_trace(self, station, det, channel,
         """
 
         station_id = station.get_id()
-        if channel_id is None or not trigger_adc:
-            channel_id = channel.get_id()
-
+        channel_id = channel.get_id()
         det_channel = det.get_channel(station_id, channel_id)
 
+        field_prefix = 'trigger_' if trigger_adc else ''
```
anelles, line 228: Does this need to be so compressed?

fschlueter: I agree to that in some extend. I think there is a reason to use some neat python syntax over making the code the most easiest to read. I am happy to change it here. (Although in a new PR I am moving this code to a private function and I do not think that newcomers have to read this function.)

fschlueter:
@anelles I incorporated your suggestion.

anelles (approved):
Lots of code beautification, so thank you for the attention to detail!

fschlueter:
> Lots of code beautification, so thank you for the attention to detail!

The functional changes come in #826 ;)

PR #826, opened 2025-02-07
Updating the analogToDigitalConverter and the RNOG/triggerBoardResponse modules

fschlueter:
This PR updates a couple of modules related to the accurate simulation of triggers running on digitised signals. Notably it changes the convention used to describe an ADC in the detector description. We deprecate a keyword to make the user aware that things have changed. I am posting here the error message which is returned:
```
                error_msg = (
                    f"The field \"{field_prefix}adc_reference_voltage\" is present in channel {channel_id}. "
                    "This field is deprecated. Please use the field adc_voltage_range instead. However, "
                    "be aware that the definition of the two fields are different. The \"adc_reference_voltage\" "
                    "referred to the maximum voltage V_max of the maximum ADC count 2^n - 1 with n being \"adc_nbits\", "
                    "assuming that the ADC operates from -V_max to V_max. As a consquence the voltage per ADC count "
                    "was calculated using: dV = adc_reference_voltage / (2^(n - 1) - 1). The \"adc_voltage_range\" "
                    "refers to the maximum voltage range V_range = V_max - V_min where V_max corresponds to 2^n - 1 "
                    "and V_min to 0. The voltage per ADC count was calculated using: dV = adc_voltage_range / (2^n - 1). "
                    "Example: If you want to simulate a ADC with a dynamic range from -1 V to 1 V, you should set have set "
                    "the adc_reference_voltage to 1 V. Now you have to set the adc_voltage_range to 2 V."
                )
```
What is currently problematic, I had to change the `adc_nbits` and `adc_noise_nbits` for two tests to make the pass.

This PR also changes the trigger condition in the "high-low" trigger from using `<` and `>` to using `<=` `>=`. This is relevant when using  discrete thresholds. Shall we change that in other modules too?


fschlueter:
Currently the analogToDigitalConverter allows to apply a `trigger_filter` to the trace. As argument a complex vector is passed which is multiplied to the channel spectra (to all the same filter). I would suggest to remove it. I do not think it is used by any of the examples/tests (I will check that) and it is better to apply the operation transparent in a user module (or use a hardware response incorperator module for this, or a triggerBoardResponse module).

fschlueter:
What currently causes issues is when the `adc_voltage_range` is calculated using the (noise) vrms*:
Old:
```
adc_ref_voltage = Vrms * (2 ** (adc_n_bits - 1) - 1) / (2 ** (adc_noise_n_bits - 1) - 1)
```
New:
```
adc_voltage_range = vrms * (2 ** adc_n_bits - 1) / (2 ** (adc_noise_n_bits - 1))
```
Changing the new to:
```
adc_voltage_range = vrms * (2 ** adc_n_bits - 1) / (2 ** (adc_noise_n_bits) - 1)
```
Will not help. Any suggestions?

* I think we actually also use this in simulations without noise in which is does not make much sense.

ryankrebs016:
I suggest defining the noise counts explicitly in the detector files rather than using the noise bits. So that line would look like `adc_voltage_range = vrms * (2 ** adc_n_bits - 1) / adc_noise_counts` . It breaks current detector files but it should be a quick fix.

NuRadioReco/modules/RNO_G/triggerBoardResponse.py
```diff
@@ -39,16 +35,17 @@ def begin(self, adc_input_range=2 * units.volt, clock_offset=0.0, adc_output="vo
...
         """
 
-        self._adc = analogToDigitalConverter()
+        self._adc = analogToDigitalConverter(log_level=self._log_level)
         self._clock_offset = clock_offset
         self._adc_output = adc_output
 
         # the fields that need to exist in the detector description for this module to work
-        self._mandatory_fields = ["trigger_adc_nbits", "trigger_adc_noise_nbits"]
+        self._mandatory_fields = ["trigger_adc_nbits"]
```
ryankrebs016, line 43: Should `"trigger_adc_voltage_range"` and `"trigger_adc_noise_count"` be present?

NuRadioReco/modules/analogToDigitalConverter.py
```diff
@@ -171,25 +163,101 @@ class analogToDigitalConverter:
...
-                          adc_output='voltage',
-                          trigger_filter=None,
-                          channel_id=None):
+    def __init__(self, log_level=logging.NOTSET):
+        logger.setLevel(log_level)
+        self._adc_types = {
+            'perfect_floor_comparator': perfect_floor_comparator,
+            'perfect_ceiling_comparator': perfect_ceiling_comparator}
+
+        self._mandatory_fields = ['adc_nbits', 'adc_sampling_frequency']
```
ryankrebs016, line 172: same question as the other mandatory fields

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -258,3 +255,9 @@ def end(self):
         dt = timedelta(seconds=self.__t)
         logger.info("total time used by this module is {}".format(dt))
         return dt
+
+def _get_threshold_channel(threshold, channel_id):
+    """ Returns channel specific threshold if threshold is a dict, otherwise returns threshold """
+    if isinstance(threshold, dict):
+        return threshold[channel_id]
```
ryankrebs016, line 262: It might be useful to throw an error if someone supplies a list or array instead of a dict or a singular value.

NuRadioReco/modules/RNO_G/triggerBoardResponse.py
fschlueter: No, because only either one of them is strictly necessary (depending on whether or not the user passes `vrms` to the module. However, before either of the one is used it is checked if it exists and an appropriate error emitted otherwise.

NuRadioReco/modules/trigger/highLowThreshold.py
fschlueter: Isn't that over engineering it a little bit? but I can do it.

fschlueter:
With the latest change we broke the CI. Instead of converting traces like this:
```
resolution = v_range / (2**n - 1)
adc = volt / resolution
```
We converting it now with:
```
resolution = (v_max - v_min) / (2**n - 1)  # same as before
adc = (volt - v_min) / resolution
# applying saturation
adc = adc + v_min / resolution
```
I am omitting the floors here! The problem is that `v_min / resolution` is not a round integer. This is because of the `-1` in `2**n - 1`.

fschlueter:
instead of `floor((volt - v_min) / resolution)` we could do `floor(volt / resolution) - floor(v_min / resolution)` (without testing it this should fix the issue). But I am not sure if that is the correct thing to do...

fschlueter:
This is the result when running the `Atmospheric Aeff` test 256 times. I have updated the reference values accordingly.
image

fschlueter:
I updated 3 tests. All used a phasedarray trigger and were affected by the changes in the ADC module.

fschlueter:
Speaking about these changes. Christian and I agreed that this is the correct thing to do. Previously the module assumed that the ADC is symmetric around the 0 which is is not in reality.

NuRadioReco/modules/trigger/highLowThreshold.py
fschlueter: Not sure what is wrong with the webinterface but it should be this now:
```
def _get_threshold_channel(threshold, channel_id):
    """ Returns channel specific threshold if threshold is a dict, otherwise returns threshold """
    if isinstance(threshold, dict):
        return threshold[channel_id]
    elif isinstance(threshold, (int, float)):
        return threshold
    else:
        raise TypeError(f"Threshold must be a int/float or dict, not {type(threshold)}")
```

NuRadioReco/modules/analogToDigitalConverter.py
```diff
@@ -19,91 +23,90 @@ def perfect_comparator(trace, adc_n_bits, adc_ref_voltage, mode='floor', output=
...
 
-    if (mode == 'floor'):
-        digital_trace = np.floor(trace / lsb_voltage).astype(int)
-    elif (mode == 'ceiling'):
-        digital_trace = np.ceil(trace / lsb_voltage).astype(int)
-    else:
-        raise ValueError('Choose floor or ceiing as modes for the comparator ADC')
+    assert mode_func in [np.floor, np.ceil], "Choose floor or ceiing as modes for the comparator ADC"
+
+    digital_trace = mode_func((trace - adc_voltage_range[0]) / lsb_voltage).astype(int)
```
ryankrebs016, line 59: Is the offset here functionally different than before or is this future proofing in case Vmin and Vmax aren't symmetric?

fschlueter: Yes, this change effected the results. I talked to Christian and Cosmin and we all agreed that this is closer to the correct implementation of an ADC as it was before (assuming the adc is perfectly symmetrical)

ryankrebs016:
nothing else other than the question above

fschlueter:
> nothing else other than the question above

Would you like to approve this PR than?

ryankrebs016 (approved):

PR #828, opened 2025-02-09
Update IC EHE limit

cozzyd:
This updates the IceCube EHE limit to the values from https://arxiv.org/abs/2502.01963 (data release at https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/JHK49D). 

The old limit is still available if you ask for it specifically.


anelles (approved):
Thanks, Cosmin. You were faster than I was. I had added this locally for our review, but happy to approve yours!

PR #830, opened 2025-02-14
generalization of triggered thermal noise generation 

cg-laser:
allow different noise levels, filters, and Vrms values per channel in noise generation

### DIFF
--- a/NuRadioReco/utilities/noise.py
+++ b/NuRadioReco/utilities/noise.py
@@ -236,17 +236,22 @@ class thermalNoiseGenerator():
                             number_of_triggers += 1
                             t_bins[iCh] = triggered_bins
                             trace_to_keep = trace if not self.keep_full_band else trace_copy
-                            if iCh == 0:
+                            if number_of_triggers == 1:
+                                print(f"Channel {iCh} triggered")
                                 n_traces[iCh] = np.roll(trace_to_keep, self.trigger_bin - np.argwhere(triggered_bins == True)[0])
                             else:
                                 tmp = np.random.randint(self.trigger_bin_low, self.trigger_bin)
+                                print(f"Channel {iCh} triggered, random trigger bin {tmp}")
                                 n_traces[iCh] = np.roll(trace_to_keep, tmp - np.argwhere(triggered_bins == True)[0])
+                if number_of_triggers == self.n_majority:
+                    # this additional break is needed because another channel might trigger
+                    # within the channel loop before the while statement is checked
+                    break
 
         traces = np.zeros((self.n_channels, self.n_samples))
         for iCh in range(self.n_channels):
             if(n_traces[iCh] is not None):
                 traces[iCh] = n_traces[iCh]
-                print(f"Channel {iCh} triggered")
             else:
                 spec = self.noise.bandlimited_noise(self.min_freq, self.max_freq, self.n_samples, self.sampling_rate,
                                                     self.amplitude[iCh], type=self.noise_type, time_domain=False)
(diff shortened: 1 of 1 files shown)


NuRadioReco/utilities/noise.py
```diff
@@ -185,44 +215,46 @@ def generate_noise(self):
...
+                        triggered_bins = get_high_low_triggers(trace, self.threshold[iCh], -self.threshold[iCh], self.time_coincidence, self.dt)
+                        if np.any(triggered_bins):
+                            number_of_triggers += 1
+                            t_bins[iCh] = triggered_bins
+                            trace_to_keep = trace if not self.keep_full_band else trace_copy
+                            if iCh == 0:
+                                n_traces[iCh] = np.roll(trace_to_keep, self.trigger_bin - np.argwhere(triggered_bins == True)[0])
+                            else:
+                                tmp = np.random.randint(self.trigger_bin_low, self.trigger_bin)
+                                n_traces[iCh] = np.roll(trace_to_keep, tmp - np.argwhere(triggered_bins == True)[0])
```
fschlueter, line 243: Did you check if the code works? Because unlike with the old code it is not given that the first channel which triggers is `iCh ==0`

cg-laser:
good point, I fixed it. 
However, I'm not sure how to test that it does exactly the same because the random sequence changes slightly. So I can't compare the previous version to the current one on an event-by-event basis.

fschlueter:
> good point, I fixed it. However, I'm not sure how to test that it does exactly the same because the random sequence changes slightly. So I can't compare the previous version to the current one on an event-by-event basis.

I think that is okay

NuRadioReco/utilities/noise.py
```diff
@@ -236,17 +236,22 @@ def generate_noise(self):
...
+                                print(f"Channel {iCh} triggered")
                                 n_traces[iCh] = np.roll(trace_to_keep, self.trigger_bin - np.argwhere(triggered_bins == True)[0])
                             else:
                                 tmp = np.random.randint(self.trigger_bin_low, self.trigger_bin)
+                                print(f"Channel {iCh} triggered, random trigger bin {tmp}")
                                 n_traces[iCh] = np.roll(trace_to_keep, tmp - np.argwhere(triggered_bins == True)[0])
+                if number_of_triggers == self.n_majority:
+                    # this additional break is needed because another channel might trigger
+                    # within the channel loop before the while statement is checked
+                    break
```
fschlueter, line 249: should we just change the condition of the while to `while True:`?

cg-laser: I don't think this would be sufficient. The `break` only breaks the for loop. The while loop still needs to be terminated.

fschlueter: ah good point.

fschlueter (approved):

NuRadioReco/utilities/noise.py
thoglu: I think we should add a quick check if the filter is ndarray, that its shape is len(1), and otherwise that its type is a list (list of arrays) .. in the current code, if we would pass all filters as a combined 2d array it would not complain but fail in the noise  generation

thoglu:
I think we should add a quick check if the filter is ndarray, that its shape is len(1), and otherwise that its type is a list (list of arrays) .. in the current code, if we would pass all filters as a combined 2d array it would not complain but fail in the noise generation

NuRadioReco/utilities/noise.py
thoglu: oops I think I resolved something and it removed parts of the previous chat.. not sure about it

thoglu (approved):

PR #832, opened 2025-02-18
Deprecate Python 3.6

sjoerd-bouma:
This PR 'officially' deprecates Python 3.6 support by removing this version from the CI (it would be somewhat of a hassle to continue testing it, as Ubuntu 20.04 will not be available anymore, and python 3.6 is not available by default on newer Ubuntu versions). Due to different numpy versions, several tests were failing already because of different random number generators anyway.

Note that most features will continue to work on Python 3.6 (for now). The python version requirement can be overruled by using 
```
pip install --ignore-requires-python nuradiomc
```

I've also taken the liberty of cleaning up the changelog, added the missing addition of birefringence, and bumped the `-dev` version number to correspond to what we claim it is in the changelog.


fschlueter (approved):

fschlueter:
Can this be merged?

PR #834, opened 2025-02-19
Add Greenland temperature profile, example to simulate effective temperature at antenna.

fschlueter:
As the title suggests, this PR does two things:

- It adds the temperature profile of the GRIP measurement (Greenland near Summit Station) to `NuRadioMC.utilities.attenuation.py` 
- Adds an example in NuRadioMC to simulate the eff. temperature at the antenna. (Code originally by @shallmann)


fschlueter:
@shallmann would you mind reviewing this?

NuRadioMC/examples/simulate_effective_ice_temperature/ice_integrator.py
```diff
@@ -0,0 +1,258 @@
...
+    depth_interp = interp1d(distance_raytracing, depth_raytracing, fill_value="extrapolate")
+    power_interp = interp1d(distance_raytracing, reflected_power, fill_value="extrapolate")
+
+    distance = np.linspace(0, 30000, 300000) * units.m
+    d_distance = distance[1] - distance[0]
+
+    depth = depth_interp(distance)
+    ref = power_interp(distance)
+    # manually set cap stuff
+    ref[depth < z_min] = 0
```
shallmann, line 190: did you test if that has no influence on the obtained temperature at the antenna? (i.e. - quickest solution - by commenting this line and setting the bottom to something like -6k meters ... ((or anything >>3km ... that way you shouldn't have to worry about the shorter attenuation length in rock...)))

NuRadioMC/examples/simulate_effective_ice_temperature/ice_integrator.py
```diff
@@ -0,0 +1,258 @@
...
+    ax.set_xlabel("x / m")
+    ax.set_ylabel("z / m")
+    ax.set_xlim(-10, 1000)
+    fig.tight_layout()
+    plt.savefig("plot.pdf")
+
+
+if __name__ == "__main__":
+
+    get_eff_temperature(fname="eff_temperature_grip_GL3_400MHz.json")
```
shallmann, line 258: you might want to consider setting the keyword argument `plot=True` for the main

shallmann (commented):
Thanks Felix for the effort of adding this. I don't recall: is the contribution from the rock bottom negligible? If you have verified and it does not have an influence, fine! Otherwise we might want to consider extending the calculation into the rock.

NuRadioMC/examples/simulate_effective_ice_temperature/ice_integrator.py
fschlueter: Okay I removed the capping and plotted the `attenuation_factor` for downwards going rays. As you can see attenuation factor for the most vertical rays at the rock bottom is < 0.1 (i.e., a factor of 10+ attenuated. But it is not equal 0.
image

fschlueter: Linear color map but everything below 0.01 is cut out.
image

fschlueter: Here is the limit 0.03. So < 3% comes from the rock bottom (for the given attenuation model).
image

NuRadioMC/examples/simulate_effective_ice_temperature/ice_integrator.py
fschlueter: done

NuRadioMC/examples/simulate_effective_ice_temperature/ice_integrator.py
fschlueter: So what is our conclusion. Should we just pretend the ice is infinitive deep and (linearly?) extrapolating the temperature/attenuation profiles since the contributions do not matter anyhow?

fschlueter: Also my statement is incorrect. Not less than 3% come from the bedrock but 3% from the bedrock arrive at the antenna.

fschlueter: Okay, when I plot what I believe is a good indicate for how much the temp from a certain volume element contributes, than this is very similar to the plot above (again cut at 0.03). The color axis shows this:
```
        eff_temp = reflection_coef * att_factor * temp_env / l_att * d_distance

        eff_temp = 1 - np.cumsum(eff_temp) / np.sum(eff_temp)

        eff_temp[eff_temp < 0.03] = np.nan
```
I am actually surprised that we still have 3 percent contribution from the bedrock. I believe this is actually wrong. Hmmm. So probably we have to change things there.
image

shallmann: Hi Felix! Thanks for this very comprehensive follow-up work. I just don't understand: would you expect much less than 3%? Back of the envelope we should be something between 3-4 absorption lengths, i.e. I do expect e**-3...4 = 2 to 5 % of the temperature should come from there, no?

NuRadioMC/examples/simulate_effective_ice_temperature/ice_integrator.py
shallmann: thanks

NuRadioMC/examples/simulate_effective_ice_temperature/ice_integrator.py
shallmann: > So what is our conclusion. Should we just pretend the ice is infinitive deep and (linearly?) extrapolating the temperature/attenuation profiles since the contributions do not matter anyhow?

I would guess that the attenuation length of rock for radio is very short. I would therefore keep the bottom temperature fixed at the bottom instead of extrapolating. You can continue for a short bit with high attenuation (=high emissivity) or just leave the attenuation constant and continue more (that's not what is physically happening but it should give you the same result if I am not mistaken)

fschlueter: I was playing with something like this around yesterday evening but did not understand the results. Will try to revisit it now. So I am happy with receiving 2-5% of the radiation from the bedrock, but I thought this 2-5% should mean nothing for the eff. Temperature at the antenna.

shallmann: Ok, well... 2-5% means 5 to 13 degrees more in temperature, or am I mistakenb? ... so it is not huge, but nonzero

fschlueter:
Okay for documentation I am posting here some plots. After discussing with @shallmann  we decided to stop the integration at `z_min` (this is technically achieved by setting the attenuation factor to 0 beyond that) but add an ideal black body radiator with the temperature of the bedrock at the end. The following two plots show the final result where the eff temperature comes from and the difference between 3 approaches to calculate the eff. temperature.
image
image

shallmann (approved):

PR #838, opened 2025-02-20
Store shower information in the NUR header

MijnheerD:
Issue being addressed:

Currently no information about the showers (such as the SimShower for simulations) is stored in the header. This makes it cumbersome to check the basic properties of the event for simulations, as typically parameters such as the energy are only stored in the Shower object.

Did you remember to: 
- [ ] include doc strings in all commits ?
- [ ] edit the changelog.txt file describing your changes?


fschlueter:
@MijnheerD are you planning to continue working on this?

galaxyvita:
I don't mind doing this, if I understand what needs to happen.

fschlueter:
> I don't mind doing this, if I understand what needs to happen.

have a look at the existing code and how we copy all station parameters into a dict which is stored in the header. It should be rather straight forward to do the same for all (sim)showers

galaxyvita:
Yes, but I looked for existing code which does the same and couldn't find anything. So I'm a buit confused.

galaxyvita:
Oke so I think I updated the function in a way that would work, but I don't have an example to run it on.

fschlueter:
left one little comment, otherwise looks fine

fschlueter:
@galaxyvita are you still interested in working on that?

galaxyvita:
If i remember correctly i added the functionality to the _parse_event_header in NuRadioRecoio.py.
But if more needs to be done i will do it for sure.

fschlueter:
oh sorry, that slipped my attention.

NuRadioReco/modules/io/eventWriter.py
```diff
@@ -20,6 +20,15 @@ def get_header(evt):
         if station.has_sim_station():
             header['stations'][station.get_id()]['sim_station'] = {}
             header['stations'][station.get_id()]['sim_station'] = station.get_sim_station().get_parameters().copy()
+
+    for shower in evt.get_showers():
+        sh_id = shower.get_id()
+        header['showers'][sh_id] = shower.get_parameters().copy()
+        if hasattr(shower, "has_sim_shower") and shower.has_sim_shower():
```
fschlueter: i am confused, only the event call has the function `has_sim_shower`?

fschlueter:
DId you test the code? (Writing and reading a couple of files?) I can see that you took existing code for the event headers, copied it for the showers and adapted it. I am not 100% sure that is the correct way of doing this. Would you mind if we discuss this in an upcoming NuRadio call?

galaxyvita:
I remember now i never finished testing it. So i will do that. Sure i don’t know when the calls are but it might be good to inform people about this and the “nur to hdf5” stuff.

galaxyvita:
Oke i found that i updated the “read_lofar_data.ipynb” in examples and printed some of the headers there.

fschlueter:
@galaxyvita can you have a look at the failing unit test

galaxyvita:
Now the error in not in my code, it is from download_from_dataserver.

NuRadioReco/modules/io/eventWriter.py
fschlueter: @galaxyvita this if condition will be always 0, hence you will never store sim_shower data. My suggestion is that you include in the header also a key `sim_showers` and loop over all sim_showers.

NuRadioReco/modules/io/NuRadioRecoio.py
```diff
@@ -171,26 +171,25 @@ def close_files(self):
 
     def get_filenames(self):
         return self._filenames
-
+
```
fschlueter: meh

NuRadioReco/modules/io/NuRadioRecoio.py
```diff
@@ -171,26 +171,25 @@ def close_files(self):
...
-
+                        
     def _parse_event_header(self, evt_header):
         from NuRadioReco.framework.parameters import stationParameters as stnp
+        from NuRadioReco.framework.parameters import showerParameters as shwp
+        from NuRadioReco.utilities import io_utilities
+        
         self.__event_ids.append(evt_header['event_id'])
-
+
```
fschlueter: meh ... also in the following lines

NuRadioReco/modules/io/NuRadioRecoio.py
```diff
@@ -201,10 +200,29 @@ def _parse_event_header(self, evt_header):
...
-
                         self.__event_headers[station_id][key].append(station_time)
                     else:
                         self.__event_headers[station_id][key].append(value)
+                else:
+                    self.__event_headers[station_id]['sim_station'] = value
+
+        if 'showers' in evt_header:
+            if not hasattr(self, '_event_showers'):
+                self._event_showers = {}
```
fschlueter: I have not digested the code fully, but is there a reason why you are not initializing (and importantly resetting) `self._event_showers` similar to `__event_headers` in the function `openFile`?

NuRadioReco/modules/io/NuRadioRecoio.py
```diff
@@ -201,10 +200,29 @@ def _parse_event_header(self, evt_header):
...
+
+            for shower_id, shower in evt_header['showers'].items():
+                if shower_id not in self._event_showers:
+                    self._event_showers[shower_id] = {}
+    
+                for key, value in shower.items():
+                    if key not in self._event_showers[shower_id]:
+                        self._event_showers[shower_id][key] = []
+    
+                    self._event_showers[shower_id][key].append(value)
```
fschlueter: I think this motivated by the structure in the same code for the stations. I do not think that makes sense here (although I am happy to discuss that). I think the idea for stations is to store the information per station accumulating information over all events. For showers, I think it would make more sense to store them per event. Does that make sense to you?

fschlueter (changes requested):
Hey @galaxyvita, unfortunately I found a couple of things we need to fix

NuRadioReco/modules/io/NuRadioRecoio.py
galaxyvita: Because I did not think about that. This would probably keep the previous files in memory.

NuRadioReco/modules/io/NuRadioRecoio.py
galaxyvita: This does make sense.

fschlueter (approved):

PR #688, opened 2024-06-05
Refactoring coreas to include interpolation

cg-laser:
Major refactoring of the coreas interface to read the electric field from coreas hdf5 files.
A new class readCoREASDetector.py can be used to read an array and get the interpolated position.

This is largely a copy of #685 which is an undefined state because of an accidental force push. Please never force-push unless you know what you are doing.

PR #839, opened 2025-02-21
Make numpy arrays contiguous before storing them

### LINKED ISSUE #836: eventWriter broken with ElectricFields
**Describe the issue**
After the changes in #831 , the eventWriter fails to write out ElectricFields. The issue seems to be that in line 43 of `_fastnumpyio.py` a sum is done between the `array.data` and some other information. However, these are of two different types which are incompatible. Namely,
```python
>>> import numpy as np
>>> ar = np.random.uniform(size=(3,200))
>>> type(ar)
<class 'memoryview'>
```
The other elements on line 43 are of type `bytes`, which is incompatible. Hence the error:
```python
TypeError: can't concat memoryview to bytes
```

My instinct tells my to add a simple cast and `+ bytes(array.data)` at the end of line 43, but I am not sure if this messes with other things.

**Expected behavior**
Well, it should not crash :)

**Additional context**
This issue came up in the context of the refactoring of the CoREAS modules, in PR #688 . There is one weird thing about this issue, and that is that I cannot reproduce this error by manually creating an ElectricField and adding it to some Event. Somehow when saving this Event , it saves like normal. So it seems like sometimes the type of `array.data` is bytes already?

sjoerd-bouma:
Fixes #836 

@MijnheerD found that the _fastnumpyio implementation to store numpy arrays only works for contiguous arrays, and raises an error otherwise. So I've added `np.ascontiguousarray` to the _fastnumpyio function, which copies the array data to contiguous ordering if this is not already the case.

### DIFF
--- a/NuRadioReco/utilities/_fastnumpyio.py
+++ b/NuRadioReco/utilities/_fastnumpyio.py
@@ -39,8 +39,32 @@ def save(file, array):
     file.write(array.data)
 
 def pack(array):
-    size=len(array.shape)
-    return bytes(array.dtype.byteorder.replace('=','<' if sys.byteorder == 'little' else '>')+array.dtype.kind,'utf-8')+array.dtype.itemsize.to_bytes(1,byteorder='little')+struct.pack(f'<B{size}I',size,*array.shape)+array.data
+    """
+    Convert a numpy array to a storable bytes object
+
+    Parameters
+    ----------
+    array : np.ndarray
+        The numpy array to store.
+
+    Returns
+    -------
+    output_bytes : bytes
+        The bytes
+
+    """
+    # Only contiguous arrays can be stored directly as bytes
+    contiguous_array = np.ascontiguousarray(array)
+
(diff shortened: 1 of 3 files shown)


MijnheerD (approved):

MijnheerD:
I can confirm this solves the issues we encountered on the `refactor_coreas_v3` branch!

fschlueter:
@sjoerd-bouma into master?

sjoerd-bouma:
> @sjoerd-bouma into master?

Yes, this fixes a bug on master, so I would merge into master first and then into develop; I guess I could branch of this and simultaneously merge into develop rather than sequentially?

fschlueter:
yes

fschlueter:
This will be merged into develop via #688

MijnheerD (approved):

MijnheerD:
This change has been merged into `develop` now, so shall we also add it to `master` then?

PR #841, opened 2025-02-21
Add example script to simulate the noise trigger rate for RNOG

fschlueter:
This PR adds an example to simulate noise triggers for RNOG. This is not the most efficient way to simulate triggers but the only way to simulate trigger rates. It also allows for a full detector simulation with individual channel responses.

PR #842, opened 2025-02-24
add readout for digitizer config

jakobhenrichs:
We add more digitizer settings to the database. In this PR the code is adjusted so that the settings can be queried from the database.


fschlueter:
Some of the information (in particular sampling rate) is passed with the data and in principle not a unchangable property of the detector. However, for simulations we need the detector to be able to return it.

fschlueter (dismissed):

fschlueter:
It seems to work well on #841 !

fschlueter (approved):

PR #843, opened 2025-02-24
use pre-commit to perform checks instead of custom scripts

sjoerd-bouma:
Some updates to the developer workflow:

- Rather than modifying `$PYTHONPATH`, the documentation now suggests using `pip install -e .[dev]` instead to install in [editable mode](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs) instead. This simplifies things (not having to modify $PYTHONPATH) and means we use the recommended, standard pip interface rather than a hand-written `install_dev.py` script. Thanks @clark2668 for suggesting this!
- The no-large-files-added check has now also been replaced by using the pre-commit hook interface from https://pre-commit.com. **This has to be installed by running** `pre-commit install` **initially**. The advantage is this allows us/users to simply add more checks on the user end; for now I have only added the 'important error check' from flake8 which fails on incorrect syntax (the first check in our CI test suite), and an error if users try to commit `.ipynb` files with output in them (this will only become relevant for a future PR that will add some interactive examples). However, with one line we could consider also adding **trailing whitespace fixes**, **end-of-file-fixers**, and probably several others if we like. Open for discussion!


documentation/source/Introduction/pages/contributing.rst
```diff
@@ -9,6 +9,19 @@ If you intend to contribute significantly to NuRadioMC, please contact
 `@anelles <https://github.com/anelles>`_ and `@cg-laser <https://github.com/cg-laser>`_ on GitHub so that we are informed about ongoing
 activities. Both will also be able to provide commit access to the repository.
 
+Installing NuRadioMC for developers
+-----------------------------------
```
fschlueter, line 13: Does that mean that this is optional?

pyproject.toml
```diff
@@ -54,13 +54,17 @@ MCEq = {version = "*", optional = true}
 crflux = {version = "*", optional = true}
 mattak = {git = "https://github.com/RNO-G/mattak", optional = true}
 rnog-runtable = {git = "ssh://git@github.com/RNO-G/rnog-runtable.git", optional = true}
+pre-commit = {version = "*", optional = true}
```
fschlueter, line 57: can't we make this a required dependency?

documentation/source/Introduction/pages/installation.rst
```diff
@@ -34,42 +34,34 @@ can be installed by appending ``[option]``, i.e. ``pip install NuRadioMC[option]
...
-To use the development version of NuRadioMC, its installation path needs to be included in the user PYTHONPATH.
-If the installer does not succesfully do this, please manually add
-
-.. code-block:: Bash
+.. code-block:: bash
 
-  export PYTHONPATH=/path/to/NuRadioMC:$PYTHONPATH
+  cd NuRadioMC/
+  pip install -e .[dev]
+  pre-commit install
```
fschlueter, line 59: can we configure the pyproject.toml such that `pre-commit install` is always executed?

documentation/source/Introduction/pages/contributing.rst
sjoerd-bouma: If you mean installing the pre-commit hook, the answer is yes; it is (for good reason) impossible to automatically install hooks which could in principle execute arbitrary code every time you invoke `git`. I have tried to word it as strongly as possible as I don't see any downsides and many upsides to everyone installing this hook, but there is (at least on github) no way to enforce this.

documentation/source/Introduction/pages/installation.rst
sjoerd-bouma: No, see comment above.

pyproject.toml
sjoerd-bouma: It's not a required dependency for users, only for developers.

fschlueter:
@sjoerd-bouma thank you very much! This `pre-commit` library seems very (very) useful! I left a few comments inline which are all alined to the same theme - can we ensure that everyone uses `per-commit`-hooks?

pyproject.toml
fschlueter: what does that mean? Is it a problem if user install it?

sjoerd-bouma: It's not a problem but also doesn't do anything (in fact, it does nothing until you run `pre-commit install` anyway), so I don't see the benefit of adding this as a dependency for everyone.

fschlueter (approved):

documentation/source/Introduction/pages/contributing.rst
sjoerd-bouma: To your other question: pre-commit install installs a git hook, which will be invoked every time you use `git commit`. I checked and if you do this from an environment without `pre-commit` installed the hook (and therefore the commit) will fail.

fschlueter: but only for this repo right?

fschlueter: so please go ahead and merge this thing

clark2668:
Thanks so much Sjoerd! Very glad to see this change :)

Issue #530, opened 2023-04-25
RNO-G IGLU amplifier includes cable delay


sjoerd-bouma:
@christophwelling 
The measurement currently used for the RNO-G IGLU amplifier response was taken with a 50m fibre cable, and currently includes the delay due to this cable. We should probably remove that...

Plot is of the amplifier response of the IGLU (channel 0) and SURFACE (channel 12) amplifiers in the time domain (i.e. as applied to a delta pulse at t=0).
![image](https://user-images.githubusercontent.com/78698543/234321732-bfb06184-cb51-4538-8be2-b6d2c94fb9f5.png)

fschlueter:
What you get with the DB is (should be) the total delay of the system, not only the fiber cable but also the amplifiers and jumper cables, and so on. However, I wonder if that can make such a big difference of ~250 ns. I am worried that for the deep channels we still have a reference cable in the total response which we are not able to subtract. I talked about that with @philippwindischhofer. This additional cable should be not a big issue when only working with deep channel but when comparing deep with shallow it will...

fschlueter:
what is the value you get for channel 13 with the new dB? I assume it is much close to 46ns and does not add ~250ns

sjoerd-bouma:
@fschlueter Yes, this is most likely what's happening. The DB values for the shallow channels are more realistic.

@shallmann If you mean the exact delay for the 50 m cable, I don't have a value for that - I would have hoped that somewhere in the database the response for this cable could be found, but it sounds like that might not be the case.

philippwindischhofer:
Hi all!

@sjoerd-bouma Indeed we don't currently have the capabilities to measure absolute fiber delays with adequate precision. That's why (even in principle) we're not able to provide a delay measurement for the 50m fiber.

@shallmann I don't know what the context of your question is (so what I'm going to say is perhaps not going to be extremely relevant), but when I compared cable delay differences for deep-deep channel pairs, the DB and the detector description json matched pretty well. (See example plot attached.)

[CH0_CH3.pdf](https://github.com/user-attachments/files/17144702/CH0_CH3.pdf)

shallmann:
[...]

PR #844, opened 2025-02-25
Update rnog simulation example

fschlueter:
This PR:
- updates the RNO-G trigger simulation example
- modifies the efieldToVoltageConverter to provide longer efield traces.
- Improves error and warning messages in the channelReadoutWindowCutter
- Makes NuRadioMC simulations a bit more quite.
- Minor cosmetic updates. 

The commit history might be a bit dirty because this PR is based of #841 but I removed one file which #841 is adding.

### DIFF
--- a/NuRadioMC/examples/08_RNO_G_trigger_simulation/simulate.py
+++ b/NuRadioMC/examples/08_RNO_G_trigger_simulation/simulate.py
@@ -84,8 +84,8 @@ class mySimulation(simulation.simulation):
 
     def __init__(self, *args, trigger_channel_noise_vrms=None, **kwargs):
         # this module is needed in super().__init__ to calculate the vrms
-        self.rnogHarwareResponse = hardwareResponseIncorporator.hardwareResponseIncorporator()
-        self.rnogHarwareResponse.begin(trigger_channels=kwargs['trigger_channels'])
+        self.rnogHardwareResponse = hardwareResponseIncorporator.hardwareResponseIncorporator()
+        self.rnogHardwareResponse.begin(trigger_channels=kwargs['trigger_channels'])
 
         super().__init__(*args, **kwargs)
         self.logger = logger
@@ -117,7 +117,7 @@ class mySimulation(simulation.simulation):
 
     def _detector_simulation_filter_amp(self, evt, station, det):
         # apply the amplifiers and filters to get to RADIANT-level
-        self.rnogHarwareResponse.run(evt, station, det, sim_to_data=True)
+        self.rnogHardwareResponse.run(evt, station, det, sim_to_data=True)
 
     def _detector_simulation_trigger(self, evt, station, det):
 
--- a/NuRadioReco/modules/efieldToVoltageConverter.py
+++ b/NuRadioReco/modules/efieldToVoltageConverter.py
@@ -33,7 +33,7 @@ class efieldToVoltageConverter():
         self.__pre_pulse_time = None
         self.__post_pulse_time = None
         self.__antenna_provider = None
-        
+
         self.logger = logging.getLogger('NuRadioReco.efieldToVoltageConverter')
         self.logger.setLevel(log_level)
         self.begin()
@@ -129,7 +129,7 @@ class efieldToVoltageConverter():
                 cab_delay = det.get_cable_delay(sim_station_id, channel_id)
                 t0 = electric_field.get_trace_start_time() + cab_delay
 
-                # if we have a cosmic ray event, the different signal travel time to the antennas has to be taken into account 
+                # if we have a cosmic ray event, the different signal travel time to the antennas has to be taken into account
                 if sim_station.is_cosmic_ray():
                     travel_time_shift = calculate_time_shift_for_cosmic_ray(det, sim_station, electric_field, channel_id)
                     t0 += travel_time_shift
@@ -146,7 +146,7 @@ class efieldToVoltageConverter():
         times_max = np.max(times_max)
 
         # Determine the maximum length of the "readout window"
-        max_channel_trace_lenght = np.max([
+        max_channel_trace_length = np.max([
             det.get_number_of_samples(station.get_id(), channel_id) / det.get_sampling_frequency(station.get_id(), channel_id)
             for channel_id in channel_ids])
 
@@ -155,7 +155,7 @@ class efieldToVoltageConverter():
         times_max += self.__post_pulse_time
 
         # Add post_pulse_time as long as we reach the minimum required trace length
-        while times_max - times_min < max_channel_trace_lenght:
+        while times_max - times_min < max_channel_trace_length:
             times_max += self.__post_pulse_time
 
         # assumes that all electric fields have the same sampling rate
@@ -231,7 +231,7 @@ class efieldToVoltageConverter():
                         start_bin = 0
 
                     new_trace[:, start_bin:stop_bin] = tr
-                
+
                 trace_object = NuRadioReco.framework.base_trace.BaseTrace()
                 trace_object.set_trace(new_trace, 1. / time_resolution)
 


fschlueter:
Remove the `noise.py` script from this branch. Should be added with #841

fschlueter:
@cg-laser maybe?

NuRadioReco/utilities/signal_processing.py
```diff
@@ -82,3 +82,43 @@ def add_cable_delay(station, det, sim_to_data=None, trigger=False, logger=None):
...
+        The vrms of the channel
+    """
+    if bandwidth is None and response is None:
+        raise ValueError("Please specify bandwidth or response")
+
+    if response is None:
+        vrms_per_channel = (temperature * resistance * constants.k *
+                            (bandwidth[1] - bandwidth[0]) / units.Hz) ** 0.5
+    else:
+        freqs = np.arange(0, 2500, 0.1) * units.MHz
```
cg-laser, line 117: can we make this an optional argument? a hardcoded max value is not idea.

NuRadioReco/utilities/signal_processing.py
```diff
@@ -82,3 +82,43 @@ def add_cable_delay(station, det, sim_to_data=None, trigger=False, logger=None):
...
+
+    For details see https://en.wikipedia.org/wiki/Johnson%E2%80%93Nyquist_noise
+    (sec. "Maximum transfer of noise power") or our wiki
+    https://nu-radio.github.io/NuRadioMC/NuRadioMC/pages/HDF5_structure.html
+
+    Parameters
+    ----------
+    temperature: float
+        The noise temperature of the channel in Kelvin
+    bandwidth: tuple (list of 2 floats) (default: None)
```
cg-laser, line 98: can we rather have this be the "effective_bandwidth"? I.e., the integral over the squared S21? 
In the most simple case it would just be f_max - f_min. 
Then this function is directly usable from simulation.py

cg-laser (changes requested):
good idea to put `calculate_vrms_from_temperature` into its own utility function. 
Shouldn't we update the simulation.py code to use this function?

fschlueter:
I updated simuation.py to use the new utility.

fschlueter:
@cg-laser any further comments? @EnriqueHuesca reported that he receives warning in the (now already deprecated) triggerTimeAdjuster which would probably also be issued by the newer channelReadoutWindowCutter but might be fixed by setting the `post_signal_time` to 400ns.

fschlueter:
ping @sjoerd-bouma

fschlueter:
Can we get that reviewed?

fschlueter:
Maybe @anelles

fschlueter:
@cg-laser can you please approve this PR (since you request changes I need your approval)!

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -32,17 +32,15 @@ def __init__(self, log_level=logging.NOTSET):
...
         self.logger.setLevel(log_level)
         self.begin()
 
 
-    def begin(self, debug=False, uncertainty=None,
-              time_resolution=None,
-              pre_pulse_time=200 * units.ns,
-              post_pulse_time=200 * units.ns,
+    def begin(self, debug=False, uncertainty=None, time_resolution=None,
+              pre_pulse_time=200 * units.ns, post_pulse_time=400 * units.ns,
```
sjoerd-bouma, line 43: I haven't checked, but is the increase in post_pulse_time actually necessary, or is it due to #530 ?

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -126,6 +124,7 @@ def run(self, evt, station, det, channel_ids=None):
             channel_ids = det.get_channel_ids(sim_station_id)
 
         for channel_id in channel_ids:
+
```
sjoerd-bouma, line 127: I think you and I perceive indentations differently :upside_down_face: For me the indentation is sufficient to indicate a break, so I don't see the need for an empty newline here. But this is just a comment, not a request for changes.

NuRadioReco/modules/efieldToVoltageConverter.py
sjoerd-bouma: maybe (as you mentioned he reported this) @EnriqueHuesca can answer this

fschlueter: It is necessary - I tested that myself using some neutrino simulation. It might be that it became necessary when we changed `simulation.py` in the NuRadioMC 3.0 branch: Since we only pass trigger channels to this module, the variation in the pulse times at the antennas which affect the trace length before padding are smaller than when we passed a full rnog station with channels on helper strings to this module

NuRadioReco/modules/efieldToVoltageConverter.py
fschlueter: Just a comment so that you maybe understand me better - often I am not refactoring code because I feel make this my explicit target. I read code in order to understand it and in order to understand it it is enormously helpful to me to add empty lines in code. Its like reading a book which does not has any paragraphs but just one big block of text is really tiring to read. But with code I need more often a break to appreciate the different steps.

sjoerd-bouma: Yes, and in general I agree that your changes help to make the code more readable! I would just never add an empty line directly beneath a `for` loop for the same reason that I wouldn't do it underneath `def`, it is clear to me from the indentation that a new logical block starts.

NuRadioReco/utilities/signal_processing.py
```diff
@@ -82,3 +83,50 @@ def add_cable_delay(station, det, sim_to_data=None, trigger=False, logger=None):
                         f"of cable delay to channel {channel.get_id()}")
 
         channel.add_trace_start_time(add_or_subtract * cable_delay)
+
+
+def calculate_vrms_from_temperature(temperature, bandwidth=None, response=None, resistance=50, freqs=None):
```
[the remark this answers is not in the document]
sjoerd-bouma: Then something is wrong here : ) public code should never return or expect input quantities in a different unit system than the NuRadio units. You've probably absorbed/lost a conversion factor somewhere in the vrms calculation.

NuRadioReco/modules/efieldToVoltageConverter.py
sjoerd-bouma: Okay, in which case the additional 200 ns is somewhat arbitrary, but considering we generally expect neutrino signals to arrive later in shallower channels and we normally use deeper channels to trigger, the asymmetry between pre and post pulse times at least makes sense.

NuRadioReco/utilities/signal_processing.py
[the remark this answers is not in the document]
sjoerd-bouma: Looks like you never convert `constants.k` to nuradio units, and there is also a units.Hz floating around. If you remove the units.Hz and use `kb = scipy.constants.k * units.joule / units.kelvin` this should fix things.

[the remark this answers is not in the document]
sjoerd-bouma: also also - is it correct to use the term 'resistance' or should we be using 'impedance'?

NuRadioMC/examples/08_RNO_G_trigger_simulation/simulate.py
```diff
@@ -93,15 +82,14 @@ def RNO_G_HighLow_Thresh(lgRate_per_hz):
...
         self.rnogHarwareResponse = hardwareResponseIncorporator.hardwareResponseIncorporator()
-        self.rnogHarwareResponse.begin(trigger_channels=deep_trigger_channels)
+        self.rnogHarwareResponse.begin(trigger_channels=kwargs['trigger_channels'])
 
         super().__init__(*args, **kwargs)
-        self.logger = logging.getLogger("NuRadioMC.RNOG_trigger_simulation")
-        self.deep_trigger_channels = deep_trigger_channels
-
+        self.logger = logger
+        self.deep_trigger_channels = kwargs['trigger_channels']
```
sjoerd-bouma, line 92: Does this actually get used? In l.199 it looks like these are hardcoded.

sjoerd-bouma (dismissed):
I haven't run the example, but I assume that you have : ) other than that things look okay to me? Couple of typos in variable names, but up to you if you want to fix them (lenght, HarwareResponse).

NuRadioMC/examples/08_RNO_G_trigger_simulation/simulate.py
fschlueter: I think I have another PR which does change this anyway

cg-laser (approved):

PR #847, opened 2025-02-26
Add impulse response as placeholder measurement

jakobhenrichs:
The impulse response measurements performed in Chicago (by @philippwindischhofer ) improved RNO-G's shallow CR analysis by a lot. This PR adds the impulse response measurement for the surface channels as a placeholder to NuRadio so that it can be used through the hardwareResponseIncorporator. In the long term, it is planned to add the impulse response measurement to the database. This PR is a shortcut so that the impulse response can be used already. It is important to note that the measurement has not the correct overall gain (it was normalized in the time domain).


sjoerd-bouma:
The normalization looks a bit odd to me (max(f)=1 instead of max(abs(f))=1). If we treat this as a placeholder anyway wouldn't it be better to have something that at least roughly mimics the amplification of the real surface amps?

sjoerd-bouma (approved):

jakobhenrichs:
I don't have a strong preference. But I can easily add some overall gain to it so that it has the same amplification as the other surface placeholder measurement.

CamphynR:
I can also comment here that we found the impulse response for the deep channels to be very useful so I can support adding those as well

jakobhenrichs:
I added a impulse response for the deep channel.

fschlueter:
@jakobhenrichs I think it is crucial to add them to the database right away! They do not have to appear in any signal chain but it allows experts to query it and use it via the rnog detector class interface.

jakobhenrichs:
> How big are those files?

The files have a size of 100kB each

jakobhenrichs:
> @jakobhenrichs I think it is crucial to add them to the database right away! They do not have to appear in any signal chain but it allows experts to query it and use it via the rnog detector class interface.

We can add them to the database, but I am not sure how it would work that they can be queried from the database and used in the hardwareResponseIncorporator without being in a signal chain. If there is a possibility, I am happy to do it that way. If not, I think it's a quick fix to make the measurements available before we have figured out how we want to handle the different gain per channel in the database.

sjoerd-bouma:
I'm just adding a plot here for comparison:
image

To me, what happens at the band edges looks a bit dodgy (the 'bump' above 800 MHz, and some noise (?) below 50 MHz). I guess this is not that significant in practice (especially if combined with bandpass filters or antenna responses that drop off before then anyway), but may be worth investigating/correcting before this graduates to non-placeholder status.

sjoerd-bouma (approved):
Thanks @jakobhenrichs  (and @philippwindischhofer )! I left one comment that maybe is worth clearing up at some stage, but I think for now as a placeholder this is more than sufficient.

jakobhenrichs:
Hi @philippwindischhofer, thanks for the suggestion with the window. I applied a tukey window to the impulse response (in time domain) before performing the FFT and updated the placeholder measurements. The results can be seen in the plot below:
plots_for_PR_SURFACE
Like in Sjoerds plot, "old" refers to the old placeholder measurements without the RADIANT.
It can be seen that the impulse response placeholder measurements are now less noisy.
The total time delays of the different measurements are:
SURFACE (old): 5.72 ns
SURFACE (impulse): 1.07 ns
IGLU (old): 259.71 ns
IGLU (impulse): 2.22 ns
Note: For the impulse response measurement, the RADIANT was triggered externally and thus the time delay is just the difference between the signal and trigger cable).

fschlueter:
> Hi @philippwindischhofer, thanks for the suggestion with the window. I applied a tukey window to the impulse response (in time domain) before performing the FFT and updated the placeholder measurements. The results can be seen in the plot below: plots_for_PR_SURFACE Like in Sjoerds plot, "old" refers to the old placeholder measurements without the RADIANT. It can be seen that the impulse response placeholder measurements are now less noisy. The total time delays of the different measurements are: SURFACE (old): 5.72 ns SURFACE (impulse): 1.07 ns IGLU (old): 259.71 ns IGLU (impulse): 2.22 ns Note: For the impulse response measurement, the RADIANT was triggered externally and thus the time delay is just the difference between the signal and trigger cable).

The time delays you are quoting are included in the group delay plot? This feature at 700MHz in the group delay for the surface chain - is that real @philippwindischhofer maybe. Do you think it makes sense to also window the frequency spectrum? What is the frequency component above 800MHz, looks to me like electronic noise which does not averages out.

jakobhenrichs:
The time delays quoted in my comment are removed from the group delay plot (so that all the group delays are on top of each other)

philippwindischhofer:
@fschlueter I think the feature in the surface group delay at 700MHz is just a consequence of the gain being really small at this point (so that the group delay becomes meaningless). And yes, the out-of-band "noise floor" is likely to do with the RADIANT.

fschlueter:
> @fschlueter I think the feature in the surface group delay at 700MHz is just a consequence of the gain being really small at this point (so that the group delay becomes meaningless). And yes, the out-of-band "noise floor" is likely to do with the RADIANT.

so shall we window the response before uploading?

philippwindischhofer:
Personally, I don't think "windowing" in the frequency domain is going to help, for two reasons. 

First, there will never be a significant signal component out-of-band, so whatever the impulse response is doing there, it will not enter into any de-dispersion or unfolding procedure. 

Second, the "noise" / RADIANT features are also present in the passband, and we cannot do anything about those (except maybe repeat the measurements and average over more traces).

fschlueter:
> Personally, I don't think "windowing" in the frequency domain is going to help, for two reasons.
> 
> First, there will never be a significant signal component out-of-band, so whatever the impulse response is doing there, it will not enter into any de-dispersion or unfolding procedure.
> 
> Second, the "noise" / RADIANT features are also present in the passband, and we cannot do anything about those (except maybe repeat the measurements and average over more traces).

But they are just an artefact from averaging over an asymmetric distribution right?

fschlueter:
Hey guys, just thinking about this. @jakobhenrichs you scaled the new placeholder measurement to have the same normalisation as the old one, correct? We should keep in mind that the old-one only described a IGLU-fiber-DRAB pair while the new one include an additonal coax-cable (probably irrelevant) a signal-splitter (should decrease the amplification by some dB) and a radiant (another some dB). So in principe the new response should be shifted a few DB down.

fschlueter:
@jakobhenrichs, @philippwindischhofer we should come to a conclusion here

jakobhenrichs:
I think I would leave the gain at the same level as the previous placeholder measurements. The overall gain is anyway not 100% correct (e.g. channel-to-channel variations) and these measurements should only be placeholders. However, if people feel strongly about it, I can adjust it.

PR #848, opened 2025-02-26
modules.RNO_G.crRNOGTemplateCreator: fix channel overwrite bug

jakobhenrichs:
After introduction of the overwrite flag in PR #791 the eFieldToVoltageConverter raised an error. In this PR a fix for this issue is added. The fix is the same as used in PR #791 to fix the test.


fschlueter (changes requested):
The proposed solution works by why are you creating channels in the function `_create_Efield` (Lines 254-255) in the first place? I do not see a reason for this.

fschlueter:
```
det_temp = detector.generic_detector.GenericDetector(json_filename=self.__detector_file, antenna_by_depth=False, create_new=True, log_level='ERROR')
```
Why do you always create a new detector object in the inner loop? I also think this module should use the rnog detector (of course always creating a new detector object is not practical when querying from the DB)

fschlueter (approved):

PR #849, opened 2025-02-27
Adding RNO-G Extra Variables and Moving Analysis Variable Calculation Functions Around

martin01021207:
Hi All @cg-laser @fschlueter @JethroStoffels,

*** Changes in `trace_utilities.py` and in `channelSignalReconstructor.py`:
Pawan @pgiri4 and I have added those two variables **entropy** and **kurtosis**, which Pawan previously introduced to us, to the NuRadio. I propose to add them to the script `trace_utilities.py` instead of `channelSignalReconstructor.py` because I personally think that taking a trace as the input argument is more general, so we can use those methods on not only a channel waveform but also on a coherently-summed waveform (CSW). Also, Jethro has already added two methods to the script `trace_utilities.py`, they are `maximum_peak_to_peak_amplitude()` and `split_trace_noise_rms()`, I got inspired by that, so then I've also moved multiple methods to `trace_utilities.py`, such as the impulsivity and the RPR calculations, and I've added some other methods to that script too, for the same reason it's more flexible to use those methods in this way. Now, if someone is to run `channelSignalReconstructor.py`, all those analysis variables' methods will be called from `trace_utilities.py`, so we are not losing anything under this change, but if anyone disagrees with me or suggests to put those methods in a different script, please let me know.

*** Changes in Jethro's script `stationDeepCRVariables.py`:
I made some changes in this script to make the function `coherent_sum()` available for inputing multiple traces, so in this way we don't have to pass the channel object if we want to use the method independently.

*** Questions I have:

1. To @JethroStoffels, would you consider rename the script `stationDeepCRVariables.py` to something like `coherentlySummedWaveforms.py` or anything that you think it's better? Because all methods in this script are about the CSW, and also I think those variables are pretty general, they should not be just for the deep CR search.

2. To everyone, 
(1) We have a simple method `split_trace_noise_rms()` which Jethro added to `trace_utilities.py` to calculate the noise RMS, that definition is different from the noise calculations in Christian's `get_SNR()` in `channelSignalReconstructor.py`, but maybe that's okay? Or how do we standardize the calculation of the noise RMS as independent method?
(2) In Christian's `get_SNR()` we see different SNR options, so it means that it's also up to the user to choose which one to use in the end of the day, right? I also saw the method `coherent_snr(coherent_trace)` in Jethro's script, it takes a single CSW as the input, but it means that the definition is again different here for the SNR calculations. I've also added a `get_SNR(trace, noise_rms)` in `trace_utilities.py`, which can handle a channel waveform or a CSW, but we probably don't want too many different SNR methods in separate scripts, so I want everybody's opinions on this, which one should we keep, or if we can at least combine Jethro's `coherent_snr(coherent_trace)` with the `get_SNR(trace, noise_rms)` that I just added?


NuRadioReco/modules/RNO_G/stationDeepCRVariables.py
```diff
@@ -84,55 +89,59 @@ def coherent_sum_step_by_step(self, event, station):
...
             plt.tight_layout(rect = [0, 0, 1, 0.96])
             plt.savefig(f'coherent_sum_steps_reference_{ref_ch.get_id()}.png')
             plt.show()
 
-    def coherent_sum(self, event, station, ref_ch):
-        sum_chan = ref_ch.get_trace()
-        channels = [ch for ch in station.iter_channels(use_channels = self.__channel_ids) if ch.get_id() != ref_ch.get_id()]
-        for idx, ch in enumerate(channels):
-            cor = signal.correlate(np.abs(hilbert(sum_chan)), np.abs(hilbert(ch.get_trace())), mode = "full")
+    def coherent_sum(self, trace_set, ref_trace, use_envelope = False):
```
fschlueter, line 133: This function could also now be moved to `trace_utilities.py`?

NuRadioReco/modules/channelSignalReconstructor.py
```diff
@@ -274,17 +214,20 @@ def run(self, evt, station, det, stored_noise = False, rms_stage = 'amp'):
             channel[chp.P2P_amplitude] = np.max(trace) - np.min(trace)
 
             #Calculate impulsivity of the signal
-            channel[chp.impulsivity] = self.get_impulsivity(channel)
+            channel[chp.impulsivity] = trace_utilities.get_impulsivity(channel)
             
             # Use noise precalculated from forced triggers
-            signal_to_noise, noise_rms, root_power_ratio = self.get_SNR_and_RPR(
+            signal_to_noise, noise_rms = self.get_SNR_and_RPR(
                 station.get_id(), channel, det, stored_noise=stored_noise, rms_stage=rms_stage)
```
fschlueter, line 221: You still use the "old name" you changed.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -432,4 +435,206 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
     ordered_rmss = np.sort(rms_of_splits)
     lowest_rmss = ordered_rmss[:lowest]
     rms = np.mean(lowest_rmss)
-    return rms
\ No newline at end of file
+    return rms
+
+def get_SNR(trace, noise_rms):
```
fschlueter, line 440: can you write out `signal_to_noise_ratio`? I think a more descriptive name in a utility module is better.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -432,4 +435,206 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+
+    peak = trace[peak_idx]
+    p2p = np.abs(np.diff(peak))
+    p2p = np.nanmax(p2p)
+    snr = p2p / (2 * noise_rms)
+
+    del upper_peak_idx, lower_peak_idx, peak_idx, peak
+    return snr
+
+def get_RPR(trace, times, noise_rms):
```
fschlueter, line 467: same here.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -432,4 +435,206 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+    upper_peak_idx = argrelextrema(trace, np.greater_equal, order = 1)[0]
+    lower_peak_idx = argrelextrema(trace, np.less_equal, order = 1)[0]
+    peak_idx = np.unique(np.concatenate((upper_peak_idx, lower_peak_idx)))
+
+    peak = trace[peak_idx]
+    p2p = np.abs(np.diff(peak))
+    p2p = np.nanmax(p2p)
+    snr = p2p / (2 * noise_rms)
+
+    del upper_peak_idx, lower_peak_idx, peak_idx, peak
```
fschlueter, line 464: I do not think this del statement is needed?

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -432,4 +435,206 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+        the 1d array array containing trace of a channel
+    noise_rms: float
+        noise root mean square.
+    Returns:
+    --------
+    root_power_ratio: float
+        Root Power Ratio value.
+    """
+    upper_peak_idx = argrelextrema(trace, np.greater_equal, order = 1)[0]
+    lower_peak_idx = argrelextrema(trace, np.less_equal, order = 1)[0]
```
fschlueter, line 456: cant you simply use `upper_peak_idx = np.argmax(trace)` and `lower_peak_idx = np.argmin(trace)` here? Or is `argrelextrema` doing something else?

fschlueter (commented):
Hi Martin, I like your idea of moving a lot of the functions to `trace_utilities.py` very much! Thank you for doing this. I made a few comments "in-line".

fschlueter:
To your other question - Yes I think we want to allow people to use different SNR definitions. Jethro added comments to `parameters.py` explaining the currently implemented version. However, moving them all to one script is a good idea.

fschlueter:
Currently your changes break our tests. Can you fix that?

NuRadioReco/modules/RNO_G/stationDeepCRVariables.py
martin01021207: @fschlueter 
Yes, I would love to move it to `trace_utilities.py`, too.

NuRadioReco/modules/channelSignalReconstructor.py
martin01021207: Ahh, not sure how I missed it, because it is `self.get_SNR()` in the script I have on my laptop, my bad I will fix this.

NuRadioReco/utilities/trace_utilities.py
martin01021207: Okay I will change the name

NuRadioReco/utilities/trace_utilities.py
martin01021207: Will do

NuRadioReco/utilities/trace_utilities.py
martin01021207: Yeah, I can remove this line.

NuRadioReco/utilities/trace_utilities.py
martin01021207: I guess @pgiri4 can answer this for us

pgiri4: Hi Felix,

I think we don't want to replace argrelextrema with argmax/argmin. The extrema we are looking for correspond to the maximum peak-to-peak (P2P) value, which comes from the local maxima and local minima.

In contrast, argmax/argmin will simply find the absolute maximum and minimum values in the entire trace, which does not necessarily capture the true P2P value we care about.

The key idea is that if a signal is present in the waveform, the true maximum and minimum values should be coming from the local maxima and minima — not just the global extremes.

For example, in the following waveform, you can see how the P2P would be calculated if we used argmax/argmin, but that P2P value would not reflect the true P2P value if we correctly used argrelextrema to detect the local peaks.

image

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -427,9 +430,231 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+
+    envelope = get_hilbert_envelope(trace)
+    maxv = np.argmax(envelope)
+    power_indexes = np.arange(len(envelope)) ## just a list of indices the same length as the array
+    closeness = list(
+        np.abs(power_indexes - maxv)
+    )  ## create an array containing index distance to max voltage (lower the value, the closer it is)
+
+    sorted_power = [x for _, x in sorted(zip(closeness, envelope))]
+    cdf = np.cumsum(sorted_power)
```
pgiri4, line 562: We recently tested in ARA analysis that it's good to square the waveform/envelope first before calculating cdf(This approach differentiates more between noise and signal-like waveforms than our current approach).  I mean, let's do cdf = np.cumsum(sorted_power**2). And please leave the following line as a reference in the comment  https://github.com/ara-software/AraProc/blob/f7e28c03a6a0603d2ac580ab353bf70940ee97f9/araproc/analysis/impulsivity.py#L64

pgiri4 (changes requested):

NuRadioReco/utilities/trace_utilities.py
martin01021207: @pgiri4 Yes Pawan, I think we should make this change, thanks!

NuRadioReco/utilities/trace_utilities.py
fschlueter: Ah okay, I clearly did appreciate what the code is doing. Does this function only look for neighbouring bins to find the highest p2p value? There is already a method in this script which calculates the maximum p2p value for variable window size `maximum_peak_to_peak_amplitude` cant you use this? Also be aware that some of the definitions of the SNR using the p2p value actually calculated the global maximum - global minimum. While I fully agree that this is not the best way of doing this we should be careful not changing the existing code unintentionally?

fschlueter: Ah I just saw that @martin01021207 already addressed this, great! However, outstanding is the question whether the old code used the global maxima or local maxima.

fschlueter:
Also @martin01021207 could you fix the doc-string of the function `maximum_peak_to_peak_amplitude` to reflect that this function returns a float of arrays the same size as the input array?

NuRadioReco/utilities/trace_utilities.py
fschlueter: Since bryan uses not the square of the power it might be useful to keep that configurable?

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -427,9 +430,231 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+    trace : waveform of given channel
+        the 1d array array containing trace of a channel
+    noise_rms: float
+        noise root mean square.
+    window_size: int
+        coincidence window size.
+    Returns:
+    --------
+    root_power_ratio: float
+        Root Power Ratio value.
```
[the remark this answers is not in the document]
sjoerd-bouma: Also I'm wondering if "returns: Root Power Ratio" is correct?

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -427,9 +430,231 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+    return signal_kurtosis
+
+def is_trace_corrupt(trace):
+    """
+    Makes sure the trace has no NAN nor INF"
+
+
+    Parameters
+    ----------
+    trace : waveform trace of given channel
```
sjoerd-bouma, line 642: `waveform` -> `np.ndarray` on first line (see above)

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -427,9 +430,231 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+    float
+        The kurtosis (tailedness) measure of the signal.
+    """
+
+    signal_kurtosis = kurtosis(trace)
+
+    return signal_kurtosis
+
+def is_trace_corrupt(trace):
+    """
```
sjoerd-bouma, line 636: I'm not sure I like `is_trace_corrupt`, I might prefer `is_valid_trace` (which would of course flip the boolean outcomes). But I don't have a strong preference, we can keep it this way too.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -427,9 +430,231 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+
+    trace = np.array(trace)
+
+    npoints_NAN = len( np.argwhere(np.isnan(trace)) )
+    npoints_INF = len( np.argwhere(np.isinf(trace)) )
+
+    if npoints_NAN or npoints_INF:
+       is_bad_waveform = True
+
+    return is_bad_waveform
```
sjoerd-bouma, line 660: A quicker/'cleaner' solution would be 
```
is_bad_waveform = np.any(np.isnan(trace)) or np.any(np.isinf(trace))
```
`isnan`/`isinf` will automatically cast to a numpy array if necessary.

NuRadioReco/utilities/trace_utilities.py
martin01021207: @blhendricks Bryan, what do you think about this?

NuRadioReco/utilities/trace_utilities.py
martin01021207: @fschlueter I've combined those two SNR functions into one, so now when the window_size is specified, the `maximum_peak_to_peak_amplitude` will be used, if not then Pawan's method will be used.

@pgiri4 Anything to add?

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -432,4 +435,206 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+    # Step 2: Calculate the probability distribution (normalized)
+    probabilities = hist / np.sum(hist)
+
+    # Step 3: Calculate Shannon Entropy
+    # Using base = 2 for entropy in bits
+    signal_entropy = entropy(probabilities, base = 2)
+
+    return signal_entropy
+
+def get_kurtosis(trace):
```
pgiri4, line 616: @fschlueter,  to answer your question about what makes the kurtosis so low for the noise waveform, I printed all the variables used in the kurtosis calculation. I think the fourth moment is responsible for the differences. 
image

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -427,9 +430,231 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
     lowest_rmss = ordered_rmss[:lowest]
     rms = np.mean(lowest_rmss)
-    return rms
\ No newline at end of file
+    return rms
+
+def get_signal_to_noise_ratio(trace, noise_rms, window_size=None):
+    """
+    Computes the Signal to Noise Ratio (SNR) of a given trace.
+
```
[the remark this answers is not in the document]
martin01021207: Just did!

NuRadioReco/utilities/trace_utilities.py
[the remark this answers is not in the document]
martin01021207: @sjoerd-bouma 
Thanks for all of this, I've fixed those ones not following the standard

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -427,9 +430,231 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+        p2p = np.nanmax(p2p)
+
+    snr = p2p / (2 * noise_rms)
+
+    return snr
+
+def get_root_power_ratio(trace, times, noise_rms):
+    """
+    Computes the Root Power Ratio (RPR) of a given trace.
+
```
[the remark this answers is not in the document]
martin01021207: @sjoerd-bouma 
Just added a description, Pawan @pgiri4 wrote it

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -427,9 +430,231 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+    cdf = np.cumsum(sorted_power)
+    cdf = cdf / cdf[-1]
+
+    cdf_avg = (np.mean(np.asarray([cdf])) * 2.0) - 1.0
+
+    if cdf_avg < 0:
+        cdf_avg = 0.0
+    return cdf_avg
+
+def get_coherent_sum(trace_set, ref_trace, use_envelope = False):
```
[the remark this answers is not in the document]
martin01021207: Just added one

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -427,9 +430,231 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+    return sum_wf
+
+def get_entropy(trace, n_hist_bins = 50):
+    """
+    Calculate the shannon entropy of a signal.
+
+
+    Parameters
+    ----------
+    trace : waveform of given channel
```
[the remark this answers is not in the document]
martin01021207: Fixed

NuRadioReco/utilities/trace_utilities.py
martin01021207: Fixed

NuRadioReco/utilities/trace_utilities.py
martin01021207: @sjoerd-bouma 
I've just renamed it to `is_NAN_or_INF()`
Even more straightforward now ;)

NuRadioReco/utilities/trace_utilities.py
martin01021207: @sjoerd-bouma I think I wrote it in this way because at that time I was also counting how many NAN and INF points there were in my simulated data. I've just added two more return values: `npoints_NAN` & `npoints_INF`

pgiri4 (approved):

martin01021207:
@fschlueter @sjoerd-bouma 
Pawan and I have added descriptions to all the methods, I've also renamed a few of the methods. @pgiri4 Pawan's already fixed the issue that was causing the test to fail, so everything is good now! If you guys agree with all these changes please approve the review, then we will merge, thanks again for all your feedback!

Another question I would like to get an answer from you guys and @JethroStoffels is that if we should rename the script `stationDeepCRVariables.py` because it shouldn't be just CR specific, but I guess we can continue this discussion after merging this branch to `add_rnog_data_processing_script`.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -1,10 +1,13 @@
 from NuRadioReco.utilities import units, ice, geometryUtilities as geo_utl, fft
 import NuRadioReco.framework.base_trace
 
-import scipy.ndimage
 import numpy as np
-import scipy.constants
 import scipy.signal
+import scipy.ndimage
```
fschlueter, line 6: Is that import still needed?

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -404,13 +407,20 @@ def maximum_peak_to_peak_amplitude(trace, coincidence_window_size):
 
     Returns
     -------
-    maximal peak to peak amplitude of the trace
+    amplitudes: array of floats (same length as the input trace)
+        Local peak to peak amplitudes
     """
-    return scipy.ndimage.maximum_filter1d(trace, coincidence_window_size) - scipy.ndimage.minimum_filter1d(trace, coincidence_window_size)
+    amplitudes = maximum_filter1d(trace, coincidence_window_size) - minimum_filter1d(trace, coincidence_window_size)
```
fschlueter, line 413: This function does not return anything anymore?

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -404,13 +407,20 @@ def maximum_peak_to_peak_amplitude(trace, coincidence_window_size):
...
 
-def split_trace_noise_rms(trace, segments=4, lowest=2):
+    return amplitudes
+
+def get_split_trace_noise_RMS(trace, segments=4, lowest=2):
```
fschlueter, line 417: why capital RMS all of the sudden?

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -419,17 +429,281 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
     ordered_rmss = np.sort(rms_of_splits)
     lowest_rmss = ordered_rmss[:lowest]
-    rms = np.mean(lowest_rmss)
-    return rms
\ No newline at end of file
+    noise_root_mean_square = np.mean(lowest_rmss)
+
+    return noise_root_mean_square
+
+def get_signal_to_noise_ratio(trace, noise_rms, window_size=None):
```
fschlueter, line 448: The p2p calculation is still weird:
If `window_size==1` `peak_to_peak_amplitudes` is used. If `window_size==0` Pawans code is used which implicitly assumes `window_size==1` correct? Is there any advantage using Pawans code if one wants `window_size==1`? What is also weird that if `window_size==0` a window of 1 is assumed. One could also calculate the global p2p if window size is 0.

NuRadioReco/utilities/trace_utilities.py
martin01021207: @fschlueter Yes this import is needed

NuRadioReco/utilities/trace_utilities.py
martin01021207: @fschlueter It does, since I thought it didn't look good when it was returning a long line, I added a variable to equal the long line, then return the variable. And this also helps writing the "Return" in the description, it's more neat.

NuRadioReco/utilities/trace_utilities.py
martin01021207: @fschlueter I added "get_" in the front because other analysis functions also have that, then I made the RMS capital because I personally prefer using capital letters to name an acronym, like we have `get_SNR` in `channelSiganlReconstructor.py`. Also that way people can easily tell this function is for getting the RMS, I know if we were to write out the whole thing root_mean_square that would be too long...

NuRadioReco/utilities/trace_utilities.py
martin01021207: Love that! You've pointed out something I didn't notice, so now I've changed it to
`if window_size or window_size == 0:`
    `peak_to_peak_amplitudes`
`else`
    `Pawan's approach`

My thought was that when there was no `window_size` specified at all, Pawan's SNR approach would be used, kinda like a back-up way to do; so now when `window_size` is given, **including zero**, the `peak_to_peak_amplitudes()` will be used.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -419,17 +429,281 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+        Trace of a waveform
+
+    Returns
+    -------
+    signal_kurtosis: float
+        Kurtosis of the signal (trace)
+    """
+    signal_kurtosis = kurtosis(trace)
+
+    return signal_kurtosis
```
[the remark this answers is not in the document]
martin01021207: @fschlueter Good call, I've fixed this.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -419,17 +429,281 @@ def split_trace_noise_rms(trace, segments=4, lowest=2):
...
+    npoints_NAN: int
+        Number of NAN points
+    npoints_INF: int
+        Number of INF points
+    """
+    is_bad_trace = False
+
+    trace = np.array(trace)
+
+    npoints_NAN = len( np.argwhere(np.isnan(trace)) )
```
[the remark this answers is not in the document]
martin01021207: @fschlueter You are being too much now! (Just kidding, I fixed it)

martin01021207:
@fschlueter Please review those changes again :)

NuRadioReco/utilities/trace_utilities.py
fschlueter: Really? In line 10 you have `from scipy.ndimage import uniform_filter1d, maximum_filter1d, minimum_filter1d`. When I search for `scipy.ndimage` I can only find it in line 6 and 10.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -1,10 +1,13 @@
...
 import NuRadioReco.framework.base_trace
 
-import scipy.ndimage
 import numpy as np
-import scipy.constants
 import scipy.signal
+import scipy.ndimage
+import scipy.constants
+from scipy.stats import kurtosis, entropy
+from scipy.signal import hilbert, argrelextrema
```
fschlueter: importing here `hilbert` is also not needed as far as i can see because the code uses everywhere `scipy.signal.hilbert` (which personally I prefer anyway)

NuRadioReco/utilities/trace_utilities.py
fschlueter: sorry I got fooled by the git diff.

NuRadioReco/utilities/trace_utilities.py
fschlueter: I do not see that update in the code.

NuRadioReco/utilities/trace_utilities.py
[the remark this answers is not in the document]
fschlueter: However, I am not seeing any of the updates you mentioned in your last rounds of comments. Did you miss to push them?

NuRadioReco/utilities/trace_utilities.py
martin01021207: @fschlueter
I've changed many lines to things like this as you suggested above:
`amplitudes = scipy.ndimage.maximum_filter1d(trace, coincidence_window_size) - scipy.ndimage.minimum_filter1d(trace, coincidence_window_size)`
`return amplitudes`

So we only need to `import scipy.ndimage` now

NuRadioReco/utilities/trace_utilities.py
martin01021207: @fschlueter I am sure this has been changed

NuRadioReco/utilities/trace_utilities.py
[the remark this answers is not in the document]
martin01021207: @fschlueter Ummm... This has been changed too

NuRadioReco/utilities/trace_utilities.py
martin01021207: @fschlueter This has been changed too... I think you are looking at the outdated code changes

martin01021207:
@fschlueter 
You can see those changes here:
https://github.com/nu-radio/NuRadioMC/pull/849/commits/0b262ddc14b2ce5611bf7f4c2d799e0663eeb6db

fschlueter:
@martin01021207 apologise again - indeed I looked at outdated code somehome ...

fschlueter:
@martin01021207 and @pgiri4 I run some tests with `get_signal_to_noise_ratio`. When the window_size is == 0 the function throws an error now (sorry I suggested that to you). If window_size is == 1 the return is always 0. If `window_size == 3` the results are always the same as `window_size == None`. So I would check that window_size is `>= 2` and would remove the case for `None`

fschlueter:
> > @martin01021207 and @pgiri4 I run some tests with `get_signal_to_noise_ratio`. When the window_size is == 0 the function throws an error now (sorry I suggested that to you). If window_size is == 1 the return is always 0. If `window_size == 3` the results are always the same as `window_size == None`. So I would check that window_size is `>= 2` and would remove the case for `None`
> 
> @fschlueter Thanks for checking! But I kinda don't get what you mean fully (or what I should do here). Would you mind make a change on this one?

This is what I meant. Fine?

martin01021207:
> > > @martin01021207 and @pgiri4 I run some tests with `get_signal_to_noise_ratio`. When the window_size is == 0 the function throws an error now (sorry I suggested that to you). If window_size is == 1 the return is always 0. If `window_size == 3` the results are always the same as `window_size == None`. So I would check that window_size is `>= 2` and would remove the case for `None`
> > 
> > 
> > @fschlueter Thanks for checking! But I kinda don't get what you mean fully (or what I should do here). Would you mind make a change on this one?
> 
> This is what I meant. Fine?

@fschlueter Okay, thanks a lot for that! But I've just changed the default `window_size` to 3, so it would be more convenient for users. If you think everything is good to go now please approve this PR, then we will merge it to branch `add_rnog_data_processing_script`

fschlueter (approved):

fschlueter:
@martin01021207 I resolved the merge conflict. Once the CI passes feel free to merge it. Thank you both for your work!

PR #850, opened 2025-02-27
Added  interactive LOFAR example

karenterveer:
This adds an example of how to read & do something with LOFAR data. It runs out of the box if the user is on the coma cluster (where LOFAR data is stored). Maybe we can think of a way (e.g. by storing an example LOFAR event somewhere) that it can also run off coma.


anelles (approved):
Very nice work! We will only find how well it works, after people use it. So let’s merge it!

PR #851, opened 2025-02-28
Add ARIANNAparameters to Station by default to support older nur files

sjoerd-bouma:
Adds ARIANNAparameters as a default allowed parameter type to `Station`. Currently, .nur files predating the new parameter_storage structure that contain ARIANNAparameters cannot be read in, because they are not an allowed parameter type.

The 'downside' is that this will become the default going forward, so that maybe the error message when passing invalid keys to `set_parameter` becomes slightly less helpful to new users. But the alternative of implementing a workaround in ParameterStorage.deserialize seems messier to me.

### DIFF
--- a/NuRadioReco/framework/station.py
+++ b/NuRadioReco/framework/station.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 import NuRadioReco.framework.base_station
 import NuRadioReco.framework.sim_station
 import NuRadioReco.framework.channel
+import NuRadioReco.framework.parameters
 from six import iteritems
 import pickle
 import logging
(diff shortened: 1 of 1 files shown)


cg-laser:
I think we want to have this as default parameters. Shouldn't also the RNO-G parameters be there? 
All station level parameters defined in `parameters.py` should be activated.

sjoerd-bouma:
> Hi Sjoerd, thanks for fixing this so quickly. I'm surprised that they weren't in yet. I think we want to have this as default parameters. Shouldn't also the RNO-G parameters be there? All station level parameters defined in `parameters.py` should be activated. Or am I missing something?

I agree that they should be! But the RNO-G parameters haven't been merged into develop yet, so I can't add them here, that should be done in the relevant PR instead.

(Maybe one more comment/question on the implementation here - I assumed ARIANNAparameters are only defined in Station, and never in a SimStation, which is why I added them here; if this is not correct I can add them to BaseStation instead).

fschlueter (approved):

fschlueter:
The failing CI test can be ignored. This was a mattak issue which is fixed in the current master branch.

PR #852, opened 2025-03-01
Add a RNO-G data processing example notebook

CamphynR:
My first attempt at including an interactive RNO-G data processing example. It requires the user to set the environment variable "RNO_G_DATA" to point to a RNO-G data directory. It might be convenient to store an event file in the repo but these files are quite big so I am not sure what a good solution might be. I'm not experienced with querying data over http but could taka a crack at it for this application.


fschlueter:
@sjoerd-bouma is there any good way to review notebooks from the webbrowser?

sjoerd-bouma:
> @sjoerd-bouma is there any good way to review notebooks from the webbrowser?

Github does a decent job of rendering .ipynb (or also markdown, for example... it even tries rst!) when you just click `view file`

sjoerd-bouma (approved):
Thanks @CamphynR !

PR #853, opened 2025-03-03
Include option in channelGalacticNoiseAdder to select only specified …

PhilippLaub:
…channels. Useful when interpolating only selected channels in the readCoREASDetector module.

### DIFF
--- a/NuRadioReco/modules/channelGalacticNoiseAdder.py
+++ b/NuRadioReco/modules/channelGalacticNoiseAdder.py
@@ -216,7 +216,7 @@ class channelGalacticNoiseAdder:
         """
         if selected_channel_ids is None:
             selected_channel_ids = station.get_channel_ids()
-            logger.info(f"Using all channel ids")
+            logger.info(f"Using all channels: {selected_channel_ids}")
         else:
             logger.info(f"Using selected channel ids: {selected_channel_ids}")
 
@@ -227,9 +227,8 @@ class channelGalacticNoiseAdder:
 
         # check that or all channels channel.get_frequencies() is identical
         last_freqs = None
-        for channel in station.iter_channels():
-            if channel.get_id() not in selected_channel_ids:
-                continue
+        for channel_id in selected_channel_ids:
+            channel = station.get_channel(channel_id)
             if last_freqs is not None and (
                     not np.allclose(last_freqs, channel.get_frequencies(), rtol=0, atol=0.1 * units.MHz)):
                 logger.error("The frequencies of each channel must be the same, but they are not!")
@@ -269,9 +268,8 @@ class channelGalacticNoiseAdder:
         c_vac = scipy.constants.c * units.m / units.s
 
         channel_spectra = {}
-        for channel in station.iter_channels():
-            if channel.get_id() not in selected_channel_ids:
-                continue
+        for channel_id in selected_channel_ids:
+            channel = station.get_channel(channel_id)
             channel_spectra[channel.get_id()] = channel.get_frequency_spectrum()
 
         for i_pixel in range(healpy.pixelfunc.nside2npix(self.__n_side)):
@@ -315,9 +313,8 @@ class channelGalacticNoiseAdder:
 
             channel_noise_spec = np.zeros_like(noise_spectrum)
 
-            for channel in station.iter_channels():
-                if channel.get_id() not in selected_channel_ids:
-                    continue
+            for channel_id in selected_channel_ids:
+                channel = station.get_channel(channel_id)
 
                 channel_pos = detector.get_relative_position(station.get_id(), channel.get_id())
                 if channel_pos[2] < 0:
@@ -373,9 +370,8 @@ class channelGalacticNoiseAdder:
                 channel_spectra[channel.get_id()] += channel_noise_spectrum
 
         # store the updated channel spectra
-        for channel in station.iter_channels():
-            if channel.get_id() not in selected_channel_ids:
-                continue
+        for channel_id in selected_channel_ids:
+            channel = station.get_channel(channel_id)
             channel.set_frequency_spectrum(channel_spectra[channel.get_id()], "same")
 
 
--- a/NuRadioReco/modules/channelGalacticNoiseAdder.py
+++ b/NuRadioReco/modules/channelGalacticNoiseAdder.py
@@ -43,6 +43,7 @@ except IndexError: # this is a common error if cfitsio is not found... there are
         "pylfmap import failed. This might be because you do not have a working "
         "installation of cfitsio. See https://github.com/F-Tomas/pylfmap/issues/2 for potential tips")
 
+
 class channelGalacticNoiseAdder:
     """
     Class that simulates the noise produced by galactic radio emission
@@ -190,7 +191,8 @@ class channelGalacticNoiseAdder:
             event,
             station,
             detector,
-            passband=None
+            passband=None,
+            selected_channel_ids=None
     ):
 
         """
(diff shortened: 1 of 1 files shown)


NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -207,7 +209,16 @@ def run(
...
             Lower and upper bound of the frequency range in which noise shall be
             added. The default (no passband specified) is [10, 1000] MHz
+        selected_channel_ids: list, default=None
+            A list containing the channels IDs to apply noise per station.
+            If None, all channels of the selected station in the detector are simulated.
+            To select all channels, set this to None.
         """
+        if selected_channel_ids is None:
+            selected_channel_ids = station.get_channel_ids()
+            logger.info(f"Using all channel ids")
```
MijnheerD, line 219: Maybe it would be more useful to print out all the channel IDs here, for debugging purposes

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -217,6 +228,8 @@ def run(
         # check that or all channels channel.get_frequencies() is identical
         last_freqs = None
         for channel in station.iter_channels():
+            if channel.get_id() not in selected_channel_ids:
+                continue
```
MijnheerD, line 232: Why do you need the `if`-statement? Can't you simply loop over `selected_channel_ids`?

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -257,6 +270,8 @@ def run(
 
         channel_spectra = {}
         for channel in station.iter_channels():
+            if channel.get_id() not in selected_channel_ids:
+                continue
```
MijnheerD, line 274: Same comment as above

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -301,6 +316,8 @@ def run(
             channel_noise_spec = np.zeros_like(noise_spectrum)
 
             for channel in station.iter_channels():
+                if channel.get_id() not in selected_channel_ids:
+                    continue
```
MijnheerD, line 320: Same comment as above

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -357,6 +374,8 @@ def run(
 
         # store the updated channel spectra
         for channel in station.iter_channels():
+            if channel.get_id() not in selected_channel_ids:
+                continue
```
MijnheerD, line 378: And lastly, same comment as above 🙃

NuRadioReco/modules/channelGalacticNoiseAdder.py
PhilippLaub: This was easier :D But I can adjust it to remove the if-statements if you want.

fschlueter:
@PhilippLaub I am not sure I understand the use case for this feature. Can you elaborate?

sjoerd-bouma:
Just for the record - what is the reason to use `selected_channel_ids` instead of `excluded_channels` as used by the channelGenericNoiseAdder? I would normally have a slight preference for the latter as it would be 1) more consistent and 2) generally, forgetting to 'exclude' a channel is not as bad as forgetting to 'include' it.

fschlueter:
> > @PhilippLaub I am not sure I understand the use case for this feature. Can you elaborate?
> 
> We have options for selecting specific channels in the readCoREASDetector and the efieldToVoltageConverter (and an option to exclude channels in the channelGenericNoiseAdder) so it made sense to me to also have this in the channelGalacticNoiseAdder.

Yes because both modules create channel objects from efield pulses/resp efield pulses for specific channels. It make sense there. The noise adder only work on the channels which are existing in a station. E.g., if you used either the `readCoREASDetector` to only simulate efields for the channels you desire or `efieldToVoltageConverter` to only convert the efield for channels you desire you do not have to specify anything in a noise adder module. One could add an exclude parameter as Sjoerd suggested but I do not see a good reason to do that.

PhilippLaub:
Hm I see. You are right, it is not necessary.

sjoerd-bouma:
> > > @PhilippLaub I am not sure I understand the use case for this feature. Can you elaborate?
> > 
> > 
> > We have options for selecting specific channels in the readCoREASDetector and the efieldToVoltageConverter (and an option to exclude channels in the channelGenericNoiseAdder) so it made sense to me to also have this in the channelGalacticNoiseAdder.
> 
> Yes because both modules create channel objects from efield pulses/resp efield pulses for specific channels. It make sense there. The noise adder only work on the channels which are existing in a station. E.g., if you used either the `readCoREASDetector` to only simulate efields for the channels you desire or `efieldToVoltageConverter` to only convert the efield for channels you desire you do not have to specify anything in a noise adder module. One could add an exclude parameter as Sjoerd suggested but I do not see a good reason to do that.

What about for RNO-G: say I want to simulate galactic noise for the upwards LPDA's only? This shouldn't really impact the other channels much and improves performance significantly.

fschlueter:
ah I see your point - you still want to simulate the hole event (with generic / thermal noise added to all channels) but not add a insignificant contribution which takes time to the deep channels.

fschlueter:
But than I would agree that we should use a blacklist instead of a whitelist (implementation is also cleaner).

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -207,7 +209,16 @@ def run(
         passband: list of float, optional
             Lower and upper bound of the frequency range in which noise shall be
             added. The default (no passband specified) is [10, 1000] MHz
+        excluded_channel_ids: list, default=None
+            A list containing the channels IDs to exclude per station.
+            If None, all channels of the selected station in the detector are used.
         """
+        if excluded_channel_ids is None:
+            selected_channel_ids = station.get_channel_ids()
+            logger.info(f"Using all channels: {selected_channel_ids}")
```
fschlueter, line 218: Can you make them debug messages. This would produce a lot of "noise" in simulations.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -216,7 +227,8 @@ def run(
 
         # check that or all channels channel.get_frequencies() is identical
         last_freqs = None
-        for channel in station.iter_channels():
+        for channel_id in selected_channel_ids:
+            channel = station.get_channel(channel_id)
```
fschlueter, line 231: you can use `for channel in station.iter_channels(use_channels= selected_channel_ids):` instead.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -190,7 +191,8 @@ def run(
             event,
             station,
             detector,
-            passband=None
+            passband=None,
+            excluded_channel_ids=None
```
sjoerd-bouma, line 195: Can we use `excluded_channels` (also in the docstring), just because that's what the genericNoiseAdder does as well?

NuRadioReco/modules/channelGalacticNoiseAdder.py
sjoerd-bouma: ...why? because it save one line?

fschlueter: Yes. In this file actually 3 lines because we have this loop three times. Counter question - if we have such a argument - why not using it. What I want to avoid is that every module does it differently.

fschlueter: I guess this is in line with your own comment :)

fschlueter: for me not.

PhilippLaub: > you can use `for channel in station.iter_channels(use_channels= selected_channel_ids):` instead.

Thanks! I was not aware of this feature.

sjoerd-bouma (approved):
Thanks @PhilippLaub !

PR #854, opened 2025-03-05
Add data driven noise

CamphynR:
This pull request aims to add another option to the channelGenericNoiseAdder moduled titled "data-driven". The goal is to use scale parameters derived from data to define rayleigh distributions per frequency and draw a noise spectrum. The scale parameters are found by looking at forced trigger events. The forced trigger events are run through a CW filter (the sinesubtraction filter that can be found in the SineSub branch) and the spectral amplitude distributions per frequency are fitted to obtain a scale paramater value. More info can be found in the attached slides [link](https://radio.uchicago.edu/wiki/images/7/72/Thermal_noise_ruben_v3.pdf) The generation of these scale parameter files is done outside of the scope of this PR. I attach a scale parameter file for s11, season23 to this PR. Similar files for all stations and seasons are being produced.

A rudimentary test is added at the bottom of the channelGenericNoiseAdder module. A current problem is that the noise amplitudes seem to appear too low. This pull request is definitely still a work in progress so any input is appreciated!
[thermal_noise_scale_parameters_s11_season23.json](https://github.com/user-attachments/files/19091562/thermal_noise_scale_parameters_s11_season23.json)


NuRadioReco/modules/channelGenericNoiseAdder.py
```diff
@@ -145,6 +160,27 @@ def bandlimited_noise(self, min_freq, max_freq, n_samples, sampling_rate, amplit
...
+                self.logger.error("When selecting data-driven noise, the station and channel ids should be passed to bandlimeted noise")
+                raise ValueError
+            
+            if station_id in [11]:
+                scale_parameter_path = self.scale_parameter_dir + "/" + f"thermal_noise_scale_parameters_s{station_id}_season23.json"
+            else:
+                raise NotImplementedError("Other station parameters are being generated")
+            
+            scale_parameter_dictionary = load_scale_parameters(scale_parameter_path)
+            scale_parameters = np.array(scale_parameter_dictionary["scale_parameters"])
```
fschlueter, line 176: minor comment. the function `load_scale_parameters` could directly return `scale_parameters`?

NuRadioReco/modules/channelGenericNoiseAdder.py
```diff
@@ -145,6 +160,27 @@ def bandlimited_noise(self, min_freq, max_freq, n_samples, sampling_rate, amplit
         elif type == 'rayleigh':
             fsigma = amplitude * sigscale / np.sqrt(2.)
             ampl[selection] = self.__random_generator.rayleigh(fsigma, nbinsactive)
+        elif type == "data-driven":
+            self.logger.warning("You have selected to generate data-driven noise NOTE THAT THE HARDWARE RESPONSE IS ALREADY INLUDED IN THIS NOISE and hence this module should be run AFTER incorporating hardware response.")
```
fschlueter, line 164: I agree, releasing this warning all the time is not a suitable way. There are better ways! Can you remove this?

NuRadioReco/modules/channelGenericNoiseAdder.py
```diff
@@ -145,6 +160,27 @@ def bandlimited_noise(self, min_freq, max_freq, n_samples, sampling_rate, amplit
...
+            
+            if station_id is None or channel_id is None:
+                self.logger.error("When selecting data-driven noise, the station and channel ids should be passed to bandlimeted noise")
+                raise ValueError
+            
+            if station_id in [11]:
+                scale_parameter_path = self.scale_parameter_dir + "/" + f"thermal_noise_scale_parameters_s{station_id}_season23.json"
```
fschlueter, line 171: can you add a check that the file is present?

fschlueter (commented):
Hey Ruben, thanks for doing this. I am not sure the way you have implemented it is the best way. Fave you considered using `bandlimited_noise_from_spectrum` or `precalculate_bandlimited_noise_parameters` in combination with `bandlimited_noise_from_precalculated_parameters`?

CamphynR:
This module should now be in a "first working prototype" state. I include here another scale parameter file with increased resolution. Further scale parameter files will be linked on the RNO-G wiki until a more permanent solution is discussed
[thermal_noise_scale_parameters_s11_season23.json](https://github.com/user-attachments/files/19129559/thermal_noise_scale_parameters_s11_season23.json)

NuRadioReco/modules/channelGenericNoiseAdder.py
```diff
@@ -462,3 +537,67 @@ def run(self, event, station, detector,
...
+
+    def create_sim_event(station_id, channel_id, detector, frequencies, sampling_rate):
+        event = Event(run_number=-1, event_id=-1)
+        station = Station(station_id)
+        station.set_station_time(detector.get_detector_time())
+        channel = Channel(channel_id)
+        channel.set_frequency_spectrum(np.zeros_like(frequencies, dtype=np.complex128), sampling_rate)
+        station.add_channel(channel)
+        event.set_station(station)
+        return event, station
```
fschlueter, line 564: FYI, you can use the module also like this which is a bit more efficient:
```
noise = noiseAdder.bandlimited_noise(..., time_domain=False)
channel.set_frequency_spectrum(noise, sampling_rate)
```

CamphynR: Thanks that is indeed more convenient

NuRadioReco/modules/channelGenericNoiseAdder.py
```diff
@@ -145,14 +180,34 @@ def bandlimited_noise(self, min_freq, max_freq, n_samples, sampling_rate, amplit
         elif type == 'rayleigh':
             fsigma = amplitude * sigscale / np.sqrt(2.)
             ampl[selection] = self.__random_generator.rayleigh(fsigma, nbinsactive)
+        elif type == "data-driven":
+            if station_id is None or channel_id is None:
+                self.logger.error("When selecting data-driven noise, the station and channel ids should be passed to bandlimeted noise")
+                raise ValueError
+            
+            scale_parameter_path = f"thermal_noise_scale_parameters_s{station_id}_season23.json"
+            if scale_parameter_path in self.scale_parameter_paths:
```
fschlueter, line 189: if `scale_parameter_dir` was not passed in the begin function this line will through an error because `self. scale_parameter_paths` is not initalized (correct?).

NuRadioReco/modules/channelGenericNoiseAdder.py
```diff
@@ -145,14 +180,34 @@ def bandlimited_noise(self, min_freq, max_freq, n_samples, sampling_rate, amplit
         elif type == 'rayleigh':
             fsigma = amplitude * sigscale / np.sqrt(2.)
             ampl[selection] = self.__random_generator.rayleigh(fsigma, nbinsactive)
+        elif type == "data-driven":
+            if station_id is None or channel_id is None:
+                self.logger.error("When selecting data-driven noise, the station and channel ids should be passed to bandlimeted noise")
+                raise ValueError
+            
+            scale_parameter_path = f"thermal_noise_scale_parameters_s{station_id}_season23.json"
```
fschlueter, line 188: For now fine, but in principle we should on mid term we probably want to get ride of the `_season23` suffix being hardcoded.

NuRadioReco/modules/channelGenericNoiseAdder.py
```diff
@@ -353,11 +413,31 @@ def bandlimited_noise_from_spectrum(self, n_samples, sampling_rate, spectrum, am
...
+            if scale_parameter_path in self.scale_parameter_paths:
+                scale_parameter_full_path = self.scale_parameter_dir + "/" + scale_parameter_path
+            else:
+                raise NotImplementedError("Other station parameters are being generated")
+            
+            scale_parameters = load_scale_parameters(scale_parameter_full_path)
+            fsigma = scale_parameters[channel_id](frequencies[selection])
+            ampl[selection] = self.__random_generator.rayleigh(fsigma, n_samples_freq)
```
fschlueter, line 429: since this code is a copy from the one in the other function you could consider putting this into its own function.

CamphynR:
As a check I include here the mean spectra obtained by generating 100000 data-driven noise spectra
image

fschlueter (approved):

CamphynR:
As a final comment here is the comparison between noise generated by this module in a range of 0-1000MHz, shown on blue, and the average frequency spectrum taken over station 11 season 2023 data, shown in orange. This plot shows channel 0. Note that the curves were taken with a butterworth bandpassfilter between 0.1-0.8 GHz, hence the difference with the previously posted curve
image

PR #855, opened 2025-03-05
Add strides to high low trigger and remove extra +1 from window sizes

ryankrebs016:
Really minor changes to the high low trigger that add striding for the high low logic and the coincidence logic. It also adds an option to extend the traces (in case simulating with the final number of samples) so that the ends have a chance to trigger since it would overlap with a window that extends before and after the trace. Also removes the +1 which extended the windows by a sample.


ryankrebs016:
Wait, I made a few mistakes

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -28,27 +29,38 @@ def get_high_low_triggers(trace, high_threshold, low_threshold,
...
-        raise TypeError(f"The trace ({trace.dtype}) and the threshold ({type(high_threshold)}) must have the same type")
+    n_bins_coincidence = int(np.round(time_coincidence / dt))
 
-    c2 = np.array([1, -1])
-    m1 = np.convolve(trace >= high_threshold, c, mode='full')[:-(n_bins_coincidence - 1)]
-    m2 = np.convolve(trace <= low_threshold, c, mode='full')[:-(n_bins_coincidence - 1)]
-    return np.convolve(m1 & m2, c2, mode='same') > 0
+    if extend_trace:
+        #hack it by assuming noise at ends so that beginning and end of trace aren't completely neglected, trace time may
+        trace=np.concatenate((trace,trace[:n_bins_coincidence]))
```
fschlueter, line 48: I am not sure we want this. It adds unphysical noise to the trace. Furthermore it is not needed? The simulated traces on which we trigger should always be longer of what we actually write out.

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -28,27 +29,38 @@ def get_high_low_triggers(trace, high_threshold, low_threshold,
...
+    if extend_trace:
+        #hack it by assuming noise at ends so that beginning and end of trace aren't completely neglected, trace time may
+        trace=np.concatenate((trace,trace[:n_bins_coincidence])) 
 
+    num_frames = int(np.floor((len(trace) - n_bins_coincidence) / step))
+    logger.debug("length of trace {} samples, coincidence window {}, num window bins {}".format(len(trace), n_bins_coincidence, num_frames))
 
-def get_majority_logic(tts, number_of_coincidences=2, time_coincidence=32 * units.ns, dt=1 * units.ns):
+    trace_windowed = np.lib.stride_tricks.as_strided(trace, (num_frames, n_bins_coincidence),
+                                                        (trace.strides[0] * step, trace.strides[0]))
```
fschlueter, line 54: Can you add a comment in the text what the function does?

fschlueter (commented):
Hey Ryan, a number of tests are failing. Do you expect these changes to effect the test outcome?

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -28,27 +29,38 @@ def get_high_low_triggers(trace, high_threshold, low_threshold,
...
     triggered bins: array of bools
         the bins where the trigger condition is satisfied
     """
-    n_bins_coincidence = int(np.round(time_coincidence / dt)) + 1
-    c = np.ones(n_bins_coincidence, dtype=bool)
-    logger.debug("length of trace {} bins, coincidence window {} bins".format(len(trace), len(c)))
 
-    if trace.dtype != type(high_threshold):
-        logger.error(f"The trace ({trace.dtype}) and the threshold ({type(high_threshold)}) must have the same type")
-        raise TypeError(f"The trace ({trace.dtype}) and the threshold ({type(high_threshold)}) must have the same type")
```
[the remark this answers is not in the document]
ryankrebs016: Do you have a recommendation for type coercion when it gets numpy.int64 and built-in int?

NuRadioReco/modules/trigger/highLowThreshold.py
ryankrebs016: For some scripts I'm working with they only generate 1024 samples and then passes these to the trigger so then windows cause a bias against the edge samples. I have no issue removing this

ryankrebs016:
Tests are failing because the trigger time slightly shifted. Rather than corresponding to the first sample that triggered, it points to the start of the window that triggered. The difference is pretty small

NuRadioReco/modules/trigger/highLowThreshold.py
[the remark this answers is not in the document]
ryankrebs016: Def thought this was throwing an error on those. Maybe not

fschlueter:
@ryankrebs016 I just checked a few tests - I do not think that differences are actually small. I have seen difference of 200ns and more. The tests do report the relative differences which are on the level of ~1-2% but this is not a good metric for the trigger time (when the time changes from 10200ns to 10400ns the absolute change is relevant the relative does not look significant)

ryankrebs016:
Trigger times on nu events in SingleEvents test are now close. Testing with step=5 are close enough to pass too

ryankrebs016:
Dumb question, should I be using rebase instead of merge when pulling in changes from develop?

ryankrebs016:
Okay, so I'm happier with this now. It will correctly get the trigger time for triggers appearing less than a window into the trace. Some noise tests do fail as it changes the trigger logic slightly, partly from reducing the window size by 1 sample, but also by allowing the get_high_low_triggers function to see the full window of "this channel is triggering here" instead of the transition from "triggering to not triggering". The behavior of get_majority_logic stays the same the except for the reduced window size by 1.

fschlueter:
@ryankrebs016 thank you for all your work! I will be on vacation next week put I will be reviewing the first thing I am back

fschlueter:
> Dumb question, should I be using rebase instead of merge when pulling in changes from develop?

Its not a dumb question. As far as I understand merging has the benefit of better understanding the history of changes while it "pollutes" the commit history with merge commits. Rebasing on the other hand can cause problems if people work on the same branch together. Hence when working on branches which are already pushed to the remote server and other people might work on them it is saver to use merge. However, in cases where all changes are still on local branches it makes sense to rebase for a cleaner history

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -72,19 +93,33 @@ def get_majority_logic(tts, number_of_coincidences=2, time_coincidence=32 * unit
     triggered_times: array of floats
         the trigger times
     """
+
+    assert np.abs((time_coincidence / dt) % step - step/2) < 0.9*step , "Trigger times may be inaccurate due to windowing"
+
+    # Coincidence bins needs reduced by step since the output traces of get_high_low_triggers is already binned by step
+    n_bins_coincidence = int(np.round(time_coincidence / dt / step))
```
fschlueter, line 100: I do not understand why we have to divide by `steps` here. But also I am not sure if we should divide by `dt == 1 / sampling_rate` here. Shouldn't we divide by `time_coincidence` passed to `get_high_low_triggers`?

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -72,19 +93,33 @@ def get_majority_logic(tts, number_of_coincidences=2, time_coincidence=32 * unit
...
+    stride_step = 1
```
fschlueter, line 101: Why not allowing a striding for this coincidence condition as well?

fschlueter (commented):
@ryankrebs016 I left some comments. I am also wondering if the padding is actually necessary (and currently it is not optional as far as I can see). Is it only to get the trigger time ride? What edge case is effected by it?

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -28,27 +28,46 @@ def get_high_low_triggers(trace, high_threshold, low_threshold,
...
-    c2 = np.array([1, -1])
-    m1 = np.convolve(trace >= high_threshold, c, mode='full')[:-(n_bins_coincidence - 1)]
-    m2 = np.convolve(trace <= low_threshold, c, mode='full')[:-(n_bins_coincidence - 1)]
-    return np.convolve(m1 & m2, c2, mode='same') > 0
+    n_bins_coincidence = int(np.round(time_coincidence / dt))
+
+    # Pad trace end enough so that the end isn't cut off after striding.
+    padded_trace=np.pad(trace, (n_bins_coincidence-1,1), "constant")
+    num_frames = int(np.floor((len(padded_trace) - n_bins_coincidence) / step))
+    num_real_frames = int(np.floor((len(trace)) / step))
```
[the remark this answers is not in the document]
fschlueter: I just changed it - let see if it works.

ryankrebs016:
> @ryankrebs016 I left some comments. I am also wondering if the padding is actually necessary (and currently it is not optional as far as I can see). Is it only to get the trigger time ride? What edge case is effected by it?

It's to get the trigger times right for triggers < 1 window length into the trace - beginning or end. It pads by 0's so it shouldn't affect the behavior of the trigger

NuRadioReco/modules/trigger/highLowThreshold.py
ryankrebs016: yeah so there probably should be an option because this function gets called by the simple trigger too... Though when it's called in the high low trigger it is already strided (binned) by the step size. 

an example: a trace (length 1024), with a high low window of 6 and a step of 4 is strided into a new array (256, 6) in `get_high_low_triggers`. The (256,6) array is passed to `get_majority_logic`, so when calculating the coincidence window it needs reduced by step length.

NuRadioReco/modules/trigger/highLowThreshold.py
ryankrebs016: see below comment for some clarification.

NuRadioReco/modules/trigger/highLowThreshold.py
fschlueter: okay I agree - sorry I for got that windows typically overlap if they are not strided.

fschlueter: And if stride step is 1 the trace out of get_high_low_triggers maintains its time resolution (this was the other error I made in my head).

fschlueter: Okay I would actually pass `dt=dt * step` to the function and use the step parameter to acutally allow for striding in this function as well.

ryankrebs016:
Had to add function arguments in order to 1. generalize the `get_majority_logic` function so it could be called from the simple trigger and do striding and 2. to make the striding start from the beginning of the trace (no padding) as if the trigger is called on real detector data

NuRadioReco/modules/trigger/highLowThreshold.py
ryankrebs016: can't change the dt since the function wouldn't know if the input is already strided or not. so added an argument which changes the stride to 1 or step

ryankrebs016:
Added an argument so that striding will start from the beginning of the trace (without padding) for if the trigger gets called on real detector data

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -72,19 +109,42 @@ def get_majority_logic(tts, number_of_coincidences=2, time_coincidence=32 * unit
...
     """
+
+    if already_strided:
+        # Coincidence bins needs reduced by step since the output traces of get_high_low_triggers is already binned by step
+        n_bins_coincidence = int(np.round(time_coincidence / dt / step))
+        stride_step = 1
+    else:
+        n_bins_coincidence = int(np.round(time_coincidence / dt))
+        stride_step = step
+
```
fschlueter, line 120: Maybe I am overlooking something but we only need to know whether the single channel traces were strided to set the coincidence window correct, i.e., to set the correct dt: `dt_stride = dt * step_high_low`. This is irrelevant for the striding in the majority_logic function which can have its own `step` argument?

fschlueter:
> Added an argument so that striding will start from the beginning of the trace (without padding) for if the trigger gets called on real detector data

why do we know that this is happening on real data? Because we readout the waveforms in blocks and the striding is always aligned with the block beginning I guess. 

I do not like the fact that we have a `is_data` argument buried so deep in a trigger module. On the other hand I do not see a huge risk of doing something wrong... but maybe @cg-laser or @sjoerd-bouma have a better idea to archive the correct behaviour and not relying on such a flag? Or maybe we can set the flag automatically by checking the presents of a sim_shower or sim_station?

cg-laser:
I think this is rather an RNO-G specific setting. For ARIANNA for example, the high-low trigger is implemented via an analog circuit. And for this, this module was originally developed. 
Wouldn't a name that actually explains the purpose, something like "align_xxx_"?

NuRadioReco/modules/trigger/highLowThreshold.py
ryankrebs016: True, the two step arguments don't need to be the same. So changing dt first is probably right

ryankrebs016:
> > Added an argument so that striding will start from the beginning of the trace (without padding) for if the trigger gets called on real detector data
> 
> why do we know that this is happening on real data? Because we readout the waveforms in blocks and the striding is always aligned with the block beginning I guess.

Yeah, the waveforms are built from units of 4 samples so the striding lines up with the beginning. the trigger logic runs on a clock 1/4 of the sampling rate.

fschlueter (approved):
I do not quite understand why the doc-string for the argument `step` says 
> ... and clock rate mismatch in trigger logic

And I wonder if you should pass the option for the `step` parameter for the majority logic to pass as the module argument instead of setting it to 1. But I am happy with that PR regardless

fschlueter:
@ryankrebs016 do you want to merge this? you can ignore the failing test.

fschlueter:
@ryankrebs016, are you planing to continue working on this PR?

PR #857, opened 2025-03-07
set the correct efield position (relative antenna position)

PhilippLaub:
Before that the positions of the efields were not set and would default to `[0,0,0]`. Now they are set to the antenna positions relative to the station. Fixes a bug in `efieldToVoltageConverter` where travel time shifts are calculated based on the difference between efield and antenna position.

### DIFF
--- a/NuRadioReco/modules/io/coreas/readCoREASDetector.py
+++ b/NuRadioReco/modules/io/coreas/readCoREASDetector.py
@@ -290,7 +290,8 @@ class readCoREASDetector:
                     # Store the trace in an ElecticField object
                     coreas.add_electric_field_to_sim_station(
                         sim_station, channel_ids_for_group_id, smooth_res_efield.T, res_trace_start_time,
-                        sim_shower[shp.zenith], sim_shower[shp.azimuth], self.coreas_interpolator.sampling_rate)
+                        sim_shower[shp.zenith], sim_shower[shp.azimuth], self.coreas_interpolator.sampling_rate,
+                        efield_position=antenna_position_rel)
 
                 sim_station.set_parameter(stnp.zenith, sim_shower[shp.zenith])
                 sim_station.set_parameter(stnp.azimuth, sim_shower[shp.azimuth])


PhilippLaub:
One thing that should be discussed here is whether the `efield.get_position()` function should return the _relative position_ or the _absolute position_ of the efield. In the docs it says that it is the relative position, but @MijnheerD raised the point that it should return the absolute position.

fschlueter:
> One thing that should be discussed here is whether the `efield.get_position()` function should return the _relative position_ or the _absolute position_ of the efield. In the docs it says that it is the relative position, but @MijnheerD raised the point that it should return the absolute position.

Initially, I agreed with Mitja - efields are somwhat a property of the physical event and not associated to a station. However, `efields` are stored in a `sim_station` which is stored in a `station`. Hence, an efield is alsways associated to a station. So I am fine either way, we should only properly document it and check that it is everywhere done the same way.

fschlueter:
@PhilippLaub thanks for doing this. The function `add_electric_field_to_sim_station` is also called in other reader modules. Can you make sure that there the position is also set?

fschlueter:
> > One thing that should be discussed here is whether the `efield.get_position()` function should return the _relative position_ or the _absolute position_ of the efield. In the docs it says that it is the relative position, but @MijnheerD raised the point that it should return the absolute position.
> 
> Initially, I agreed with Mitja - efields are somwhat a property of the physical event and not associated to a station. However, `efields` are stored in a `sim_station` which is stored in a `station`. Hence, an efield is alsways associated to a station. So I am fine either way, we should only properly document it and check that it is everywhere done the same way.

Looking a bit through the code I think we should keep it relative. E.g. `simulation.py` sets its to the relative value. Not worth the effort.

PhilippLaub:
> @PhilippLaub thanks for doing this. The function `add_electric_field_to_sim_station` is also called in other reader modules. Can you make sure that there the position is also set?

I'm not sure if I understand the other CoREAS readers correctly but it seems to me that for `readCoREASShower` and `readCoREASStation` the default position of  `[0,0,0]` is intended if we keep it relative.

fschlueter (approved):

MijnheerD:
I guess the thing I don't like about the position being relative, is that you always need the `Detector` object to get the full position. To me that feels like a very akward design. But if you say it is too deeply baked into the codebase, I agree it is not worth the effort of changing it everywhere.

MijnheerD:
Would it be worth to change the function name to `efield.get_relative_position()` ? It would make it much clearer what the property is supposed to represent. And I think that refactoring this name everywhere should not be too difficult.

PhilippLaub:
> Would it be worth to change the function name to `efield.get_relative_position()` ? It would make it much clearer what the property is supposed to represent. And I think that refactoring this name everywhere should not be too difficult.

Hm, after searching for the function in the repo, I think it might not be too much of an effort to change the name. I will look into this on Monday.

fschlueter:
If you change the name please keep the old name with a deprecation warning. The problem is not only what is in the repo but what other people have in their private code - we do not want to make their lives harder. For the same reason changing what a function returns (relative to absolute position) is dangerous.

fschlueter:
> I guess the thing I don't like about the position being relative, is that you always need the `Detector` object to get the full position. To me that feels like a very akward design. But if you say it is too deeply baked into the codebase, I agree it is not worth the effort of changing it everywhere.

I agree that this feels a bit awkward but I think there are enough good reasons to have it this way too and also to keep it that way.

PhilippLaub:
> @PhilippLaub typically the author merges the PR after a 24h period after approval

I know, but I wanted to do the renaming of the function in this PR (but I did not find the time for that so far). But I can also open a new PR for this.

PR #860, opened 2025-03-10
Add module for Galactic noise on Efields

MijnheerD:
This is another addition that was made in the context of @twybo master thesis. For MC studies, he needed an easy way of adding noise to electric field traces (in the RIT method you work with the Efields instead of the voltages). Because of how the `ElectricField` objects are stored, it was not straightforward to adapt the `channelGalacticNoiseAdder` to include this use case. To avoid too much code duplication (and maintenance overhead) I made this a subclass of the `channelGalacticNoiseAdder` with only the `run` method being changed.

If there is a cleaner way of implementing this, let me know!


NuRadioReco/modules/efieldGalacticNoiseAdder.py
```diff
@@ -0,0 +1,153 @@
...
+        if self.__caching:
+            if self.__freqs is None:
+                self.__freqs = freqs
+            else:
+                if not np.allclose(self.__freqs, freqs, rtol=0, atol=0.01 * units.MHz):
+                    self.__freqs = freqs
+                    self._get_cached_antenna_response.cache_clear()
+                    logger.warning(
+                        "Frequencies have changed. Clearing antenna response cache. "
+                        "(If this happens often, something might be wrong...")
```
fschlueter, line 80: This is not needed right? In the original module it is used to cache the antenna pattern which is not used here?

NuRadioReco/modules/efieldGalacticNoiseAdder.py
```diff
@@ -0,0 +1,153 @@
...
+                    1j * delta_phases) * np.cos(polarizations)
+                channel_noise_spec[2][passband_filter] = noise_spectrum[2][passband_filter] * np.exp(
+                    1j * delta_phases) * np.sin(polarizations)
+
+                # add noise spectrum from this to field freq spectrum
+                field_spectra[field.get_unique_identifier()] += np.sum(channel_noise_spec, axis=0)
+
+        # Store updated fields spectra
+        for field in station.get_electric_fields():
+            field.set_frequency_spectrum(field_spectra[field.get_unique_identifier()], "same")
```
fschlueter, line 153: Why du you store a single polarisation as the sum of all 3 polarisations instead of each polarisation independently?

NuRadioReco/modules/efieldGalacticNoiseAdder.py
```diff
@@ -0,0 +1,153 @@
...
+                                 * freqs[passband_filter] ** 2 * noise_temperature * self.solid_angle / c_vac ** 2)
+            spectral_radiance[np.isnan(spectral_radiance)] = 0
+
+            # calculate radiance per energy bin
+            spectral_radiance_per_bin = spectral_radiance * d_f
+
+            # calculate electric field per energy bin from the radiance per bin
+            efield_amplitude = np.sqrt(
+                spectral_radiance_per_bin / (c_vac * scipy.constants.epsilon_0 * (
+                        units.coulomb / units.V / units.m))) / d_f
```
fschlueter, line 122: Shall we add this code block in the utilities? Would be also useful for modules which Ruben is working on.

fschlueter (commented):
A few comments, I am fine with that being its on module as it fits/obeys our naming conventions.

NuRadioReco/modules/efieldGalacticNoiseAdder.py
MijnheerD: Indeed! I was bringing over the latest changes from the channelGalacticNoiseAdder and did not think this through

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -111,6 +111,50 @@ def get_electric_field_energy_fluence(electric_field_trace, times, signal_window
 
     return f_signal * dt * conversion_factor_integrated_signal
 
+
+def get_electric_field_from_radiance(frequencies, noise_temperature, solid_angle):
```
fschlueter, line 115: Thanks for doing this. Isnt a better name `get_electric_field_from_temperature`?

NuRadioReco/modules/efieldGalacticNoiseAdder.py
```diff
@@ -0,0 +1,130 @@
+from NuRadioReco.modules.channelGalacticNoiseAdder import *
```
fschlueter, line 1: is that needed? Can't you just import the class type?

NuRadioReco/modules/efieldGalacticNoiseAdder.py
```diff
@@ -0,0 +1,130 @@
...
+            )
+
+            # assign random phases to electric field
+            noise_spectrum = np.zeros((3, freqs.shape[0]), dtype=np.complex128)
+            phases = self.__random_generator.uniform(0, 2. * np.pi, len(efield_amplitude))
+
+            noise_spectrum[1][passband_filter] = np.exp(1j * phases) * efield_amplitude
+            noise_spectrum[2][passband_filter] = np.exp(1j * phases) * efield_amplitude
+
+            channel_noise_spec = np.zeros_like(noise_spectrum)
```
fschlueter, line 108: can we rename the variable? the name is confusing.

NuRadioReco/modules/efieldGalacticNoiseAdder.py
MijnheerD: No, I also need the `get_local_coordinates` function from that module. And most of the imports are also identical to the ` `channelGalacticNoiseAdder` module, so this construction saves a lot of spae as all the imports (like healpy, pygdsm, etc) are also resolved as well

fschlueter (approved):
Thanks Mitja!

fschlueter:
this is maybe a more general question/comment - do we want to have unit tests for every new module we add? @MijnheerD this would problem require some more time and effort so it is completely optional. A test which would run on a dummy event and even compares the output would be a neat addition.

MijnheerD:
I don't have the bandwidth right now to implement such a thing, so I will just merge this without the test. I think we should be careful with requiring more effort to contribute to this repo, cause it will scare people off (we also talked about this in the context of the typing discussion). Maybe if there was a clear test template it could work. In my head this would consist of one dummy event everyone can use, with a clear description of what is in the NUR file and an example script where the developer could just slot in their module and have something running in 30min.

NuRadioReco/modules/efieldGalacticNoiseAdder.py
```diff
@@ -0,0 +1,129 @@
...
+
+        passband_filter = (freqs > passband[0]) & (freqs < passband[1])
+
+        site_latitude, site_longitude = detector.get_site_coordinates(station.get_id())
+        station_time = station.get_station_time()
+
+        local_coordinates = get_local_coordinates((site_latitude, site_longitude), station_time, self.__n_side)
+
+        # n_ice = ice.get_refractive_index(-0.01, detector.get_site(station.get_id()))
+        n_air = ice.get_refractive_index(depth=1, site=detector.get_site(station.get_id()))
```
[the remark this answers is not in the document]
MijnheerD: This is a copy-paste from the `channelGalacticNoiseAdder` , with just the ice part commented out for when someone wants to add it later down the line. How would you do it differently?

pyproject.toml
```diff
@@ -63,7 +64,7 @@ pre-commit = "*"
 [tool.poetry.extras]
 dev = ["pre-commit", "Sphinx", "sphinx-rtd-theme", "numpydoc"]
 proposal = ["proposal"]
-galacticnoise = ['pygdsm', 'pylfmap']
+galacticnoise = ['pygdsm', 'pylfmap', 'healpy']
```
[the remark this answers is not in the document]
MijnheerD: Hmm, then maybe I can just revert this change? I added it because my IDE was complaining about `healpy` not being installed (but not the other packages), so I thought that it somehow did slip through. But probably I did not have `pygdsm` installed either, it just did not trigger the checker because it is inside this `try` block.

NuRadioReco/modules/efieldGalacticNoiseAdder.py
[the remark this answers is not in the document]
sjoerd-bouma: I would have done it the same way as in the channelGalacticNoiseAdder, because I would assume this module does the same thing : ) but if it's additional effort then I don't think it's the end of the world if this doesn't work for in-ice detectors at all. Would maybe be nice to raise an explicit warning in that case, though.

NuRadioReco/modules/efieldGalacticNoiseAdder.py
```diff
@@ -108,6 +108,11 @@ def run(
 
             for field in station.get_electric_fields():
                 field_pos = field.get_position()
+                if field_pos[2] < 0:
+                    logger.warning(
+                        f"ElectricField {field.get_unique_identifier()} has field position {field_pos} below the surface."
+                        f"This is not yet supported, as ice properties are not implemented in this module.")
+
```
sjoerd-bouma, line 115: I think this might raise a warning for LOFAR as well, because some antennas (and therefore efields) are at negative z compared to the station? Alternatively, if it's not urgent I'm happy to take a look and copy over the same logic as in the channelGalacticNoiseAdder but as Felix knows this means this PR might lie around for a bit.

fschlueter: efields/antennas are relative to the station. In the "detector layout" files I got from artur are the channel positions only in the x-y plane. But we can also change this to `< -0.1` to be save?

sjoerd-bouma: That's SKA, probably, not LOFAR? For LOFAR we definitely have some antennas at negative z relative to the station height of 7.6 m. This will only raise spurious warnings, which I think is not as bad as not raising a warning when the module is doing something 'wrong', but it would obviously be better to either include a check to see if we are at an in-air site, or just copy over the logic from the channelGalacticNoiseAdder.

MijnheerD: The lowest `ant_position_z` I could find in the LOFAR json file is -0.0074, so Felix's suggestion would indeed solve this. Is there a better way to find if a site is in-air?

MijnheerD: Because actually the check in `channelGalacticNoiseAdder` is also on the channel's relative position, so how does it not trigger for LOFAR in that module?

MijnheerD:
Maybe an idea to discuss in a different PR: should we add in-air/in-ice attributes to our `Detector` class? Just like we have `set_is_neutrino()` and `set_is_cosmic_ray()` for `Event`s, I think it would make sense to simply define the `Detector` to be above/below ground.

fschlueter:
> Maybe an idea to discuss in a different PR: should we add in-air/in-ice attributes to our `Detector` class? Just like we have `set_is_neutrino()` and `set_is_cosmic_ray()` for `Event`s, I think it would make sense to simply define the `Detector` to be above/below ground.

I am not sure I get the point. We can define whether an detector is above below ground by the z-coordinate. I am not a big fan fo `is_neutrion/is_cosmic_ray` in the event class

NuRadioReco/modules/efieldGalacticNoiseAdder.py
sjoerd-bouma: because `ice` knows that lofar is not an in-ice site, so all fresnel coefficients etc. will just be 1.

sjoerd-bouma:
> Maybe an idea to discuss in a different PR: should we add in-air/in-ice attributes to our `Detector` class? Just like we have `set_is_neutrino()` and `set_is_cosmic_ray()` for `Event`s, I think it would make sense to simply define the `Detector` to be above/below ground.

Right now for in-ice detectors we store ice models and detectors separately, but there is a use case there also for tying them together. For in-air detectors, we currently 'rely' on the [ice](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.utilities.ice.html#module-NuRadioReco.utilities.ice) utility, which knows based on the 'site' if it is in-air or in-ice. But a) this interface should be deprecated at some point and b) this is error-prone, not least because the list of `site`s needs to be maintained manually. So I agree storing this information in the `Detector` might make more sense.

sjoerd-bouma:
For now I've just copy-pasted the in-ice code from the channelGalacticNoiseAdder to the efieldGalacticNoiseAdder. As long as the site is known, this should not change any behaviour compared to how it was previously, and I prefer it this way because at least the two modules are now consistent. I hope this works for you @MijnheerD, otherwise I'm happy to revert things and discuss an alternative that would work better.

Medium to long-term, there is a lot of duplicated code now, because internally the channelGalacticNoiseAdder already calculated efields, it just also folded in the antenna response. Maybe if we split those two things up we can save ourselves some trouble in the future if/when development on the two modules diverges...

MijnheerD:
> @MijnheerD if you're happy with my changes, feel free to merge. I think it might be useful to revisit this at some point in the future and just separate out the efield part in the channelGalacticNoiseAdder so we don't need to maintain duplicate code in the two modules, but this doesn't need to be on your plate.

Very well, I will merge the changes as they are. Concerning the consolidation of the two classes, my first thought would be to invert the relationship, ie make `channelGalacticNoiseAdder` a child of `efieldGalacticNoiseAdder` and pull out the electric field generation into a separate function. The `run()` function of `efielfGalacticNoiseAdder` would then become quite simple, but it would allow `channelGalacticNoiseAdder`s one to just call it and apply the antenna model on it.

PR #862, opened 2025-03-11
Add caching to the efieldToVoltageConverter

PhilippLaub:
Implements a caching of the antenna pattern in the `efieldToVoltageConverter` module (as already implemented in `channelGalacticNoiseAdder`). Also includes a check for array length when comparing frequency arrays.


fschlueter:
Thumbs up for using the labels :)

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -86,7 +91,18 @@ def begin(self, debug=False, uncertainty=None,
             for iCh in self.__uncertainty['sys_amp']:
                 self.__uncertainty['sys_amp'][iCh] = np.random.normal(1, self.__uncertainty['sys_amp'][iCh])
 
-        self.antenna_provider = antennapattern.AntennaPatternProvider()
```
fschlueter, line 89: I am not sure if we want to add a warning that we deprecated this interface or if we want to keep it. For example like this:
```
@property
def antenna_provider(self):
    return self.__antenna_provider
```

NuRadioReco/utilities/geometryUtilities.py
```diff
@@ -169,3 +169,36 @@ def get_fresnel_r_s(zenith_incoming, n_2=1.3, n_1=1.):
     n = n_2 / n_1
     return np.conjugate((np.cos(zenith_incoming) - SM.sqrt(n**2 - np.sin(zenith_incoming)**2)) / \
                         (np.cos(zenith_incoming) + SM.sqrt(n**2 - np.sin(zenith_incoming)**2)))
+
+
+def fresnel_factors_and_signal_zenith(detector, station, channel_id, zenith):
```
fschlueter, line 174: Can you add a doc-string?

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -219,9 +234,38 @@ def run(self, evt, station, det, channel_ids=None):
...
+                            self.__freqs = ff
+                            self._get_cached_antenna_response.cache_clear()
+                            self.logger.warning(
+                                "Frequencies have changed (values). Clearing antenna response cache. "
+                                "(If this happens often, something might be wrong...")
+
+
+                if self.__caching:
+                    zenith_antenna, t_theta, t_phi = geo_utl.fresnel_factors_and_signal_zenith(det, station, channel_id, zenith)
+                    antenna_model = det.get_antenna_model(station.get_id(), channel_id, zenith_antenna)
```
fschlueter, line 259: I think you need here `sim_station` instead of `station`

PhilippLaub: Hm, I'm not sure about that. The station object is used only to get the station ID and to check if it is set as a CR event.

NuRadioReco/modules/efieldToVoltageConverter.py
PhilippLaub: Good point! I will add this with a warning.

NuRadioReco/modules/efieldToVoltageConverter.py
fschlueter: yes. and as you can see the tests are mostly failing because of the latter right? Also the original code used `sim_station`.

NuRadioReco/utilities/geometryUtilities.py
fschlueter: Thanks for doing this @PhilippLaub but can you add a `
```
Parameters
-----------
```
and
```
Returns
-------
```
section like in (most) other functions.

PhilippLaub: Sure. While I'm at it I can also add these sections to the other functions in the file.

fschlueter: <3

fschlueter (approved):
@PhilippLaub thanks a lot!

NuRadioReco/utilities/geometryUtilities.py
```diff
@@ -113,9 +145,23 @@ def get_fresnel_t_p(zenith_incoming, n_2=1.3, n_1=1.):
     this polarization corresponds to the eTheta polarization
 
     parallel and perpendicular refers to the signal's polarization with respect
-    to the 'plane of incident' which is defindes as: "the plane of incidence
+    to the 'plane of incident' which is definded as: "the plane of incidence
```
sjoerd-bouma, line 148: I also prefer the new typo :upside_down_face:

NuRadioReco/utilities/geometryUtilities.py
```diff
@@ -162,10 +236,87 @@ def get_fresnel_r_s(zenith_incoming, n_2=1.3, n_1=1.):
...
+    -------
+    float
+        Fresnel coefficient r for phi (perpendicular) polarization
     """
     n = n_2 / n_1
     return np.conjugate((np.cos(zenith_incoming) - SM.sqrt(n**2 - np.sin(zenith_incoming)**2)) / \
                         (np.cos(zenith_incoming) + SM.sqrt(n**2 - np.sin(zenith_incoming)**2)))
+
+
+def fresnel_factors_and_signal_zenith(detector, station, channel_id, zenith):
```
sjoerd-bouma, line 262: I'm a little bit torn about this function - does it actually get use outside of `trace_utilities`? I guess for in-air it is useless, and for in-ice it actually **shouldn't** be used because it relies on the to-be-deprecated `ice` interface.

NuRadioReco/utilities/geometryUtilities.py
PhilippLaub: ups :)

NuRadioReco/utilities/geometryUtilities.py
PhilippLaub: If the `ice` interface is going to be deprecated in the future, what will replace it?

fschlueter: And for FAERIE we do to. I had to set `is_neutrino` for FAERIE simulations to avoid using this function. I would argue we should abandon the concept of `is_cosmic_ray`/`is_neutrino` and rather check if the efield position is at the antenna position.

fschlueter: @sjoerd-bouma we introduced this function since it is now used in directly in the efieldToVoltage converter too if `caching==True`. Furthermore it fits better in geometryUtilities than trace_utilities. But I agree a better doc string and warning would be good. @PhilippLaub shall I do this?

PhilippLaub: I'm on it

NuRadioReco/utilities/geometryUtilities.py
```diff
@@ -162,10 +236,88 @@ def get_fresnel_r_s(zenith_incoming, n_2=1.3, n_1=1.):
...
+        zenith_antenna = get_fresnel_angle(zenith, n_ice, 1)
+        t_theta = get_fresnel_t_p(zenith, n_ice, 1)
+        t_phi = get_fresnel_t_s(zenith, n_ice, 1)
+        logger.debug(("Channel {:d}: electric field is refracted into the firn. "
+                        "theta {:.0f} -> {:.0f}. Transmission coefficient p (eTheta) "
+                        "{:.2f} s (ePhi) {:.2f}".format(
+                            channel_id, zenith / units.deg, zenith_antenna / units.deg, t_theta, t_phi)))
+    elif position[2] > 0:
+        # now the signal is coming from below, do we have an antenna above the surface?
+        zenith_antenna = get_fresnel_angle(zenith, 1., n_ice)
```
sjoerd-bouma, line 317: Without having looked at this in a lot of detail - I don't think this does the same as before, at least the `if/elif` doesn't cover the same cases as `if`/`else` before (e.g. position[2] < 0 but no cosmic ray / signal from above), right?

NuRadioReco/utilities/geometryUtilities.py
```diff
@@ -162,10 +236,88 @@ def get_fresnel_r_s(zenith_incoming, n_2=1.3, n_1=1.):
...
+    n_ice = ice.get_refractive_index(-0.01, detector.get_site(station.get_id()))
+
+    # no reflection/refraction at the ice-air boundary
+    zenith_antenna = zenith
+    t_theta = 1.
+    t_phi = 1.
+
+    position = detector.get_relative_position(station.get_id(), channel_id)
+    if position[2] < -3 * units.m:
+        logger.warning("This function might return inaccurate results. Consider using raytracing instead")
```
sjoerd-bouma, line 303: Thanks for adding a warning! Can you make it slightly more specific and say `This function might return inaccurate results for deep in-ice antennas`?

sjoerd-bouma (approved):
Thanks @PhilippLaub !

fschlueter:
@PhilippLaub I want to use these changes hence I take the liberty to merge it, thank you!

fschlueter:
I think something went wrong here. The warning we implemented with this PR is swamping the command line in neutrino simulations. I guess we have to move the warning into the if condition?

PhilippLaub:
Do you mean the deprecation warning?

PR #863, opened 2025-03-11
Implement data drive noise in simulation

fschlueter:
Coming soon.

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -603,6 +603,9 @@ def apply_det_response(
                 # normalize noise level to the bandwidth its generated for
                 Vrms[channel_id] = Vrms_per_channel[station.get_id()][channel_id] / (norm / max_freq) ** 0.5
 
+            # Here we can only add "rayleigh" noise and not "data-driven" noise because we
+            # apply the hardware response afterwards with detector_simulation_filter_amp.
+            # If you want to use data-driven noise, you need to define detector_simulation_part2
             channelGenericNoiseAdder.run(
                 evt, station, det, amplitude=Vrms, min_freq=0 * units.MHz,
                 max_freq=max_freq, type='rayleigh',
@@ -1221,6 +1224,11 @@ class simulation:
             logger.error(err_msg)
             raise ValueError(err_msg)
 
+        # If you want to simulate data-driven noise you have to provide the detector_simulation_part2 function
+        if self._config["noise_type"] == "data-driven" and self.detector_simulation_part2 is None:
+            logger.error("Data-driven noise is enabled but no detector_simulation_part2 function is provided.")
+            raise ValueError("Data-driven noise is enabled but no detector_simulation_part2 function is provided.")
+
         # Initialize detector
         self._antenna_pattern_provider = antennapattern.AntennaPatternProvider()
         if det is None:
@@ -1678,7 +1686,8 @@ class simulation:
                     # because we need to add noise to traces where the amplifier response
                     # was already applied to.
                     if bool(self._config['noise']):
-                        self.add_filtered_noise_to_channels(evt, station, non_trigger_channels)
+                        self.add_filtered_noise_to_channels(
+                            evt, station, non_trigger_channels, noise_type=self._config['noise_type'])
 
                     channelSignalReconstructor.run(evt, station, self._det)
                     self._set_event_station_parameters(evt)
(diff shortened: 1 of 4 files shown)


fschlueter:
Currently this PR is based on #854 and #844. The diff will simplify once these PRs are merged.

fschlueter:
There is (at least) one thing we might want to consider which would imply larger changes. The NuRadioReco modules used in `simulation.py` are not a property of the `simulation` class. Hence if we have to create a second object of those modules if we want to use them in functions like `simulate_detector_part2`. I think this is potentially a source of nasty bugs.

fschlueter:
I have been fighting quite a bit to get the normalisation for the data-driven noise correct. To be honest I am not sure that the normalization I found is correct. I only convinced my self that this is (maybe) correct by verifying that the data driven noise spectra scale the same way as the rayleigh simulated noise when you vary the sampling rate or number of samples for a fixed temperature (since you change the bandpass with the sampling rate you change amplitude you are passing to the module)! 

That the spectra change as a function of the number of samples might be unintuitive on first glance but is due to our FFT normalisation convention (see our [documentation](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.utilities.fft.html#module-NuRadioReco.utilities.fft)), i.e., what we calculate with our FFT is not the power density spectrum but the energy density spectrum. Hence - it is expected that the energy density as a function of the frequency when the traces are getting longer (which is the effect when varying either).

@sjoerd-bouma or anybody else - I would appreciate if you would critically check this change.

image

NuRadioMC/simulation/simulation.py
```diff
@@ -1417,8 +1425,12 @@ def get_distance_cut(shower_energy):
                                                     self._propagator.get_number_of_raytracing_solutions(),
                                                     particle_mode=particle_mode)
 
-        efieldToVoltageConverter.begin()
-        channelGenericNoiseAdder.begin(seed=self._config['seed'])
+
+        # Only needed for data-driven noise
+        scale_dir = os.path.join(os.path.dirname(__file__), "../examples/RNO_G_trigger_simulation/data_driven_noise_files")
```
sjoerd-bouma, line 1430: egh... does this have to be in simulation.py?

NuRadioReco/modules/RNO_G/dataProviderRNOG.py
```diff
@@ -16,10 +16,10 @@ class dataProviderRNOG:
...
     This module is a wrapper around the following modules (in this order):
 
-    #. `NuRadioReco.modules.io.RNO_G.readRNOGDataMattak`
-    #. `NuRadioReco.modules.RNO_G.channelGlitchDetector`
-    #. `NuRadioReco.modules.RNO_G.channelBlockOffsetFitter`
-    #. `NuRadioReco.modules.channelAddCableDelay`
+    - `NuRadioReco.modules.io.RNO_G.readRNOGDataMattak`
+    - `NuRadioReco.modules.RNO_G.channelGlitchDetector`
+    - `NuRadioReco.modules.RNO_G.channelBlockOffsetFitter`
+    - `NuRadioReco.modules.channelAddCableDelay`
```
sjoerd-bouma, line 22: Why?

sjoerd-bouma (commented):
Haven't really followed this, but what is data driven noise? Is it just [bandlimited_noise_from_spectrum](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.modules.channelGenericNoiseAdder.html#NuRadioReco.modules.channelGenericNoiseAdder.channelGenericNoiseAdder.bandlimited_noise_from_spectrum) with a spectrum parameterised from RNO-G data? I'm not a big fan of how there's RNO-G specific code in simulation.py and the channelGenericNoiseAdder now, though part of that already seems to have been introduced by #854.

NuRadioReco/modules/RNO_G/dataProviderRNOG.py
fschlueter: This list could be rendered by my IDE, the one before not. Since I am using VSCode I assume the experience is the same for may users.

NuRadioMC/simulation/simulation.py
fschlueter: We can probably make it configureable - I can do that.

fschlueter:
> Haven't really followed this, but what is data driven noise? Is it just [bandlimited_noise_from_spectrum](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.modules.channelGenericNoiseAdder.html#NuRadioReco.modules.channelGenericNoiseAdder.channelGenericNoiseAdder.bandlimited_noise_from_spectrum) with a spectrum parameterised from RNO-G data? I'm not a big fan of how there's RNO-G specific code in simulation.py and the channelGenericNoiseAdder now, though part of that already seems to have been introduced by #854.

No it is not and it is (in principle*) also not RNO-G specific. It is using a parameterization of the rayleigh parameter to generate noise. This means it could be used by any experiment. 

*Currently It is using a hardcoded 3.2 GHz and 2048 in the normalisation. We can change that. 

Would you be happier if we renamed the noise_type to something like `parameterized`?

PR #867, opened 2025-03-12
Adding Hit Filter to NuRadioReco

martin01021207:
Adding the script of Hit Filter to NuRadioReco/modules/RNO_G/


NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,462 @@
...
+        complete_time_check: bool (default: False)
+            If users want the time checker to run through all channel groups
+        complete_hit_check: bool (default: False)
+            If users want the high hit checker to run through all channels
+        time_window: float (default: 10.0*units.ns)
+            Coincidence window for two adjacent channels
+        is_multi_thresholds: bool (default: False)
+            If users want to check with multiple hit thresholds
+        threshold_multipliers: 1-D list (default: [6.5, 6.0, 5.5, 5.0, 4.5])
+            Different hit threhold levels, if not multiple threholds only the first element is used
```
fschlueter, line 40: typo

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,462 @@
...
+from NuRadioReco.utilities import trace_utilities, units
+
+import numpy as np
+import logging
+
+logger = logging.getLogger('NuRadioReco.RNO_G.stationHitFitter')
+
+class stationHitFilter:
+
+    def __init__(self, complete_time_check = False, complete_hit_check = False, time_window = 10.0*units.ns, is_multi_thresholds = False, threshold_multipliers = [6.5, 6.0, 5.5, 5.0, 4.5]):
```
fschlueter, line 16: A style comment. We try to follow the pep8 style guide lines which suggest to not add spaces around the `=` in function definitions. See https://peps.python.org/pep-0008/#other-recommendations or a screenshot of the relevant recommandation:
image

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,462 @@
...
+        """
+        self._complete_time_check = complete_time_check
+        self._complete_hit_check = complete_hit_check
+        self._dT = time_window
+        self._is_multi_thresholds = is_multi_thresholds
+
+        if self._is_multi_thresholds:
+            self._threshold_multipliers = threshold_multipliers
+        else:
+            self._threshold_multipliers = [ threshold_multipliers[0] ]
```
fschlueter, line 50: can you also remove the spaces here?

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,462 @@
...
+        self._n_channel_pairs_in_PA = len(self._channel_pairs_in_PA)
+
+        self._envelope_max_time_index = np.array( [None] * self._n_channels_in_ice )
+        self._envelope_max_time = np.array( [None] * self._n_channels_in_ice )
+        self._noise_RMS = np.array( [None] * self._n_channels_in_ice )
+        self._times = [ [] for i in range(self._n_channels_in_ice) ]
+        self._trace = [ [] for i in range(self._n_channels_in_ice) ]
+        self._envelope_trace = [ [] for i in range(self._n_channels_in_ice) ]
+        self._hit_thresholds = [ [] for i in range(self._n_channels_in_ice) ]
+        self._is_over_hit_threshold = [ [] for i in range(self._n_channels_in_ice) ]
```
fschlueter, line 68: and here

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,462 @@
...
+            Output channel number
+        """
+        if channel_in >= 21:
+            channel_out = channel_in - 9
+        else:
+            channel_out = channel_in
+        return channel_out
+
+
+    def _set_up(self, set_of_traces, set_of_times, RMS):
```
fschlueter, line 117: no variable names starting with capital letters please

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,462 @@
...
+        if not self._complete_hit_check:
+            if self.is_passed_time_checker():
+                self._passed_hit_filter = True
+                return self._passed_hit_filter
+            else:
+                self._hit_checker()
+                if self.is_passed_hit_checker():
+                    self._passed_hit_filter = True
+        else:
+            self._hit_Checker()
```
fschlueter, line 274: typo in function name

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,462 @@
...
+        self._time_checker()
+
+        if not self._complete_hit_check:
+            if self.is_passed_time_checker():
+                self._passed_hit_filter = True
+                return self._passed_hit_filter
+            else:
+                self._hit_checker()
+                if self.is_passed_hit_checker():
+                    self._passed_hit_filter = True
```
fschlueter, line 272: I am a bit confused. In l265 it says `if not self._complete_hit_check`. Why would you still run it when the time_checker fails?

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,462 @@
...
+        and another coincident pair in any group. When the time check fails, the Hit Filter will
+        see if there's any in-ice channel that has a high hit (maximum > 6.5*RMS), which is the
+        hit checker, and whenever there's a high hit it passed the Hit Filter.
+
+        Parameters
+        ----------
+        complete_time_check: bool (default: False)
+            If users want the time checker to run through all channel groups
+        complete_hit_check: bool (default: False)
+            If users want the high hit checker to run through all channels
```
fschlueter, line 34: I might not quite understand what these bools do. Can you elaborate?

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,462 @@
...
+            for i in range(self._n_thresholds):
+                self._hit_thresholds[i_channel].append(None)
+                self._is_over_hit_threshold[i_channel].append(None)
+
+        self._passed_time_checker = None
+        self._passed_hit_checker = None
+        self._passed_hit_filter = None
+
+
+    def _channel_mapping(self, channel_in):
```
fschlueter, line 91: is that function really needed? Cant we always have list/arrays of length 24 and only access the indexes we are interested in?

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,462 @@
...
+        if self._is_multi_thresholds:
+            self._threshold_multipliers = threshold_multipliers
+        else:
+            self._threshold_multipliers = [ threshold_multipliers[0] ]
+
+        self._n_thresholds = len(self._threshold_multipliers)
+
+        self._in_ice_channels = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 21, 22, 23]
+        self._in_ice_channel_groups = ([0, 1, 2, 3], [9, 10], [23, 22], [8, 4], [5], [6], [7], [11], [21])
+        self._channel_pairs_in_PA = ([0, 1], [0, 2], [0, 3], [1, 2], [1, 3], [2, 3])
```
fschlueter, line 56: In order to make the module more general, those could be passed to the module?

fschlueter: A conceptional why are channels 4 and 8 not in one group with the PA? And why are the 3 antennas on the helper stings not each one group (i.e., why are you splitting up between vpols and hpols)?

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,462 @@
...
+        self._noise_RMS = np.array( [None] * self._n_channels_in_ice )
+        self._times = [ [] for i in range(self._n_channels_in_ice) ]
+        self._trace = [ [] for i in range(self._n_channels_in_ice) ]
+        self._envelope_trace = [ [] for i in range(self._n_channels_in_ice) ]
+        self._hit_thresholds = [ [] for i in range(self._n_channels_in_ice) ]
+        self._is_over_hit_threshold = [ [] for i in range(self._n_channels_in_ice) ]
+
+        self._in_time_sequence = []
+        for i_group, group in enumerate(self._in_ice_channel_groups):
+            if i_group == 0:
```
fschlueter, line 72: Group 0 has a special importance right? It would be more general if one would pass another list with a flag which tells you which group is important (this way you can also have non or several).

fschlueter (commented):
Hi @martin01021207, thank you very much for this module! I have to admit I have not yet fully understood everything. But I though I leave a first round of comments and get back to it after you replies.

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: Fixed

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: Just fixed this.

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: Fixed

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: Fixed

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: Fixed

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: Thank you for this, I fixed it.

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter I personally think that skipping those 9 surface channels would be easier and faster, because I am using the Hilbert Transform only for in-ice channels so the length of the array is 15, and same for other output arrays, they are consistent in length. If I were to use all channels and then select the 15, it might cause confusions on those outputs (some channels were processed and some were not), or I might need to add a condition to select the 15 in many places if that's the case.

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter I made it private because I don't want users to select the wrong channels or to be able to change it, it's strict to that order.

I thought it would be good to separate Vpols and Hpols, no? Hpols are usually more noisy, right? I don't want to increase the probability of random noise passing filter, it's currently 5%.

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter 
`self._in_time_sequence` is a 2-D list, so groups are separate in there, but you just raised a very good point, thank you! I think I can add one more getter function to only return the list for PA.

martin01021207:
@fschlueter 
I fixed all the typos and made sure the coding style is following the standard way.

And I've also added an extra getter function to return the coincidence for the PA group, it returns a dictionary, so one can specify the pair by using, for example:
is_in_time_sequence_PA[(0,1)]
Then it will give us True or False
Those keys are: (0,1), (0,2), (0,3), (1,2), (1,3), (2,3)

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: That is not what I meant but okay :)

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: I am not objecting that they are private but that they are hardcoded. Okay I can follow this argument. Can you post in this PR a link to one of your presentation where you describe the hit filter? Are it anyway useful to use the hpols than?

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: Can you improve the doc-string?

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,482 @@
...
+                    self._in_time_sequence[i_group].append(None)
+            elif len(group) > 1:
+                self._in_time_sequence.append([])
+                for i_channel_pair in range(len(group)-1):
+                    self._in_time_sequence[i_group].append(None)
+
+        for i_channel in range(self._n_channels_in_ice):
+            for i in range(self._n_thresholds):
+                self._hit_thresholds[i_channel].append(None)
+                self._is_over_hit_threshold[i_channel].append(None)
```
fschlueter, line 85: Cant you just write:
```
        self._hit_thresholds = [[None for _ in range(self._n_thresholds)] for i in range(self._n_channels_in_ice)]
        self._is_over_hit_threshold = [[None for _ in range(self._n_thresholds)]  for i in range(self._n_channels_in_ice)]
```

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: Can you maybe a comment about the structure and purpose of the 2d list?

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,482 @@
...
+        hit_thresholds = self.get_hit_thresholds()
+        n_counts_above_threshold = np.zeros(self._n_thresholds)
+
+        for i_channel in range(self._n_channels_in_ice):
+            for i in range(self._n_thresholds):
+                if np.amax(envelopes[i_channel]) > hit_thresholds[i_channel][i]:
+                    self._is_over_hit_threshold[i_channel][i] = True
+                    n_counts_above_threshold[i] += 1
+                else:
+                    self._is_over_hit_threshold[i_channel][i] = False
```
fschlueter, line 230: try something like this. Should be much faster
```
self._is_over_hit_threshold = np.max(envelopes, axis=-1) > hit_thresholds
n_counts_above_threshold += int(self._is_over_hit_threshold)
passed = np.any(self._is_over_hit_threshold)
```

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter Okay I will post a document here. To your question on using the Hpols, channels 4&8 are the only Hpol pair that we check for coincidence (they are the only Hpol pair anyway), and sometimes they are coincident, so I think it's still worthy to keep them. Other Hpols will only be checked for high hits (maximum > 6.5*noise_RMS), and when there's a high hit in one of the Hpols it's most likely not thermal noise, so I think the current structure should be fine.

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter Just added more details on that.

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter Just added a comment

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter Okay I changed it. Thanks for this, it's made the code look cleaner.

martin01021207:
Reference to the Hit Filter in Wiki:
https://radio.uchicago.edu/wiki/index.php/Deep_CR_search#Hit_Filter

martin01021207:
@fschlueter Please let me know what else needs to be changed, or if you think it looks okay now feel free to approve it, then I should merge this branch to develop after PR #790 is merged.

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: put in `_in_ice_channel_groups` the single hpols appear (as well as single vpols)

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: @martin01021207 I did not forget about the indexing and the for-loops. I do not think you need them. If `hit_thresholds` and `envelopes` are numpy arrays you do not have to loop over the individual elements. As a rule of thumb, you want to avoid using for-loops in python as much as possible (in big contrast to c++). Try to get rid as many for-loops as you can and replace them with (numpy) array operations.

fschlueter: just getting rid of the for-loop over the channels speed up the code for me by factor ~10, both for-loops where factor ~50

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,482 @@
...
+        envelope_max_time = self.get_envelope_max_time()
+        is_coincident_in_PA = [False, False, False]
+
+        for i_group, group in enumerate(self._in_ice_channel_groups):
+            if i_group == 0:
+                for i_channel_pair, channel_pair in enumerate(self._channel_pairs_in_PA):
+                    if abs(envelope_max_time[channel_pair[1]] - envelope_max_time[channel_pair[0]]) <= self._dT * abs(channel_pair[1] - channel_pair[0]):
+                        self._in_time_sequence[i_group][i_channel_pair] = True
+                    else:
+                        self._in_time_sequence[i_group][i_channel_pair] = False
```
fschlueter, line 181: Can be simplified

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,482 @@
...
+                        is_coincident_in_PA[2] = True
+                if sum(is_coincident_in_PA) >= 2:
+                    self._passed_time_checker = True
+                    if not self._complete_time_check:
+                        break
+            elif len(group) > 1:
+                if abs(envelope_max_time[self._channel_mapping(group[0])] - envelope_max_time[self._channel_mapping(group[1])]) <= self._dT:
+                    self._in_time_sequence[i_group][0] = True
+                else:
+                    self._in_time_sequence[i_group][0] = False
```
fschlueter, line 196: can be simplified

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,482 @@
...
+
+                if self._in_time_sequence[i_group][0] and sum(is_coincident_in_PA) >= 1:
```
fschlueter, line 198: Since you require a coincidence in the PA, you could break the code if you do not find a coincidence in the PA and only look for coincidences in other channel groups if you found a PA coincidence. If you do that you do not need this check here: `sum(is_coincident_in_PA) >= 1:`

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter Could you please elaborate? I don't quite get it what you mean here.

martin01021207: I am guessing that you are saying those single channels can be taken out, if yes then I agree. In my early code I didn't have `self._in_ice_channels = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 21, 22, 23]`, so I put in those single channels in `self._in_ice_channel_groups` and I didn't (I forgot to) remove them after I added `self._in_ice_channels` which makes things more convenient. Now, I just took those single channels out of `self._in_ice_channel_groups`.

fschlueter: yes

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: Fixed

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: Fixed

martin01021207:
@fschlueter Just fixed those parts, please take a look :)

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter I've just realized that we might still need that condition (`sum(is_coincident_in_PA) >= 1`) when `self._complete_hit_check=True` and `self._complete_time_check=True`, but you were absolutely right we could break the loop when there was no coincident pair in PA when `self._complete_time_check=False`.

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,479 @@
...
+        """
+        for channel in self._in_ice_channels:
+            mapped_channel = self._channel_mapping(channel)
+            if noise_RMS.size != 0:
+                self._noise_RMS[mapped_channel] = noise_RMS[channel]
+            else:
+                self._noise_RMS[mapped_channel] = trace_utilities.get_split_trace_noise_RMS(set_of_traces[channel])
+
+            for i in range(self._n_thresholds):
+                self._hit_thresholds[mapped_channel][i] = self._noise_RMS[mapped_channel] * self.get_threshold_multipliers()[i]
```
fschlueter, line 148: get rid of for-loop

martin01021207: @fschlueter It doesn't work when I take it out of the loop, I got errors

martin01021207: @fschlueter Okay I fixed it

fschlueter: Yes you have to convert `_hit_thresholds` to an array :)

martin01021207: @fschlueter Yes I just did

martin01021207:
@fschlueter Please take another look at the code and see if it's okay :)
I also added an example script, which was modified from `data_analysis_example_advanced.py`, applying the Hit Filter and saving events that passed the filter.

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: Hey Martin, I was suggesting something like this:
```
        envelopes = self.get_envelope_trace()
        hit_thresholds = self.get_hit_thresholds()

        self._is_over_hit_threshold = np.amax(envelopes, axis=-1) > hit_thresholds
        self._passed_hit_checker = np.any(self._is_over_hit_threshold)

        return self._passed_hit_checker
 ```

File: NuRadioReco/modules/RNO_G/stationHitFilter.py [hunk 18]
```diff
@@ -0,0 +1,478 @@
...
+            if self.is_passed_time_checker():
+                self._passed_hit_filter = True
+                return self._passed_hit_filter
+            else:
+                self._hit_checker()
+                if self.is_passed_hit_checker():
+                    self._passed_hit_filter = True
+        else:
+            self._hit_checker()
+            if self.is_passed_time_checker() or self.is_passed_hit_checker():
```
fschlueter, line 272: This is inefficient because now both are always calculated.

fschlueter:
@martin01021207 I pushed some changes myself - I hope that is okay for you, please let me know if you want to discuss the changes. They are aimed at improving performance and making the code more readable.

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: But why do we need this object - its there for book-keeping but we do not really need it right? In anyway I reduced it to a one-liner. Let me know if you are happy with that.

fschlueter:
@martin01021207 okay I went through the module and change quite a bit. I tested it to the level that it runs through. Can you check that it does what it is suppose to do?

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter But that only happens when` self._complete_hit_check=Ture`, and that's when the user wants to calculate both.

NuRadioReco/examples/RNOG/apply_hit_filter_example.py
```diff
@@ -0,0 +1,108 @@
+"""
```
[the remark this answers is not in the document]
martin01021207: @fschlueter But there are already many lines in the standard example, I think that adding the Hit Filter to the standard example would overwhelm newcomers more, it's supposed to be simple, no? What I actually want is a **ready to use script**, kinda like what I am doing myself with my analysis, but I save filtered events in a root file, the script here saves filtered events in a nur file.

martin01021207:
> @martin01021207 I pushed some changes myself - I hope that is okay for you, please let me know if you want to discuss the changes. They are aimed at improving performance and making the code more readable.

@fschlueter I agree with most of your changes, thank you for helping optimize the code! But I think one feature is missing now and the code doesn't work when `is_multi_thresholds=True`.

The missing feature is:

> is_coincident_in_PA = [False, False, False]
> ......
> if channel_pair[0] == 0 and self._in_time_sequence[i_group][i_channel_pair]:
>     is_coincident_in_PA[0] = True
> elif channel_pair[0] == 1 and self._in_time_sequence[i_group][i_channel_pair]:
>     is_coincident_in_PA[1] = True
> elif channel_pair[0] == 2 and self._in_time_sequence[i_group][i_channel_pair]:
>     is_coincident_in_PA[2] = True

This way it doesn't just choose any random two pairs, but it needs to be in sequence.

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -67,26 +61,21 @@ def __init__(self, complete_time_check=False, complete_hit_check=False, time_win
...
         # Each sub-list contains bool(s) indicating the COINCIDENCE of channel pair(s) in each group
         # self._in_time_sequence[0] is the PA, it has 6 pairs (6 bools)
         # Initiating each element with None, it will later be replaced with bool when running the time checker
-        self._in_time_sequence = []
-        for i_group, group in enumerate(self._in_ice_channel_groups):
-            if i_group == 0:
-                self._in_time_sequence.append([None for _ in range(self._n_channel_pairs_in_PA)])
-            else:
-                self._in_time_sequence.append([None for _ in range(len(group)-1)])
+        self._in_time_sequence = [[None for _ in range(int((len(group) - 1)**2))] for group in self._in_ice_channel_groups]
```
martin01021207: @fschlueter This way there are 9 pairs for Group 1 (PA), but it should be only 6 pairs

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter `is_multi_thresholds=True` stops working after this change

fschlueter:
> > @martin01021207 I pushed some changes myself - I hope that is okay for you, please let me know if you want to discuss the changes. They are aimed at improving performance and making the code more readable.
> 
> @fschlueter I agree with most of your changes, thank you for helping optimize the code! But I think one feature is missing now:
> 
> > is_coincident_in_PA = [False, False, False]
> > ......
> > if channel_pair[0] == 0 and self._in_time_sequence[i_group][i_channel_pair]:
> > is_coincident_in_PA[0] = True
> > elif channel_pair[0] == 1 and self._in_time_sequence[i_group][i_channel_pair]:
> > is_coincident_in_PA[1] = True
> > elif channel_pair[0] == 2 and self._in_time_sequence[i_group][i_channel_pair]:
> > is_coincident_in_PA[2] = True
> 
> This way it doesn't just choose any random two pairs, but it needs to be in sequence.

Oh, I did not get this from your slide, don’t you use in your slide an example where two pairs are not next to each other?

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: UPS, I guess the equation is 2**n-1 and not (n-1)**2

fschlueter: No that is also not working …

fschlueter:
> > @martin01021207 okay I went through the module and change quite a bit. I tested it to the level that it runs through. Can you check that it does what it is suppose to do?
> 
> @fschlueter The code doesn't work when `is_multi_thresholds=True`, I am still looking into it now. I got different results too: **Before:** Station 23 Run 1070 Number of forced trigger events: 718 Number of total RF events: 2599 Number of RF events passed the hitFilter: 140 **After:** Station 23 Run 1070 Number of forced trigger events: 718 Number of total RF events: 2599 Number of RF events passed the hitFilter: 2599
> 
> I will need to change back some features though.

sorry for this - it’s not your job to debug my code. But if you do it I would be happy. But I can take another look at it tmr

martin01021207:
> > > @martin01021207 okay I went through the module and change quite a bit. I tested it to the level that it runs through. Can you check that it does what it is suppose to do?
> > 
> > 
> > @fschlueter The code doesn't work when `is_multi_thresholds=True`, I am still looking into it now. I got different results too: **Before:** Station 23 Run 1070 Number of forced trigger events: 718 Number of total RF events: 2599 Number of RF events passed the hitFilter: 140 **After:** Station 23 Run 1070 Number of forced trigger events: 718 Number of total RF events: 2599 Number of RF events passed the hitFilter: 2599
> > I will need to change back some features though.
> 
> sorry for this - it’s not your job to debug my code. But if you do it I would be happy. But I can take another look at it tmr

@fschlueter I've changed it back to the original structure for major parts, although I would love to make the code more concise as much as you want it to be, you've already helped me a lot, and I think the code does look better now than how it was before (at least not too many nested loops now), so I think we should just go ahead and merge this branch if there's no more obvious changes that we can make, I don't think we should keep messing with the structure because it might take longer to review and re-test the code; furthermore, if we are going to make the Hit Filter more general in the future, many things will have to change too at that point.

martin01021207:
The script `apply_hit_filter.py` (I removed the word `_example`) is ready to use for users to load a run of data, exclude forced triggers, process RF events in the standard way, and then apply the Hit Filter and save only events that passed. So with this script, anyone can just use it without having to write themselves another script to apply the Hit Filter.

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: > There should be a equation

The only way is to use combination, that's the actual formula here, but I don't think that will make it easier to read and understand.

fschlueter:
> The script `apply_hit_filter.py` (I removed the word `_example`) is ready to use for users to load a run of data, exclude forced triggers, process RF events in the standard way, and then apply the Hit Filter and save only events that passed. So with this script, anyone can just use it without having to write themselves another script to apply the Hit Filter.

And why not adding it to the existing script?

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: `math.factorial(n-1)` works?

fschlueter: But maybe another question - why do we have to object?

fschlueter:
> > > > @martin01021207 I pushed some changes myself - I hope that is okay for you, please let me know if you want to discuss the changes. They are aimed at improving performance and making the code more readable.
> > > 
> > > 
> > > @fschlueter I agree with most of your changes, thank you for helping optimize the code! But I think one feature is missing now:
> > > > is_coincident_in_PA = [False, False, False]
> > > > ......
> > > > if channel_pair[0] == 0 and self._in_time_sequence[i_group][i_channel_pair]:
> > > > is_coincident_in_PA[0] = True
> > > > elif channel_pair[0] == 1 and self._in_time_sequence[i_group][i_channel_pair]:
> > > > is_coincident_in_PA[1] = True
> > > > elif channel_pair[0] == 2 and self._in_time_sequence[i_group][i_channel_pair]:
> > > > is_coincident_in_PA[2] = True
> > > 
> > > 
> > > This way it doesn't just choose any random two pairs, but it needs to be in sequence.
> > 
> > 
> > Oh, I did not get this from your slide, don’t you use in your slide an example where two pairs are not next to each other?
> 
> @fschlueter They don't have to be right next to each other but it needs to be in time sequence, for example: If the first pair is (0,2), then the second pair has to be one of the (1,2) (1,3) pairs or (2,3) to be True. If two pairs are (0,1)=True (0,3)=True but non of the (1,2) (1,3) (2,3) is True then it doesn't count 2 pairs. I didn't write this in the slides because it's too much details but essentially the 2 pairs in PA should be in time sequence.

Why?

fschlueter:
> > > The script `apply_hit_filter.py` (I removed the word `_example`) is ready to use for users to load a run of data, exclude forced triggers, process RF events in the standard way, and then apply the Hit Filter and save only events that passed. So with this script, anyone can just use it without having to write themselves another script to apply the Hit Filter.
> > 
> > 
> > And why not adding it to the existing script?
> 
> @fschlueter I replied to your same question above, I am quoting it here:
> 
> > @fschlueter But there are already many lines in the standard example, I think that adding the Hit Filter to the standard example would overwhelm newcomers more, it's supposed to be simple, no? What I actually want is a ready to use script, kinda like what I am doing myself with my analysis, but I save filtered events in a root file, the script here saves filtered events in a nur file.
> 
> Because this script is now ready to use for hit filtering specifically, if we put it with the standard example, then people will probably get confused with it.

Why would people get confused with it? Think about a situation where we have 5, 6 or 7 examples for RNO-G. This will be very confusing and difficult to maintain.

fschlueter:
@martin01021207 I do not think the module is currently in a situation where we should merge it. Can you tell me which run you use to verify the module?

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter `math.factorial(n-1)` can work but then we will need to import math for this, but if you think this is better then we can do it.

martin01021207: > But maybe another question - why do we have to object?

@fschlueter Because the number of elements is fixed for it to be used in the `time_checker()`, so we call it first, and then what happens next is in the `time_checker()`. Or if you think there's a better way please feel free to do it, I am not that advanced of a code writer.

martin01021207:
> @martin01021207 I do not think the module is currently in a situation where we should merge it. Can you tell me which run you use to verify the module?
@fschlueter 
S23 Run 1070

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: We do not need that 2d-list at all? What purpose has it?

martin01021207:
If you use the example script to run, these are the results:

> Total: 3577 events
> Forced Triggers: 718 events
> RF Triggers: 2859 events
> RF Passed Hit Filter: 158 events

fschlueter:
> > > > > @martin01021207 I pushed some changes myself - I hope that is okay for you, please let me know if you want to discuss the changes. They are aimed at improving performance and making the code more readable.
> > > > 
> > > > 
> > > > @fschlueter I agree with most of your changes, thank you for helping optimize the code! But I think one feature is missing now:
> > > > > is_coincident_in_PA = [False, False, False]
> > > > > ......
> > > > > if channel_pair[0] == 0 and self._in_time_sequence[i_group][i_channel_pair]:
> > > > > is_coincident_in_PA[0] = True
> > > > > elif channel_pair[0] == 1 and self._in_time_sequence[i_group][i_channel_pair]:
> > > > > is_coincident_in_PA[1] = True
> > > > > elif channel_pair[0] == 2 and self._in_time_sequence[i_group][i_channel_pair]:
> > > > > is_coincident_in_PA[2] = True
> > > > 
> > > > 
> > > > This way it doesn't just choose any random two pairs, but it needs to be in sequence.
> > > 
> > > 
> > > Oh, I did not get this from your slide, don’t you use in your slide an example where two pairs are not next to each other?
> > 
> > 
> > @fschlueter They don't have to be right next to each other but it needs to be in time sequence, for example: If the first pair is (0,2), then the second pair has to be one of the (1,2) (1,3) pairs or (2,3) to be True. If two pairs are (0,1)=True (0,3)=True but non of the (1,2) (1,3) (2,3) is True then it doesn't count 2 pairs. I didn't write this in the slides because it's too much details but essentially the 2 pairs in PA should be in time sequence.
> 
> Why?
> 
> @fschlueter This way we can make sure those two pairs are close enough with each other, they shouldn't just be two pairs separately.

The question why was targeted at: Why to we require the pairs to be next to each other. I do not see why (0, 1) and (2, 3) should pass but not (0, 1) and (0, 3) should not, In fact I believe for a neutrino the latter is more likely

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: In the newest update I've renamed it to `self._is_in_time_window` to avoid the confusion with the later `coincidences_in_time_sequence_PA = [False, False, False]` in the `time_checker()`. `self._is_in_time_window` is needed to save True or False for those channel pairs in groups if they are in the coincidence window or not. `coincidences_in_time_sequence_PA` is used in the `time_checker()` to check if the two pairs are in time sequence.

martin01021207:
> > > > > > @martin01021207 I pushed some changes myself - I hope that is okay for you, please let me know if you want to discuss the changes. They are aimed at improving performance and making the code more readable.
> > > > > 
> > > > > 
> > > > > @fschlueter I agree with most of your changes, thank you for helping optimize the code! But I think one feature is missing now:
> > > > > > is_coincident_in_PA = [False, False, False]
> > > > > > ......
> > > > > > if channel_pair[0] == 0 and self._in_time_sequence[i_group][i_channel_pair]:
> > > > > > is_coincident_in_PA[0] = True
> > > > > > elif channel_pair[0] == 1 and self._in_time_sequence[i_group][i_channel_pair]:
> > > > > > is_coincident_in_PA[1] = True
> > > > > > elif channel_pair[0] == 2 and self._in_time_sequence[i_group][i_channel_pair]:
> > > > > > is_coincident_in_PA[2] = True
> > > > > 
> > > > > 
> > > > > This way it doesn't just choose any random two pairs, but it needs to be in sequence.
> > > > 
> > > > 
> > > > Oh, I did not get this from your slide, don’t you use in your slide an example where two pairs are not next to each other?
> > > 
> > > 
> > > @fschlueter They don't have to be right next to each other but it needs to be in time sequence, for example: If the first pair is (0,2), then the second pair has to be one of the (1,2) (1,3) pairs or (2,3) to be True. If two pairs are (0,1)=True (0,3)=True but non of the (1,2) (1,3) (2,3) is True then it doesn't count 2 pairs. I didn't write this in the slides because it's too much details but essentially the 2 pairs in PA should be in time sequence.
> > 
> > 
> > Why?
> > @fschlueter This way we can make sure those two pairs are close enough with each other, they shouldn't just be two pairs separately.
> 
> The question why was targeted at: Why to we require the pairs to be next to each other. I do not see why (0, 1) and (2, 3) should pass but not (0, 1) and (0, 3) should not, In fact I believe for a neutrino the latter is more likely

(0,1) & (0,3) can pass if (1,3) also satisfies it, and it usually should. But it can be random noise, right? It then won't guarantee that (1,3) also satisfies.

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: Sorry I do not see that.

fschlueter:
> If you use the example script to run, these are the results:
> 
> > Total: 3577 events
> > Forced Triggers: 718 events
> > RF Triggers: 2859 events
> > RF Passed Hit Filter: 158 events
> 
> These are slightly different from the results that I showed you above because I got those results from my actual analysis script where burn sample events can be selected or excluded, but the results here came straight from running the example script (no burn events).
```
Total: 3577 events
Forced Triggers: 718 events
RF Triggers: 2859 events
RF Passed Hit Filter: 158 events
```
Okay, I got the same results as you (with my version of the code). Was just a little tweak. I will push it to another branch for you to checkout.

fschlueter:
@martin01021207 I pushed my version of your module to a new branch and create a PR. Please have a look. However, I texted you on slack because I have a hard time understanding the pair criteria you are using.

martin01021207:
> > If you use the example script to run, these are the results:
> > > Total: 3577 events
> > > Forced Triggers: 718 events
> > > RF Triggers: 2859 events
> > > RF Passed Hit Filter: 158 events
> > 
> > 
> > These are slightly different from the results that I showed you above because I got those results from my actual analysis script where burn sample events can be selected or excluded, but the results here came straight from running the example script (no burn events).
> 
> ```
> Total: 3577 events
> Forced Triggers: 718 events
> RF Triggers: 2859 events
> RF Passed Hit Filter: 158 events
> ```
> 
> Okay, I got the same results as you (with my version of the code). Was just a little tweak. I will push it to another branch for you to checkout.

@fschlueter It didn't work the way it was when `is_multi_thresholds=False` was used, but I fixed it following your code logic.

martin01021207:
> @martin01021207 I pushed my version of your module to a new branch and create a PR. Please have a look. However, I texted you on slack because I have a hard time understanding the pair criteria you are using.

@fschlueter Felix, I really appreciate your time reviewing the code and helping it look better than it was, I've learnt much from you in this PR, so now the Hit Filter is using your code logic (more advanced than my elementary level code writing). But I've made some small changes, they are:
1. `complete_time_check=False` (default)
    The original idea was that when we wanted to check all channel groups we would do
    `complete_time_check=True`. By default is False when we just want the time checker to be quick (not
    necessarily going through all groups). So now this part is as before.
2. Some changes in the `_hit_checker()`:
    It worked the same for just one threshold but didn't do the same for multiple thresholds, so I added a few
    lines to fix it (following your logic), now it works.
3. set_up() function:
    It now takes traces and times instead of the station object so one can use the Hit Filter outside of the
    standard NuRadio structure.
4. Some other small changes.

Please take a look at the code again, and feel free to make more changes if needed, or if it's all good now please approve it :)

martin01021207:
@fschlueter I deleted `apply_hit_filter.py` and added the Hit Filter application to those two example scripts.

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,484 @@
...
+        envelopes = self.get_envelope_traces()
+        hit_thresholds = self.get_hit_thresholds()
+
+        max_envelopes = np.amax(envelopes, axis=-1)
+
+        self._is_over_hit_threshold = max_envelopes[:, None] > hit_thresholds
+
+        n_counts_above_threshold = np.sum(self._is_over_hit_threshold, axis=0)
+
+        self._passed_hit_checker = np.any(n_counts_above_threshold > np.arange(self._n_thresholds))
```
fschlueter, line 188: I do not understand these two lines. `self._passed_hit_checker` is probably `True` if `n_counts_above_threshold > 0` but why making it so complicated?

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,484 @@
...
+
+        self.set_up(traces, times, noise_RMS)
+        self.apply_hit_filter()
+
+        return self.is_passed_hit_filter()
+
+
+    def end(self):
+        """(Unused)"""
+        pass
```
fschlueter, line 307: @martin01021207, I saw that you added some statistics to print in the example scripts. Can you add that to the end method?

NuRadioReco/examples/RNOG/data_analysis_example.py
```diff
@@ -108,24 +117,30 @@
...
             plt.close(fig)
 
+        # Check the trigger type in data, we want to skip forced triggers and apply the Hit Filter only to non-FT events
+        is_FT = info[idx].get('triggerType') == "FORCE"
+        n_events_FT += int(is_FT)
+        is_passed_HF = False
+        if not is_FT:
+            # Apply the Hit Filter to non-FT events
+            is_passed_HF = stationHitFilter.run(evt, station, det)
+            n_events_passed += int(is_passed_HF)
```
fschlueter, line 127: I actually think it would be interesting to also apply the hit filter to FT events but do the book keeping for them separately.

NuRadioReco/examples/RNOG/data_analysis_example.py
```diff
@@ -49,15 +51,22 @@
...
+    n_events_FT = 0
+    n_events_passed = 0
 
     # Loop over all events (the reader module has options to select events -
     # see class documentation or module arguements in config file)
     for idx, evt in enumerate(dataProviderRNOG.run()):
 
         if (idx + 1) % 50 == 0:
-            logger.info(f'"Processing events: {idx + 1}\r')
+            print(f'Processed events: {idx + 1}')
```
fschlueter, line 69: Please change this back

NuRadioReco/examples/RNOG/data_analysis_example.py
```diff
@@ -22,15 +23,15 @@
...
+    parser.add_argument('filenames', type=str, nargs="*",
                         help='Specify root data files if not specified in the config file')
     parser.add_argument('--outputfile', type=str, required=True, help='Specify the output file')
     parser.add_argument('--detectorfile', type=str, nargs=1, default=None,
                         help="Specify detector file. If you do not specified a file. "
                         "the description is queried from the database.")
 
     args = parser.parse_args()
-    nulogging.set_general_log_level(logging.WARNING)
+    nulogging.set_general_log_level(logging.ERROR)
```
fschlueter, line 34: Why changing this?

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter Having these two lines makes sure the code (your version) do the same as what my version was doing. So when `is_multi_thresholds=True`, we will have multiple thresholds going from high to low following the multipliers: [m0, m1, m2, m3, m4], so we have thresholds: [th0, th1, th2, th3, th4]. Here are the scenarios where the hit checking will be `True`:

> `1 hit > th0 (same as when we have only single threshold)`
> `2 hits > th1, but < th0`
> `3 hits > th2, but < th1`
> `4 hits > th3, but < th2`
> `5 hits > th4, but < th3`


It means more events can pass when that's the case, here are the results for multiple thresholds are used (same run as before):

> Total: 3577 events
> Forced Triggers: 718 events
> RF Triggers: 2859 events
> RF Passed Hit Filter: 472 events

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter Yeah but how do we count events in this end function?

NuRadioReco/examples/RNOG/data_analysis_example.py
martin01021207: @fschlueter Sure, we can do that

NuRadioReco/examples/RNOG/data_analysis_example.py
martin01021207: > Please change this back

@fschlueter I changed it because it wasn't printing things out or maybe there were just too much printout information so it got pushed away too quickly to be seen...

NuRadioReco/examples/RNOG/data_analysis_example.py
martin01021207: @fschlueter Have you tried to run the original version? There was so much printout information, it was very hard to keep up, also the block offset fitter kept printing out the status, it was overwhelming.

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter Take counts as an input argument and then use end() to print? But there can be different situations though, like different types of events that you want to count separately. I thought it would be easier to just count them in the working script.

NuRadioReco/examples/RNOG/data_analysis_example.py
martin01021207: @fschlueter I've made this change, so now the Hit Filter works on all events but count them separately.

NuRadioReco/examples/RNOG/data_analysis_example.py
fschlueter: With `WARNING` status messages should be be printed. Let me try it.

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: Sorry for the late reply. So the idea is that you could events, inside the module with "member variables" (e.g., `self._n_events` and `self._n_passed_events`). See this module for example: https://github.com/nu-radio/NuRadioMC/blob/develop/NuRadioReco/modules/io/RNO_G/readRNOGDataMattak.py

martin01021207:
@fschlueter Okay, I've added some logger information in the Hit Filter.

NuRadioReco/modules/RNO_G/stationHitFilter.py
fschlueter: Two things: If you define multiple thresholds the code will return `True` if the found maximum amplitude is above any of the thresholds which means if it is above the lowest of them. T**he information above which threshold the amplitude is is not currently stored. (I.e., you do not store the number of hits in your example but only that at least one hit occured)** Hence defining several thresholds does not make much sense. Unless you implement an option where you actually store what thresholds are passed and what not (for book-keeping purposes) the mulit-threshold option is equivalent to just setting a single threshold to the value of the lowest level in the list. And second, what the code currently does could be written in one line which is easier to understand and faster.

fschlueter:
To get the trigger type via the NR data structure do: `event.get_station().get_first_trigger().get_name()`

NuRadioReco/modules/RNO_G/stationHitFilter.py
martin01021207: @fschlueter You raised a really good point, so after thinking through it I've decided to just discard the multi-threshold feature, it was not helping much but only made the whole thing more complicated, and I am currently using only one threshold in the analysis myself anyway.

martin01021207:
> @martin01021207 thanks that is exactly what I had in mind. Two things though (sorry): We also have RADIANT (surface) trigger). You should either treat them the same way as Forced triggers or you skip them or treat them independently but I would not treat them the same way as LT trigger. Second thing: You are using the interface of the reader module to get the event information from mattak. The trigger type is also stored in the NuRadio data structure. So no need to do that.

@fschlueter Thanks for that, I didn't think about the surface trigger and didn't know I could just directly get the event info from the event object. I've made these changes.

martin01021207:
@fschlueter Please take another look at the code, now there's just one threshold, and no multi-thresholds anymore. The logger has been updated too.

NuRadioReco/modules/RNO_G/stationHitFilter.py
```diff
@@ -0,0 +1,525 @@
...
+            self.__count_passed_RADIANT += int(self.is_passed_hit_filter())
+            self.__is_wanted_trigger_type = False
+        elif trigger_type == "UNKNOWN":
+            self.__count_UNKNOWN += 1
+            self.__count_passed_UNKNOWN += int(self.is_passed_hit_filter())
+            self.__is_wanted_trigger_type = True
+        else:
+            self.__count_RF += 1
+            self.__count_passed_RF += int(self.is_passed_hit_filter())
+            self.__is_wanted_trigger_type = True
```
fschlueter, line 314: I thought an `"UNKNOWN"` trigger is always an RADIANT trigger @shallmann  or @cozzyd can you confirm? If so you can add in l290 `self.__is_wanted_trigger_type = False` and only set it to true if trigger same is `"LT"`

fschlueter: I think in principle the day can also produce "PPS" trigger although I have never seen them. I would make an explicit `trigger_type == "LT"` and have an `else` which deals with `"UNKNOWN"` and everything else.

fschlueter: I took the liberty of changing it

fschlueter (approved):
@martin01021207 thank you for your hard work. I did some changes myself - please have a look if you agree with them. If you do please wait 24h before merging.

fschlueter:
FYI the errors can be ignored and you can merge this PR after the 24h

PR #870, opened 2025-03-13
Remove artificial dependency on "is_cosmic_ray" in efieldToVoltageConverter modules

fschlueter:
This PR does three things:
- It replace the `sim_station.is_cosmic_ray()` condition with `dist_antenna_efield < 0.01 * units.mm` to add an additional time shift to the signals.
- It lets the module `efieldToVoltageConverterPerEfield` use the function `calculate_time_shift_for_cosmic_ray` defined in `efieldToVoltageConverter` to calculate the time shift
- It modifies `calculate_time_shift_for_cosmic_ray` in two ways. Instead of using the antenna position relative to the efield position to define the refractive index, it now uses the antenna position relative to the station and instead of the signal direction in sim_station it now uses the signal direction in the electric field.

### DIFF
--- a/NuRadioReco/modules/efieldToVoltageConverter.py
+++ b/NuRadioReco/modules/efieldToVoltageConverter.py
@@ -11,7 +11,7 @@ from NuRadioReco.framework.parameters import stationParameters as stnp
 
 from NuRadioReco.modules.base.module import register_run
 from NuRadioReco.detector import antennapattern
-from NuRadioReco.utilities import units, fft, ice, trace_utilities, geometryUtilities as geo_utl
+from NuRadioReco.utilities import units, fft, ice, signal_processing, geometryUtilities as geo_utl
 
 
 class efieldToVoltageConverter():
@@ -130,7 +130,7 @@ class efieldToVoltageConverter():
                 cab_delay = det.get_cable_delay(sim_station_id, channel_id)
                 t0 = electric_field.get_trace_start_time() + cab_delay
 
-                # if we have a cosmic ray event, the different signal travel time to the antennas has to be taken into account 
+                # if we have a cosmic ray event, the different signal travel time to the antennas has to be taken into account
                 if sim_station.is_cosmic_ray():
                     travel_time_shift = calculate_time_shift_for_cosmic_ray(det, sim_station, electric_field, channel_id)
                     t0 += travel_time_shift
@@ -223,7 +223,7 @@ class efieldToVoltageConverter():
                         start_bin = 0
 
                     new_trace[:, start_bin:stop_bin] = tr
-                
+
                 trace_object = NuRadioReco.framework.base_trace.BaseTrace()
                 trace_object.set_trace(new_trace, 1. / time_resolution)
 
@@ -269,7 +269,7 @@ class efieldToVoltageConverter():
                     VEL = [np.array([vel_tmp['theta'] * t_theta, vel_tmp['phi'] * t_phi])]
                 else:
                     # get antenna pattern for current channel
-                    VEL = trace_utilities.get_efield_antenna_factor(
+                    VEL = signal_processing.get_efield_antenna_factor(
                         sim_station, ff, [channel_id], det, zenith, azimuth, self.__antenna_provider)
 
                 if VEL is None:  # this can happen if there is not signal path to the antenna
--- a/NuRadioReco/modules/efieldToVoltageConverter.py
+++ b/NuRadioReco/modules/efieldToVoltageConverter.py
@@ -130,8 +130,10 @@ class efieldToVoltageConverter():
                 cab_delay = det.get_cable_delay(sim_station_id, channel_id)
                 t0 = electric_field.get_trace_start_time() + cab_delay
 
-                # if we have a cosmic ray event, the different signal travel time to the antennas has to be taken into account
-                if sim_station.is_cosmic_ray():
+                # If the efield is not at the antenna (as possible for a simulated cosmic ray event),
+                # the different signal travel time to the antennas has to be taken into account
+                dist_channel_efield = np.linalg.norm(det.get_relative_position(sim_station_id, channel_id) - electric_field.get_position())
+                if dist_channel_efield / units.mm > 0.01:
                     travel_time_shift = calculate_time_shift_for_cosmic_ray(det, sim_station, electric_field, channel_id)
                     t0 += travel_time_shift
 
@@ -189,11 +191,14 @@ class efieldToVoltageConverter():
                 # calculate the start bin
                 if not np.isnan(electric_field.get_trace_start_time()):
                     cab_delay = det.get_cable_delay(sim_station_id, channel_id)
-                    if sim_station.is_cosmic_ray():
+
+                    dist_channel_efield = np.linalg.norm(det.get_relative_position(sim_station_id, channel_id) - electric_field.get_position())
+                    if dist_channel_efield / units.mm > 0.01:
                         travel_time_shift = calculate_time_shift_for_cosmic_ray(
                             det, sim_station, electric_field, channel_id)
                     else:
                         travel_time_shift = 0
+
                     start_time = electric_field.get_trace_start_time() - times_min + cab_delay + travel_time_shift
                     start_bin = int(round(start_time / time_resolution))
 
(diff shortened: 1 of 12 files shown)


cg-laser:
> This PR does three things:
> 
> * It replace the `sim_station.is_cosmic_ray()` condition with `dist_antenna_efield < 0.01 * units.mm` to add an additional time shift to the signals.
> * It lets the module `efieldToVoltageConverterPerEfield` use the function `calculate_time_shift_for_cosmic_ray` defined in `efieldToVoltageConverter` to calculate the time shift
> * It modifies `calculate_time_shift_for_cosmic_ray` in two ways. Instead of using the antenna position relative to the efield position to define the refractive index, it now uses the antenna position relative to the station and instead of the signal direction in sim_station it now uses the signal direction in the electric field. This is potentially a bug fix

I didn't look though the code changes yet, but the first change might introduce a bug. We have this `cosmic_ray` flag also to determine if we need to refract the signal into the ice, as far as I recall. Where do we get this information from now?

fschlueter:
> > This PR does three things:
> > 
> > * It replace the `sim_station.is_cosmic_ray()` condition with `dist_antenna_efield < 0.01 * units.mm` to add an additional time shift to the signals.
> > * It lets the module `efieldToVoltageConverterPerEfield` use the function `calculate_time_shift_for_cosmic_ray` defined in `efieldToVoltageConverter` to calculate the time shift
> > * It modifies `calculate_time_shift_for_cosmic_ray` in two ways. Instead of using the antenna position relative to the efield position to define the refractive index, it now uses the antenna position relative to the station and instead of the signal direction in sim_station it now uses the signal direction in the electric field. This is potentially a bug fix
> 
> I didn't look though the code changes yet, but the first change might introduce a bug. We have this `cosmic_ray` flag also to determine if we need to refract the signal into the ice, as far as I recall. Where do we get this information from now?

Okay, I was under the impression that if the efield is at the antenna position no refracting into the ice is necessary - but if that is not the case this indeed is problematic.

fschlueter:
> > > This PR does three things:
> > > 
> > > * It replace the `sim_station.is_cosmic_ray()` condition with `dist_antenna_efield < 0.01 * units.mm` to add an additional time shift to the signals.
> > > * It lets the module `efieldToVoltageConverterPerEfield` use the function `calculate_time_shift_for_cosmic_ray` defined in `efieldToVoltageConverter` to calculate the time shift
> > > * It modifies `calculate_time_shift_for_cosmic_ray` in two ways. Instead of using the antenna position relative to the efield position to define the refractive index, it now uses the antenna position relative to the station and instead of the signal direction in sim_station it now uses the signal direction in the electric field. This is potentially a bug fix
> > 
> > 
> > I didn't look though the code changes yet, but the first change might introduce a bug. We have this `cosmic_ray` flag also to determine if we need to refract the signal into the ice, as far as I recall. Where do we get this information from now?
> 
> Okay, I was under the impression that if the efield is at the antenna position no refracting into the ice is necessary - but if that is not the case this indeed is problematic.

But is that ever the case? I would argue that is a problem in it self.

cg-laser (changes requested):
I went through the code in more detail. I agree, the things you changed are fine, but there is another problem that will cause problems (I think). 

In this function https://github.com/nu-radio/NuRadioMC/blob/develop/NuRadioReco/utilities/geometryUtilities.py#L306 the `is_cosmic_ray` flag is still needed for standard CoREAS. However, for FAERIE, the efield direction should already take into account the refraction into the ice. 

It almost seems as if a bug was already introduced when the `__caching` was implemented: https://github.com/nu-radio/NuRadioMC/pull/870/files#diff-b46191c8a58b5a7d6754d2dd51d04f08a21e8478b473c3b1a88a3ff296d0c3f0R268-R278
If I see it correctly, the refraction into the ice is only calculated when caching is activated?

fschlueter:
> It almost seems as if a bug was already introduced when the `__caching` was implemented: https://github.com/nu-radio/NuRadioMC/pull/870/files#diff-b46191c8a58b5a7d6754d2dd51d04f08a21e8478b473c3b1a88a3ff296d0c3f0R268-R278
> If I see it correctly, the refraction into the ice is only calculated when caching is activated?

@cg-laser no I think so. If caching is not used, the function `get_efield_antenna_factor` is used which internally calls `fresnel_factors_and_signal_zenith`!

fschlueter:
> In this function https://github.com/nu-radio/NuRadioMC/blob/develop/NuRadioReco/utilities/geometryUtilities.py#L306 the `is_cosmic_ray` flag is still needed for standard CoREAS. However, for FAERIE, the efield direction should already take into account the refraction into the ice.

Apparently, I forgot to change that call of `is_comic_ray` there too. Why would it still be needed for standard CoREAS? In standard CoREAS the `electric_field` will not be at the same position as the `channel` right? In this case we would still perform the additional porpagation.

fschlueter:
Okay - removing `is_cosmic_ray` from `geometryUtilities.py` requires quite a few code changes so I would like to ask first if everyone is happy with the concept before doing it.

fschlueter:
> Ok let me try to summarize the different use cases:
> 
> 1. air shower radio signals simulated with CoREAS (or similar) at ground level. We need to know transmit the efield into the ice which affects the signal strength and the propagation direction. So far this is done by changing the antenna response pattern. The property `station.is_cosmic_ray()` is used to identify these cases.
>    We also need to add additional propagation times.
> 2. air shower radio signals that are already propagated into the ice. Identifyable by having the efield at the same position as the antenna.
> 3. in-ice signals. They can be upgoing or downgoing. Also here, the electric field should be at the position of the antenna.
> 
> Did I miss any use case?
> 
> I agree that it would be good to get rid of the `is_cosmic_ray`/`is_neutrino` flag, because we need to use `is_neutrino` for all in-ice showers, e.g., pulser simulations, which is confusing.
> 
> The confusion probably arises from the `signal_processing.get_efield_antenna_factor()` function, which adjusts the antenna response for refraction. So only for 1).
> 
> Can we discuss a solution in the next simulation call?

No I think that summarises it. My question regarding case 1) is: Would there ever be the situation where we want to simulate an antenna in the ice and we want to simulate the effect of the reflection layer with `signal_processing.get_efield_antenna_factor()` but the electric field which is simulate at the surface (i.e., z=0) is to close to the antenna? I.e., would we ever put in-ice antennas with a depth of `|z| < 0.1 mm`?

Sure lets discuss this in the next call.

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -268,32 +268,35 @@ def run(self, evt, station, det, channel_ids=None):
...
+                        ff, zenith_antenna, azimuth, *antenna_orientation)
 
-                if VEL is None:  # this can happen if there is not signal path to the antenna
+                if vel is None:  # this can happen if there is not signal path to the antenna
                     voltage_fft = np.zeros_like(efield_fft[1])  # set voltage trace to zeros
                 else:
                     # Apply antenna response to electric field
-                    VEL = VEL[0]  # we only requested the VEL for one channel, so selecting it
-                    voltage_fft = np.sum(VEL * np.array([efield_fft[1], efield_fft[2]]), axis=0)
+                    vel = np.array([vel['theta'] * t_theta, vel['phi'] * t_phi])
```
cg-laser, line 305: is it a problem if you override `vel`?

cg-laser (commented):
great, this is much cleaner! Thanks a lot Felix.

NuRadioReco/modules/efieldToVoltageConverter.py
fschlueter: No I do not think that this is a problem

NuRadioReco/utilities/geometryUtilities.py
```diff
@@ -301,25 +315,31 @@ def fresnel_factors_and_signal_zenith(detector, station, channel_id, zenith):
...
-                        "{:.2f} s (ePhi) {:.2f}".format(
-                            channel_id, zenith / units.deg, zenith_antenna / units.deg, t_theta, t_phi)))
-    elif position[2] > 0:
+        logger.debug((
+            "Channel {:d}: electric field is refracted into the firn. "
+            "theta {:.0f} -> {:.0f}. Transmission coefficient p (eTheta) "
+            "{:.2f} s (ePhi) {:.2f}".format(channel_id, zenith / units.deg,
+            zenith_antenna / units.deg, t_theta, t_phi)))
+
+    elif position[2] > 0 and zenith > 0.5 * np.pi:
```
fschlueter, line 335: FYI I added the condition that the signal is upgoing (coming from below)

cg-laser (approved):

PR #871, opened 2025-03-14
Simplify function convert_obs_positions_to_nuradio_on_ground

fschlueter:
The benefit of this change:
- we do not have to initalize coordinatesystems
- It was somewhat confusing that you zenith and azimuth for this function (which you actually do not)

@MijnheerD is this function in the CI, I have not extensively tested the change.

### DIFF
--- a/NuRadioReco/modules/io/coreas/coreas.py
+++ b/NuRadioReco/modules/io/coreas/coreas.py
@@ -217,7 +217,7 @@ def convert_obs_positions_to_nuradio_on_ground(observer_pos, declination=0):
 
     obs_positions = np.array([hp.rotate_vector_in_2d(obs_pos, -declination) for obs_pos in obs_positions])
 
-    return obs_positions
+    return np.squeeze(obs_positions)
 
 # READER FUNCTIONS
 def read_CORSIKA7(input_file, declination=None, site=None):
--- a/NuRadioReco/modules/io/coreas/coreas.py
+++ b/NuRadioReco/modules/io/coreas/coreas.py
@@ -183,7 +183,7 @@ def convert_obs_to_nuradio_efield(observer, zenith, azimuth, magnetic_field_vect
     return efield_on_sky, efield_times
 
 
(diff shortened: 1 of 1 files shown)


NuRadioReco/modules/io/coreas/coreas.py
```diff
@@ -223,12 +213,11 @@ def convert_obs_positions_to_nuradio_on_ground(observer_pos, zenith, azimuth, ma
         observer_pos[:, 1] * -1,
         observer_pos[:, 0],
         observer_pos[:, 2]
-    ]) * units.cm
+    ]).T * units.cm
 
-    # second to last dimension has to be 3 for the transformation
-    obs_positions_geo = cs.transform_from_magnetic_to_geographic(obs_positions)
+    obs_positions = np.array([hp.rotate_vector_in_2d(obs_pos, -declination) for obs_pos in obs_positions])
```
MijnheerD, line 218: Isn't there a way to vectorise this step? Or would this require a change in radiotools itself?

MijnheerD (commented):
I don't think this function is tested in CI no, unless there is a test using `readCoREASDetector` (which would implicitly depend on this function to read in the Event). Though I think @lpyras added starshapes to the server for testing, so it should be possible make a script for it

MijnheerD:
Actually, I now see that `readCoREASDetector` is used in scripts like `SimpleMCReconstruction.py` . I believe that one is part of CI?

NuRadioReco/modules/io/coreas/coreas.py
fschlueter: Yes I think it would require changes in radiotools. Feel free to do that. Otherwise I think this is acceptable.

fschlueter:
> Actually, I now see that `readCoREASDetector` is used in scripts like `SimpleMCReconstruction.py` . I believe that one is part of CI?

Yes I think it is.

fschlueter:
@MijnheerD ?

MijnheerD (approved):
I just made it such that the list comprehension is not necessary. In my testing this did not change the result, and the fact that the unit tests did not fail makes me happy :)

PR #872, opened 2025-03-17
Efield fluence rice dist estimator

MartinRavn:
S. Martinelli et al. have recently proposed a method for estimating electric field energy fluences based on Rice distributions in this paper: https://arxiv.org/pdf/2407.18654. This method is a replacement for the noise subtraction we do here https://github.com/nu-radio/NuRadioMC/blob/10902608d54e723977371be43f920dc6e69c1170/NuRadioReco/utilities/trace_utilities.py#L100 and should give a smaller bias at low SNR and correctly estimate uncertainties.

The code was published alongside the paper. I have ported it to NuRadioReco and implemented it in the `electricFieldSignalReconstructor`.

### DIFF
--- a/NuRadioReco/modules/electricFieldSignalReconstructor.py
+++ b/NuRadioReco/modules/electricFieldSignalReconstructor.py
@@ -33,7 +33,7 @@ class electricFieldSignalReconstructor:
         logger.setLevel(log_level)
 
     @register_run()
-    def run(self, evt, station, det, fluence_method="noise_subtraction", signal_search_window=None, debug=False):
+    def run(self, evt, station, det, signal_search_window=None, fluence_method="noise_subtraction", debug=False):
         """
         reconstructs quantities for electric field
 
@@ -44,6 +44,12 @@ class electricFieldSignalReconstructor:
         station: station
 
         det: detector
+
+        signal_search_window: tuple
+            search window for signal in ns
+
+        fluence_method: str
+            method to calculate fluence ("noise_subtraction" or "rice_distribution")
 
         debug: bool
             set debug
@@ -99,9 +105,10 @@ class electricFieldSignalReconstructor:
                 dt = 1. / electric_field.get_sampling_rate()
                 ax.plot(tt / units.ns, trace[1] / units.mV * units.m)
                 ax.plot(tt / units.ns, trace[2] / units.mV * units.m)
-                ax.plot(tt / units.ns, envelope_mag / units.mV * units.m)
-                ax.vlines([low_pos * dt, up_pos * dt], 0, envelope_mag.max() / units.mV * units.m)
+                ax.plot(times_masked / units.ns, envelope_mag / units.mV * units.m)
+                ax.vlines([signal_search_window[0] + low_pos * dt, signal_search_window[0] + up_pos * dt], 0, envelope_mag.max() / units.mV * units.m, linestyles='dotted')
                 ax.vlines([signal_time - self.__signal_window_pre, signal_time + self.__signal_window_post], 0, envelope_mag.max() / units.mV * units.m, linestyles='dashed')
+                plt.savefig("trace_debug.png")
 
             times = electric_field.get_times()
             mask_signal_window = (times > (signal_time - self.__signal_window_pre)) & (times < (signal_time + self.__signal_window_post))
@@ -110,7 +117,7 @@ class electricFieldSignalReconstructor:
                 # set the noise window to the first "self.__noise_window" ns of the trace. If this cuts into the signal window, the noise window is reduced to not overlap with the signal window
                 mask_noise_window = times < min(times[0] + self.__noise_window, signal_time - self.__signal_window_pre)
 
-            signal_energy_fluence, signal_energy_fluence_error = trace_utilities.get_electric_field_energy_fluence(trace, times, mask_signal_window, mask_noise_window, return_error=True, method=fluence_method)
+            signal_energy_fluence, signal_energy_fluence_error = trace_utilities.get_electric_field_energy_fluence(trace, times, mask_signal_window, mask_noise_window, return_uncertainty=True, method=fluence_method)
             electric_field.set_parameter(efp.signal_energy_fluence, signal_energy_fluence)
             electric_field.set_parameter_error(efp.signal_energy_fluence, signal_energy_fluence_error)
 
--- a/NuRadioReco/modules/electricFieldSignalReconstructor.py
+++ b/NuRadioReco/modules/electricFieldSignalReconstructor.py
@@ -33,7 +33,7 @@ class electricFieldSignalReconstructor:
         logger.setLevel(log_level)
 
     @register_run()
-    def run(self, evt, station, det, debug=False):
+    def run(self, evt, station, det, fluence_method="noise_subtraction", debug=False):
         """
         reconstructs quantities for electric field
 
@@ -102,13 +102,7 @@ class electricFieldSignalReconstructor:
                 # set the noise window to the first "self.__noise_window" ns of the trace. If this cuts into the signal window, the noise window is reduced to not overlap with the signal window
                 mask_noise_window = times < min(times[0] + self.__noise_window, signal_time - self.__signal_window_pre)
 
-            signal_energy_fluence = trace_utilities.get_electric_field_energy_fluence(trace, times, mask_signal_window, mask_noise_window)
-            dt = times[1] - times[0]
-            signal_energy_fluence_error = np.zeros(3)
-            if(np.sum(mask_noise_window)):
-                RMSNoise = np.sqrt(np.mean(trace[:, mask_noise_window] ** 2, axis=1))
-                signal_energy_fluence_error = (4 * np.abs(signal_energy_fluence / self.__conversion_factor_integrated_signal) * RMSNoise ** 2 * dt + 2 * (self.__signal_window_pre + self.__signal_window_post) * RMSNoise ** 4 * dt) ** 0.5
-            signal_energy_fluence_error *= self.__conversion_factor_integrated_signal
+            signal_energy_fluence, signal_energy_fluence_error = trace_utilities.get_electric_field_energy_fluence(trace, times, mask_signal_window, mask_noise_window, return_error=True, method=fluence_method)
             electric_field.set_parameter(efp.signal_energy_fluence, signal_energy_fluence)
             electric_field.set_parameter_error(efp.signal_energy_fluence, signal_energy_fluence_error)
 
--- a/NuRadioReco/modules/electricFieldSignalReconstructor.py
+++ b/NuRadioReco/modules/electricFieldSignalReconstructor.py
@@ -33,7 +33,7 @@ class electricFieldSignalReconstructor:
         logger.setLevel(log_level)
 
     @register_run()
-    def run(self, evt, station, det, fluence_method="noise_subtraction", debug=False):
+    def run(self, evt, station, det, fluence_method="noise_subtraction", signal_search_window=None, debug=False):
         """
         reconstructs quantities for electric field
 
@@ -52,11 +52,19 @@ class electricFieldSignalReconstructor:
         for electric_field in station.get_electric_fields():
             trace_copy = copy.copy(electric_field.get_trace())
 
+            if signal_search_window is not None:
+                times = electric_field.get_times()
+                signal_search_window_mask = (times > signal_search_window[0]) & (times < signal_search_window[1])
+                times_masked = times[signal_search_window_mask]
+                trace_copy = trace_copy[:, signal_search_window_mask]
+            else:
+                times_masked = electric_field.get_times()
+
             # calculate hilbert envelope
             envelope = np.abs(signal.hilbert(trace_copy))
             envelope_mag = np.linalg.norm(envelope, axis=0)
             signal_time_bin = np.argmax(envelope_mag)
-            signal_time = electric_field.get_times()[signal_time_bin]
+            signal_time = times_masked[signal_time_bin]
             electric_field[efp.signal_time] = signal_time
 
     #
--- a/NuRadioReco/utilities/trace_utilities.py
+++ b/NuRadioReco/utilities/trace_utilities.py
@@ -104,7 +104,25 @@ def get_channel_voltage_from_efield(station, electric_field, channels, detector,
         return np.real(voltage_trace)
 
 
-def get_electric_field_energy_fluence(electric_field_trace, times, signal_window_mask=None, noise_window_mask=None, RMSNoise=None, return_error=False, method="noise_subtraction"):
+def get_electric_field_energy_fluence(electric_field_trace, times, signal_window_mask=None, noise_window_mask=None, return_uncertainty=False, method="noise_subtraction"):
+    """
+    Returns the energy fluence of each component of a 3-dimensional electric field trace.
+
+    Parameters
+    ----------
+    electric_field_trace : numpy.ndarray
+        The electric field trace to calculate the energy fluence for
+    times : numpy.ndarray
+        The time grid for the electric field trace
+    signal_window_mask : numpy.ndarray
+        A boolean mask that selects the signal window
+    noise_window_mask : numpy.ndarray
+        A boolean mask that selects the noise window. Only used if method is "noise_subtraction"
+    return_uncertainty : bool
+        If True, the uncertainty of the energy fluence is returned
+    method : str
+        The method to use for the energy fluence calculation. Can be either "noise_subtraction" or "rice_distribution" (also "rice_disttribution_sara" but this will be removed in future commits)
+    """
 
     dt = times[1] - times[0]
 
@@ -113,22 +131,25 @@ def get_electric_field_energy_fluence(electric_field_trace, times, signal_window
             f_signal = np.sum(electric_field_trace ** 2, axis=1)
         else:
             f_signal = np.sum(electric_field_trace[:, signal_window_mask] ** 2, axis=1)
+
         if noise_window_mask is not None and np.sum(noise_window_mask) > 0:
             f_noise = np.sum(electric_field_trace[:, noise_window_mask] ** 2, axis=1)
             f_signal -= f_noise * np.sum(signal_window_mask) / np.sum(noise_window_mask)
             f_signal[f_signal < 0] = 0
 
-            if RMSNoise is None:
-                RMSNoise = np.sqrt(np.mean(electric_field_trace[:, noise_window_mask] ** 2, axis=1))
+            # calculate RMS noise for error estimation
+            RMSNoise = np.sqrt(np.mean(electric_field_trace[:, noise_window_mask] ** 2, axis=1))
+        else:
+            RMSNoise = None
 
         signal_energy_fluence = f_signal * dt * conversion_factor_integrated_signal
 
         # calculate error if RMSNoise is known:
-        if RMSNoise is not None and return_error:
+        if RMSNoise is not None and return_uncertainty:
             signal_window_duration = sum(signal_window_mask) * dt if signal_window_mask is not None else len(times) * dt
             signal_energy_fluence_error = (4 * np.abs(signal_energy_fluence / conversion_factor_integrated_signal) * RMSNoise ** 2 * dt + 2 * signal_window_duration * RMSNoise ** 4 * dt) ** 0.5  * conversion_factor_integrated_signal
         else:
-            signal_energy_fluence_error = np.zeros(3)
+            pass #signal_energy_fluence_error = np.zeros(3)
     
     elif method == "rice_disttribution":
         signal_energy_fluence = np.zeros(len(electric_field_trace))
@@ -198,7 +219,7 @@ def get_electric_field_energy_fluence(electric_field_trace, times, signal_window
                 )
 
             #sample frequency (after the windowing) in MHz
-            delta_f = frequencies_window[1] - frequencies_window[0]
+            delta_f = (frequencies_window[1] - frequencies_window[0]) / units.MHz
 
             #to convert the amplitudes squared into energy fluence units
             conversion_factor = conversion_factor_integrated_signal #scipy.constants.epsilon_0 * scipy.constants.c / scipy.constants.e
@@ -215,10 +236,10 @@ def get_electric_field_energy_fluence(electric_field_trace, times, signal_window
             #get the fluence uncertainty as the root square of the variance
             fluence_freq_error = np.sqrt(fluence_freq_variance)
 
-            signal_energy_fluence[i_pol] = fluence_freq * 1000
-            signal_energy_fluence_error[i_pol] = fluence_freq_error * 1000
+            signal_energy_fluence[i_pol] = fluence_freq
+            signal_energy_fluence_error[i_pol] = fluence_freq_error
 
-    if return_error:
+    if return_uncertainty:
         return signal_energy_fluence, signal_energy_fluence_error
     else:
         return signal_energy_fluence
--- a/NuRadioReco/utilities/trace_utilities.py
+++ b/NuRadioReco/utilities/trace_utilities.py
@@ -137,8 +137,6 @@ def get_electric_field_energy_fluence(electric_field_trace, times, signal_window
                 trace = electric_field_trace[i_pol,:],
                 times = times,
                 t_peak = np.mean(times[signal_window_mask]), # This is not necessarily the peak time
-                f_low = 30 * units.MHz,
-                f_high = 80 * units.MHz,
                 spacing_noise_signal = 20,
                 window_length_tot = sum(signal_window_mask) * dt,
                 relative_taper_width = 0.142857143,
@@ -149,8 +147,6 @@ def get_electric_field_energy_fluence(electric_field_trace, times, signal_window
                 times = times,
                 t_peak = np.mean(times[signal_window_mask]),
                 noise_estimators = noise_estimators,
-                f_low = 30 * units.MHz,
-                f_high = 80 * units.MHz,
                 window_length_tot = sum(signal_window_mask) * dt,
                 relative_taper_width = 0.142857143,
                 )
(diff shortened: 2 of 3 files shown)


NuRadioReco/utilities/trace_utilities.py
```diff
@@ -12,6 +12,12 @@
 # see Phys. Rev. D DOI: 10.1103/PhysRevD.93.122005
 # to convert V**2/m**2 * s -> J/m**2 -> eV/m**2
 
+from NuRadioReco.utilities.fluence_rice_dist_estimator import get_signal_fluence_estimators, get_noise_fluence_estimators
+
+import sys
+sys.path.append("/home/mravn/fluence-and-uncertainty-estimation-based-on-rice-distribution")
+from signal_estimator import get_signal_fluence_estimators as get_signal_fluence_estimators_sara
+from noise_estimator import get_noise_fluence_estimators as get_noise_fluence_estimators_sara
```
anelles, line 20: I am not a big fan of having first names in here. Please chose a more generic name. 
If you cannot think of anything, please at least go for the last name.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -98,18 +104,146 @@ def get_channel_voltage_from_efield(station, electric_field, channels, detector,
...
+            #get the variance of the trace fluence summing up the frequency variances and converting in (eV/m^2)^2
+            fluence_freq_variance = np.sum(variances) * (((dt * one_side_spectrum_corr_factor) **2) * delta_f * conversion_factor) **2
+
+            #get the fluence uncertainty as the root square of the variance
+            fluence_freq_error = np.sqrt(fluence_freq_variance)
+
+            signal_energy_fluence[i_pol] = fluence_freq
+            signal_energy_fluence_error[i_pol] = fluence_freq_error
+
+    elif method == "rice_disttribution_sara":
```
anelles, line 196: See above.

anelles:
Overall, thank you very much for working on this. I hoping for people more involved in daily analysis to review more in depth.

anelles:
> Seems @anelles is reviewing this simultaneously, but I already wrote some comments, so...

Just had a brief comment -- I am counting on @sjoerd-bouma to do the hard work ;)

NuRadioReco/utilities/trace_utilities.py
MartinRavn: Sorry. I didn't expect such a quick review. I already removed it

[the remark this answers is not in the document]
MartinRavn: Already done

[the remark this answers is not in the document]
MartinRavn: I just had this code here for cross-checks. I will post some plots shortly.

MartinRavn:
Here are some plots of the relative error of the theta component of 10000 random electric fields generated with this function

MartinRavn:
Here are some plots of the relative error of the theta component of 10000 electric fields generated with [this](https://github.com/nu-radio/NuRadioMC/blob/10902608d54e723977371be43f920dc6e69c1170/NuRadioReco/utilities/analytic_pulse.py#L20) function. I have used the same SNR definition as in the paper, so these plots should be equivalent to Figure 9. 
image

MartinRavn:
I however see that the Rice distribution method provides correct uncertainties image

MartinRavn:
The main conclusions from the plots above are:
- My port of Sara's code to NuRadioReco gives exactly the same results as using Sara's code directly
- The Rice distribution method seems to have a bias at low SNR, which is contrary to what is shown in the paper
- The Rice distribution method provides correct uncertainty estimates

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -98,18 +97,234 @@ def get_channel_voltage_from_efield(station, electric_field, channels, detector,
...
+        signal_energy_fluence = np.zeros(len(electric_field_trace))
+        signal_energy_fluence_error = np.zeros(len(electric_field_trace))
+        for i_pol in range(len(electric_field_trace)):
+            noise_estimators, frequencies_window = get_noise_fluence_estimators(
+                trace = electric_field_trace[i_pol, :],
+                times = times,
+                signal_window_mask = signal_window_mask,
+                spacing_noise_signal = 20 * units.ns,
+                relative_taper_width = 0.142857143,
+                use_median_value = False
```
MartinRavn, line 165: Should we make these parameters controllable when calling `get_electric_field_energy_fluence`? I believe these settings work well for most analyses, but a user may want to change them.

sjoerd-bouma: You could add an additional optional argument `estimator_kwargs={}` or similar with arguments that are passed on to `get_noise_fluence_estimators` and/or `get_signal_fluence_estimators`. I probably wouldn't define them explicitly here, especially as the values you provide here are just the default values of the function.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -98,18 +97,234 @@ def get_channel_voltage_from_efield(station, electric_field, channels, detector,
...
+            spectrum_window = np.abs(fft.time2freq(windowed_trace, 1/dt))
+
+            list_ffts_squared.append(spectrum_window**2)
+            noise_start = noise_stop
+
+        elif noise_stop > signal_start and noise_start <= signal_start:
+            noise_start = signal_stop
+
+        else:
+            print("Your peak is at zero or negative time...! ")
```
sjoerd-bouma, line 265: Replace print with logger statement.

NuRadioReco/modules/electricFieldSignalReconstructor.py
```diff
@@ -91,9 +105,10 @@ def run(self, evt, station, det, debug=False):
                 dt = 1. / electric_field.get_sampling_rate()
                 ax.plot(tt / units.ns, trace[1] / units.mV * units.m)
                 ax.plot(tt / units.ns, trace[2] / units.mV * units.m)
-                ax.plot(tt / units.ns, envelope_mag / units.mV * units.m)
-                ax.vlines([low_pos * dt, up_pos * dt], 0, envelope_mag.max() / units.mV * units.m)
+                ax.plot(times_masked / units.ns, envelope_mag / units.mV * units.m)
+                ax.vlines([signal_search_window[0] + low_pos * dt, signal_search_window[0] + up_pos * dt], 0, envelope_mag.max() / units.mV * units.m, linestyles='dotted')
                 ax.vlines([signal_time - self.__signal_window_pre, signal_time + self.__signal_window_post], 0, envelope_mag.max() / units.mV * units.m, linestyles='dashed')
+                plt.savefig("trace_debug.png")
```
sjoerd-bouma, line 111: Looks like this is not your fault, but we should call `plt.close()` (or `plt.show()`) after plotting, to prevent clogging the memory with loads of figures.

sjoerd-bouma (commented):
I haven't tested the module myself yet, but are you and Sara still debugging why the RICE method results in a large(r) bias on the estimated signal? It might be good to make sure this is what we expect before merging this.

NuRadioReco/utilities/trace_utilities.py
MartinRavn: I implemented this, but I had to add two unused keyword arguments to `get_signal_fluence_estimators` so that it has the same keywords as `get_noise_fluence_estimators`. Let me know if you know of a better way to do this.

MartinRavn: Do you also think these keywords should be controllable from `electricFieldSignalReconstructor`, or can we use the default values there?

NuRadioReco/utilities/trace_utilities.py
MartinRavn: I implemented a proper error here so we don't risk having an infinite while loop.

NuRadioReco/modules/electricFieldSignalReconstructor.py
MartinRavn: It looks like we in most cases don't save debug plots and just call `plt.show()`. So ill replace this line with `plt.show()`

MartinRavn:
I am in contact with Sara and we improved the bias somewhat by changing the settings of the methods. These improved settings are used for the plots above. I am not sure why they don't see this bias in their analysis. The only differences between my simulation and their simulation is that they use unfolded noise from Auger instead of unfolded thermal noise, and they use electric field pulses from CoREAS instead of the simple analytic electric field model we have in NuRadioReco. Maybe one of these things are responsible for the difference, however, I am not sure.

I believe the bias comes from [this line](https://github.com/nu-radio/NuRadioMC/blob/31a94f94f656c234ae79d0a72a0e2ed5d0f4fca0/NuRadioReco/utilities/trace_utilities.py#L139), i.e., setting unphysical negative fluences to zero before summing over frequencies. This will introduce a bias towards larger fluences. I tried commenting out this line and instead setting negative fluences to zero after summing over frequencies. This gives something that looks very similar to the Noise subtraction method (shown below), however, with correct uncertainty estimates. So I am wondering if the two methods are equivalent in this case.
image

MartinRavn:
And here is a plot where I used CoREAS simulations instead.
image

anelles:
Given that this is planned to be used by the wider community, doesn't it make sense to involved Sara and maybe also Tim here directly? To discuss this once openly with everyone? It seems a bit odd to not have the experts comment here. I will go and track down their github handle, unless someone beats me to it ...

anelles:
I just invited @TimH54. I hope this works!

cg-laser:
> Given that this is planned to be used by the wider community, doesn't it make sense to involved Sara and maybe also Tim here directly? To discuss this once openly with everyone? It seems a bit odd to not have the experts comment here. I will go and track down their github handle, unless someone beats me to it ...

we have already been discussing with them via email for a while but it might be good to move this discussion to this PR. I asked them to join. 
Martin wrote a stand-alone test code and my understanding is that so far, no one could find a bug in it. He'll upload it here as well.

MartinRavn:
Here is the link to the box folder with a simplified script and toy model data. The script only depends on a clone of [Sara's code](https://gitlab.iap.kit.edu/saramartinelli/fluence-and-uncertainty-
estimation-based-on-rice-distribution) and NuRadioMC for the units. This script should recreate the central plot I showed above, labeled "Rice Distribution (Sara's code)":
https://uppsala.box.com/s/bj532merpmdo8o5yprr9lfm3b6rqok9a

MartinRavn:
I have written down some math explaining why I believe the Rice distribution method should give the same estimates as the noise subtraction method when we allow negative frequency fluence estimators before the sum over frequencies. I hope it makes sense
image

fschlueter:
> I believe the bias comes from [this line](https://github.com/nu-radio/NuRadioMC/blob/31a94f94f656c234ae79d0a72a0e2ed5d0f4fca0/NuRadioReco/utilities/trace_utilities.py#L139), i.e., setting unphysical negative fluences to zero before summing over frequencies. This will introduce a bias towards larger fluences. I tried commenting out this line and instead setting negative fluences to zero after summing over frequencies. This gives something that looks very similar to the Noise subtraction method (shown below), however, with correct uncertainty estimates. So I am wondering if the two methods are equivalent in this case.

I am a bit confused the line you are referencing is of the traditional "noise subtraction" method which does not have a strong bias to begin with right? Or is something similarly done in the rice method?

MartinRavn:
> > I believe the bias comes from [this line](https://github.com/nu-radio/NuRadioMC/blob/31a94f94f656c234ae79d0a72a0e2ed5d0f4fca0/NuRadioReco/utilities/trace_utilities.py#L139), i.e., setting unphysical negative fluences to zero before summing over frequencies. This will introduce a bias towards larger fluences. I tried commenting out this line and instead setting negative fluences to zero after summing over frequencies. This gives something that looks very similar to the Noise subtraction method (shown below), however, with correct uncertainty estimates. So I am wondering if the two methods are equivalent in this case.
> 
> I am a bit confused the line you are referencing is of the traditional "noise subtraction" method which does not have a strong bias to begin with right? Or is something similarly done in the rice method?

Sorry, I linked to the wrong line. This is the line I meant: https://github.com/nu-radio/NuRadioMC/blob/31a94f94f656c234ae79d0a72a0e2ed5d0f4fca0/NuRadioReco/utilities/trace_utilities.py#L332

MartinRavn:
> > The main conclusions from the plots above are:
> > 
> > * My port of Sara's code to NuRadioReco gives exactly the same results as using Sara's code directly
> > * The Rice distribution method seems to have a bias at low SNR, which is contrary to what is shown in the paper. However, I use only thermal noise and they use unfolded noise from Auger, which has non-thermal components.
> > * The Rice distribution method provides correct uncertainty estimates
> 
> @MartinRavn do you think it would help if we test the method with different noise? You could either use the new `data-driven` mode of the `channelGenericNoiseAdder` which draws noise from a rayleigh distribution without directly using an antenna pattern but the `sigma` parameter is parameterized from data and thus contains electronic noise as well as ambient thermal radiation which is folded through the antenna. Or you may could use the module which @CamphynR is developing where you take a temperature profile at the antenna and fold in the electric field produced from this temperature (via Rayleigh-Jeans) through the antenna and later add electronic noise via the `channelGenericNoiseAdder`?

Yes, I can try this. I am taking a course this week, so I can get around to it on Friday. However, I don't expect it to change much, since this methods doesn't assume anything about the noise, except that it is rayleigh distributed, or the signal.

fschlueter:
> > > The main conclusions from the plots above are:
> > > 
> > > * My port of Sara's code to NuRadioReco gives exactly the same results as using Sara's code directly
> > > * The Rice distribution method seems to have a bias at low SNR, which is contrary to what is shown in the paper. However, I use only thermal noise and they use unfolded noise from Auger, which has non-thermal components.
> > > * The Rice distribution method provides correct uncertainty estimates
> > 
> > 
> > @MartinRavn do you think it would help if we test the method with different noise? You could either use the new `data-driven` mode of the `channelGenericNoiseAdder` which draws noise from a rayleigh distribution without directly using an antenna pattern but the `sigma` parameter is parameterized from data and thus contains electronic noise as well as ambient thermal radiation which is folded through the antenna. Or you may could use the module which @CamphynR is developing where you take a temperature profile at the antenna and fold in the electric field produced from this temperature (via Rayleigh-Jeans) through the antenna and later add electronic noise via the `channelGenericNoiseAdder`?
> 
> Yes, I can try this. I am taking a course this week, so I can get around to it on Friday. However, I don't expect it to change much, since this methods doesn't assume anything about the noise, except that it is rayleigh distributed, or the signal.

Yes I am also missing the creativity to believe this is the reason but maybe still worth a shot.

fschlueter:
Could the bias be related to the fact that you are using shorter traces / less or shorter noise windows and thus have a larger statistical uncertainty in the noise estimation?

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -98,18 +97,242 @@ def get_channel_voltage_from_efield(station, electric_field, channels, detector,
...
+
+            signal_energy_fluence[i_pol] = fluence_freq
+            signal_energy_fluence_error[i_pol] = fluence_freq_error
+
+    if return_uncertainty:
+        return signal_energy_fluence, signal_energy_fluence_error
+    else:
+        return signal_energy_fluence
+
+def get_noise_fluence_estimators(trace, times, signal_window_mask, spacing_noise_signal=20*units.ns, relative_taper_width=0.142857143, use_median_value=False):
```
fschlueter, line 199: Is this and the following function `get_signal_fluence_estimators` meant to be used outside of `get_electric_field_energy_fluence`? If not you could consider making them private.

TimH54:
Hi all, sorry I only found the time to look in here right now! I cannot answer why Martin gets significantly different results from the ones Sara got. Maybe Sara has more insight, but she is currently in a crunch to get her PhD thesis written up, which she has to deliver mid-April. That is of course rather unfortunate, but I hope everybody understands. So it might be best to let this lie for a moment and hopefully she can inject some more thoughts once she is available again? If the problem boils down to mathematically quantifying the equivalence (or not) of noise subtraction and the Rice-based estimator, we might also want to discuss with Diego Ravignani, who was involved in developing this part. But we should also then wait for Sara to be available.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -98,18 +97,242 @@ def get_channel_voltage_from_efield(station, electric_field, channels, detector,
...
+                signal_window_mask = signal_window_mask,
+                noise_estimators = noise_estimators,
+                **estimator_kwargs
+                )
+
+            #sample frequency (after the windowing) in MHz
+            delta_f = frequencies_window[1] - frequencies_window[0]
+
+            #to convert the amplitudes squared into energy fluence units
+            conversion_factor = conversion_factor_integrated_signal
```
fschlueter, line 177: Why are these lines necessary?

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -98,18 +97,242 @@ def get_channel_voltage_from_efield(station, electric_field, channels, detector,
...
+
+            #correcting for selecting positive frequencies
+            one_side_spectrum_corr_factor = 1 #np.sqrt(2)
```
fschlueter, line 180: same

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -98,18 +97,242 @@ def get_channel_voltage_from_efield(station, electric_field, channels, detector,
...
+            mask_time = np.all([np.round(times, 5) >= np.round(noise_start, 5), np.round(times, 5) < np.round(noise_stop, 5)], axis=0)
+            time_trace_clipped = trace[mask_time]
+
+            #applying the Tukey window
+            windowed_trace = time_trace_clipped * window
 
-    if signal_window_mask is None:
-        f_signal = np.sum(electric_field_trace ** 2, axis=1)
+            #calculating the spectrum and frequencies
+            frequencies_window = np.fft.rfftfreq(len(windowed_trace), d=dt)
```
fschlueter, line 259: can you use the method `fft.freqs` - does the same thing but as a plus it caches the result which should speed up things.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -98,18 +97,242 @@ def get_channel_voltage_from_efield(station, electric_field, channels, detector,
...
+            delta_f = frequencies_window[1] - frequencies_window[0]
+
+            #to convert the amplitudes squared into energy fluence units
+            conversion_factor = conversion_factor_integrated_signal
+
+            #correcting for selecting positive frequencies
+            one_side_spectrum_corr_factor = 1 #np.sqrt(2)
+
+            #get the fluence of the trace summing up the frequency estimators and converting in eV/m^2
+            fluence_freq = np.sum(estimators) * ((dt * one_side_spectrum_corr_factor) **2) * delta_f * conversion_factor
```
fschlueter, line 183: Why do you have to multiply the sum with `dt**2`? I though that is not necessary according to our [docu](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.utilities.fft.html#module-NuRadioReco.utilities.fft)

fschlueter:
@MartinRavn the plots here are made using the mean as an estimator for the noise power in each frequency bin? I assume you have tried using the median? (I would expect that the median is typically a bit lower than the median (for rayleigh distributed noise) and hence the change to have a signal < 0 is lower?

I guess the problem, in comparison to the noise subtraction method, can be boiled down to the fact that for the rice method the estimated noise amplitude (mean over x noise windows) is subtracted from each amplitude (or square amplitude - does not matter) in the signal window before summing over the all amplitudes in the frequency domain while in the noise subtraction method the noise estimate is subtracted after summing. So instead of:
```
signal_estimators = spectrum_window**2 - noise_estimators
```
we can calculate:
```
signal_estimators = np.sum(spectrum_window**2) - np.sum(noise_estimators)
```
This will not inherently affect the variance estimation? And I guess this is equivalent with the suggestion which Martin made already that we drop this line:
```
signal_estimators[signal_estimators < 0] = 0
```

TimH54:
Ok, we just had an extensive discussion about your findings between Sara, Harm, Diego and myself. Also, Sara has done lots of additional plots and studies in the meantime to try and see if there is any fundamental problem - she did not find any. I try to summarize here what we concluded. Sorry, it's complex. ;)

First of all, there is one significant difference in what you do and Sara did in the paper: You evaluate the signal at the true time-bin, whereas Sara, trying to mimick what one would do in an experiment, took the signal as the highest peak in a +/- 30 ns window around the true time-bin. The latter is likely to introduce additional bias, as you systematically pick up upward-fluctuations. This might explain why your "noise subtraction" plots look better than those of Sara. (It does not explain why your "Rice method" plots look worse than those of Sara.)

Second, we think that even though you find larger biases than Sara for the "Rice method", the result is still reasonable. Keep in mind that we are looking at ratios of very very small numbers ("almost zero estimated signal" divided by "almost zero true signal"). The ratios will blow up by construction, but this is not actually really a problem if the uncertainties are estimated as large enough (which they are). A better way to visualize what is going on at these very low signal-to-noise ratios, where we are looking for signals smaller than the noise level and everything becomes very sensitive to details, would be to calculate the linear fluence difference rather than the ratio. So, even though your "Rice plot" looks worse than that of Sara, we think that the result is still "fine". It would be interesting to understand why you get a different result from Sara, but Sara can really not look into this right now as she has to urgently finish her thesis. So if we deem this a useful study (I am not entirely convinced that it is, as it might have little practical relevance), it would have to be done without Sara's help for the time being. We acknowledge that setting the per-frequency signals to zero is not ideal, but Sara looked into the distributions and they are highly asymmetric, so we currently do not see a simple solution on how to improve on that step; this would really be an additional study to do.

Third, the "noise subtraction" method also just "looks better" on these plots than it actually is because of the way the ratios are visualized. These plots have the range from -1 to 0 on one side but 0 to infinity on the other side. This kind of visualization was reasonable for Sara's paper, where the bias for the Rice method only reached 15% or so. But when you get factors of 10 or so this is not a good visualization any more. The clipped values at -1 for the "noise subtraction" method actually represent "factors of infinity" underestimation. They "look benign" on these plots, but they are actually a much bigger problem than upwards factors of 10. We actually are of the opinion that one should exclude such data from analysis (which means that the "noise subtraction" method cannot be applied any more below some SNR threshold, in Sara's analysis around approximately SNR values of 5, as almost all data would have to be rejected).

In summary, we have not found any fundamental problem or mistake, and we are still convinced that the "Rice method" is a step-up from the "noise subtraction method", as it a) gives you correct uncertainties and b) at very low signals gives you a reasonable estimate (biased, yes, but with a bias smaller than the uncertainty and no need to drop the data as one should for the "noise subtraction method"). Certainly, things can be improved further; keep in mind that the Rice method originally estimates amplitudes and on the way to fluences additional biases arise, which might be improved upon; especially it would be nicer to not have to set frequency bins to zero). But we consider it a good step forward and see no reason to stick with the "noise subtraction method".

NuRadioReco/utilities/trace_utilities.py
MartinRavn: They are not. They are just a relic from copying Sara's code. I'll remove them.

NuRadioReco/utilities/trace_utilities.py
MartinRavn: You are right. This is actually a bug. I didn't see it because I only tested with `dt = 1 ns`.

anelles:
So I just talked to @TimH54 about this. I think the conclusion holds that there is no obvious bug, but that at the smallest signals things remain tricky. I would suggest that we merge this here. @MartinRavn can you resolve the conflicts and then we can move forward. 
People also want to use it for LOFAR / SKA, so it should not be hidden in a branch.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -98,18 +97,249 @@ def get_channel_voltage_from_efield(station, electric_field, channels, detector,
...
+    trace : np.ndarray
+        Trace to estimate the noise fluence from.
+    times : np.ndarray
+        Time grid for the trace.
+    signal_window_mask : np.ndarray
+        Boolean mask for the signal window.
+    spacing_noise_signal : float (optional)
+        Spacing between noise windows and signal window. Makes sure no signal leaks into the noise windows.
+    relative_taper_width : float (optional)
+        Width of the taper region for the Tukey window relative to the full window length.
```
anelles, line 214: Same thing, please explain how that number came to be and what it means.

MartinRavn:
> So I just talked to @TimH54 about this. I think the conclusion holds that there is no obvious bug, but that at the smallest signals things remain tricky. I would suggest that we merge this here. @MartinRavn can you resolve the conflicts and then we can move forward. People also want to use it for LOFAR / SKA, so it should not be hidden in a branch.

Sounds good. @fschlueter, @sjoerd-bouma do you know why the unit tests fail?

sjoerd-bouma:
> Sounds good. @fschlueter, @sjoerd-bouma do you know why the unit tests fail?

Apologies @MartinRavn, I must have missed this. The unit tests failed for reasons unrelated to this PR, and have been fixed on `develop` already. I agree with @anelles that as there are no clear outstanding problems it would be nice to get this merged : )

fschlueter (approved):

fschlueter (dismissed):

fschlueter:
Thinking about it, I think this new feature deserves an entry to the changelog..

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -98,18 +97,249 @@ def get_channel_voltage_from_efield(station, electric_field, channels, detector,
...
+    Parameters
+    ----------
+    trace : np.ndarray
+        Trace to estimate the noise fluence from.
+    times : np.ndarray
+        Time grid for the trace.
+    signal_window_mask : np.ndarray
+        Boolean mask for the signal window.
+    spacing_noise_signal : float (optional)
+        Spacing between noise windows and signal window. Makes sure no signal leaks into the noise windows.
```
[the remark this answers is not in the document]
MartinRavn: I have added this now

NuRadioReco/utilities/trace_utilities.py
MartinRavn: These default values are taken directly from Sara's code. I have no better explanation than that it is 1/7 of the full length. I believe Sara has thoroughly tested this.

MartinRavn:
Thanks @fschlueter for fixing the merge conflict. I have responded to @anelles comments and updated the changelog. I also ran my test script again, and everything looks fine.

PR #875, opened 2025-03-18
Make add_to_trace compatible with 3d traces

fschlueter:
This PR implement compatibility for 3d-traces (i.e., those used for electric_fields) with the method `BaseTrace.add_to_trace`.

### DIFF
--- a/NuRadioReco/framework/base_trace.py
+++ b/NuRadioReco/framework/base_trace.py
@@ -409,10 +409,8 @@ class BaseTrace:
 
         # Add the trace to the original trace:
         original_trace = self.get_trace()
-        if original_trace.ndim == 1:
-            original_trace[i_start_readout:i_end_readout] += trace_to_add[i_start_channel:i_end_channel]
-        else:
-            original_trace[:, i_start_readout:i_end_readout] += trace_to_add[:, i_start_channel:i_end_channel]
+        # arr[..., start:stop] works for any dimension: 1, 2, 3, ...
+        original_trace[..., i_start_readout:i_end_readout] += trace_to_add[..., i_start_channel:i_end_channel]
 
         self.set_trace(original_trace, sampling_rate_readout)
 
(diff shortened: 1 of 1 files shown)


NuRadioReco/framework/base_trace.py
```diff
@@ -393,23 +393,27 @@ def ceil(x):
...
         if i_end_readout - i_start_readout != i_end_channel - i_start_channel:
             logger.error("The traces do not have the same length. This should not happen.")
             raise ValueError('The traces do not have the same length. This should not happen.')
 
         # Add the trace to the original trace:
         original_trace = self.get_trace()
-        original_trace[i_start_readout:i_end_readout] += trace_to_add
+        if original_trace.ndim == 1:
+            original_trace[i_start_readout:i_end_readout] += trace_to_add[i_start_channel:i_end_channel]
+        else:
```
MartinRavn, line 414: You can consider having `elif original_trace.ndim == 2:` here and throw an error if the dimension in neither 1 or 2. If the original trace is a 3D array the current implementation would still run and probably give an unintended result. However, I think this is an unlikely scenario, so you can probably just ignore this comment.

MartinRavn (dismissed):
Looks good. I only have one comment which can maybe be ignored

sjoerd-bouma (commented):
I think you could use an Ellipsis `...` to generalize this (and save three lines of code!). See https://numpy.org/doc/stable/user/basics.indexing.html#dimensional-indexing-tools, i.e.
`trace_to_add[..., i_start_channel:i_end_channel]` covers both cases (as well as higher dimensions, though we don't have a use case for those yet).

fschlueter:
Well

> I think you could use an Ellipsis `...` to generalize this (and save three lines of code!). See https://numpy.org/doc/stable/user/basics.indexing.html#dimensional-indexing-tools, i.e. `trace_to_add[..., i_start_channel:i_end_channel]` covers both cases (as well as higher dimensions, though we don't have a use case for those yet).

Thanks @sjoerd-bouma, done!

sjoerd-bouma (approved):

PR #876, opened 2025-03-18
Fix/correct station14 station11 geometries

shallmann:
The description for station 14 in the json is odd. With this PR I am trying to fix:

- LPDA positions/numbering
- LPDA orientations
- shallow HPol -> shallow VPol

and incorrect numbering of station 11 LPDA channels.

This is my starting point:
Screenshot 2025-03-18 at 14 03 11
(black text is channel number in json. Blue are GPS position "ID"s


shallmann:
After fixing positions:
Screenshot 2025-03-18 at 14 10 39


... still need to put in correct orientations...

shallmann:
Screenshot 2025-03-18 at 23 38 02
I extended the visualization script to be able to select a specific station and antenna type (in this case `python visualize_detector.py RNO_G/RNO_season_2024.json 14 LPDA`)

shallmann:
Looks roughly ok now.
- I fixed all the LPDA channel numbering and positions
- I set the downward-LPDA centers to the mean of the tine ends

-  I calculated the phi rotation from the via the `arctan2(dy, dx)` of the two GPS points of the downward-facing tine ends and took that also for the upward facing on the same side.
- I set the theta orientation to 0 for upward and 180 for downward
- I set all phi orientations to 0 (at 0 and 180 deg it should not make a difference, please verify)

- I replaced the additional shallow H-Pol by a V-Pol as should be

- I fixed the station 11 channel numbering for the LPDAs which had been found to be wrong a while ago.

@sanyukta13 can you please double-check if these changes are correct? Thanks!

shallmann:
@fschlueter the CSV values are correct, I double checked with Timo why the plot in the [deployment notes](https://docs.google.com/document/d/1Q7q9QJXi0X263zZ3p9BYqFvO3b4u9FckLtXWnqu6JrY/edit?tab=t.0#heading=h.wc033icd7vh9) looks rotated and his guess was that he used a quick online conversion from lat/lon to northing/easting that might be imperfect.

shallmann:
@fschlueter yes I have... I used my favorite C17 and the "multistation cluster" events reconstruction matches the other stations! ... so the ordering of channel looks correct!

Screenshot 2025-03-19 at 10 29 58
(figure is reconstructed zenith/azimuth using the upward-LPDAs, trigger time is color-coded)

fschlueter (approved):

sanyukta13:
Hi,

I might be a little late in reviewing due to spring break at KU. Sorry for
the inconvenience. Will get back to you as soon as I can.

Best Regards
Sanyukta Agarwal

shallmann:
I also swapped the 2nd and 3rd arm of the surface LPDAs for station 11 in the older json files. I attach a visual proof that all jsons are consistent.

The fact that they are swapped is known since a while but it seems has never made it into the jsons (eg the June 5, 2024 Calibration call if you are looking for slides) and has been verified by various sources (pulsers, station solar panel positions via noise, aircrafts) by Sjoerd and myself.

@JethroStoffels as station 11 expert, can you please verify or appoint a responsible person to verify the LPDA positions are correct in the DB detector?

Screenshot 2025-03-20 at 13 51 10
Screenshot 2025-03-20 at 13 51 00

sanyukta13 (approved):
Thanks for updating the json @shallmann, looks good to me!

PR #878, opened 2025-03-21
Implement core resampling method (from BaseTrace class) in utilities. Make warning smarter

fschlueter:
This implements the resampling method in utilities thus it can be used without creating a BaseTrace object to check if the resampling would yield the correct number of samples.

@sjoerd-bouma did you intentionally not calculate the final number of samples to be more efficient or are you fine with that?


fschlueter:
Urg circular imports - a solution would be to move the `add_cable_delay` function back to the module (this keeping it outside of the class). @sjoerd-bouma would be probably even happier with that?

PR #880, opened 2025-03-24
Use NuRadio units in io/LOFAR

sjoerd-bouma:
I noticed that `readLOFARData` still occasionally returns quantities in a different unit system than NuRadio, especially e.g. `get_calibration_delays` which is actually expected to be called by the user to calibrate these delays out. 

As far as I'm concerned, there are two possible solutions:
1. We make all of the `rawTBBio*` modules private, use the LOFAR unit conventions in those, and only convert to NuRadio units in the relevant functions in `readLOFARData` (`get_calibration_delays`, `get_metadata`). This may potentially make it harder to maintain the `rawTBBio` modules, but on the other hand explicitly introducing NuRadio units everywhere will potentially make them diverge (more) from usage outside NuRadio by Brian et al. (not sure that this is something we should worry about).
2. We consistently change the inputs and outputs of all methods in all modules under `io/LOFAR` to use NuRadio units.

I have tried to do the latter, but I'm happy to have a discussion about whether this is the correct approach. I have checked that for a test file (from 2012) the resulting .nur file is unchanged, though it seems not all functions in the `rawTBBio` modules are always used, so it might be good to test it on a couple of other files.

For reviewers: commit 550f2d3a9ef4fb53f75622687646190fb5f33ec9 implements the relevant changes, commit  5fb64ae12bdb46efb589a05168e466aa1e5e094c only fixes/improves some docstrings.


NuRadioReco/modules/io/LOFAR/rawTBBio.py
```diff
@@ -129,7 +131,8 @@ def read_antenna_delays(fname):
 
     def parse_line_v1(line):
         ant_name, pol_E_delay, pol_O_delay = line.split()[0:3]
-        additional_ant_delays[ant_name] = [float(pol_E_delay), float(pol_O_delay)]
+        additional_ant_delays[ant_name] = [float(pol_E_delay) * units, float(pol_O_delay) * units.s]
```
PhilippLaub, line 134: should be `units.s` in `float(pol_E_delay) * units`, right?

MijnheerD:
I am more in favor of option 1. The `rawTBBio` functions are LOFAR internal and should be used directly. They will probably not be changed, unless we decide to incorporate some of Brian's latest changes in which case it would be more difficult if our version had different conventions than his. I would just make sure all docstrings are very clear about what units are returned (if this is not already the case) and focus on the NRR compatibility of the user facing module, which is `readLOFARData`.

NuRadioReco/modules/io/LOFAR/__init__.py
```diff
@@ -1 +1,10 @@
+"""
+Package with modules related to the read-in of LOFAR (.root) data files
```
MijnheerD, line 2: LOFAR files are not ROOT files?

NuRadioReco/modules/io/LOFAR/readLOFARData.py
```diff
@@ -256,7 +270,7 @@ def __init__(
         Parameters
         ----------
         time_s: int
-            Event trigger timestamp in seconds
+            Event trigger timestamp in (UNIX) seconds
```
MijnheerD, line 273: We can at the same time also change the variable name to `time_unix`, to keep it more consistent?

NuRadioReco/modules/io/LOFAR/readLOFARData.py
```diff
@@ -692,7 +715,7 @@ def run(self, detector, trace_length=65536):
                 channel_group: int = detector.get_channel_group_id(station_id, channel_id)
 
                 channel = NuRadioReco.framework.channel.Channel(channel_id, channel_group_id=channel_group)
-                channel.set_trace(this_trace, station_dict['metadata'][4] * units.Hz)
+                channel.set_trace(this_trace, station_dict['metadata'][4] * units.GHz)
```
MijnheerD, line 718: I don't understand this change? In the new `rawTBBio` the sample frequency is already in internal units, so there should be no multiplication here?

sjoerd-bouma:
> I am more in favor of option 1. The `rawTBBio` functions are LOFAR internal and should not be used directly. They will probably not be changed, unless we decide to incorporate some of Brian's latest changes in which case it would be more difficult if our version had different conventions than his. I would just make sure all docstrings are very clear about what units are returned (if this is not already the case) and focus on the NRR compatibility of the user facing module, which is `readLOFARData`.

Sure, I'm happy to change this, it's a straightforward enough change. Just for my own understanding: to what extent are the `rawTBBio` modules just straight copies of code by Brian et al.? Can you send me a link to the original repository? They don't seem to be in PyCRTools (though the file claims to be closely related).

sjoerd-bouma:
I also had another question re: calibration delays. It seems right now the user should **always** call `get_calibration_delays` and shift the traces manually. I am wondering if it wouldn't make more sense to store these values in the detector description and expect the user to always use the `channelAddCableDelay` class to remove this - not because this approach is inherently less error-prone (although it has the advantage of being easier to track if the delays have been corrected for or not, through the event module list), but because this is a standard data processing step for in-ice detectors. Alternatively, we could also consider applying the delays by default in the `run` function? I guess what I'm worried about is that it's relatively easy to miss this step for a new user.

MijnheerD:
> > I am more in favor of option 1. The `rawTBBio` functions are LOFAR internal and should not be used directly. They will probably not be changed, unless we decide to incorporate some of Brian's latest changes in which case it would be more difficult if our version had different conventions than his. I would just make sure all docstrings are very clear about what units are returned (if this is not already the case) and focus on the NRR compatibility of the user facing module, which is `readLOFARData`.
> 
> Sure, I'm happy to change this, it's a straightforward enough change. Just for my own understanding: to what extent are the `rawTBBio` modules just straight copies of code by Brian et al.? Can you send me a link to the original repository? They don't seem to be in PyCRTools (though the file claims to be closely related).

The file is basically a copy from an earlier version of this one, with only some changes to the logging: https://github.com/Bhare8972/LOFAR-LIM/blob/master/LoLIM/IO/raw_tbb_IO.py

NuRadioReco/modules/io/LOFAR/rawTBBio.py
```diff
@@ -64,18 +64,20 @@
...
-    "": 1.0,
-    "kHz": 1000.0,
-    "MHz": 10.0 ** 6,
-    "GHz": 10.0 ** 9,
-    "THz": 10.0 ** 12,
+    "": units.Hz,
+    "kHz": units.kHz,
+    "MHz": units.MHz,
+    "GHz": units.GHz,
+    "THz": units.tera * units.Hz,
```
fschlueter, line 80: I am completely out of context here but this is not a 1-to-1 change!

MijnheerD (approved):
All the changes look good! From my side you can merge, if the units pass of course :)

sjoerd-bouma:
Thanks @MijnheerD ! I will run a couple more checks - I used @karenterveer's LOFAR example to make sure nothing has changed, and while the unit changes seem fine, moving the calibration delays to before the RFIfilter and amplitude calibration etc. seems to have changed things, though I'm not yet sure to what extent.

sjoerd-bouma:
It seems for the test file the only difference is in a single channel, which has a calibration delay of -5114 ns (!); the difference in the 'test' is actually just at the trace edges, because the tapering window was applied before the calibration delays previously.

So unless there's some other objection, I will merge this today.

PR #881, opened 2025-03-24
Do not try to load run table by default

fschlueter:
It seems nobody actually uses this feature - hence do not even try to load it (and potentially cause a warning for the user).


sjoerd-bouma (approved):
I always set it to False anyway, so I'm happy with the new default : )

PR #883, opened 2025-03-25
Minor speedup of simulation.py

fschlueter:
I replaced a call of `channelSignalReconstructor.run(evt, sim_station, det)` with a custom function which only calculates `maximum_amplitude_envelope` `signal_time`.

Recently the functionality of `channelSignalReconstructor` increased but with that also the time it takes. This call is in `apply_det_response_sim` and crucially runs before the trigger on almost all events. I benchmarked it a bit and the relative improvement is greater than a factor of 10 and the absolute improvement is on average between 2 and 3 ms and up to 10 ms.


fschlueter:
The channelSignalResonstructor runs on the triggered events later. We could also run it on the sim_station later again?

NuRadioMC/simulation/simulation.py
```diff
@@ -1861,3 +1863,21 @@ def integrated_channel_response(self):
     @integrated_channel_response.setter
     def integrated_channel_response(self, value):
         self._integrated_channel_response = value
+
+def _calculate_amp_per_ray_solution(station):
+    """ Calculate the max amplitude and time of the ray solutions
+
+    We are using this function instead of the channelSignalReconstructor
```
anelles, line 1870: Can you be a bit more verbose in your description here. We don't want people to use this function instead of channelSIgnalReconstructor all the time.

anelles (commented):
One tiny thing and then happy to approve

fschlueter:
> One tiny thing and then happy to approve

I updated the doc-string. Also keep in mind that this is a private function indicated by the leading `_`. Hence, this function is not displayed in the documenation.

anelles:
Always assume the dumbest possible ... you know what ;)

anelles (approved):

PR #886, opened 2025-03-26
Docstring

philippwindischhofer:
This slightly extends some docstrings in `rnog_detector` to clarify the proper handling of the time delays introduced by detector channels.

### DIFF
--- a/NuRadioReco/detector/RNO_G/rnog_detector.py
+++ b/NuRadioReco/detector/RNO_G/rnog_detector.py
@@ -1302,11 +1302,11 @@ class Detector():
         time_delay: float
             Sum of the time delays of all components in the signal chain for one channel
 
-        Important note
-        --------------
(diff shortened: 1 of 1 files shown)


NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -1302,9 +1302,15 @@ def get_time_delay(self, station_id, channel_id, use_stored=True, trigger=False)
         time_delay: float
             Sum of the time delays of all components in the signal chain for one channel
 
+        Important note
```
fschlueter, line 1305: According to our documentation we should only use sections which are defined in the [numpydoc docs](https://numpydoc.readthedocs.io/en/latest/format.html). So in this case you could use `Notes`. 
Maybe do:
```
Notes
------
IMPORTANT: bla bla
```

fschlueter (dismissed):

sjoerd-bouma (approved):

Issue #884, opened 2025-03-26
Storing traces in Fourier domain


fschlueter:
Right now any `BaseTrace` object will be written to a `.nur` file in the time domain. Is there a good reason for this? Allowing to store them in the Fourier domain allow to skip two fft operations and only makes the (de)serialize slightly longer (of course we make sure that we do not store the data twice). But maybe I am missing something

sjoerd-bouma:
What's the use case? It slightly speeds up data writing (and data access) for data where you're only interested in the frequency spectrum?

My 2 cents - I don't think we want to worry too much about speed of data writing, and data reading is only sped up if the future user that reads the data wants it in the frequency domain, otherwise it **slows down** data reading. So for consistency and simplicity I think the current implementation is fine.

CamphynR:
It would be convenient for me since I'd like to store a lot of filtered frequency spectra for the thermal noise analysis (on an event per event basis). We want to avoid performing a lot of FFTs in this case.

cg-laser:
we discusses this in the inicemc call, and the main reservation is that the frequency domain representation is ambiguous and depends on our FFT definition. Problems will arise if we change the FFT definition in the future (it already happened once) or if we change the data output format, which can be read stand-alone. 
Did you benchmark how much time you actually save by this?

fschlueter:
I will close this Issue as the corresponding PR is also closed

PR #887, opened 2025-03-26
Allow for storing frequency spectra directly to nur files

CamphynR:
This PR aims to allow for storing frequency spectra directly to nur files. This allows for speedup in use cases in which a lot of filtered/cleaned frequency spectra are to be stored. This PR is based on issue https://github.com/nu-radio/NuRadioMC/issues/884 and should help in avoiding unnecessary FFTs

### DIFF
--- a/NuRadioReco/framework/base_trace.py
+++ b/NuRadioReco/framework/base_trace.py
@@ -298,18 +298,23 @@ class BaseTrace:
         self.set_trace(resampled_trace, sampling_rate)
 
     def serialize(self):
-        time_trace = self.get_trace()
-        # if there is no trace, the above will return np.array(None).
-        if not time_trace.shape:
+        if (self._time_trace) is None and (self._frequency_spectrum is None):
             return None
+        if (self._time_trace is not None) == (self._frequency_spectrum is not None):
+            raise ValueError("Both time trace and frequency spectrum are stored only one of the two should be kept")
         data = {'sampling_rate': self.get_sampling_rate(),
-                'time_trace': time_trace,
+                'time_trace' : self._time_trace,
+                'frequency_spectrum' : self._frequency_spectrum,
                 'trace_start_time': self.get_trace_start_time()}
         return pickle.dumps(data, protocol=4)
 
     def deserialize(self, data_pkl):
         data = pickle.loads(data_pkl)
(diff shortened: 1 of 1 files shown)


sjoerd-bouma:
With our current file structure (just pickling everything), I think this is mostly fine, storing mixed-format data (spectra and traces) shouldn't cause any issues. Long-term, I still think it would be preferable to move to a different file format, where this would probably be a bad idea. But maybe we just cross that bridge if and when we get to it.

I would like someone else to have a think about whether there is a good fundamental reason why we shouldn't do this (maybe we briefly flag it at the meeting tomorrow?), otherwise we can probably merge this.

fschlueter:
@CamphynR we just discussed it in the NuRadio call. There are two arguments against this feature. The first one is mentioned above by Sjoerd the second is that if we ever change our FFT convention it would be very difficult to keep track of this if we read old files which contain the fft spectra. Hence, we need a strong reason to do that. Can you benchmark how much time we are actually saving?

CamphynR:
By looking at the time it takes to process around 1k events, once with the spectra saved and once with the traces saved it seems the reading (using eventReader) takes 1.45 ms per event to load the spectra of all channels for the spectra directly saved and 1.98 ms per event for the traces directly saved. So when reading it leads to a difference of roughly 0.5 ms

fschlueter:
> By looking at the time it takes to process around 1k events, once with the spectra saved and once with the traces saved it seems the reading (using eventReader) takes 1.45 ms per event to load the spectra of all channels for the spectra directly saved and 1.98 ms per event for the traces directly saved. So when reading it leads to a difference of roughly 0.5 ms per event

I also run some superficial tests. I just tested how much the ffts cost. In them most crucial case you would save two ffts per channel, i.e., 48 ffts per RNO-G event. For 1000 events I am getting a run time of ~0.5s:
```
In [11]: %timeit [[fft.time2freq(tr, 3.2) for _ in range(48)] for _ in range(1000)]
434 ms ± 5.09 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
So only if we run over sizeable data (+100k events) we are saving minutes.

fschlueter:
I will close this PR. @CamphynR feel free to reopen it if you get new performance metrics or want to bring this up again.

PR #888, opened 2025-03-26
Fix interferometry.get_time_shifts_plane

fschlueter:
`Fix interferometry.get_time_shifts_plane` did not return the same as `geometryUtilities.get_time_delay_from_direction` for observers which are not on a horizontal plane. Now they do. 

MWE:
```
from NuRadioReco.utilities import geometryUtilities, interferometry
from matplotlib import pyplot as plt
import numpy as np
pos = np.random.normal(0, 100, (2000, 3))


t1 = geometryUtilities.get_time_delay_from_direction(np.deg2rad(30), np.deg2rad(45), pos)

t2 = interferometry.get_time_shifts_plane(positions=pos, zenith=np.deg2rad(30), azimuth=np.deg2rad(45), n0=1.000293)

fig, ax = plt.subplots(1, 2, subplot_kw=dict(projection='3d'))

ax[0].scatter(*pos.T, c=t1)

sct = ax[1].scatter(*pos.T, c=t2)

print(np.allclose(t1, t2))
plt.colorbar(sct)

plt.show()
```

This leaves the question if we should remove one of the functions. The interferometry  method is actually much faster which is why I would go with this implementation (we can keep the function in the more general module `geometryUtilities`.
```
In [16]: %timeit geometryUtilities.get_time_delay_from_direction(np.deg2rad(30), np.deg2rad(45), pos)
711 µs ± 3.71 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [17]: %timeit interferometry.get_time_shifts_plane(positions=pos, zenith=np.deg2rad(30), azimuth=np.deg2rad(45), n0=1.000293)
12.5 µs ± 38 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
What do you think @sjoerd-bouma

### DIFF
--- a/NuRadioReco/utilities/interferometry.py
+++ b/NuRadioReco/utilities/interferometry.py
@@ -2,7 +2,9 @@ import numpy as np
 import sys
 from scipy import signal, constants
 from radiotools import helper as hp
-from NuRadioReco.utilities import units
+from NuRadioReco.utilities import units, geometryUtilities as geo
+import warnings
+
 
 # to convert V**2/m**2 * ns -> V**2/m**2 * s -> J/m**2 -> eV/m**2
 conversion_factor_integrated_signal = 1 / units.s * \
@@ -239,8 +241,8 @@ def get_time_shifts_plane(positions, zenith, azimuth, n0):
                     [0, 0, 1]])
 
     # Rotation around x-axis -> rotation into "shower plane"
-    c = np.cos(-zenith)
-    s = np.sin(-zenith)
+    c = np.cos(zenith + np.pi)
+    s = np.sin(zenith + np.pi)
     e2 = np.matrix([[1, 0, 0],
                     [0, c, -s],
                     [0, s, c]])


sjoerd-bouma:
Without looking at this in more detail - the difference is you vectorize the dot product for more than one position (instead of the for loop in geometryUtilities)? I don't know how common that use case is but I'm happy if you optimise the function in geometryUtilities. I agree maintaining a separate function in the interferometry module that just does the same thing is probably unnecessary.

sjoerd-bouma:
Sorry, cheeky commit, but I couldn't help it : ) try the speed comparison again, if you like, and consider replacing/removing the function in interferometry.

fschlueter:
> Sorry, cheeky commit, but I couldn't help it : ) try the speed comparison again, if you like, and consider replacing/removing the function in interferometry.

You did not run a speed comparison yet?

sjoerd-bouma:
> You did not run a speed comparison yet?

What, you think I don't test my code? Of course I did ;)

fschlueter:
> > You did not run a speed comparison yet?
> 
> What, you think I don't test my code? Of course I did ;)

Than please enlighten us. I liked your implementation more and would have suggested it if I did not find the speed difference.

sjoerd-bouma:
On my end the geometry_utilities function is about 3-4 times faster than the one in interferometry, both for a single position and for an array of positions as in your test case.

fschlueter:
my results:
```
In [1]: from NuRadioReco.utilities import geometryUtilities, interferometry

In [2]: pos = np.random.normal(0, 100, (2000, 3))

In [3]: %timeit  geometryUtilities.get_time_delay_from_direction(np.deg2rad(30), np.deg2rad(45), pos)
6.28 µs ± 12.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [4]: %timeit interferometry.get_time_shifts_plane(positions=pos, zenith=np.deg2rad(30), azimuth=np.deg2rad(45), n0=1.000293)
12.4 µs ± 29.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
not a factor 3 to 4 as @sjoerd-bouma promised but non the less faster (and the better implementation)

sjoerd-bouma:
Actually I just realized there is an issue with my implementation, in that it doesn't return the right shape if `pos.ndim > 2`. This is not officially a supported use case (according to the docstring), but it's a straightforward fix, so I've fixed it.

For Felix
```
%timeit geometryUtilities.get_time_delay_from_direction(np.deg2rad(30), np.deg2rad(45), pos)
11.3 µs ± 536 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
%timeit interferometry.get_time_shifts_plane(positions=pos, zenith=np.deg2rad(30), azimuth=np.deg2rad(45), n0=1.000293)
33.6 µs ± 1.35 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

sjoerd-bouma (approved):
now I'm thinking about whether this generalizes to the case where `zenith` and `azimuth` are also arrays - it doesn't, but probably could be made to do so fairly straightforwardly. But we also probably all have better things to do!

PR #889, opened 2025-03-26
Small fix and improvement to RNO-G detector class

philippwindischhofer:
This fixes:
* Convert device ID back into an integer when an exported detector description is loaded back in

This adds:
* The ability to add "comments" to the detector description upon export, which then gets printed when the resulting file is read back in. This is useful to e.g. keep track of a "changelog" when a calibrated (and hence modified w.r.t. the GPS result) geometry is provided.


philippwindischhofer:
Example output
image

fschlueter (approved):

PR #891, opened 2025-03-27
Remove module NuRadioReco/modules/efieldAirToIcePropagator.py 

fschlueter:
as it misses the actual implementation


anelles (approved):
Thanks for cleaning up ancient left-overs!

fschlueter:
I will merge this right away, I think that is uncontroversial and it may allow us to release a new version today with this already included

PR #892, opened 2025-03-28
Make merge_hdf5.py more robust

fschlueter:
For all non str attributes replaced `np.all(a == b)` with `np.allclose(a, b)` to avoid warnings from numerical noise.


anelles (approved):
Indeed more general! Thanks!

PR #893, opened 2025-03-28
Avoid zero-division in channelSignalReconstructor. Disable VEL caching in simulation.py

fschlueter:
…L caching in simulation.py. Add better warning to efieldToVoltageConverter

These change are already included in PR #863 (but I for see that this will take a while to be merged). This will avoid unnecessary warnings and math errors when running neutrino simulations.


sjoerd-bouma (approved):
Thanks Felix

PR #895, opened 2025-03-31
Ice summit 5th exponential polynomial

oeyenbob:
Pull request to include my empirical 5th-degree exponential ice model for Summit Station based on density data, calibration and weather balloon measurements.

### DIFF
--- a/NuRadioMC/SignalProp/analyticraytracing.py
+++ b/NuRadioMC/SignalProp/analyticraytracing.py
@@ -210,7 +210,7 @@ def get_y_turn( C_0, x1, n_ice, b, delta_n, z_0):
     c = n_ice ** 2 - C_0 ** -2
     gamma_turn, z_turn = get_turning_point(c, b, z_0, delta_n)
     C_1 = x1[0] - get_y_with_z_mirror(x1[1], C_0, n_ice, b, delta_n, z_0)[0]
-    y_turn = get_y(gamma_turn[0], C_0, C_1, n_ice, b, z_0) 
+    y_turn = get_y(gamma_turn[0], C_0, C_1, n_ice, b, z_0)
     return y_turn
 
 def get_delta_y(C_0, x1, x2, n_ice, b, delta_n, z_0, medium_reflection, C0range=(-1.0,-1.0), reflection=0, reflection_case=2):
@@ -402,16 +402,16 @@ class ray_tracing_2D(ray_tracing_base):
             (default: False)
         overwrite_speedup: bool
             The signal attenuation is calculated using a numerical integration
-            along the ray path. This calculation can be computational inefficient depending on the details of 
+            along the ray path. This calculation can be computational inefficient depending on the details of
             the ice model. An optimization is implemented approximating the integral with a discrete sum with the loss
-            of some accuracy, See PR #507. This optimization is used for all ice models listed in 
+            of some accuracy, See PR #507. This optimization is used for all ice models listed in
             speedup_attenuation_models (i.e., "GL3"). With this argument you can explicitly activate or deactivate
             (True or False) if you want to use the optimization. (Default: None, i.e., use optimization if ice model is
             listed in speedup_attenuation_models)
         use_cpp: bool
             if True, use CPP implementation of minimization routines
             default: True if CPP version is available
-            
+
         """
         self.medium = medium
         if(not hasattr(self.medium, "reflection")):
@@ -439,7 +439,7 @@ class ray_tracing_2D(ray_tracing_base):
                 global get_reflection_point,obj_delta_y_square,get_delta_y
                 global get_y_turn, get_y_with_z_mirror,get_turning_point
                 global get_gamma, get_y, get_C0_from_log
-                global  get_z_unmirrored, get_y_diff, n
+                global get_z_unmirrored, n
                 try:
                     get_reflection_point = jit(get_reflection_point, nopython=True, cache=True)
                     obj_delta_y_square = jit(obj_delta_y_square, nopython=True, cache=True)
@@ -453,7 +453,7 @@ class ray_tracing_2D(ray_tracing_base):
                     get_z_unmirrored = jit(get_z_unmirrored, nopython=True, cache=True)
                     n = jit(n, nopython=True, cache=True)
                     self.use_cpp = False
-                except: 
+                except:
                     self.__logger.warning("Error in compiling methods using jit - proceeding without numba")
                     compile_numba = False
 
@@ -502,7 +502,7 @@ class ray_tracing_2D(ray_tracing_base):
                 x1 = x1t
             else:
                 x11, x1, x22, x2, C_0, C_1 = segment
-                
+
             # first treat special case of ice to air propagation
             z_int = None
             points = None
@@ -542,7 +542,7 @@ class ray_tracing_2D(ray_tracing_base):
                 x1 = x1t
             else:
                 x11, x1, x22, x2, C_0, C_1 = segment
-                
+
             # first treat special case of ice to air propagation
             z_int = None
             points = None
@@ -890,8 +890,8 @@ class ray_tracing_2D(ray_tracing_base):
                 x1 = x1t
             else:
                 x11, x1, x22, x2, C_0, C_1 = segment
-                
-            # treat special ice to air case. Attenuation in air can be neglected, so only 
+
+            # treat special ice to air case. Attenuation in air can be neglected, so only
             # calculate attenuation until the ray reaches the surface
             if x2[1] > 0:
                 z_turn = 0
@@ -987,7 +987,7 @@ class ray_tracing_2D(ray_tracing_base):
                             idx -= 1
                         elif idx == -1:
                             idx = 0
-                        
+
                         integrand = integrate.quad(self.ds, segments[idx], segments[idx + 1], args=(C_0), epsrel=1e-2, points=[z_turn])[0]
                         attenuation = (attenuation_util.get_attenuation_length(z_turn, f, self.attenuation_model) for f in freqs)
                         att_int = np.array(
@@ -1001,7 +1001,7 @@ class ray_tracing_2D(ray_tracing_base):
                 else:
                     points = None
                     if x1[1] < z_turn and z_turn < x2_mirrored[1]:
-                        points = [z_turn]                        
+                        points = [z_turn]
                     attenuation_exp = np.array([integrate.quad(dt, x1[1], x2_mirrored[1], args=(
                         C_0, f), epsrel=1e-2, points=points)[0] for f in freqs])
                 tmp = np.exp(-1 * attenuation_exp)
@@ -1481,8 +1481,8 @@ class ray_tracing_2D(ray_tracing_base):
 
         zs = np.linspace(x1[1], x1[1] + np.abs(x1[1]) + np.abs(x2[1]), 1000)
         yz = get_y_with_z_mirror(zs, C_0, self.medium.n_ice, self.__b, self.medium.delta_n, self.medium.z_0, C_1)
-        yy = yz[0] 
-        zz = yz[1]        
+        yy = yz[0]
+        zz = yz[1]
         ax.plot(yy, zz, '-', label='C0 = {:.3f}'.format(C_0))
         ax.plot(x1[0], x1[1], 'ko')
         ax.plot(x2[0], x2[1], 'd')
@@ -1877,13 +1877,13 @@ class ray_tracing(ray_tracing_base):
         ----------
         medium: medium class
             class describing the index-of-refraction profile
-        
+
         attenuation_model: string
             signal attenuation model
-        
+
         log_name:  string
             name under which things should be logged
-        
+
         log_level: logging object
             specify the log level of the ray tracing class
 
@@ -1893,15 +1893,15 @@ class ray_tracing(ray_tracing_base):
             * logging.DEBUG
 
             default is NOTSET (global control)
-        
+
         n_frequencies_integration: int
             the number of frequencies for which the frequency dependent attenuation
             length is being calculated. The attenuation length for all other frequencies
             is obtained via linear interpolation.
-        
+
         n_reflections: int (default 0)
             in case of a medium with a reflective layer at the bottom, how many reflections should be considered
-        
+
         config: dict
             a dictionary with the optional config settings. If None, the config is intialized with default values,
             which is needed to avoid any "key not available" errors. The default settings are
@@ -1913,14 +1913,14 @@ class ray_tracing(ray_tracing_base):
                 * self._config['propagation']['birefringence'] = False
 
         detector: detector object
-        
+
         ray_tracing_2D_kwards: dict
             Additional arguments which are passed to ray_tracing_2D
-            
+
         use_cpp: bool
             if True, use CPP implementation of minimization routines
             default: True if CPP version is available
-            
+
         """
         self.__logger = logging.getLogger('NuRadioMC.ray_tracing_analytic')
         self.__logger.setLevel(log_level)
@@ -1938,7 +1938,7 @@ class ray_tracing(ray_tracing_base):
                          config=config,
                          detector=detector)
         self.set_config(config=config)
-        
+
         self.use_cpp = use_cpp
         if use_cpp:
             self.__logger.status(f"using CPP version of ray tracer")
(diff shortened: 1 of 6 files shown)


anelles:
@philippwindischhofer should definitely take a look at this!

anelles:
While I am happy to review this: @philippwindischhofer should definitely take a look at this!

philippwindischhofer:
Hi @boboeyen! Thanks for this!

I think `IceModel_Polynomial` is not currently implemented in the `master` branch of RadioPropa, but still sits in your branch [here](https://github.com/nu-radio/RadioPropa/tree/ice_model/exponential_polynomial). Are you planning to open a corresponding PR for this on the RadioPropa side?

boboeyen:
> Hi @boboeyen! Thanks for this!
> 
> Will take a closer look at the ice model in a bit. For now, one general question below.
> 
> I think `IceModel_Polynomial` is not currently implemented in the `master` branch of RadioPropa, but still sits in your branch [here](https://github.com/nu-radio/RadioPropa/tree/ice_model/exponential_polynomial). Are you planning to open a corresponding PR for this on the RadioPropa side?

Ah, woops, thanks for noticing that. I have started a pull-request there as well!

boboeyen:
> Hi @boboeyen! Thanks for this!
> 
> Will take a closer look at the ice model in a bit. For now, one general question below.
> 
> I think `IceModel_Polynomial` is not currently implemented in the `master` branch of RadioPropa, but still sits in your branch [here](https://github.com/nu-radio/RadioPropa/tree/ice_model/exponential_polynomial). Are you planning to open a corresponding PR for this on the RadioPropa side?

Ah, woops, thanks for noticing that. I have started a pull-request there as well!

NuRadioMC/utilities/medium_base.py
```diff
@@ -420,6 +420,211 @@ def get_birefringence_index_of_refraction(self, position):
...
+        The latter can be added using the `add_reflective_layer` function.
+
+        The z_shift is a variable introduced to be able to shift the model
+        up or down along the z direction when this is needed to account for
+        differences along station for example.
+
+        Parameters
+        ----------
+        a:  (n,) np.array of floats, NuRadio density units
+            coefficients for the exponential polynomial describing the density profile
```
anelles, line 443: Should "n" be a fixed number of parameters? I think a bit of math would help here ...

boboeyen:
I saw one test failed in the analytical ray tracer, but I didn't change anything in this file. Is this a problem in the develop branch?

fschlueter:
Yes ignore the rnog reader test.

fschlueter:
I will let @philippwindischhofer review this PR as he is much more involved with ice models. However, I have a general comment - is it curical that these functions are fast (I believe so but I am not sure)? If so I think there is quite some potential in removing for-loops (over `self._a` for example even if a is only 5 elements long).

philippwindischhofer:
I sent this to Bob yesterday, but also posting here to have more eyes look at it. In short, I can't get the Poly5 model to give me sensible raytraced propagation times.

Here is a little toy example that traces a short ray through different ice models, calculates the travel time, and compares it to the expectation.

```
import numpy as np                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                             
import NuRadioMC.SignalProp.analyticraytracing as analytic_ray_tracing                                                                                                                                                                                                       
import NuRadioMC.SignalProp.radioproparaytracing as radiopropa_ray_tracing                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                             
def test():                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                             
    ice_simple = medium.greenland_simple()                                                                                                                                                                                                                                   
    raytracer_simple = analytic_ray_tracing.ray_tracing(ice_simple)                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                             
    ice_firn = medium.greenland_firn()                                                                                                                                                                                                                                       
    raytracer_firn = radiopropa_ray_tracing.radiopropa_ray_tracing(ice_firn)                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                             
    ice = medium.greenland_poly5()                                                                                                                                                                                                                                           
    raytracer = radiopropa_ray_tracing.radiopropa_ray_tracing(ice)                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                             
    for label, ice, raytracer in zip(["greenland_simple", "greenland_firn", "greenland_poly5"],                                                                                                                                                                              
                                     [ice_simple, ice_firn, ice],                                                                                                                                                                                                            
                                     [raytracer_simple, raytracer_firn, raytracer]):                                                                                                                                                                                         
                                                                                                                                                                                                                                                                             
        z_deep = -100 * units.m                                                                                                                                                                                                                                              
        ior_deep = ice.get_index_of_refraction(np.array([0, 0, z_deep]))                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                             
        src_pos = np.array([0.0, 0.0, z_deep])                                                                                                                                                                                                                               
        dest_pos = np.array([30.0, 0.0, z_deep])                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                             
        raytracer.set_start_and_end_point(src_pos, dest_pos)                                                                                                                                                                                                                 
        raytracer.find_solutions()                                                                                                                                                                                                                                           
        num_solutions = raytracer.get_number_of_solutions()                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                             
        time_direct = np.linalg.norm(src_pos - dest_pos) / 0.3 * ior_deep                                                                                                                                                                                                    
        time_direct_rt = np.min([raytracer.get_travel_time(iS) for iS in range(num_solutions)])                                                                                                                                                                              
                                                                                                                                                                                                                                                                             
        print(f"{label}: direct = {time_direct:.2f}, raytraced = {time_direct_rt:.2f}" +                                                                                                                                                                                     
              (" !!!" if abs((time_direct_rt - time_direct) / time_direct) > 1e-3 else ""))                                                                                                                                                                                  
                                                                                                                                                                                                                                                                             
if __name__ == "__main__":                                                                                                                                                                                                                                                   
    test() 
```

Here is the output that I get:

```
greenland_simple: direct = 174.52, raytraced = 174.64                                                                                                                                                                                                                        
greenland_firn: direct = 174.13, raytraced = 171.35 !!!                                                                                                                                                                                                                      
greenland_poly5: direct = 174.70, raytraced = 98.40 !!! 
```

RadioPropa seems to be a bit fiddly, numerically, and I had to pick "the right" ray start and end points to make it work for the `greenland_firn` model (otherwise it would only find the reflected ray but not the direct one). However, I couldn't get the `Poly5` model to give me in-ice travel times ...

I know other people here are way more proficient with RadioPropa than I am ... so any input is welcome!

NuRadioMC/utilities/medium_base.py
```diff
@@ -420,6 +419,221 @@ def get_birefringence_index_of_refraction(self, position):
...
+    Predefined ice model (to inherit from) with polynomial exponential shape of degree n
+    """
+    def __init__(self, a, z_0, z_shift=0 * units.meter, z_air_boundary=0 * units.meter, z_bottom=None,
+                 density_factor=0.8506 * (units.cm**3 / units.gram)):
+        """
+        Initiation of an exponential polynomial ice model.
+
+        .. math::
+
+            n(z) = \sum_{i=0}^{n} a_{i} \cdot \exp^{i}(z/z_0)
```
fschlueter, line 433: can you actually write the full and correct equation here:
`n(z) = 1 + density_factor * sum_{i=0}^{n} a_i * exp(i * (z-z_shift) / z_0)`

fschlueter: @boboeyen I fixed the equation. Please check if I understood everything correct.

fschlueter:
@philippwindischhofer it might be that the underlying RadioPropa function `getPropagationTime` not take into account the refractive index. There was a change by Ilse 5 years ago which removed that: https://github.com/nu-radio/RadioPropa/pull/11/files. So at the very least the NuRadio wrapper returns something which is not what you want.

fschlueter:
I looked a bit deeper into the history of RadioPropa. The original author Tobias Winchen did not include the refractive index in the calculation of `PropagationTime`. Than a Carina Kanitz implemented it and Ilse deleted it again. A few thoughts about this:
- The calculation of the `PropagationTime` in RadioPropa should obey the refractive index and the changes made by Ilse are wrong. OR internally the `PropagationTime` should not take into account the refractive index (for some reason unknown to me). If the latter is the case at the very least the wrapper in NuRadio is wrong as in NuRadio the function `get_travel_time` certainly should obey the refractive index

fschlueter:
Ray path for firm model
image
ray path for poly5 model:
image
clearly something is off.

NuRadioMC/utilities/medium_base.py
boboeyen: Yes, my bad. The original equation is not wrong, but then a_i represent the coefficients of the fitted density polynomial translated to ice which is confusing as the translation is done by the class itself.

Would it make more sense to give the function the polynomial coefficient of the ice translation directly, rather than making the class do the translation itself?

boboeyen:
> I looked a bit deeper into the history of RadioPropa. The original author Tobias Winchen did not include the refractive index in the calculation of `PropagationTime`. Than a Carina Kanitz implemented it and Ilse deleted it again. A few thoughts about this:
> 
> * The calculation of the `PropagationTime` in RadioPropa should obey the refractive index and the changes made by Ilse are wrong. OR internally the `PropagationTime` should not take into account the refractive index (for some reason unknown to me). If the latter is the case at the very least the wrapper in NuRadio is wrong as in NuRadio the function `get_travel_time` certainly should obey the refractive index

The `PropagationTime` in RadioPropa is implemented correctly. The refractive index is taken into account in the `PropagationCK` function through the scaling of some unit vector.

boboeyen:
I have inspected it just know and I have found and resolved the issue. The initiation of the RadioPropa scalar field of the polynomial had the wrong translation of the density factor (I was using the inverse of the units stupidly enough)

When running the script on my PC it seems to work alright now with some side nodes.
- Close-by and at the same height, the radiopropa ray tracer seems to have some accuracy issues. You can clearly see this when replacing the analytical raytracer with the radiopropa raytracer for the simple ice model. When I tested it with the source a bit deeper (z=-300m) and a bit further away (x=150m), the accuracy is more what I expect based on the study in my thesis.
- If you use for all ice models the radiopropa raytracer, timings are similar with small variations due to the ice model itself, so the ice model is now correctly implemented in both NuRadio and RadioPropa (I checked this also by simply plotting the profile using both objects)

src = (0,0,-300) & dst = (150,0,-100)
```
greenland_simple: direct = 1454.33, raytraced = 1478.97 (analytic)
greenland_simple: direct = 1454.33, raytraced = 1477.17 (radiopropa)
greenland_firn: direct = 1451.08, raytraced = 1476.02 (radiopropa)
greenland_poly5: direct = 1455.86, raytraced = 1479.23 (radiopropa)
```

fschlueter:
@boboeyen thanks! Somewhat unrelated but I am curious, do you understand, by any chance, this function in RadioPropa (CRPropa):
```
void PropagationCK::tryStep(const Y &y, Y &out, Y &error, double h,
		ParticleState &particle, double z) const {
	std::vector<Y> k;
	k.reserve(6);

	out = y;
	error = Y(0);

	// calculate the sum of b_i * k_i
	for (size_t i = 0; i < 6; i++) {

		Y y_n = y;
		for (size_t j = 0; j < i; j++)
			y_n += k[j] * a[i * 6 + j] * h;

		// update k_i
		k[i] = dYdt(y_n, particle, z);

		out += k[i] * b[i] * h;
		error += k[i] * (b[i] - bs[i]) * h;
	}
}
```

boboeyen:
> @boboeyen I did what you said but still get weird results: (with your changes)
> ```
> greenland_simple (ana): direct = 1454.33, raytraced = 1478.97 !!!
> greenland_simple (rp): direct = 1454.33, raytraced = 1477.17 !!!
> greenland_firn (rp): direct = 1451.08, raytraced = 1476.02 !!!
> greenland_poly5 (rp): direct = 1455.86, raytraced = 833.96 !!!
> ```

Weird!

Could you pull polynomial brach in radiopropa again and recompile/rebuild radiopropa to make sure it's up to date. That caused also a problem for me earlier

boboeyen:
> @boboeyen thanks! Somewhat unrelated but I am curious, do you understand, by any chance, this function in RadioPropa (CRPropa):
> ```
> void PropagationCK::tryStep(const Y &y, Y &out, Y &error, double h,
> 		ParticleState &particle, double z) const {
> 	std::vector<Y> k;
> 	k.reserve(6);
> 
> 	out = y;
> 	error = Y(0);
> 
> 	// calculate the sum of b_i * k_i
> 	for (size_t i = 0; i < 6; i++) {
> 
> 		Y y_n = y;
> 		for (size_t j = 0; j < i; j++)
> 			y_n += k[j] * a[i * 6 + j] * h;
> 
> 		// update k_i
> 		k[i] = dYdt(y_n, particle, z);
> 
> 		out += k[i] * b[i] * h;
> 		error += k[i] * (b[i] - bs[i]) * h;
> 	}
> }
> ```

I once understood it better. But it has something to do with the cash-karp method (a Runge-kutta method) that is used to integrate the eikonal equation. The step used in this method depends on some internal checks

fschlueter:
> > @boboeyen I did what you said but still get weird results: (with your changes)
> > ```
> > greenland_simple (ana): direct = 1454.33, raytraced = 1478.97 !!!
> > greenland_simple (rp): direct = 1454.33, raytraced = 1477.17 !!!
> > greenland_firn (rp): direct = 1451.08, raytraced = 1476.02 !!!
> > greenland_poly5 (rp): direct = 1455.86, raytraced = 833.96 !!!
> > ```
> 
> Weird!
> 
> Could you pull polynomial brach in radiopropa again and recompile/rebuild radiopropa to make sure it's up to date. That caused also a problem for me earlier

That is what I did but I got a compilation error. Also I had to add a cmake flag because apparently the cmake standard in RadioPropa is deprecated by now?

fschlueter:
alright - that did the trick. Thanks @boboeyen:
```
greenland_simple (ana): direct = 1454.33, raytraced = 1478.97 !!!
greenland_simple (rp): direct = 1454.33, raytraced = 1477.17 !!!
greenland_firn (rp): direct = 1451.08, raytraced = 1476.02 !!!
greenland_poly5 (rp): direct = 1455.86, raytraced = 1479.23 !!!
```

sjoerd-bouma:
> greenland_simple (ana): direct = 1454.33, raytraced = 1478.97 !!!
> greenland_simple (rp): direct = 1454.33, raytraced = 1477.17 !!!

I don't know if this is the time or the place for this discussion, but an almost 2 ns discrepancy between radiopropa and the analytic ray tracer for the same ice model is quite large, no? Is this expected? And why doesn't this cause issues in e.g. calibration studies?

boboeyen:
> > greenland_simple (ana): direct = 1454.33, raytraced = 1478.97 !!!
> > greenland_simple (rp): direct = 1454.33, raytraced = 1477.17 !!!
> 
> I don't know if this is the time or the place for this discussion, but an almost 2 ns discrepancy between radiopropa and the analytic ray tracer for the same ice model is quite large, no? Is this expected? And why doesn't this cause issues in e.g. calibration studies?

This might just be a problem of step resolution in the propagation function of RadioPropa. You could try to limit the maximal step size that the PropagationCK module uses ([line 298 in the radioproparaytracing.py](https://github.com/nu-radio/NuRadioMC/blob/488703bcd889eee732dbc85a6d6c854f758fde91/NuRadioMC/SignalProp/radioproparaytracing.py#L298)) to see this makes a difference for close-by vertices. At the moment the min and max stepsize or 1mm and 1 meter which might be to large for close-by vertices.

fschlueter:
> > greenland_simple (ana): direct = 1454.33, raytraced = 1478.97 !!!
> > greenland_simple (rp): direct = 1454.33, raytraced = 1477.17 !!!
> 
> I don't know if this is the time or the place for this discussion, but an almost 2 ns discrepancy between radiopropa and the analytic ray tracer for the same ice model is quite large, no? Is this expected? And why doesn't this cause issues in e.g. calibration studies?

Not that I am particularly happy about the difference but I think what actually meters are differences in the relative time delay between the propagation times of one source and two receivers. I hope this is much lower than 2ns.

PR #896, opened 2025-03-31
Added rno-g cal pulser waveform models

sanyukta13:
Cal pulser waveform measurements with DAQbox at temperature -3C and -18C, taken in lab at Chicago (ref: /mnt/labdata/chicago-rnog-system-response-2024/cal_pulse_temp_dependence)


sanyukta13:
test
pulse waveform at different temperatures - plan to majorly use 4C but good to have -3C and -18C on for testing

PR #898, opened 2025-04-03
Optimize logging in simulation.py

fschlueter:
Small guilty pleasure project. Main change is reformatting the logging output in the init function from something like this
image
to something like this:
image
Also reduces code dublication by slightly restructuring the code.


anelles:
Explain to me, how is this prettier to output more text during logging?

fschlueter:
> Explain to me, how is this prettier to output more text during logging?

Ups I swapped the pictures. The table-style output is the "new" formatting this PR tries to implement.

anelles:
Ah, that makes much more sense!

anelles (approved):
I like it! And otherwise, I see just general clean-up in here.

PR #908, opened 2025-04-14
Fix bug in analyticraytracing.py

fschlueter:
Bug introduced with PR #660 (only causes failure of the raytracing - no wrong physics result). Affected the "optimised" calculation of the attenuation factor with the python raytracer. The optimised is, by default, only used for the GL3 model. 

Additionally, this PR includes some variable renaming. This renaming helped me immensely in understanding what the code does.

### DIFF
--- a/NuRadioMC/SignalProp/analyticraytracing.py
+++ b/NuRadioMC/SignalProp/analyticraytracing.py
@@ -33,7 +33,7 @@ logger = logging.getLogger("NuRadioMC.analytic_ray_tracing")
 cpp_available = False
 
 try:
-    from NuRadioMC.SignalProp.CPPAnalyticRayTracing import wrapper
+    from NuRadioMC.SignalProp.CPPAnalyticRayTracing import cpp_wrapper
     cpp_available = True
     logger.status("CPP version of ray tracer is available")
 except:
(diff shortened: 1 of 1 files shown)


NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -963,61 +967,64 @@ def dt(t, C_0, frequency):
...
                             idx -= 1
                         elif idx == -1:
                             idx = 0
 
-                        integrand = integrate.quad(self.ds, segments[idx], segments[idx + 1], args=(C_0), epsrel=1e-2, points=[z_turn])[0]
-                        attenuation = (attenuation_util.get_attenuation_length(z_turn, f, self.attenuation_model) for f in freqs)
-                        att_int = np.array(
-                            [integrand, attenuation])
+                        integrand = integrate.quad(self.ds, sub_segments[idx], sub_segments[idx + 1], args=(C_0), epsrel=1e-2, points=[z_turn])[0]
+                        attenuation = np.array([attenuation_util.get_attenuation_length(z_turn, f, self.attenuation_model) for f in freqs])
```
fschlueter, line 997: These lines are the relevant bug fix. The rest are only renamings of variables

cg-laser (approved):

fschlueter:
Some tests are failing. I suspect the reason is that the CPP version of the raytracer is not compiling anymore and the python version does not produce the same results anymore?

cg-laser:
yes but why would the C++ compilation fail?

fschlueter:
The CI uses the `ubuntu-latest` OS, I suspect that some dependencies recently changed which causes this to fail. Not sure if it is related but on my Mac it also started to fail very recently (on my list to look into it).

fschlueter:
Nice theory of mine but probably completely wrong. I introduced a bug in the PR which probably cased the CPP RT from being not available. Hopefully fixed it...

fschlueter:
@cg-laser sorry I dismissed your approval.

sjoerd-bouma (approved):

cg-laser (approved):

fschlueter:
Given that the PR was already approved yesterday I will go a head and merge it.

PR #910, opened 2025-04-15
This PR change the method used to integrate over the attenuation length in the cpp raytracer

fschlueter:
The CPP ray tracer is known to occasionally produce Nans for the attenuation length/factor. See #541. This PR changes the (default) method used for integration. The method seems to not produce the Nans and return the same/very similar results otherwise. 

This solution was though of by @EnriqueHuesca. He compiled a long document giving more details: 
[DESY_application_assingment_EHS.pdf](https://github.com/user-attachments/files/19762847/DESY_application_assingment_EHS.pdf)


fschlueter:
Hm, this change seems to not fix the same issue for the ice model GL3 (as attempted as well in this PR #521).

Issue #911, opened 2025-04-16
self.__modules_station keyerror after serializing


CamphynR:
This is probably a very niche issue but I run into a crash in which the __modules_station attribute of the Event class yields a key error. To sketch this problem I include my steps below:

I first create event instances in such a way:
```
import numpy as np
from NuRadioReco.framework.event import Event
from NuRadioReco.framework.station import Station
from NuRadioReco.framework.channel import Channel

event = Event(run_number=-1, event_id=-1)
station = Station(station_id)
station.set_station_time(detector.get_detector_time())
for channel_id in channel_ids:
    channel = Channel(channel_id)
    channel.set_frequency_spectrum(np.zeros_like(frequencies, dtype=np.complex128), sampling_rate)
    station.add_channel(channel)
event.set_station(station)
```
after which I save them to disk using the EventWriter class from NuRadioReco.modules.io.eventWriter. In another step I open these files with NuRadioReco.modules.io.eventReader. If you then loop over the events and apply a module like for example channelBandpassFilter the code will crash because in the register_module_station function of Event  (in NuRadioReco.framework.event) the final line yields a keyError

 ```
def register_module_station(self, station_id, instance, name, kwargs):
        """
        registers modules applied to this event

        Parameters
        ----------
        station_id: int
            the station id
        instance: module instance
            the instance of the module that should be registered
        name: module name
            the name of the module
        kwargs:
            the key word arguments of the run method
        """
        iE = len(self.__modules_event)
        self.__modules_station[station_id].append([iE, name, instance, kwargs])
```
I believe this is due to the serialization of self.__modules_station. In the Event init function this is defined with a collections.defaultdict object

 `self.__modules_station = collections.defaultdict(list)`

But when serializing the self.__modules_station is reverted to a normal python dictionary (as can be seen in the serialize function in NuRadioReco.framework.event.py). If no module was applied before and the self.__modules_station dict is empty the final line of register_module_station will crash with a KeyError.

PR #912, opened 2025-04-16
Initilaize station key in register_module_station if dict is empty

CamphynR:
This is a simple fix to try and resolve Issue https://github.com/nu-radio/NuRadioMC/issues/911


fschlueter:
@CamphynR, thanks for the very detailed issue and this PR.

I am not quite sure if I like the solution because it does not fix the actual problem you are describing: A defaultdict is not surviving the searilization/deserialization. 

What about modifying the `deserialize` function to use this:
```
self.__modules_station.update(data['__modules_station'])
```
This will keep the type of `self.__modules_station` as a defaultdict i think

fschlueter:
This is being addressed in #927

PR #913, opened 2025-04-17
remove logger setting level in db_mongo read and write

CamphynR:
I noticed that these two files set their log levels which caused my script's logger levels to print everything. I removed the setting such that these modules only print the log level of the root logger (unless there was a reason these levels had to be set explicitly?).


sjoerd-bouma (approved):
Looks good to me, thanks Ruben!

PR #916, opened 2025-04-19
Update trace_utilities.py

martin01021207:
Improving the cross correlation calculations in the function **get_coherent_sum()** in the module **trace_utilities**. It's good to do the z-scoring normalization for the traces before calculating the cross correlation to get a more reliable outcome:
(trace - np.mean(trace)) / np.std(trace)

This CSW failed:
Screenshot 2025-04-19 at 5 15 51 PM

With the proper z-scoring normalization, the CSW came out right:
Screenshot 2025-04-19 at 5 12 09 PM

### DIFF
--- a/NuRadioReco/utilities/trace_utilities.py
+++ b/NuRadioReco/utilities/trace_utilities.py
@@ -387,7 +387,7 @@ def get_coherent_sum(trace_set, ref_trace, use_envelope = False):
     def process(trace):
         if use_envelope:
             trace = get_hilbert_envelope(trace)
-        return (trace - np.mean(trace, axis=-1)) / np.std(trace, axis=-1)
+        return (trace - np.mean(trace, axis=-1, keepdims=True)) / np.std(trace, axis=-1, keepdims=True)
 
     n_samples = len(ref_trace)
     ref_processed = process(ref_trace)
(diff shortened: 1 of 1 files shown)


martin01021207:
Also added an extra analysis function to apply the Teager-Kaiser Energy Operator (TKEO) on traces.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -383,19 +383,26 @@ def get_coherent_sum(trace_set, ref_trace, use_envelope = False):
...
-
-        aligned_trace = np.roll(trace, lag)
+            trace = get_hilbert_envelope(trace)
+        return (trace - np.mean(trace)) / np.std(trace)
+
+    n_samples = len(ref_trace)
+    ref_processed = process(ref_trace)
+
+    # Process all traces
+    traces_processed = np.array([process(trace) for trace in trace_set])
```
fschlueter, line 396: if you do:
```
def process(traces):
    return (traces - np.mean(traces, axis=-1)) / np.std(traces, axis=-1)
```
I think you can just write `traces_processed = process(trace_set)`

martin01021207:
@fschlueter Good suggestion and I just made the change, please take a look. If you think it all looks okay now feel free to approve it.

martin01021207:
@fschlueter So the normalization doesn’t affect the actual original waveforms, we use the processed/normalized waveforms to find when they are the most correlated, then we use that time lag to roll the original waveforms and sum them up. This normalization doesn’t affect the waveform itself but it helps better finding the time lag of best correlation. In the failed case, the best time lag was wrong so that’s why those pulses don’t line up.

fschlueter (approved):
Thanks @martin01021207, and sorry I missed that detail

PR #917, opened 2025-04-23
Ska detector compatibility and galactic noise scaling

PhilippLaub:
This PR adds some functions to the SKA detector class to ensure compatibility with `efieldToVoltageConverter`. Furthermore, a scaling parameter is added to `channelGalacticNoiseAdder` to allow noise scaling by a chosen factor.

### DIFF
--- a/NuRadioReco/detector/SKA/detector.py
+++ b/NuRadioReco/detector/SKA/detector.py
@@ -11,7 +11,8 @@ class Detector:
 
     def __init__(
             self, position_path=None, channel_file=None,
-            detector_altitude=460 * units.m, maximum_radius=600 * units.m):
+            detector_altitude=460 * units.m, maximum_radius=600 * units.m,
+            n_samples=1024, sampling_frequency=0.8*units.GHz):
         """
         Simple class to describe an ideal SKA detector.
 
@@ -42,11 +43,17 @@ class Detector:
             Altitude of the detector in meters.
         maximum_radius: float (Default: 600 * units.m)
             Maximum radius of stations to be included when reading from file.
+        n_samples: int (Default: 1024)
+            Number of ADC samples per channel.
+        sampling_frequency: float (Default: 0.8 * units.GHz)
+            Sampling frequency of the channels.
         """
 
         self.logger = logging.getLogger("NuRadioReco.detector.SKA.detector")
         self.detector_altitude = detector_altitude
         self.maximum_radius = maximum_radius
+        self.__n_samples = n_samples 
+        self.__sampling_frequency = sampling_frequency # TODO: move to .json file once these values are confirmed 
 
         if channel_file is None:
             channel_file = os.path.join(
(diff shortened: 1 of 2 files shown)


NuRadioReco/detector/SKA/detector.py
```diff
@@ -42,11 +43,17 @@ def __init__(
...
+            Number of ADC samples per channel.
+        sampling_frequency: float (Default: 0.8 * units.GHz)
+            Sampling frequency of the channels.
         """
 
         self.logger = logging.getLogger("NuRadioReco.detector.SKA.detector")
         self.detector_altitude = detector_altitude
         self.maximum_radius = maximum_radius
+        self.__n_samples = n_samples 
+        self.__sampling_frequency = sampling_frequency # TODO: move to .json file once these values are confirmed
```
fschlueter, line 56: THis is a practical solution but as you indicated in your comment the better place for this would be in the json file. Why not putting it already there and having the getter function taking it from there?

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -361,6 +365,9 @@ def run(
                     + antenna_response['phi'] * channel_noise_spec[2]
                 )
 
+                # scale noise spectrum:
+                channel_noise_spectrum *= self.scaling
+
```
fschlueter, line 370: @cg-laser are you fine with that? As Philipp wrote in the doc-string this is quite useful for us in SKA but it is somewhat "hacky" which I would like to get your approval.

NuRadioReco/detector/SKA/detector.py
PhilippLaub: I wanted to have an easy way of changing these values in order to be able to experiment with them (at least as long as we are still not entirely sure about them).

fschlueter: I see, okay

fschlueter (dismissed):

fschlueter:
Failing tests can be ignored

sjoerd-bouma:
At first glance these changes feel like 'hacks' that make a fairly specific use case easier. What's the long term plan? If it is to eventually deprecate most of these parameters again... I'm not sure `develop` is the place to put these.

PhilippLaub:
I mean it is a bit hacky, but it does not break/change anything for people that don't use it. Also it should stay useful for extremely dense arrays where simulating the whole array is not efficient.

sjoerd-bouma (approved):
> I mean it is a bit hacky, but it does not break/change anything for people that don't use it. Also it should stay useful for extremely dense arrays where simulating the whole array is not efficient.

If two of our most active SKA developers consider this a useful addition, I won't stop you : ) One optional nitpicky comment.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -105,6 +106,9 @@ def begin(
         caching: bool, default: True
             If True, the antenna response is cached for each channel. This can speed up this module
             by a lot. If the frequencies of the channels change, the cache is cleared.
+        scaling: float, default: 1.0
+            Scaling factor for the noise. This is useful when doing interferometry with every 
+            nth antenna (in this case one would set the scaling to 1/\sqrt{n}).
```
sjoerd-bouma, line 111: This docstring could be slightly more precise - I guess the use case is not when you do interferometry, but as a shortcut to what you would get from interferometry without the computational overhead of actually simulating all antennas, or am I wrong?

PhilippLaub: I see your point. I will make it clearer.

sjoerd-bouma (approved):

sjoerd-bouma:
The documentation failing is unrelated to this PR and due to https://github.com/sphinx-doc/sphinx/issues/13533, which I imagine will get fixed soon, shouldn't stop you from merging this.

PR #918, opened 2025-04-24
Updates attenuation calculation for the cpp wrapper. Implements numba support for the python version of the GL3 attenuation model

fschlueter:
Most significant changes in this PR:

- Changes the integration method and max tolerated error in integration (important for GL3) for attenuation calculation in cpp wrapper
- changes variable names in analytic raytracer + make one function numba compatiable
- Support GL3 attenuation model in python with numba
- Some more cleanup
- Add python script to compare attenuation calculations

### DIFF
--- a/NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp
+++ b/NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp
@@ -297,9 +297,18 @@ double get_attenuation_along_path(double pos[2], double pos2[2], double C0,
 	F.params=&params;
 
 	double result, error;
-	double epsabs = 1.e-4; // small initial absolute error
-	double epsrel = 1.e-5; // small initial relative error
-	double max_epsrel = 1.e-3; // max excepted relative error
+	double epsabs = 0; // 0 -> only take into account relative error
+	double epsrel = 1.e-7; // small initial relative error
+	double max_epsrel = 1.e-5; // max excepted relative error
+
+	// We have to tolerate larger uncertainties for the GL3 model because it's rougthness is
+	// a challange for the integration
+	if (model == 5) {
+		epsabs = 1.e-4;
+		epsrel = 1.e-5;
+		max_epsrel = 1.e-3;
+	}
+	
 	int max_badfunc_tries = 5;
 	int num_badfunc_tries = 0;
 	double delta_epsrel = (max_epsrel - epsrel) / max_badfunc_tries; // small initial relative error
--- a/NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp
+++ b/NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp
@@ -287,14 +287,14 @@ double dt_freq (double t, void *p){
 double get_attenuation_along_path(double pos[2], double pos2[2], double C0,
 		double frequency, double n_ice, double delta_n, double z_0, int model) {
 
-	double x2_mirrored[2]={0.};
-	get_z_mirrored(pos,pos2,C0,x2_mirrored, n_ice, delta_n, z_0);
+	double x2_mirrored[2] = {0.};
+	get_z_mirrored(pos, pos2, C0, x2_mirrored, n_ice, delta_n, z_0);
 
 	gsl_integration_workspace *w = gsl_integration_workspace_alloc(2000);
 	gsl_function F;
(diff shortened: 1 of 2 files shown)


fschlueter:
The plots are generated with the new example script. They compare the attenuation factor for the model GL3 calculated with the cpp wrapper and the optimized version of the python implementation (+ time of calculations).  
image
image

image

fschlueter:
For comparison, this is for the model GL1 and the python impl uses the `sicpy.integrate.quad` function:
image
image
image

NuRadioMC/utilities/attenuation.py
```diff
@@ -259,9 +245,63 @@ def get_attenuation_length(z, frequency, model):
     else:
         att_length_f[att_length_f < min_length] = min_length
         att_length_f[z > 0] = np.inf
+
     return att_length_f
 
 
+try:
```
sjoerd-bouma, line 252: Just to understand - these are just jitted versions of the `scipy.interpolate` classes defined at the top of the file? I would have complained about the depth sign convention but if this is just to ensure these functions have the same output as `__call__` on the already-defined classes, I think it's fine to overwrite them. Just feels a bit odd (I guess one could consider making all of these private, too).

sjoerd-bouma: ...and I guess you've checked that using a jitted version of `np.interpolate` is faster than the scipy class (I don't need exact numbers, just would be good that there's a point to doing this)?

NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp
```diff
@@ -285,20 +285,34 @@ double dt_freq (double t, void *p){
...
+	if (model == 5) {
+		epsabs = 1.e-4;
+		epsrel = 1.e-5;
+		max_epsrel = 1.e-3;
+	}
+	
+	int max_badfunc_tries = 5;
+	int num_badfunc_tries = 0;
+	double delta_epsrel = (max_epsrel - epsrel) / max_badfunc_tries; // small initial relative error
+
```
sjoerd-bouma, line 315: I think there's something to say for the previous logarithmic increase in the maximum error, rather than the linear one here - you're going from 1e-7 to 2e-6 immediately, now, which is quite a large jump. But this is probably a bit nitpicky.

sjoerd-bouma (commented):
I guess some changes to the channelGalacticNoiseAdder snuck in unintentionally? Looks like you might have to solve a merge conflict for those, too. But they're harmless enough that I don't really care.

Other than that I think this looks good. I'm a bit surprised to see some rare O(1%) differences for GL1 - while this is unlikely to really cause issues, I had hoped that any 'numerical issues' here would just be orders of magnitude below stuff that we might have to care about in the future. On a similar note, it might be interesting to look at the specific cases where GL3 has a difference of 20% or more to see if we understand where this comes from, just to make sure we are justified in not worrying about this.

NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp
fschlueter: done

NuRadioMC/utilities/attenuation.py
sjoerd-bouma: depth being positive, I'm not a fan because we generally work with `z<0`. But seeing as this is (or should be) an internal function I can live with it.

sjoerd-bouma (dismissed):
If you still want to look into the 3 vertices where Python and C++ disagree by >15% for GL3, I think this might be interesting (or if it's straightforward to identify the source/emitter positions for those you can just post them here), but other than that I think this looks okay and can be merged.

fschlueter:
another example

fschlueter:
Maybe we have to fall back to #521?

fschlueter:
Anyway this calculation is significantly slower as PR #521. I suggest to merge this despite the issue and in practice use 521 for GL3.

fschlueter:
These are results with the integration method `gsl_integration_qags` (instead of `gsl_integration_qag`). Results seem a bit better maybe? 
image
image
image
image

fschlueter:
okay I merged #521 into this branch.

fschlueter:
After #521 has been merged this PR solely reimplements the logarithmic steps for the relative error. + a fix in the python script. @sjoerd-bouma would you approve this?

fschlueter:
Feel free to merge it in my absense.

PR #920, opened 2025-04-28
Patch RNO-G simulation example

fschlueter:
This PR is based on #863 but does not include the implementation of the "data-driven noise". 

This PR does:
- renames an example folder + changes in the example script simulate.py
- Changes to the logger (specifically in ARZ.py)
- Minor clean up

sorry for the messy diff.

### DIFF
--- a/NuRadioReco/modules/trigger/highLowThreshold.py
+++ b/NuRadioReco/modules/trigger/highLowThreshold.py
@@ -171,7 +171,6 @@ class triggerSimulator:
             trigger_adc=True,
             clock_offset=0,
             adc_output='voltage',
-            pre_trigger_time=55 * units.ns,
             step=1,
             align_strides_to_start=False,
             pre_trigger_time=None):
@@ -217,8 +216,6 @@ class triggerSimulator:
             Options:
             * 'voltage' to store the ADC output as discretised voltage trace
             * 'counts' to store the ADC output in ADC counts
-        pre_trigger_time: float (default: 55 ns)
-            The time data before the trigger occured is stored (i.e., defines the readout window).
         step: int
             stride length for sampling rate and clock rate mismatch in trigger logic
         align_strides_to_start: bool (default: False)


fschlueter:
@avijai2000 maybe?

avijai2000:
Sorry I don't know if I'm missing a part of the conversation, all I can see is "maybe?". Would you like me to check this or something? @fschlueter

fschlueter:
any updates?

avijai2000:
Oh sorry yeah this looks good to me

anelles (approved):
I trust that @avijai2000 meant to approve this.

avijai2000:
Yeah sorry I didn't know how to do that

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1942,8 +1942,10 @@ class describing the index-of-refraction profile
             if True, use CPP implementation of minimization routines
             default: True if CPP version is available
 
+        compile_numba: bool (default: None)
+            Only relevant if `use_cpp` is False. If None, the default is True (if `use_cpp` is False).
```
sjoerd-bouma, line 1946: I'm not sure how I feel about `None` instead of `True`, usually explicit is better than implicit, but I guess there is some ambiguity because it gets overwritten by `use_cpp`

NuRadioMC/examples/08_RNO_G_trigger_simulation/simulate.py
sjoerd-bouma, line 1: Not super relevant, but please use `git mv` in future rather than deleting and readding files so the git history is kept.

NuRadioMC/simulation/simulation.py
```diff
@@ -583,7 +583,7 @@ def apply_det_response(
     station = evt.get_station()  # will raise an error if there are more than one station, but this should never happen
     # convert efields to voltages at digitizer
     if detector_simulation_part2 is not None:
-        detector_simulation_part2(evt, station, det)
+        detector_simulation_part2(evt, station, det, add_noise=add_noise)
```
sjoerd-bouma, line 586: This seems like a bug fix? (Not complaining, just making sure this is intentional)

NuRadioReco/modules/trigger/highLowThreshold.py
```diff
@@ -215,12 +216,14 @@ def run(self, evt, station, det,
             Options:
             * 'voltage' to store the ADC output as discretised voltage trace
             * 'counts' to store the ADC output in ADC counts
+        pre_trigger_time: float (default: 55 ns)
+            The time data before the trigger occured is stored (i.e., defines the readout window).
```
sjoerd-bouma, line 220: I don't mind whether this or #940 gets merged first, but in either case please document also the option to use a `dict` to set the pre trigger times, and/or link to the documentation of this feature in `framework.trigger.Trigger`.

sjoerd-bouma (approved):
I'll admit I haven't tested the example, probably I should, but all the minor changes to core files look good.

sjoerd-bouma:
> @sjoerd-bouma why is that test failing? Due to the change in the default inelasticity model, no event is triggering anymore?

Yeah, I did run some checks (without proposal) where the Veff decreased only by the expected 10-20%, but the webinar test has gone from ~10 triggering events to ~3, and it's not seeded, so pretty regularly no events trigger and cause the tests to fail. We can have a bit more of a look or just set a favourable seed for this test.

PR #922, opened 2025-04-30
Fix calculation of frequency vector for attenuation calculation

fschlueter:
This PR rewrites the function `__get_frequencies_for_attenuation`. Previously the function would not obey the `max_detector_freq` argument and produce a frequency vector with a maximum frequency identical to the input vector.

A MWE for this:
```
import numpy as np

def get_frequencies_for_attenuation(frequency, max_detector_freq, n_frequencies_integration=100):
    mask = frequency > 0
    nfreqs = min(n_frequencies_integration, np.sum(mask))
    freqs = np.linspace(frequency[mask].min(), frequency[mask].max(), nfreqs)
    if(nfreqs < np.sum(mask) and max_detector_freq is not None):

        if nfreqs < np.sum(mask) and max_detector_freq is not None:
            mask2 = frequency <= max_detector_freq
            nfreqs2 = min(n_frequencies_integration, np.sum(mask2 & mask))

            freqs = np.linspace(frequency[mask2 & mask].min(), frequency[mask2 & mask].max(), nfreqs2)
            if(np.sum(~mask2) > 1):
                freqs = np.append(freqs, np.linspace(frequency[~mask2].min(), frequency[~mask2].max(), nfreqs // 2))

    return freqs

# For a simulated sampling rate of 5 GHz, abitrary long trace
frequencies = np.linspace(0, 2.5, 2000)

# RNO-G nyquist frequency of 1.2 GHz, 100 points
freqs = get_frequencies_for_attenuation(frequencies, 1.6, n_frequencies_integration=100)
print(freqs[-1])
```

In particular I do not understand these two lines in the original code:
```
            if(np.sum(~mask2) > 1):
                freqs = np.append(freqs, np.linspace(frequency[~mask2].min(), frequency[~mask2].max(), nfreqs // 2))
```
@cg-laser do I miss something here?

This caused issues for the CPP version of the raytracer + GL3.

### DIFF
--- a/NuRadioMC/SignalProp/analyticraytracing.py
+++ b/NuRadioMC/SignalProp/analyticraytracing.py
@@ -852,17 +852,35 @@ class ray_tracing_2D(ray_tracing_base):
         return np.sqrt(1/f_inverse_squared)
 
     def __get_frequencies_for_attenuation(self, frequency, max_detector_freq):
-            mask = frequency > 0
-            nfreqs = min(self.__n_frequencies_integration, np.sum(mask))
-            freqs = np.linspace(frequency[mask].min(), frequency[mask].max(), nfreqs)
-            if(nfreqs < np.sum(mask) and max_detector_freq is not None):
-                mask2 = frequency <= max_detector_freq
-                nfreqs2 = min(self.__n_frequencies_integration, np.sum(mask2 & mask))
-                freqs = np.linspace(frequency[mask2 & mask].min(), frequency[mask2 & mask].max(), nfreqs2)
-                if(np.sum(~mask2) > 1):
-                    freqs = np.append(freqs, np.linspace(frequency[~mask2].min(), frequency[~mask2].max(), nfreqs // 2))
-            self.__logger.debug(f"calculating attenuation for frequencies {freqs}")
-            return freqs
+        """ Returns a frequency vector for the attenuation calculation.
+
+        It takes the frequency vector of a simulated electric field and makes it sparser
+        and limits the maximum frequency. This function is used to reduce the number of
+        frequencies for which the attenuation is calculated (which is time consuming).
+        Afterwards the attenuation factors for the missing frequencies can be interpolated.
+        The maximum frequency is set to the maximum frequency of the detector (if given).
+
+        Parameters
+        ----------
+        frequency: array
+            Frequency vector of the simulated electric field
+        max_detector_freq: float
+            Maximum frequency of the detector (the nyquist frequency)
+
+        Returns
+        -------
+        freqs: array
+            Sparse frequency vector for the attenuation calculation
+        """
+        if max_detector_freq is None:
+            max_detector_freq = np.max(frequency)
+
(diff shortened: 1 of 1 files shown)


cg-laser (changes requested):
This changes the behaviour. 
The previous (admittedly obscure) logic was to also calculate the attenuation lengths for frequencies outside of the detector sampling rate, but at a coarser resolution. Within the range from 0 to max_detector_frequency, the attenuation length is calculated in small step sizes. For the higher frequencies, the stepsizes are larger.
I the current implementation, I think, the attenuation length will be zero for all frequencies larger than max_detector_freq ,corresponding to a hard unphysical rectangular lowpass filter.

cg-laser:
In the second if statement, the frequency array is extended up to the highest requested frequency.

fschlueter:
but isn't that okay, having a rectangular bandpass filter at the nyquist frequency? Typical detectors will  apply another filter with a lower bandwidth. In any case with the current implementation my simulations would crash because the attenuation would introduce INF or very large numbers.

fschlueter:
For the ice models GL1 and GL3 the attenuation length is becoming negative at ~1.5GHz (in the python implementation we set a minimum attenuation length of 1m. But I am not sure how useful the attenuation factor in this range is.
image

fschlueter:
@christophwelling maybe you can comment on the validity of the attenuation length (for GL1, 2, 3) above ~ 1 GHz and what your suggested course of action is.

christophwelling:
I would not trust this model above 1GHz at all. Our measurements only went to about 350MHz, and upper limits up to 600MHz, from which we did a linear fit. That fit should hold to ~1GHz or so, but not much further (figure 4 here: https://eprints.lib.hokudai.ac.jp/dspace/bitstream/2115/32469/1/P185-212.pdf, the imaginary permittivity corresponds to attenuation). So above that, the values from this model (and the other Greenland models too, since they are based on the same measurements and the same interpolation) should not be trusted in my opinion.

fschlueter:
@cg-laser what do we want to do here?

cg-laser:
I don't think that it was tried to extrapolate to higher frequencies. Negative attenuation lengths are unphysical. This is just a result of fitting to limited frequency range and not caring how the function develops to higher frequencies (at least this is my suspicion).  The attenuation length decreases with frequency, but it will never become negative. 

Just take the Phased array as an example. The sampling rate is just 500MHz but the filter we use still lets through substantial power above 250MHz. So this PR will produce unphysical artifacts. 

The best fix would be to improve the extrapolations of the attenuation lenght models. 
In addition, we should define a minimal attenuation length (as we should already be doing) of e.g. 1m.

fschlueter:
> The best fix would be to improve the extrapolations of the attenuation lenght models.
> In addition, we should define a minimal attenuation length (as we should already be doing) of e.g. 1m.

we are doing this already.

fschlueter:
Reverted the changes and only slightly modify the function + adding doc-strings.

cg-laser (approved):
thanks Felix

PR #923, opened 2025-04-30
Patch attenuation.h

fschlueter:
Add wrapper for the cpp implementation of `get_attenuation_length`. Implement a `min_length` for all models in the cpp wrapper - same as in the python implementation.


fschlueter:
@christophwelling you can see the attenuation length for GL3 for high frequencies (without the `min_length`) in https://github.com/nu-radio/NuRadioMC/pull/922 

At some point you included a cap of the frequency for this model of 600MHz. I removed this because I think this is the cleaner solution, what do you think?

christophwelling:
That's okay with me. Should we include a warning, since the attenuation model is not really well-difined at these frequencie?

fschlueter:
Do we need a warning if we set the length to 1? We are discussing that currently in #922, this warning would actually fire all the time!

fschlueter:
@christophwelling if you are happy with that can you approve it ?

fschlueter:
or maybe @sjoerd-bouma

anelles (approved):
Urg, dusting off the reading C++ skills. But I think this is fine to merge.

PR #924, opened 2025-04-30
Read git hash on import

fschlueter:
Currently, when you start a bunch of simulations on a batch system, they are start running, some are already finished others are not and you - without thinking about anything evil - change the branch, you create yourself a problem when you want to merge them because the git hash for the files differ. 

This solution has the problem that at import it might make the code slower... 

Not sure if I am happy with that, feel free to comment.

### DIFF
--- a/NuRadioReco/utilities/version.py
+++ b/NuRadioReco/utilities/version.py
@@ -1,3 +1,6 @@
+import NuRadioMC
+import NuRadioReco
+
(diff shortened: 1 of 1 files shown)


sjoerd-bouma (approved):
Thanks Felix! Another one of those tiny things that you seem to be much better at actually fixing than me (it had been on my to-do list too... this should also speed up event writing in some cases!).

PR #925, opened 2025-04-30
Adding Dr. Sjoerd as identity to the release

fschlueter:
Apparently an identity has to be set for the release command.

PR #927, opened 2025-05-01
Hotfix/event deserialization

sjoerd-bouma:
Fixes two minor, obscure issues:
- Opening an older .nur file used to result in `self.__modules_station` **not** being a defaultdict, which could raise a KeyError if subsequently running modules on it
- One debug message in `geometryUtilities` explicitly formats the channel id as `:d`, which raises an error if the channel id is not an integer. While it might be a good idea to make sure that channel ids are always integers in the future, this is not currently enforced, and I don't think this is where we would want to raise an error.

I've also co-opted #925, which this PR is based on. This should fix the automatic release to Github (I tested that giving an arbitrary identity works on a private repository), thanks @fschlueter! For v3.0.0 I made the release manually already.

### DIFF
--- a/NuRadioReco/framework/event.py
+++ b/NuRadioReco/framework/event.py
@@ -688,4 +688,5 @@ class Event(NuRadioReco.framework.parameter_storage.ParameterStorage):
         if "__modules_event" in data:
             self.__modules_event = data['__modules_event']
         if "__modules_station" in data:
-            self.__modules_station = data['__modules_station']
+            for key, value in data['__modules_station'].items():
+                self.__modules_station[key] = value
--- a/NuRadioReco/utilities/geometryUtilities.py
+++ b/NuRadioReco/utilities/geometryUtilities.py
@@ -327,7 +327,7 @@ def fresnel_factors_and_signal_zenith(detector, station, channel_id, zenith):
         t_theta = get_fresnel_t_p(zenith, n_ice, 1)
         t_phi = get_fresnel_t_s(zenith, n_ice, 1)
         logger.debug((
-            "Channel {:d}: electric field is refracted into the firn. "
+            "Channel {}: electric field is refracted into the firn. "
             "theta {:.0f} -> {:.0f}. Transmission coefficient p (eTheta) "
             "{:.2f} s (ePhi) {:.2f}".format(channel_id, zenith / units.deg,
             zenith_antenna / units.deg, t_theta, t_phi)))
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -1,6 +1,6 @@
 name: Build and publish
 
-on: 
+on:
   pull_request:
   workflow_dispatch:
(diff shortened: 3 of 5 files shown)


sjoerd-bouma:
Ah, I didn't see @CamphynR had already reported this issue in #911 and fixed it in #912... sorry for not commenting there. But I think this would also close #911.

NuRadioReco/framework/event.py
```diff
@@ -688,4 +688,5 @@ def deserialize(self, data_pkl):
         if "__modules_event" in data:
             self.__modules_event = data['__modules_event']
         if "__modules_station" in data:
-            self.__modules_station = data['__modules_station']
+            for key, value in data['__modules_station'].items():
+                self.__modules_station[key] = value
```
fschlueter, line 692: why not `self.__modules_station[key].update(data['__modules_station'])`?

NuRadioReco/utilities/geometryUtilities.py
```diff
@@ -327,7 +327,7 @@ def fresnel_factors_and_signal_zenith(detector, station, channel_id, zenith):
         t_theta = get_fresnel_t_p(zenith, n_ice, 1)
         t_phi = get_fresnel_t_s(zenith, n_ice, 1)
         logger.debug((
-            "Channel {:d}: electric field is refracted into the firn. "
+            "Channel {}: electric field is refracted into the firn. "
```
fschlueter, line 330: shouldn't we start enforcing it and also convert any potential string to an integer in the reader module?

fschlueter:
We also want to patch develop, right? Do we have to open two PR or how does that work?

NuRadioReco/utilities/geometryUtilities.py
sjoerd-bouma: Sure, but I'm happy for that to happen in a different PR (probably to develop), and we'll have to be careful not to break all files that have channel ids (or e.g. run numbers, etc.) in the wrong format.

sjoerd-bouma:
> We also want to patch develop, right? Do we have to open two PR or how does that work?

The easiest is to just merge master into develop after each PR into master. I can have a look into automating this at some point, but for now I would just suggest merging this and then making a manual PR `master -> develop`.

NuRadioReco/framework/event.py
fschlueter: The benefit is not out weighing the effort

fschlueter (approved):

NuRadioReco/framework/event.py
sjoerd-bouma: Sorry, maybe I got my tone wrong on that response - the answer was no reason, other than I wasn't 100% sure how `dict.update` works. Having read the documentation it would be equivalent to what I wrote, so I can in principle replace it still. I think I got it confused in my mind with a pandas `update` method that contrary to my expectation did not overwrite existing keys (in my memory, at least).

Issue #792, opened 2025-01-17
Store `antenna_by_depth` in detector json


sjoerd-bouma:
The `Detector` class on initialization has the annoying default argument `antenna_by_depth=True` even though this results in unusable antenna names for most detectors I have seen in the past 4 years. While we could change this and probably save a lot of inexperienced users some headaches at the cost of technically breaking backwards-compatibility, whether or not antennas are stored as full antenna descriptions or as short strings where the antenna depth suffix should be automatically added is a property of the json, not something that is dynamic at class initialization. I therefore think we should move towards just storing this value inside the detector jsons.

This should be a relatively straightforward change.

fschlueter:
I absolutely agree. I think it caused me headaches twice (because I forgot it after the first time). Would be great to avoid it in the future!

cg-laser:
I agree, we can change the default parameter.

PR #928, opened 2025-05-02
Change of the default value for the detector class argument  from  to

fschlueter:
Partially addressing #792. Changing default value - this will break backwards compatibility.


sjoerd-bouma:
I'm not dead set against this - the new default is definitely better than the old one - wouldn't it be a similar amount of effort to just add the key `antenna_by_depth` to all json files as suggested in #792 ?

fschlueter:
> I'm not dead set against this - the new default is definitely better than the old one - wouldn't it be a similar amount of effort to just add the key `antenna_by_depth` to all json files as suggested in #792 ?

what exactly do you mean by this? That you read the value of `antenna_by_depth` directly from the json file?

fschlueter:
I agree it is probably not much more work...

fschlueter:
@sjoerd-bouma, the truth is that I can not (and should not) spend more time on this. I agree that your solution would have been probably the best. However, do you think we can merge this to at least improve the situation slightly?

fschlueter:
@sjoerd-bouma I think we are at a point where we either merge or close this. I have not the time to work an your proposed solution.

fschlueter:
?

sjoerd-bouma:
> ?

I wanted to avoid a merge conflict in case #1049 was more or less ready to merge, but that doesn't seem to be the case. I will try to give it a go next week, I think it should be relatively straightforward to implement this.

fschlueter:
ping

fschlueter:
@sjoerd-bouma can you give us an honest estimate for when you think you can address #1049

fschlueter:
Okay I will close this PR

Issue #882, opened 2025-03-24
Implement bookkeeping and time logging into channelBlockOffsetFitter


fschlueter:
It would be a nice feature if the module would report how often a significant offset was removed and how much time it takes per event.

fschlueter:
@sanyukta13 any update?

sanyukta13:
Hi Felix,

Waiting for the iglu-cal measurements to finalize the pulse out of
cal_board + iglu_cal and then can be used in simulations to feed into the
tx antenna.

Best Regards,
Sanyukta Agarwal
Contact: +1-785-424-2512
sites.google.com/view/sanyuktaagarwal
<http://sites.google.com/view/sanyuktaagarwal>

sanyukta13:
HI Felix, didn't get the time to look into this yet, will look it over the weekend!

sanyukta13:
HI @fschlueter , the function in [channelBlockOffsetFitter](https://github.com/nu-radio/NuRadioMC/blob/add-bookkeeping-channelBlockOffsetFitter/NuRadioReco/modules/RNO_G/channelBlockOffsetFitter.py) which prints the timing info is `end()` which is not being called in either of the rno-g data reading modules so it will not be displayed during data reading. I am not sure if you mean it to be that way? 

Data reading modules I am referring to: dataProviderRNOG and readRNOGDataMattak

fschlueter:
Yes, please do that!

PR #929, opened 2025-05-05
Add bookkeeping channel block offset fitter

sanyukta13:
Resolution of Issue #882 
- added time logging in function remove_offsets (in ms)
- will log anytime the offset being applied is >50% of the vrms, added in function add_offsets

### DIFF
--- a/NuRadioReco/modules/RNO_G/channelBlockOffsetFitter.py
+++ b/NuRadioReco/modules/RNO_G/channelBlockOffsetFitter.py
@@ -17,6 +17,7 @@ from NuRadioReco.modules.base.module import register_run
 import numpy as np
 import scipy.optimize
 import logging
+import time
 
 logger = logging.getLogger('NuRadioReco.RNO_G.channelBlockOffsetFitter')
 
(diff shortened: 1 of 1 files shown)


fschlueter:
@sanyukta13 thank you very much for getting this started. When I made the issue I actually had a different solution in mind (sorry for not making the issue more descriptive). 

What I thought would be very helpful is getting a kind of summary statistic at the end of the program. This can be achieved by - for example - using class member variable  (e.g.`self._channel_counter`) to count the number or fraction of events where a large blockoffset has been removed. Than in the end we can see whether a run was particularly problematic or not. I would also log the combined processing time of the module.

sanyukta13:
HI @fschlueter, a few questions on large offset statistics:
- summary statistic at the end of the program would have the check for offset value (>0.5*vrms) in the following every class where the offset is being applied to the trace? So `add_offsets`, `fit_block_offsets` and `_calculate_block_offsets`? 
- It will then count and store the number of large offsets in self._event_counter? 
- do we want to count and report on large offsets in all the channels combined in a given run or do a channel-wise statistic on large offset percentage in that run

fschlueter:
Hey @sanyukta13, I think I did not get across idea I had in mind with the related issue (it is also purely described - sorry for that). I went ahead and made a first implementation as I had imagined it. Could you validate it by running over a few events? Didnt have the time to do that yet.

fschlueter:
@sanyukta13 did you had time to do that?

fschlueter:
any update on this?

fschlueter:
@sanyukta13

fschlueter:
@sanyukta13

PR #930, opened 2025-05-06
Updating phasedarray trigger modules

fschlueter:
Isolating changes from #799 related to the trigger modules.


NuRadioReco/utilities/signal_processing.py
```diff
@@ -108,7 +108,82 @@ def resample(trace, sampling_factor):
     return resampled_trace
 
 
-def upsampling_fir(trace, original_sampling_frequency, int_factor=2, ntaps=2**7):
+def digital_upsampling(
```
fschlueter, line 111: Why is the function called "digital"_upsampling? Is there something inherent to the upsampling methods only working on digital traces? I am not sure that I like the fact that this function allows to "digitize" the trace after upsampling based on the passed arguments. How about making this automatic - if the input trace is digitial (i.e., has only integer values) we automatically round the trace and if that is not the case we do nothing (at least as an default behaviour?)

NuRadioReco/framework/trigger.py
```diff
@@ -14,14 +14,16 @@ def deserialize(triggers_pkl):
             trigger = SimpleThresholdTrigger(None, None)
         elif(trigger_type == 'high_low'):
             trigger = HighLowTrigger(None, None, None, None, None)
-        elif(trigger_type == 'simple_phased'):
-            trigger = SimplePhasedTrigger(None, None)
```
fschlueter, line 18: Can we simply change this trigger_type? This would cause a `ValueError` if we read an old file with a `simple_phased` trigger. (same argument for `envelope_phased`)

NuRadioReco/utilities/signal_processing.py
ryankrebs016: Digital (sorry it's kind of a non descriptive name) in the sense that it's aimed at upsampling only by integer factors rather than the fractional upsampling/downsampling done in the resample function. It's more suited to something an FPGA could do. It doesn't do full digitization (like resampling to adc frequency and discretizing the voltage), it only forces the values back to ints if the adc_output is counts.

NuRadioReco/framework/trigger.py
ryankrebs016: I guess it needs to keep a deprecated copy of the SimplePhasedTrigger and the EnvelopePhasedTrigger trigger classes for reading old data?

NuRadioReco/utilities/signal_processing.py
fschlueter: Are you okay with the changes I just made @ryankrebs016? Also I am wondering if rounding is actually the correct thing to do. Would an digital upsampling not floor values?

ryankrebs016: These functions look okay to me! Rounding should be kept as that's what I'm doing in firmware. It only added a quick comparison over truncation

anelles:
@fschlueter these will wait until simulations confirm that the implementation works?

anelles:
To me, the validation would be there, if we can reproduce the trigger efficiency curve as measured in the field.

anelles:
But that probably requires the cal pulser implementation to be ready, right?

ryankrebs016:
> To me, the validation would be there, if we can reproduce the trigger efficiency curve as measured in the field.

I think that needs a completed calibration (geometry, cable delays, noise shape, signal chain) to really get right for this trigger. But I could isolate the trigger by running the RTL simulations (which are the gate-level, maybe slightly higher level, simulations of firmware) and extract the waveforms along each stage of the trigger and then compare 1:1 to the trigger modules.

NuRadioReco/framework/trigger.py
```diff
@@ -1,34 +1,39 @@
...
         elif trigger_type == 'int_power':
             trigger = IntegratedPowerTrigger(None, None, None)
-        elif trigger_type == 'envelope_phased':
-            trigger  = EnvelopePhasedTrigger(None, None, None, None)
-        elif(trigger_type == 'rnog_surface_trigger'):
+        elif trigger_type == 'digital_envelope_phased':
+            trigger  = AnalogEnvelopePhasedTrigger(None, None, None, None)
+        elif trigger_type == 'analog_envelope_phased':
+            trigger  = DigitalEnvelopePhasedTrigger(None, None, None)
+        elif trigger_type == 'rnog_surface_trigger':
```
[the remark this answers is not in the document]
ryankrebs016: The thresholding is done differently between the two so yes, I think so. btw it looks analog/digital got swapped in the names

NuRadioReco/modules/phasedarray/phasedArrayBase.py
```diff
@@ -0,0 +1,530 @@
...
+
+        time_step = 1. / sampling_frequency
+
+        ant_z = self._get_antenna_positions(station, det, triggered_channels, 2)
+
+        self.check_vertical_string(station, det, triggered_channels)
+        ref_z = np.max(ant_z)
+
+        # Need to add in delay for trigger delay
+        cable_delays = np.array([det.get_cable_delay(station.get_id(), channel_id) for channel_id in triggered_channels])
```
ryankrebs016, line 97: The delays gained from the measured signal chains matter now. Stealing code from Philipp here https://github.com/RNO-G/antenna-positioning/blob/f67eb61c2591bbe85ccf0003d1491802139d2360/AntPosCal/position_fitting/global_fit/delays.py#L96 We should add this unless there's something in the works that will account for it when calling get_cable_delay()?
```
    def group_delay(freqs, s21):
        omega = 2 * np.pi * freqs
        phase_rad = np.unwrap(np.angle(s21))
```

NuRadioReco/utilities/signal_processing.py
```diff
@@ -224,10 +224,11 @@ def upsampling_fir(trace, original_sampling_frequency, upsampling_factor=2, ntap
     if coeff_gain != 1:
         up_filt = np.round(up_filt * coeff_gain) / coeff_gain
 
+    up_filt = np.trim_zeros(up_filt)
     zero_padded_sig = np.zeros(len(trace) * upsampling_factor)
     zero_padded_sig[::upsampling_factor] = trace
     upsampled_trace = np.convolve(zero_padded_sig, up_filt, mode='full')[
-        len(up_filt) // 2 : len(zero_padded_sig) + len(up_filt) // 2] * upsampling_factor
+        (len(up_filt) // 2) -1 : len(zero_padded_sig) + (len(up_filt) // 2) -1] * upsampling_factor
```
fschlueter, line 231: is that correct? should be the line 227 in an else block @ryankrebs016

NuRadioReco/modules/phasedarray/phasedArrayBase.py
```diff
@@ -83,32 +84,71 @@ def calculate_time_delays(
...
         beam_rolls = []
-        for angle in phasing_angles:
-
-            delays = -(ant_z - ref_z) / cspeed * ref_index * np.sin(angle) - cable_delays
-
-            delays -= np.max(delays)
-            roll = np.array(np.round(delays / time_step)).astype(int)
 
+        for angle in phasing_angles:
+            delays = (ant_z - ref_z) / cspeed * ref_index * np.sin(angle) - cable_delays - group_delays
```
fschlueter, line 102: did you remove the global `-` on purpose?

NuRadioReco/modules/phasedarray/phasedArrayBase.py
```diff
@@ -83,32 +84,71 @@ def calculate_time_delays(
...
+            delays -= np.min(delays)
+            roll = np.array(np.round(delays * sampling_frequency)).astype(int)
             subbeam_rolls = dict(zip(triggered_channels, roll))
             beam_rolls.append(subbeam_rolls)
 
+        self.buffered_delays[station.get_id()] = beam_rolls
+
         return beam_rolls
 
+    def get_response_group_delays(self, station, det, triggered_channels, fmin=0.150, fmax=0.200):
```
fschlueter, line 112: as mentioned above, there is a class function defined for the response class. I think I would rather improve this function instead of defining a new one.

NuRadioReco/modules/phasedarray/phasedArrayBase.py
ryankrebs016: Yeah. When the phasing/rolling happens I wanted the beam traces and powers to look like this. Beam 0 points up so ch 3 needs to physically wait longer until it's samples are needed. For beam 11, that points down, ch 3 is used the soonest so it should happen first.
image

NuRadioReco/utilities/signal_processing.py
ryankrebs016: Moving it to under the `if coeff_gain != 1` statement so the trim actually does something

NuRadioReco/modules/phasedarray/phasedArrayBase.py
ryankrebs016: I'll update this and use it. The group delay will be relevant unless the responses are normalized so they have 0 group delay (ie if the cab_delay_time is actually the full cable delay time without another hidden time delay)

NuRadioReco/modules/phasedarray/phasedArrayBase.py
fschlueter: Sorry not quite sure I understand the plot. I am a bit confused though. only one thing is correct right (having the - sign or not). Why did it work before?

NuRadioReco/detector/response.py
```diff
@@ -480,31 +480,35 @@ def get_time_delays(self):
...
-
-        time_delay = -np.diff(np.unwrap(phase)) / delta_freq / 2 / np.pi
-        mask = np.all([195 * units.MHz < freqs, freqs < 250 * units.MHz], axis=0)[:-1]
-        time_delay1 = np.mean(time_delay[mask])
+        freqs = np.linspace(fmin, fmax, 1000)
+        response = self(freqs) #make sure trigger = true does happen too
+        phase_angle = np.angle(response)
+        unwrapped = np.unwrap(phase_angle)
+        group_delays = -np.gradient(unwrapped) / (2 * np.pi * np.gradient(freqs))
+        avg_delay = np.mean( group_delays[np.logical_and(freqs>fmin, freqs<fmax)] )
```
fschlueter, line 511: thinking about it - it is probably easier to create a frequency vector between `fmin` and `fmax` and not a larger one and than mask it.

fschlueter: upos you are doing that already, hm, why the mask then?

NuRadioReco/modules/phasedarray/phasedArrayBase.py
ryankrebs016: I just wanted the relative times that the beams are "triggering" to match how it happens. As in beam 11 (points down) happens first because all signals arrive at the trigger board at about the same time, but for beam 0 it needs to wait longer for ch 0 to receive a signal. - and max on line 103 will make all channels roll to where ch3 is in the trace. + and min will make all channels roll to where ch0 is in the traces.

fschlueter: ahhh I did not noticed that you change the max to a min.... no I can see that this does not change the actual trigger efficiency. sorry for that

ryankrebs016:
The stuff shown on Monday used this branch. The trigger times might need double checked. I never looked into how generated events looked like in the radiant. An extra thing that would be nice is wrapping the trigger module into a RNO-G phased array trigger so run doesn't look like this 
```
    powerTrigger.run(
        evt, station, det, Vrms=None, trigger_name="power", threshold=power_thresh, triggered_channels=[0,1,2,3],
        phasing_angles=np.arcsin(np.linspace(np.sin(np.deg2rad(60)),np.sin(np.deg2rad(-60)),12)),
        ref_index=1.75, trigger_adc=True, apply_digitization=False, adc_output="counts", upsampling_factor=4,
        upsampling_method="fir", window=24, averaging_divisor=32, step=4, saturation_bits=8, filter_taps=45,
        coeff_gain=256 )
```
but I think this can come later (or get contained in a kwarg dict in an example)

fschlueter:
@ryankrebs016 this are the 4 radiant WFs + trigger time for one example event
image
And that are the corresponding flower waveforms
image

fschlueter:
This plot has the correct y-axis in the trace plot and I added `+1` to all traces which puts the baseline to 0 (meaning the baseline is simulated with -1..., I guess this is something to debug @ryankrebs016 ?)
image

fschlueter:
Flower waveforms zommed in with noise. Looks good?
image

fschlueter:
@ryankrebs016 I am not sure if we should pay much attention to the AI's comments regarding the upsampling. The proposed changes seem not significant? If you are happy, please approve.

fschlueter:
Sorry @ryankrebs016, I merged the branch from #985 to allow the CI to pass module changes introduced in this PR.

fschlueter:
So you have to reapprove. But we can also wait until we see what CI tests fail and update them if the difference seem resonable.

ryankrebs016 (approved):

ryankrebs016:
The beam delays might have changed slightly which could affect it. I would need to probe the values the code calculates. Otherwise I think it's the same

ryankrebs016:
Beam delays were as expected, they were on opposite and negative, so the failing trigger times on the one test is on purpose. Is the Veff test critical since the input files are randomly generated?

ryankrebs016:
### First the trigger times on the failing trigger tests
On the develop branch, this plot is the first and last channel traces with all the beam traces. 
image
The beam traces get moved into the past from rolling, rather than being causally linked to when signals would appear in a daq. The trigger times are then calculated with https://github.com/nu-radio/NuRadioMC/blob/1636d2bc8495b0dfbc3d9560a8d51e60be978d2e/NuRadioReco/modules/phasedarray/triggerSimulator.py#L442, where the trigger delays look like `[[0,-10,-19,-29] ... [-29,-19,-10,0]]* time step`. Since delays of ch0 are all <=0, the trigger times get pushed further before the channel trace. So the calculated trigger times precede the actual signal in the channel traces.

Then on this branch we have 
image.
Beam traces are now rolled into the future so that they are limited to after when the earliest signal arrives. Then the trigger times are calculated with https://github.com/nu-radio/NuRadioMC/blob/4cfdc3217e6763c61ac1d055316435130d9cc77a/NuRadioReco/modules/phasedarray/phasedArrayBase.py#L528 with trigger delays of `[[0,10,19,29] ... [29,19,10,0] * time_step`. The trigger delays don't change anything since the overall shift is built into the beam delays, so the trigger times are now causally okay and follow how firmware would see each beam at different times.

ryankrebs016:
> okay, that explains the changing trigger times. But does it explain the changed Veff (i.e., number of triggered events with fixed seeds?)

I think so. If I revert the beam delay calculations, the test passes which suggests it's from how the beam traces are aligned with respect to the coarsely stepped integration windows of the trigger in the test.

ryankrebs016:
>  In principle it would be good to maintain backwards compatibility but probably not for every price. Did you change anything else? Do you think it is easy for you to make this module backwards compatible again?

This was just updated. I think this happened when I was trying to combine the analog and digital enveloping and didn't cleanup the older one.

> Have you validated the current module?

For the normal phased array or the analog envelope?

fschlueter:
> > In principle it would be good to maintain backwards compatibility but probably not for every price. Did you change anything else? Do you think it is easy for you to make this module backwards compatible again?
> 
> This was just updated. I think this happened when I was trying to combine the analog and digital enveloping and didn't cleanup the older one.
> 
> > Have you validated the current module?
> 
> For the normal phased array or the analog envelope?

the latter

ryankrebs016:
Haven't checked those. I'll try to update those example scripts for the current simulation.py framework today

fschlueter:
You can extend the interface to allow more flexible configuration but I would keep the defaults the same. Or, we can implement a standard class and provide ARA arguments?

fschlueter:
And can you also an entry to the `changelog.txt` and summerize all the changes implemented with this PR? I know it is annoying but can be super helpful in the future.

ryankrebs016:
I'm holding off on updating "Single event test (South Pole)" reference files since this test passes locally on develop and this branch.

ryankrebs016 (approved):

fschlueter:
merging this, since i only dismissed ryans approval due to updating the branch

PR #932, opened 2025-05-08
Use custom numpy pickling only inside NuRadioReco.framework

### LINKED ISSUE #919: New io_utilities break pickling numpy arrays containing objects and dictionaries
I get a segmentation fault with no error message if I try to save a numpy array containing objects or dictionaries after importing NuRadioMC. This happens because of the new `_pickle_numpy_array` and `_unpickle_numpy_array` in `NuRadioReco/utilities/io_utilities.py` that overwrite the standard pickling mechanism for numpy arrays. I have made this minimal script that can be run with `mode = "save"` or `mode = "load"` that reproduces the error:

```
import numpy as np
import pickle
import NuRadioMC

mode = "save"

filename = "test_pickle.pkl"

if mode == "save":
    results = np.zeros([5,10], dtype=object)
    with open(filename, "wb") as f:
        pickle.dump(results, f)

elif mode == "load":
    with open(filename, "rb") as f:
        results = pickle.load(f)

print(results)
```

sjoerd-bouma:
Restricts the custom pickling of numpy objects to use within `NuRadioReco.framework` only. As highlighted by @MartinRavn in #919, currently we overwrite the pickling of all numpy objects globally upon importing NuRadioReco, also affecting other user code. This PR introduces a custom version of `pickle.dumps` that is explicitly imported and used instead of the global `pickle.dumps` inside all modules in `NuRadioReco.framework`. Any other pickling code remains unaffected (notably, we also unfortunately use pickle to store numpy arrays in some other places, e.g. for antenna models, ARZ profiles, etc - this may cause issues in the future, though my recommendation would probably be to replace pickle in those instances altogether with something more stable, e.g. np.savez or hdf5. But this is for a future PR).

Note that the unpickling mechanism is not affected by this change, and therefore in particular `.nur` files created with any previous version of NuRadioMC should remain readable.

Fixes #919

### DIFF
--- a/NuRadioReco/__init__.py
+++ b/NuRadioReco/__init__.py
@@ -1,7 +1,4 @@
 import os
-import copyreg
-import numpy as np
-from .utilities.io_utilities import _pickle_numpy_array, _pickle_numpy_scalar
 import logging
 
 from NuRadioReco.utilities.logging import NuRadioLogger, _setup_logger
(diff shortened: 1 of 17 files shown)


fschlueter (approved):

fschlueter:
I trust that sjoerd has done the due diligence on this change.

sjoerd-bouma:
I still need to update the changelog and version number before merging, and can run some additional tests before then - though of course reviewers can also do this, regardlessly. There's a test script for io in `NuRadioReco/test/benchmarks/test-io.py` (which I should at some point integrate into the CI...). 

But I think the main question for reviewers was if they're happy with this changed implementation. One reason we originally implemented the pickling change as a global modification was to prevent the scenario where someone extends `framework` and forgets to use the custom pickling mechanism. But on balance it is probably better to do things this way.

anelles (approved):
Looks like a clean change.

PR #936, opened 2025-05-22
[DOC] Fix typo in far-field condition equation

sjoerd-bouma:
By chance happened upon (I think?) a typo in the documentation about the far-field condition - the scaling with `1/R` instead of `R` agrees with [wikipedia](https://en.wikipedia.org/wiki/Fresnel_number) and common sense, I think. I also wonder if the 'third' condition mentioned in words in the text isn't equivalent to the Fresnel condition.


EnriqueHuesca (approved):
After a short but productive discussion, there is not much to improve over the existing writing.

PR #939, opened 2025-06-02
Add download of sample data to LOFAR interactive example

sjoerd-bouma:
I added the code to download the sample data for the LOFAR interactive example, so it should run through even for people without access to coma.


karenterveer:
nice work!

karenterveer (approved):

PR #940, opened 2025-06-03
add option to set pre trigger times

cg-laser:
This PR fixes two things:

- adds option to the high/low and PA trigger modules to configure the pre_trigger_times. The new logic is that the pre_trigger_time is a property of the trigger itself. This property is used by the windowCutter module. However, we forgot to make this configurable. This PR introduces an optional kward to the run methods of the trigger module to make this configurable. 

- Return the number of triggered events in the run method of simulation.py. We forgot to do this during the refactoring. Therefore, the runner.py utility broke. It is useful to know how many event trigger, e.g., to keep simulating until a fixed number of events were triggered.

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -1442,6 +1442,8 @@ class simulation:
         logger.status("Starting NuRadioMC simulation")
         self.__time_logger.reset_times()
 
+        i_triggered_events = 0 # counter for triggered events
+
         particle_mode = "simulation_mode" not in self._fin_attrs or self._fin_attrs['simulation_mode'] != "emitter"
         event_group_ids = np.array(self._fin['event_group_ids'])
         unique_event_group_ids = np.unique(event_group_ids)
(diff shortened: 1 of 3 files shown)


fschlueter:
I dont quite understand the purpose of passing the `pre_trigger_times` via `**kwargs`, why not explicitly setting it to `pre_trigger_times=pre_trigger_time,`

fschlueter:
ahh, because you want to use the default in the class definition. I would rather do this: 
```
class Trigger:
    """
    base class to store different triggers
    """

    def __init__(self, name, channels=None, trigger_type='default', pre_trigger_times=None):
        ....
        self._pre_trigger_times = pre_trigger_times or 55 * units.ns
````

fschlueter (approved):
Your call if you want to use my suggestion ...

sjoerd-bouma (approved):
One minor comment about the docstring, otherwise this is a nice improvement. I prefer your solution with `kwargs` over Felix' suggestion :smiling_imp: which obscures the default value in `Trigger` somewhat.

PR #945, opened 2025-06-10
Fix order such that kwargs override the defaults

MijnheerD:
**Issue being addressed** 
The order of the interpolation options merging was wrong, which meant that the defaults were always executed. This fix ensures that the options can be overridden via the function call. 

I did not edit the changelog.txt file because it seems to minor to document, but if you want me to change this then I can add it.


fschlueter (approved):

sjoerd-bouma (approved):
Thanks Mitja! In principle I think this would be worth noting in the changelog as well because there's been a previous release that still has this bug. But the main things is that you've found it and fixed it : )

PR #949, opened 2025-06-16
Report n_triggered_events also for emitter simulations

sjoerd-bouma:
Since the new simulation workflow, the logging output with the number of triggered events and effective volume were not reported for emitter simulations. While the effective volume still does not make sense, this PR makes the logger report the number of triggered events also for emitter simulations, as requested by @ikrav.

### DIFF
--- a/NuRadioMC/simulation/output_writer_hdf5.py
+++ b/NuRadioMC/simulation/output_writer_hdf5.py
@@ -532,7 +532,9 @@ class outputWriterHDF5:
         Returns:
             float: The calculated effective volume (Veff)
         """
-        # calculate effective volume
+        if not self._mout: # if no events triggered, this is an empty dict
+            return None
+
         triggered = remove_duplicate_triggers(self._mout['triggered'], self._mout['event_group_ids'])
         n_triggered = np.sum(triggered)
         try:


anelles:
I think this is for @ikrav to review, whether this solves the reported bug.

ikrav:
> I think this is for @ikrav to review, whether this solves the reported bug.

Martin Liu tesetd this for me, and it works as intended. When a pulser is simulated as the source of the signal, we observe the expected log messages:

"STATUS - 2025-06-25 22:29:55,332 - NuRadioMC.HDF5OutputWriter - fraction of triggered events = 5/5 (sum of weights = 5.00)"

Thanks, Sjoerd!

sjoerd-bouma:
> @sjoerd-bouma One of the unit tests fails, as far as I see, do you know what is wrong there? I don't work routinely with github and I am not sure whether I should not submit my review with approval until all tests pass, or if my review and the tests are independent. I am happy with the change because it successfully gives the requested behavior. Still, it is not easy for me to make an expert judgement about the code changes because they are not trivial and may interact with other parts of the code in a way I can't predict easily. From looking at the diff between old and new code, it looks good, but I don't understand it in full complexity in relation to other parts.

Ironically, if our tests had been working properly, this would not have been caught... but if there are no triggers, trying to access the `triggered` property of an empty output file raised an error. Should be fixed now!

ikrav (approved):
Looks good to me, approving.

PR #954, opened 2025-06-27
Uses non-numba function when passing an array to attenuation function

fschlueter:
I encountered an numba error when running the `simulate_effective_ice_temperature`. This fixes this error. I did not dig to deep to find a better solution ...


fschlueter:
@sjoerd-bouma or @CamphynR can someone approve this PR?

CamphynR (approved):

PR #957, opened 2025-07-04
Minor fixes to docstrings and efield fluence estimator

sjoerd-bouma:
- Fixes a misspelled check for `method == rice_disttribution` in `trace_utilities.get_electric_field_energy_fluence`; just use `rice` instead and raise an error if a non-supported method is provided instead
- Assortment of smaller docstring fixes which I was too lazy to put in a separate PR, but shouldn't affect anything.

### DIFF
--- a/NuRadioMC/utilities/birefringence_models/model_description.txt
+++ b/NuRadioMC/utilities/birefringence_models/model_description.txt
@@ -4,22 +4,24 @@ In order to use the data, spline interpolated functions were fitted to the data 
 
 The previously used models are presented here:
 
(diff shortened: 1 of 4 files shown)


fschlueter (dismissed):

sjoerd-bouma:
Apologies @fschlueter, I was too lazy to make a new PR just to add the RNO-G and LOFAR notebooks to the CI (I had missed that they hadn't been added before), so I decided to abuse this PR to add them. I promise I will merge this the next time you approve : )

fschlueter:
I agree symlinks are fishy - just go with the README at this point

PR #958, opened 2025-07-10
Fix sign in H_theta in WIPL-D antenna models (introduce createLPDA_100MHz_v2_InfFirn_n1.4)

sjoerd-bouma:
After a very long time spent debugging cosmic-ray polarizations, I found an issue with the WIPL-D antenna models in NuRadio. Because zenith `theta` is defined to go from `90 -> -90` degrees from the positive to the negative z-axis in WIPL-D, instead of `0 -> 180` degrees in NuRadio, in addition to the conversion of the zenith, there is also a minus sign when converting the theta-component of the vector effective length `H_theta`. This sign had so far been missing. 

I have updated the `preprocess_WIPLD` functions to include this sign, and corrected it for the `createLPDA_100MHz_InfFirn` and `createLPDA_100MHz_InfFirn_n1.4` antenna models in the new `createLPDA_100MHz_v2_InfFirn` and `createLPDA_100MHz_v2_InfFirn_n1.4` models, respectively. If requested I can correct the other WIPL-D models as well.

The effect of this sign mistake is to swap polarization angles of `x` degrees with angles of `180-x` degrees. This is **not** negligible for incoming electric fields which are not purely theta- or phi-polarized and antenna responses where both the sensitivities to both components are non-negligible. Here's a simple example plot for a realistic parameterized cosmic ray signal demonstrating this (the blue pulse is for the old antenna model, orange shows the updated one)
image

However, this sign mistake does **not** affect previous ARIANNA analyses, as far as I am aware, which have usually defined the polarization angle from the electric field fluences in theta and phi, which are always positive and hence not sensitive to a sign in theta or phi. Neither is it expected to affect the overall expectation for the cosmic-ray trigger rate significantly. However, the distribution as a function of incoming azimuth **will** change.

There is one more cross-check one can do: naively one expects the theta and phi components to contribute coherently to the measured signal if their projections onto the antenna tine direction point in parallel directions, and interfere destructively if they point in opposite directions. This plot shows the relative phases of the theta and phi components as a function of incoming direction for an upward-pointing LPDA with its tines aligned along the x-axis, using the **old** antenna model.
image

In the NuRadio convention, at an azimuth of 45 degrees, the x-component of eTheta is positive, but the x-component of ePhi is negative, so one would expect a phase difference between the two components of ~180 degrees, with the situation reversed for an azimuth of 135 degrees. This is indeed the **opposite** of what is shown for the old antenna model above, and correct for the new antenna models.

### DIFF
--- a/NuRadioReco/detector/antenna_models_hash.json
+++ b/NuRadioReco/detector/antenna_models_hash.json
@@ -25,6 +25,8 @@
 	"createLPDA_100MHz_InfAir.pkl": "380f3c36e46bfeacf337bf68923c42cb435712d6",
 	"createLPDA_100MHz_InfFirn.pkl": "75539869378cf145b0735159fb59e1cb86206b6f",
   "createLPDA_100MHz_InfFirn_n1.4.pkl": "44826dc663ec712ca3f642314c688ce52c314387",
+	"createLPDA_100MHz_v2_InfFirn.pkl": "70efa8ef54742057d9d8c3e0045aa469b4214717",
+  "createLPDA_100MHz_v2_InfFirn_n1.4.pkl": "2feedcc4b0fb562a693f450c520d900688267ccc",
   "createLPDA_100MHz_z1cm_InFirn_RG.pkl": "14157207338067dd363153b5ec761694cae483ad",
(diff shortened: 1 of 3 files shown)


jakobhenrichs:
Yes, thanks for the heads-up! We will change it in the database.

zsselcuk:
Hi, Sjoerd. We have updated the antenna models in the detector DB.

fschlueter (approved):

PR #959, opened 2025-07-11
Feature/impulsive signal reconstruction

sjoerd-bouma:
Some tools for the reconstruction of 'generic' impulsive signals, e.g. wind-induced events at RNO-G, as used in chapter 6 of my thesis ([link](https://ecap.nat.fau.de/wp-content/uploads/2025/05/Final_Thesis_Bouma.pdf)).

I did some tidying up but probably should test this code once more and/or write a small example; I'm also not sure to what extent things will work if using >3 antennas in reconstruction, so this either needs to be improved or a caveat should be added. But I'm opening a draft PR already as a reminder to myself, and also because we have a student who wanted to use the code.


NuRadioReco/modules/impulsiveSignalReconstructor.py
```diff
@@ -0,0 +1,565 @@
...
+        elif method == 'xcorr':
+            dt = get_dt_correlation(channels, pos, **kwargs)
+        elif method == 'simple_threshold':
+            dt = find_threshold_crossing(channels, **kwargs)
+            
+        zenith, azimuth = geometryUtilities.analytic_plane_wave_fit(dt, pos)
+
+        if np.isnan(zenith):
+            logger.error(f'No valid analytic solution, direction reconstruction failed')
+            return
```
fschlueter, line 557: `return None, None`

annanayjaitly:
can we merge this now @fschlueter? I have been using it a lot for my analysis so far and it works well. your last comment acknowledges the only bug i was having

NuRadioReco/modules/impulsiveSignalReconstructor.py
fschlueter: is that `None, None` implied or would that throw an error

fschlueter (dismissed):

fschlueter:
sure, if you say it works. In principle it would be nice to have it included in a little example script. Its used to plot e.g. the direction from where wind noise is comming?

NuRadioReco/modules/impulsiveSignalReconstructor.py
sjoerd-bouma: I think returning (and storing?) nans is fine? It's basically an error.

NuRadioReco/modules/impulsiveSignalReconstructor.py
```diff
@@ -0,0 +1,578 @@
...
+                dt = find_threshold_crossing_from_stft(channels, **kwargs)
+        elif method == 'xcorr':
+            dt = get_dt_correlation(channels, pos, **kwargs)
+        elif method == 'simple_threshold':
+            dt = find_threshold_crossing(channels, **kwargs)
+        else:
+            raise ValueError(f"Invalid value for method ({method}). Options are 'stft', 'xcorr' or 'simple_threshold'.")
+
+            
+        zenith, azimuth = geometryUtilities.analytic_plane_wave_fit(dt, pos)
```
sjoerd-bouma, line 567: @PascalSchr pointed out in #974 that this implicitly uses the refractive index for air. This works if all antennas are at the same depth (the factor `sin(theta)` drops out when converting between the in-air and in-ice angles), but not if they are at different depths.

sanyukta13:
I can maybe write a couple of lines which ensures this same depth
condition, otherwise it'll simply return the in ice zenith, does that sound
good?

Although the ideal expectation is all 3 uplpda are at the same depth but
something with ensuring from the dB. Thanks!

Best Regards
Sanyukta Agarwal

PR #960, opened 2025-07-24
Add ice thermal noise

CamphynR:
This is the initial pull request to add a module that can add thermal noise due to the surrounding ice to thermal noise traces. This module is used in the absolute amplitude calibration and is a modified version of channelGalacticNoiseAdder. We make use of effective temperatures that can be constructed using Steffen's code, which was modified and added to NuRadio by Felix under NuRadioMC/examples/simulate_effective_ice_temperature. These effective temperatures are stored in several json files.

A question I would like to ask is how to add these effective temperatures to NuRadio? Generating the json files is not straightforward since it requires the use of RadioPropa. The json files are around 5kb each and full usage of the module requires at least 3 of these files.

Feel free to comment on this module it is very much a work in progress.

### DIFF
--- a/NuRadioReco/modules/channelIceThermalNoiseAdder.py
+++ b/NuRadioReco/modules/channelIceThermalNoiseAdder.py
@@ -242,7 +242,7 @@ if __name__ == "__main__":
 
     # These files were generated using the code NuRadioMC/examples/simulate_effective_ice_temperature
     # see the README there for more details
-    eff_temperature_dir = "NuRadioReco/examples/RNOG/thermal_noise/eff_temperatures"
+    eff_temperature_dir = f"{NuRadioReco.__path__[0]}/examples/RNOG/thermal_noise/eff_temperatures"
     eff_temperature_files = ["eff_temperature_-1.0m_ntheta100_GL3.json", "eff_temperature_-40m_ntheta100.json", "eff_temperature_-100m_ntheta100_GL3.json"] 
     eff_temperature_paths = [os.path.join(eff_temperature_dir, eff_temp_f) for eff_temp_f in eff_temperature_files]
     thermal_noise_adder = channelIceThermalNoiseAdder()


CamphynR:
I will also add a noise generation example to this PR

NuRadioReco/modules/channelThermalNoiseAdder.py
```diff
@@ -0,0 +1,282 @@
...
+    """
+    Module to generate thermal noise from ice
+    This class is a stripped version of channelGalacticNoiseAdder
+    Intead of using a skymap, the module uses pre generated effective antenna temperatures,
+    generated by Steffen's effective temperature code which was modified by Felix
+    (see NuRadioMC/examples/simulate_effective_ice_temperature )
+    """
+
+    def __init__(self):
+        self.__n_side = None
```
fschlueter, line 32: needed?

NuRadioReco/modules/channelThermalNoiseAdder.py
```diff
@@ -0,0 +1,282 @@
...
+            urlretrieve(grip_url, grip_path)
+
+        grip_temp = np.loadtxt(grip_path, skiprows=39)
+        depth_values = -grip_temp[:,0]
+        temperature_values = grip_temp[:,1]
+        profile = interp1d(depth_values,
+                           temperature_values,
+                           bounds_error=False,
+                           fill_value=(temperature_values[-1],temperature_values[0]))
+        return profile(depth)
```
fschlueter, line 57: please use `NuRadioMC.utilities.attenuation.get_grip_temperature` instead. Make somewhere sure that this is only valid for greenland ice.

NuRadioReco/modules/channelThermalNoiseAdder.py
```diff
@@ -0,0 +1,282 @@
...
+    def solid_angle(self, theta, d_theta, d_phi):
+        return np.abs(np.sin(theta) * np.sin(d_theta / 2) * 2 * d_phi)
+
+    def get_temperature_from_json(self, temperature_file):
+        with open(temperature_file, "r") as file_open:
+            temperature_file_dict = json.load(file_open)
+        z_antenna = temperature_file_dict["z_antenna"]
+        theta = temperature_file_dict["theta"] * units.rad
+        eff_temperature = temperature_file_dict["eff_temperature"]
+        return z_antenna, theta, eff_temperature
```
fschlueter, line 68: can you add a doc-string to this function explaining the origin of the files.

NuRadioReco/modules/channelThermalNoiseAdder.py
fschlueter: maybe even with a warning if the detector site is not summit or so.

NuRadioReco/modules/channelThermalNoiseAdder.py
```diff
@@ -0,0 +1,282 @@
...
+#                               7 : -40, 12: -1.0, 13: -1.0}
+        self.channel_depths = {i : -100 for i in [0, 1, 2, 3, 4, 8, 9, 10, 11, 21, 22, 23]}
+        for i in [5, 6, 7]:
+            self.channel_depths[i] = -40
+        for i in [12, 13, 14, 15, 16, 17, 18, 19, 20]:
+            self.channel_depths[i] = -1.0
+
+        self.debug = debug
+        if self.debug:
+            nr_phi_bins = 16
```
fschlueter, line 110: Is that legit?

NuRadioReco/modules/channelThermalNoiseAdder.py
```diff
@@ -0,0 +1,282 @@
...
+        with open(temperature_file, "r") as file_open:
+            temperature_file_dict = json.load(file_open)
+        z_antenna = temperature_file_dict["z_antenna"]
+        theta = temperature_file_dict["theta"] * units.rad
+        eff_temperature = temperature_file_dict["eff_temperature"]
+        return z_antenna, theta, eff_temperature
+
+
+
+    def begin(self, sim_library_dir, nr_phi_bins=64, debug=False):
```
fschlueter, line 72: The entire function is quite RNO-G specific. I am okay with keeping this module RNO-G specific but than we should move it the `/modules/RNOG` or remove the hard-coded things.

NuRadioReco/modules/channelThermalNoiseAdder.py
```diff
@@ -0,0 +1,282 @@
...
+        self.freqs = freqs
+        d_f = freqs[2] - freqs[1]
+
+        if passband is None:
+            passband = [10 * units.MHz, 1600 * units.MHz]
+
+        passband_filter = (freqs > passband[0]) & (freqs < passband[1])
+
+        site_latitude, site_longitude = detector.get_site_coordinates(station.get_id())
+        station_time = station.get_station_time()
```
fschlueter, line 160: you dont need that.

NuRadioReco/modules/channelThermalNoiseAdder.py
```diff
@@ -0,0 +1,282 @@
...
+
+        self.debug = debug
+        if self.debug:
+            nr_phi_bins = 16
+        self.phis = np.linspace(0 * units.degree, 360 * units.degree, nr_phi_bins)
+        return
+
+
+    @register_run()
+    def run(
```
fschlueter, line 116: can you remove some of the commented code lines in this function

fschlueter (commented):
@CamphynR mostly minor comments. Only the comments regarding the RNO-G specifics in the begin function need so thinking.

NuRadioReco/modules/channelIceThermalNoiseAdder.py
```diff
@@ -0,0 +1,252 @@
...
+    station = Station(station_id)
+    station.set_station_time(detector.get_detector_time())
+    for channel_id in channel_ids:
+        channel = Channel(channel_id)
+        channel.set_frequency_spectrum(np.zeros_like(frequencies, dtype=np.complex128), sampling_rate)
+        station.add_channel(channel)
+    event.set_station(station)
+
+#    eff_temperature_files = ["/add/list/of/eff_temperature/files/here"] 
+    eff_temperature_files = ["/home/ruben/Downloads/eff_temperatures/eff_temperature_-1.0m_ntheta100_GL3.json", "/home/ruben/Downloads/eff_temperatures/eff_temperature_-40m_ntheta100.json", "/home/ruben/Downloads/eff_temperatures/eff_temperature_-100m_ntheta100_GL3.json"]
```
shallmann, line 236: Hi Ruben, can you add a comment with which code these files were generated and change the file paths to something available for the general user (are they small enough to put in the repo directly? otherwise one could use the downloader and the DESY storage to download on the fly - which is also used for antenna-models)

shallmann: Great thanks!

NuRadioReco/modules/channelIceThermalNoiseAdder.py
```diff
@@ -0,0 +1,252 @@
...
+    def __init__(self):
+        self.__antenna_pattern_provider = NuRadioReco.detector.antennapattern.AntennaPatternProvider()
+
+
+    @functools.lru_cache(maxsize=1024 * 32)
+    def get_cached_antenna_response(self, antenna_pattern, zen, azi, *ant_orient):
+        return antenna_pattern.get_antenna_response_vectorized(self.freqs, zen, azi, *ant_orient)
+
+    def solid_angle(self, theta, d_theta, d_phi):
+        return np.abs(np.sin(theta) * np.sin(d_theta / 2) * 2 * d_phi)
```
shallmann, line 35: Maybe you also want to change this to:
```
return np.abs(np.sin(theta+d_theta / 2) * np.sin(d_theta / 2) * 2 * d_phi)
```
... otherwise you get a small shift (which will be negligible if you have very many theta bins)
(`np.sin(theta+d_theta / 2) * np.sin(d_theta / 2) * 2` is actually equivalent to `np.cos(theta)-np.cos(theta+d_theta)` ... personally I find the latter notation more intuitive)

shallmann (changes requested):

NuRadioReco/modules/channelIceThermalNoiseAdder.py
shallmann: ... or maybe you shouldn't since thetas are not calculated in the bin centers... but that was what could have also lead to Felix's confusion, since you always give the first "bin" a zero-temperature with the current implementation

NuRadioReco/modules/channelIceThermalNoiseAdder.py
```diff
@@ -0,0 +1,263 @@
...
+    station.set_station_time(detector.get_detector_time())
+    for channel_id in channel_ids:
+        channel = Channel(channel_id)
+        channel.set_frequency_spectrum(np.zeros_like(frequencies, dtype=np.complex128), sampling_rate)
+        station.add_channel(channel)
+    event.set_station(station)
+
+    # These files were generated using the code NuRadioMC/examples/simulate_effective_ice_temperature
+    # see the README there for more details
+    eff_temperature_dir = "NuRadioReco/examples/RNOG/thermal_noise/eff_temperatures"
```
shallmann, line 245: not sure this will work when running from outside the NuRadioMC directory and/or without the NuRadioMC directory in the PATH... might it be better to replace `NuRadioReco` by the content of the `NuRadioReco.__path__` variable?

shallmann:
> I assume that the temperature files will always be generated by the provided script in your example and no-one will attempt to input hand-crafted files, but just to be on the safe side: might basic sanity checks on the input be useful? (rad vs. deg, thetas not covering the entire zenith range)

Thanks for integrating that

fschlueter:
@CamphynR feel free to merge this

PR #961, opened 2025-07-29
Change hardware db signal chain structure + enable gain calibration

jakobhenrichs:
This PR modifies the RNO-G hardware database read and write code such that it will work with a new and better database structure. Additionally, the code was adjusted to allow for gain calibration factors in the signal chain. The latter will allow to have the impulse response with a gain calibration in the database. 

The database needs to be slightly restructured before this PR can be merged (the code is tested on a test database with a new structure).

Also, before merging the response or detector class needs to be adjusted in order to allow an application of the gain calibration value.

### DIFF
--- a/NuRadioReco/detector/RNO_G/db_mongo_read.py
+++ b/NuRadioReco/detector/RNO_G/db_mongo_read.py
@@ -140,6 +140,8 @@ class Database(object):
 
         self.__digitizer_collection = "digitizer_configuration"
 
+        self.__gain_calibration_collection = "gain_calibration"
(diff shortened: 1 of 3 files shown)


fschlueter:
I think these changes will have as a consequence that detector files exported before will not be compatible anymore. I think this is okay at the current stage of distribution and so on. But for the future we should probably try to avoid that or at least try to make it backwards compatible.

fschlueter (dismissed):

zsselcuk (approved):

PR #962, opened 2025-07-29
Example of how to add thermal noise to simulation events

CamphynR:
This pull request illustrates how thermal noise can be simulated. The thermal noise components are identified to be: thermal noise from ice, galactic noise and electronic noise. This notebook also illustrates the procedure used in the absolute system response calibration of RNO-G.


fschlueter:
@CamphynR thanks for doing this! I am wondering, this examples produces 3 different sets of events, to each is added on of the three noise components. I understand where you are coming from, this is how you have to do it for your study?! However, for a realistic reconstruction (with a in the future calibrated detector), one once to simulate all three noise sources for each event. Hence I would suggest the following structure:
```
imports

def gen_events():
    ....
    
def adding_3_component_noise(event, ...):
    ....
    
events = gen_events()


for event in events:

    adding_3_component_noise(event)
```

That would also allow the function `adding_3_component_noise` to be imported elsewhere.

fschlueter:
Hey Ruben, thanks for adaption this. I have to admit I did not thought about the idea to create a class to wrap all three modules and I still think it might be cleaner to call all three modules in a more NuRadio-style way. But I do not want to delay that much further. However, can you do two things:
1. Move this example + data files in it's own subfolder?
2. Remove redandent code of simulation the three components individually?

fschlueter (approved):

PR #964, opened 2025-08-12
Update quickstart.rst: Fix path of notebook

fschlueter:


fschlueter:
This is a quick fix to make the documentation work. I can understand why this example is also good to have in NuRadioReco/examples. But it also make sense to keep in in MC... Can you work with symbolic links?

PR #970, opened 2025-08-22
Restructure event.py + add missing doc-strings + making a few functio…

fschlueter:


NuRadioReco/framework/event.py
```diff
@@ -43,101 +43,41 @@ def __init__(self, run_number, event_id):
...
-            else:
-                if len(self.__modules_event) == iE:
-                    break
+        self._id = evt_id
 
-                iE += 1
-                yield self.__modules_event[iE - 1]
+    def get_event_id(self):
+        """ Wrapper around get_id() """
+        return self.get_id()
```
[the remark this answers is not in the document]
sjoerd-bouma: @fschlueter I think I agree with our AI overlords in this case; why do you think an additional `get_event_id` method is necessary? For most (probably should be all, but I haven't checked) classes in `framework` we provide a `get_id` method, which I think is both the most straightforward and therefore the best solution.

NuRadioReco/framework/event.py
```diff
@@ -587,7 +541,156 @@ def get_hybrid_information(self):
...
+            The instance of the module that should be registered.
+        name: module name
+            The name of the module.
+        kwargs:
+            The key word arguments of the run method.
+        """
+        iE = len(self.__modules_event)
+        self.__modules_station[station_id].append([iE, name, instance, kwargs])
+
+    def _iter_modules(self, station_id=None):
```
sjoerd-bouma, line 590: I think making the `register_module*` methods private is a good idea, but although I'm not sure that `iter_modules` is a particularly helpful interface, I think there's some value in exposing some method that can display the modules applied to an event / station. So I'm inclined to keep this public for now.

sjoerd-bouma (commented):
Git hasn't done a great job at showing the changes this time (it highlights a lot of code as changed) but I'll trust the CI and you that only method names and docstrings have changed. 
Thanks for tidying things up, two nitpicky comments but other than that I think this should be pretty uncontroversial.

NuRadioReco/framework/event.py
fschlueter: In my opinion its a developer interface (and actually for somewhat advanced development). Hence, I think it is okay to keep in private. If you do not know about the register_module interface you do not need to know about the iter_modules function.

sjoerd-bouma: I don't agree, even if a user is never expected to explicitly write to the list of modules, retrieving it has value (e.g. in a reconstruction, if I want to find out which bandpass filters have been applied to my data). This method probably does not get used much, and there might be more intuitive ways to provide this information, but I think there should be _some_ public interface that provides this possibility.

NuRadioReco/framework/event.py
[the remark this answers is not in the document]
sjoerd-bouma: There's always a tradeoff between providing additional aliases so there's always one that's close to the user's first guess, vs increasing the number of methods in a class which makes it harder to see what its capabilities are. I would have probably erred on the side of not including the wrapper in this case, but seeing as you're reducing the number of unnecessary methods by 2/3 in this PR already, I'll leave it up to you (and your student) in this case.

PR #971, opened 2025-08-22
Fail with an informative error message when cpp raytracer is not avai…

fschlueter:
It seems this is not the intended behavior.

- either loudly fall back to python raytracer if cpp does not compile
- or fail with an informative error msg

I choose the latter.

### DIFF
--- a/NuRadioMC/SignalProp/analyticraytracing.py
+++ b/NuRadioMC/SignalProp/analyticraytracing.py
@@ -418,7 +418,7 @@ class ray_tracing_2D(ray_tracing_base):
                 self.__logger.info('Using C++ raytracer')
         else:
             if use_cpp:
-				msg = ('C++ raytracer was explicitly requested, but is not available (i.e. on-the-fly compilation failed). '
+                msg = ('C++ raytracer was explicitly requested, but is not available (i.e. on-the-fly compilation failed). '
 					   'Abort.... ! Either fix the compilation or set use_cpp to False. '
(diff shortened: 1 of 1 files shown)


NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -417,8 +417,12 @@ def __init__(self, medium, attenuation_model=None,
             else:
                 self.__logger.info('Using C++ raytracer')
         else:
-            self.__logger.warning('C++ raytracer is not available. Using Python raytracer.')
-            self.__logger.warning("check NuRadioMC/NuRadioMC/SignalProp/CPPAnalyticRayTracing for manual compilation")
+            if use_cpp:
+                self.__logger.error('C++ raytracer was explicitly requested, but is not available. Abort.... Either fix the compilation or set use_cpp to False')
+                raise RuntimeError('C++ raytracer was explicitly requested, but is not available. Abort.... Either fix the compilation or set use_cpp to False')
```
sjoerd-bouma, line 422: [nitpick] is it worth still pointing the user towards `NuRadioMC/NuRadioMC/SignalProp/CPPAnalyticRayTracing` in this case?

sjoerd-bouma (approved):

fschlueter:
better?

fschlueter:
> > better?
> 
> Yes and no, more informative error message, but bad formatting and everything is broken. I do recommend you _also_ install pre-commit ;) because this would have been caught.

well I use the Github GUI for this so not sure :D

fschlueter:
ping

PR #974, opened 2025-08-26
Add shallow in-ice forward-folding CR air-shower reconstruction module

sjoerd-bouma:
Adds the shallow in-ice forward-folding reconstruction code for cosmic-ray air showers described in https://ecap.nat.fau.de/wp-content/uploads/2025/05/Final_Thesis_Bouma.pdf under `NuRadioReco/modules/crEfieldReconstructor`.

This is essentially an 'improved' version of the `voltageToAnalyticEfieldConverter`, by explicitly including the signal direction in the forward folding fit (see section 7.2 of the thesis linked above). Note that this PR is based on #959, so should probably be reviewed and merged later.

Todo:
- [ ] Add reconstruction example to `examples` and to the CI.


fschlueter:
Is this RNO-G specific? There seems to be little to do to get this ready?

NuRadioReco/modules/crEfieldReconstructor.py
```diff
@@ -0,0 +1,683 @@
...
+    def get_cosmic_ray_traces(
+            self, zenith, azimuth, amplitude, pol_angle, slope,
+            phase_p0, phase_p1=0, quadratic_term=0, quadratic_term_offset=0.08):
+        """
+        Return the voltage traces
+
+        Convenience function that returns the analytically-parameterized cosmic ray spectra
+        for all channels as an array. For details of the parameterization, see [1]_.
+
+        By default, this function returns the voltage spectra (including antenna and signal chain responses);
```
PascalSchr: this function always returns the trace no?

NuRadioReco/modules/crEfieldReconstructor.py
```diff
@@ -0,0 +1,683 @@
...
+            ## and then we fit everything together
+            self.fix_parameters(False)
+            self.fix_parameters(amplitude=True) # amplitude is determined exactly, no need to fit
+            res1 = self._minimize(x0=res_quadratic.x, **minimizer_options, bounds=bounds)
+
+        if use_MC_direction:
+            res = res1
+        else:
+            # we check the direction once more - it's possible we ended up in a local minimum
+            x0 = copy.copy(res1.x)
```
PascalSchr: res1 is only set within two if statements (519 and 538) and could throw an error if neither condition is met

NuRadioReco/modules/crEfieldReconstructor.py
```diff
@@ -0,0 +1,683 @@
...
+            self.fix_parameters(True)
+            self.fix_parameters(quadratic_term=False)
+
+            x0 = list(res.x)
+            bounds[0][-1] = np.min([-2, x0[4]]) # we limit the quadratic term to be smaller than the linear one
+            bounds[1][-1] = np.max([2, -x0[4]])
+            res_quadratic = self._minimize(x0=res.x, **minimizer_options, bounds=bounds)
+
+            ## and then we fit everything together
+            self.fix_parameters(False)
```
PascalSchr, line 552: if debuging with use_MC_direction, zenith and azimuth are released here before self._minimize

NuRadioReco/modules/crEfieldReconstructor.py
```diff
@@ -0,0 +1,683 @@
...
+            t_shifts_guess, corr = get_dt_correlation(channels, self._channel_positions, templates=templates, full_output=True)
+            max_corr = np.sum(corr)
+            zenith = np.nan
+            i = 0
+
+            # the best guess (by correlation) may not have a solution; in that case, we check the
+            # 'second best' guess by taking the next-highest correlation for the lowest SNR channel
+            while np.isnan(zenith):
+                zenith, azimuth = geometryUtilities.analytic_plane_wave_fit(t_shifts_guess, self._channel_positions)
+                i += 1
```
PascalSchr: "i" is never used and analytic_plane_wave_fit just runs over and over again with no change?
also: analytic_plane_wave_fit(dt, pos, n_index=1.000293) also defaults to n_air

NuRadioReco/modules/crEfieldReconstructor.py
```diff
@@ -0,0 +1,683 @@
...
+        References
+        ----------
+        .. [1] C. Welling et al., Reconstructing the cosmic-ray energy from the radio
+           signal measured in one single station, http://dx.doi.org/10.1088/1475-7516/2019/10/075
+
+        """
+
+        antenna_response = trace_utilities.get_efield_antenna_factor(
+            self._station, self._freqs, self._channels_sorted, self._det, zenith, azimuth, self.__antenna_provider)
+        time_delays = geometryUtilities.get_time_delay_from_direction(zenith, azimuth, self._channel_positions)
```
PascalSchr: "get_time_delay_from_direction(zenith, azimuth, positions, n=1.000293)"    defaults to n_air which does not seem to be correct for our antennas.

NuRadioReco/modules/crEfieldReconstructor.py
```diff
@@ -0,0 +1,683 @@
...
+        """
+
+        n_iterations = 0
+        zenith_guess, azimuth_guess = params[:2] # zenith and azimuth are the first two entries
+        channels = [self._station.get_channel(channel_id) for channel_id in self._channels_sorted[:3]]
+
+        while n_iterations < maxiter:
+            params[0:2] = zenith_guess, azimuth_guess
+            traces_guess = self.get_cosmic_ray_traces(*params)[:3]
+            delta_t_geometry = geometryUtilities.get_time_delay_from_direction(zenith_guess, azimuth_guess, self._channel_positions)
```
PascalSchr: "get_time_delay_from_direction(zenith, azimuth, positions, n=1.000293)" defaults to n_air which does not seem to be correct for our antennas.

NuRadioReco/modules/impulsiveSignalReconstructor.py
```diff
@@ -0,0 +1,568 @@
...
+            else:
+                dt = find_threshold_crossing_from_stft(channels, **kwargs)
+        elif method == 'xcorr':
+            dt = get_dt_correlation(channels, pos, **kwargs)
+        elif method == 'simple_threshold':
+            dt = find_threshold_crossing(channels, **kwargs)
+        else:
+            raise ValueError(f"Invalid value for method ({method}). Options are 'stft', 'xcorr' or 'simple_threshold'.")
+
+        zenith, azimuth = geometryUtilities.analytic_plane_wave_fit(dt, pos)
```
PascalSchr, line 556: analytic_plane_wave_fit(dt, pos, n_index=1.000293) defaults to n_air as well

NuRadioReco/modules/crEfieldReconstructor.py
```diff
@@ -0,0 +1,683 @@
...
+
+        chisq_const = np.sum(self._traces_data**2)
+
+        def chisq(params, return_scale=False):
+            x0_tmp[fit_param] = params
+            traces_model = self.get_cosmic_ray_traces(*x0_tmp)
+            crossterm = np.sum(self._traces_data * traces_model)
+            modelterm = np.sum(traces_model**2)
+
+            scale_min = crossterm / modelterm # overall scaling factor that minimizes chisq
```
PascalSchr, line 336: can modelterm be = 0 ?

NuRadioReco/modules/crEfieldReconstructor.py
```diff
@@ -0,0 +1,700 @@
...
+from NuRadioReco.framework.parameters import electricFieldParameters as efp
+from NuRadioReco.framework.parameters import stationParameters as stnp
+from NuRadioReco.modules.base.module import register_run
+from NuRadioReco.modules.impulsiveSignalReconstructor import get_dt_correlation
+
+logger = logging.getLogger('NuRadioReco.CREfieldReconstructor')
+_REFRACTIVE_INDEX_AIR = 1.000293
+
+class CREfieldReconstructor:
+    """"
```
PascalSchr: there are 4    ' " '    here

PascalSchr (commented):
hope i didn't miss obvious stuff while reading through it

NuRadioReco/modules/crEfieldReconstructor.py
sjoerd-bouma: Not really, unless one explicitly passes an amplitude of 0, which I would consider outside the physically viable solution space.

NuRadioReco/modules/impulsiveSignalReconstructor.py
sjoerd-bouma: Thanks for flagging this - I forgot this PR is / was based on #959. I think I'm happy with the solution in this PR, but I need to think about whether I think it's a good solution for #959 also...

NuRadioReco/modules/crEfieldReconstructor.py
sjoerd-bouma: I have a vivid memory of implementing the logic described in the comment above, but git doesn't, and git is probably right. It probably was slightly overengineered anyway.

sjoerd-bouma:
Thanks for reviewing. The refractive index wasn't actually 'wrong', we reconstruct the cosmic-ray direction in air, but the implicit in-air to in-ice conversion only works for antennas at equal depth (as for ARIANNA, which the original reconstruction was written for). I've now included explicit conversion between in-ice and in-air zenith, so that in principle this reconstruction should now work also for antennas at (slightly) different depths, as long as they all see the same refractive index.

There should be no practical downside to this change; there's a very obscure edge case where the solution space for the analytic plane wave fit that is used as a first guess is now slightly larger which may lead to in-ice angles that are not allowed in-air due to total internal reflection, but this should 1) rarely happen and 2) fall back to a valid first guess for the cosmic-ray direction in this case.

The example I added to the CI fails due to a DeprecationWarning in radiotools, I will try to think of a solution for this and then ping you again.

PR #975, opened 2025-08-26
Pin pygdsm<=1.6.0 for older python versions

### LINKED ISSUE #973: pygdsm breaks compatibility with Python < 3.10
`pygdsm`, which is an optional dependency used by the `channelGalacticNoiseAdder`, has currently broken for Python versions older than 3.10. The reason is new typing specifications in pygdsm that are not compatible with older Python versions (and/or incorrect Python version requirements in pygdsm). The breaking changes have been introduced somewhere since pygdsm 1.6.0, which did work; we are currently on version 1.6.3. 

As far as I can tell, there have not been any major changes in pygdsm since 1.6.0, so the simplest solution would be to pin pygdsm to 1.6.0 for now (alternatively, someone can make a PR at pygdsm to make the typing python 3.7-compatible again).

sjoerd-bouma:
Pins `pygdsm<=1.6.0` on Python versions older than 3.10, because newer pygdsm versions use typing formats incompatible with those Python versions (without updating the minimum Python version in their package requirements).

This fixes the `channelGalacticNoiseadder` test failing on Python >=3.7, <3.10 in the CI.

Closes #973

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ sphinx-rtd-theme = {version = "*", optional = true}
 numpydoc = {version = "*", optional = true}
 proposal = {version = "7.6.2", optional = true}
 pylfmap = {version = "*", optional = true}
+pygdsm = {version = "<=1.6.0", python = "<3.10", optional = true}
 healpy = {version = "*", optional = true}
 MCEq = {version = "*", optional = true}
 crflux = {version = "*", optional = true}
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -47,6 +47,7 @@ sphinx-rtd-theme = {version = "*", optional = true}
 numpydoc = {version = "*", optional = true}
 proposal = {version = "7.6.2", optional = true}
 pylfmap = {version = "*", optional = true}
+pygdsm = {version = "<=1.6.0", python = "<3.10", optional = true}
```
fschlueter, line 50: does that mean only if you are on <=python3.9 you are using <=1.6?

fschlueter:
I am a bit confused, was pygdsm not specified in our dependency list before? And now it is made optional?

pyproject.toml
sjoerd-bouma: Yes

fschlueter:
So can we drop the `optional`?

fschlueter (approved):

PR #981, opened 2025-09-12
add option in channelSignalReconstructor to only compute SNR (faster)

sjoerd-bouma:
The `channelSignalReconstructor` by now computes quite a large number of statistics, which makes it slow-ish (O(60 ms) per RNO-G event on my laptop). I added an option to only compute the SNR in the `run` method, which takes ~1/3 of the time. 

One could consider making this option even more modular (allowing to select a list of statistics to (not) compute), or also add an optional `channel_ids` argument to run it only on some channels.


sjoerd-bouma:
> I am not sure what the best way forward is. Personally I do not like the idea of a dictionary and x different SNR definitions at the same time.
> 
> Do you think a specific module for RNO-G calculating only a few variables with some extra features (assuming common trace length and sampling rate to make the code more efficient) makes sense or creates a new problem on its own.

The SNR dictionary is legacy code at this point... I would also prefer for `parameters` to not store dictionaries, but that's probably a change to suggest in a larger setting.

I do not like the idea of having a separate module for RNO-G to compute quantities like impulsivity, if this is what you're suggesting; if you think there is a big performance uplift to gain from vectorizing over the number of channels if the trace lengths are identical I think this could be implemented as a check inside this module. (I am not so sure that the performance benefit would be that substantial, but I could well be wrong).

sjoerd-bouma:
Also, just for context on this PR - I thought it might be useful to speed up the simulation routine and/or the notebook example for the RNO-G collaboration meeting. I didn't realize until after that the `channelSignalReconstructor` is used only for triggered events, so for most simulations its speed is not very impactful. And of course this wasn't used for the tutorial in the end. 

So we can also close this PR if you think it's not a useful feature.

fschlueter (approved):

PR #984, opened 2025-09-16
Hotfix: correct specification of pygdsm / galacticnoise version requirements.

sjoerd-bouma:
Turns out @fschlueter was right and I did not specify the `pygdsm` optional dependency correctly (it would not be installed on python >= 3.10 previously).

Should be fixed now.


fschlueter (approved):

fschlueter:
this one is on you @sjoerd-bouma (#977) - it is good that we are finally getting the CI fixed.

PR #985, opened 2025-09-16
Change default `Singleton` behaviour

sjoerd-bouma:
This is essentially an extension of / fix for #977. 

# Description of the problem 
As a recap, the previous default of `Singleton` was to always returns the existing instance of a class, but silently ignore any new arguments passed to `__init__`. E.g.
```
from NuRadioReco.detector.detector import Detector

det = Detector(RNO_G/RNO_single_station.json, antenna_by_depth=True)
(...)
new_det = Detector(RNO_G/RNO_single_station.json, antenna_by_depth=False)
```
results in `new_det` ignoring the new value for `antenna_by_depth` (similarly, if I think I load a completely different detector description, I will silently get the old one instead).

#977 attempted to 'fix' this by raising an error if this was the case. However, we have at least one instance where this leads to very unintuitive behaviour: `askaryan.get_time_trace` contains a wrapper for ARZ, which was definitely not doing the right thing before, but with the behaviour introduced in #977 would have to either always call create_new internally (which would degrade performance), or implement some additional caching (which works, but is ugly). In either case, code like
```
from NuRadioMC.SignalGen import askaryan, ARZ

askaryan.get_time_trace(..., model='ARZ2019')

(...)
arz = ARZ.ARZ()
```
would raise the Singleton error because the ARZ class was already initialized internally in the askaryan module with non-default arguments. I find this not very intuitive.

# Suggested fix
The new suggested implementation of the Singleton sets `create_new=None` by default, and automatically creates a new class instance if the arguments to `__init__` have changed (i.e. implies `create_new=True` in this case). One can still recover the behaviour from #977 by explicitly setting `create_new=False`, ensuring no new instances are created.

### DIFF
--- a/NuRadioMC/EvtGen/NuRadioProposal.py
+++ b/NuRadioMC/EvtGen/NuRadioProposal.py
@@ -244,8 +244,8 @@ class ProposalFunctions(object):
             upper_energy_limit of tables that will be created by PROPOSAL, in NuRadioMC units (eV).
             There will be an error if primaries with energies above this energy will be injected.
             Note that PROPOSAL will have to regenerate tables for a new values of upper_energy_limit
-        create_new: bool (default:False)
-            Can be used to force the creation of a new ProposalFunctions object.
+        create_new: bool (default: None)
+            Set to ``True`` to force the creation of a new ProposalFunctions object.
             By default, the __init__ will only create a new object if none already exists.
             For more details, check the documentation for the
             :class:`Singleton metaclass <NuRadioReco.utilities.metaclasses.Singleton>`.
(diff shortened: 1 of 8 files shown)


fschlueter:
Hm I do not like this, I think singletons should be real singletons. This really contradicts the whole purpose. It seems ARZ is just not a real singleton... We could probably implement a caching for loading the library (that seems to be the only thing which takes time). Or we keep it like this? It is only a problem when we change between models? I would be okay if that requires an explicit `create_new=True`

fschlueter:
you can fix the test by `ARZ_object = ARZ.ARZ(create_new=True)`

sjoerd-bouma:
They're not really real singletons anyway (and for essentially all of our singletons I don't think they should be real singletons either, e.g. having two `Detector` objects simultaneously is something I use all the time).

I think this solution is a nice compromise that enables the caching benefits of singletons (not unnecessarily creating new instances) without complicating things for users or raising errors in cases where you have no reason to expect them (e.g. in the example I included above). But maybe we should discuss this on a NuRadioMC call soon.

fschlueter (dismissed):
I can live with this solution

fschlueter:
I will go ahead and merge this.

PR #991, opened 2025-09-22
Cleanup in Cython wrapper for CPP raytracer

sjoerd-bouma:
Numpy 2.3 has finally removed the deprecated `NPY_OWNDATA` flag ([link to release notes](https://numpy.org/doc/2.3/release/2.3.0-notes.html#expired-deprecations)) , which it turns out we were using in a slightly hacky fashion in the CPP wrapper. 

We could replace the `NPY_OWNDATA` flag with the `NPY_ARRAY_OWNDATA` flag. But as far as I can tell, there is actually no reason to create numpy arrays in the code at this point, so I've instead just cut out the array creation altogether. This seems to work fine locally, but let's see what the CI does : )


sjoerd-bouma:
(For context - this is why the CI fails on the latest `develop` on the newer Python versions, because those use numpy 2.3)

sjoerd-bouma:
> since we are looking at it, can we add `2> /dev/null` to the lines in the bash script removing the old wrapper library? To avoid getting annoying error messages

Your suggested solution was probably simpler and better than the one I went with, but at least I got to practise my bash : )

fschlueter (approved):

PR #994, opened 2025-10-01
Optimize readRNOGDataMattak.py. Instead of keeping datasets/files ope…

fschlueter:
…n during the entire existens of the reader, only keep it open when needed to the cost of opening a file several times


cozzyd:
yes it seemed to work. There was no obvious slow down in my testing though I didn't do any rigorous benchmarking. It's possible it may depend on filesystem too (like maybe on some network file systems, opening is expensive, though usually they'll have a file cache anyway...)

fschlueter:
I take Simons comment as formal approval!

Issue #943, opened 2025-06-06
BGR18 inelasticity distributions sampling issue at low inelasticities


sjoerd-bouma:
The new BGR18 inelasticity distributions result in discrete peaks at very low inelasticities, probably because the distribution itself is discrete and very sparse for inelasticities < 0.01 - 0.001. We should probably use some kind of interpolation to resolve this?

![Image](https://github.com/user-attachments/assets/d91fd8e8-fc15-468b-9df5-b7c48cfb848c)
('old' refers to the shower energy distribution generated with pre- #396 code)

PR #996, opened 2025-10-02
Inelasticity binning

MartinRavn:
solves #943

### DIFF
--- a/NuRadioMC/utilities/inelasticities.py
+++ b/NuRadioMC/utilities/inelasticities.py
@@ -65,13 +65,18 @@ def get_neutrino_inelasticity(n_events, model="hedis_bgr18", rnd=None,
                     inccc = np.argwhere(ncccs_ref == nccc)[0][0]
                     iE = np.argmin(np.abs(energy - nu_energies_ref))
                     get_y = intp.interp1d(np.log10(yy_ref), np.log10(dsigma_dy_ref[iF, inccc, iE]),
-                                          fill_value="extrapolate", kind="cubic")
+                                          fill_value="extrapolate", kind="linear")
 
-                    yyy = np.linspace(0, 1, 1000)
-                    yyy = 0.5 * (yyy[:-1] + yyy[1:])
+                    # The inelasticity distribution is upsampled (interpoaled and extrapolated) with logarithmic
+                    # binning below 0.1 and linear binning above 0.1:
+                    yyy_1 = np.logspace(-8, -1, 2000, endpoint=False)
+                    yyy_2 = np.linspace(1e-1, 1, 2000)
+                    yyy = np.append(yyy_1, yyy_2)
                     dsigma_dyy = 10 ** get_y(np.log10(yyy))
+                    probability_mass = dsigma_dyy * 0.5 * np.append(np.diff(yyy), yyy[-1] - yyy[-2])
+                    probability_mass /= np.sum(probability_mass)
 
-                    yy[mask] = rnd.choice(yyy, size=size, p=dsigma_dyy / np.sum(dsigma_dyy))
+                    yy[mask] = rnd.choice(yyy, size=size, p=probability_mass)
         return yy
 
     else:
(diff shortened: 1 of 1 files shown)


fschlueter:
Hey @MartinRavn, great that you took that on! I am on the go and will try to give this PR a closer look later. On first sight that looks like a much better solution but there might be an even better. From your yyy vector you could calculate the cdf and normalize it to one. Than you make an interpolation function with f(cdf, yyy). If you then draw random values between 0 and 1, you can get the inelasticity with f(x) completely bin free?!

fschlueter:
And while your are at it, can you make it such that the construction of all that is cached? Such that one have to do that only once?

fschlueter:
Sorry, my ideas above were half wrong. What I suggest is, instead of doing:
```

                    # # Sample inelasticities from the distribution and add additional uniform sub-sample smearing:
                    # indicies = rnd.choice(range(len(yyy)), size=size, p=probability_mass)
                    # yy[mask] = yyy[indicies] + np.random.uniform(0, 1, size=size) * np.append(np.diff(yyy), yyy[-1] - yyy[-2])[indicies]
```
doing:
```
                    cdf = np.cumsum(probability_mass)
                    cdf_f = intp.interp1d(cdf, yyy, fill_value="extrapolate", kind="linear")
                    randoms = rnd.uniform(0, 1, size=size)
                    yy[mask] = cdf_f(randoms)
```
What I am getting with this, is:
image
Maybe you can quickly compare the two?

MartinRavn:
Thanks! That works, but it gives a small artefact where it changes from logarithmic to linear binning. I will see if I can fix it 
image

MartinRavn:
It works well with purely logarithmic binning, which also makes the code simpler
image

MartinRavn:
It currently takes 22.20 seconds for 100000 events. If I remove parts that can be cached and define placeholders for `yyy` and `cdf` as globals, it takes 13.29 seconds.

fschlueter:
You could write a function which has the decorator `@functools.lru_cache(maxsize=256)` which takes `flavor, nccc, energy` as arguments and returns `cdf_interp` for it. see for example that I added the same decorator for `_read_differential_cross_section_BGR18`. You are right that it is probably not so relevant but it does not harm I think.

MartinRavn:
I implemented caching with `maxsize=1200`, since this is the number of combinations of flavors, interaction types, and energies. Let me know if it is too many. I also have to run `cross_sections._read_differential_cross_section_BGR18()` on import.

In addition to that, I optimized the loop a lot. We don't need to run it for every unique energy, but only the unique energy bins. Now it only takes 0.03 seconds for 100000 events with random energies.

fschlueter:
I approved but actually we still have to update the reference values and i would prefer to wait for #930 which is also already approved

NuRadioMC/utilities/inelasticities.py
```diff
@@ -64,20 +80,34 @@ def get_neutrino_inelasticity(n_events, model="hedis_bgr18", rnd=None,
...
+                    yy[mask] = cdf_interp(randoms)
 
-                    yy[mask] = rnd.choice(yyy, size=size, p=dsigma_dyy / np.sum(dsigma_dyy))
         return yy
 
     else:
         raise AttributeError(f"inelasticity model {model} is not implemented.")
 
 
+@functools.lru_cache(maxsize=int(2**(int(np.log2(100 * 6 * 2 + 1)))))
```
fschlueter, line 95: the `+1` was meant to be outside of the log2()`. The reason behind it that it is more efficient if maxsize is a power of 2 number and I wanted to find the smallest number which is larger than what we need. But maybe just use a fixed of 2048 or even only a 1024 assumeing we never need all?

fschlueter (dismissed):

NuRadioMC/utilities/inelasticities.py
MartinRavn: For 10.000 events, I get events in all (or almost all) combinations of flavor, interaction type, and energy bins (assuming they have equal probability). So I will move the +1 outside the log2 again and set it to 2048.

PR #998, opened 2025-10-07
Add set_id() method to BaseStation.

alisanozdrina:
Useful for reading event information from RNO-G stations that have not yet been assigned a stationID.

Example case: calibration pulser drop data returns 0 for station.get_id(), which causes issues later when using the Detector class. I didn’t find a quick workaround other than explicitly setting the station ID.

### DIFF
--- a/NuRadioReco/framework/base_station.py
+++ b/NuRadioReco/framework/base_station.py
@@ -81,6 +81,9 @@ class BaseStation(NuRadioReco.framework.parameter_storage.ParameterStorage):
     def get_id(self):
         return self._station_id
 
+    def set_id(self, station_id):
+        self._station_id = station_id
+
     def remove_triggers(self):
         """
         removes all triggers from the station


cozzyd:
Fixing the data is probably the right thing to do though because multiple copies exist it might not be so straightforward. Ideally it would be done on the raw data files (we'd have to write a quick and easy program to do that) and then re-rootify, but it could also be an option to rno-g-convert to override the station number (but then we'd have to make sure that any reprocessing keeps the override).

fschlueter:
Okay, lets close this PR, @alisanozdrina you can keep the interface in your local branch or you just use `station._station_id = x` (in python variables with a leading `_` are hidden but not forbidden). Can you also share the station_id and run numbers of the corrupted files?

alisanozdrina:
Ok, thanks! This specific run is the one I had issues with: https://rno-g.uchicago.edu/data/special/station12/pulser-dropdown/.
But I’d guess a similar problem might be present in other pulser-drop root files as well. There are a few such folders scattered around here: https://rno-g.uchicago.edu/data/special/.

PR #999, opened 2025-10-08
Updated changelog with inelasticity binning bugfix

MartinRavn:
Changelog for #996


sjoerd-bouma:
Hi Martin... this is the rare occasion where I think we don't need a changelog entry, because the BGR18 inelasticities are added in the same version (3.1.0) as the fix. So I would suggest closing this. Happy for @fschlueter to weigh in if he disagrees.

Issue #1001, opened 2025-10-10
UTC time mismatch between db_mongo_read.py and db_mongo_write.py


zsselcuk:
There is a mismatch between time zone handling in `db_mongo_read.py` and `db_mongo_write.py`. `db_mongo_write.add_entry_to_database()` uses a timezone-aware UTC datetime (`datetime.now(tz=UTC)`),
but `db_mongo_read.set_database_time()` compares it with a naive datetime (`datetime(2024, 11, 1)`),
causing the error: `peError: can't compare offset-naive and offset-aware datetimes`.

I will create a PR where both date times are consistent.

PR #1002, opened 2025-10-10
make both dateimetimes compatible

zsselcuk:
Issue being addressed: UTC time mismatch between db_mongo_read.py and db_mongo_write.py #1001


jakobhenrichs:
We decided to raise a warning and assume UTC as the timezone (if None is given)

Issue #1008, opened 2025-10-14
Option to choose a different primary_time to extract the signal_chain info


zsselcuk:
The function `db_mongo_read.get_channel_signal_chain` only returns the current primary measurement. We have to add the option of choosing previous ones (for example for redeployment of stations, where some old entries have to be copied). I will create a PR.

PR #1009, opened 2025-10-15
add the option to choose which primary measurement to use in function get_channel_signal_chain

zsselcuk:
Issue being addressed: #1008


fschlueter:
In general I am okay with adding that because it should not harm. On the other hand think critically about it - making the interface more complex comes at a cost. And please fix the error - it seems related to your changes.

Issue #1013, opened 2025-10-16
Detector class reverts to S parameter signal chain when using always_querry_entire_description


CamphynR:
In the RNO-G detector class, setting the always_query_entire_description to True overwrites the choice of signal_chain_measurement_name to always yield the S parameter response. Is this the expected behaviour?

I include a code example below:

```
import matplotlib.pyplot as plt
import numpy as np
from NuRadioReco.utilities import units
from NuRadioReco.detector.RNO_G.rnog_detector import Detector
from NuRadioReco.utilities import fft
from astropy.time import Time

det_time = Time("2023-08-01")
det_no_query = Detector(select_stations=13, signal_chain_measurement_name="calibrated_impulse_response_v0")
det_query_all = Detector(select_stations=13, always_query_entire_description=True, signal_chain_measurement_name ="calibrated_impulse_response_v0")
det_no_query.update(det_time)
det_query_all.update(det_time)

response_no_query = det_no_query.get_signal_chain_response(13, 1)
response_query_all = det_query_all.get_signal_chain_response(13, 1)

sampling_rate = 3.2 * units.GHz
frequencies = fft.freqs(2048, sampling_rate)

plt.plot(frequencies, np.abs(response_no_query(frequencies)), label="no query all")
plt.plot(frequencies, np.abs(response_query_all(frequencies)), label="query all")
plt.legend()
plt.show()
```

CamphynR:
The query all option already queried the default signal_chain name from the database and the code only replaces the signal name when no signal name is stored in the channel information. I added a check that overwrites this behavior when signal_chain_measurement_name is given

PR #1016, opened 2025-10-22
Fix RNO-G detector bug

CamphynR:
Solves Issue https://github.com/nu-radio/NuRadioMC/issues/1013


fschlueter (changes requested):
see comment above

CamphynR:
This should be a cleaner implementation, if query all is specified together with a signal_chain_measurement_name the name is already properly set when querying the complete station description

PR #1018, opened 2025-10-22
Update event generation to include cross sections model parameter

Skoishi:
Added a new parameter 'cross_sections_model' to the generate_eventlist_cylinder function in generator.py, defaulting to 'hedis_bgr18'. Updated calls to this function in simulate.py to pass the new parameter with a value of 'ctw'.

Issue being addressed:
generate_eventlist_cylinder function do not allow cross-section model to be set by user


fschlueter (commented):
other than my one commit looks fine

Skoishi (commented):
cancel the change on example

PR #1024, opened 2025-11-07
Improve analytic antenna responses

MartinRavn:
The current implementation of the analytic LPDA antenna response is wrong, and there are no analytic VPol and HPol models. This PR improves the analytic LPDA and implements analytic VPol and HPol approximations.

### DIFF
--- a/NuRadioReco/detector/antennapattern.py
+++ b/NuRadioReco/detector/antennapattern.py
@@ -1594,9 +1594,17 @@ class AntennaPatternAnalytic(AntennaPatternBase):
             antenna models, respectively.
         cutoff_freq: float
             Sets the low frequency cutoff for the LPDA and VPol models, and the peak
-            frequency for the HPol model.
+            frequency for the HPol model. Setting cutoff_freq=None, the default values are
+            used:
+            'analytic_LPDA': 110 MHz,
+            'analytic_VPol': 220 MHz,
+            'analytic_HPol': 500 MHz.
         max_VEL: float
             Sets the maximum value of the vector effective length for the antenna models.
+            Setting max_VEL=None, the default values are used:
+            'analytic_LPDA': 0.55 m,
+            'analytic_VPol': 0.18 m,
+            'analytic_HPol': 0.055 m.
         """
         self._notfound = False
         self._model = antenna_model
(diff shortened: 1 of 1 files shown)


MartinRavn:
Here are some plots showing the discrepancy between the **old analytic LPDA** and the antenna response of `createLPDA_100MHz_InfFirn_n1.4`:
image
image
image

MartinRavn:
Here is the **analytic VPol** compared to `RNOG_vpol_v3_5inch_center_n1.74`:
image
image
image

MartinRavn:
Here is the **analytic HPol** compared to `RNOG_hpol_v4_8inch_center_n1.74`:
image
image
image

MartinRavn:
I started following up on the things we discussed at the meeting last week:

MartinRavn:
I also compared the implemented analytic LPDA response to a SKALA antenna. It looks decent in the frequency domain, but the theta/phi dependencies are different. I think it would be best if someone implemented a different analytic model for SKALA antennas.
image

MartinRavn:
I checked that if I integrate the gain over the sphere for the LPDA model I implemented, I get less than 4 pi. This is what I get:
```
Integrated gain theta: 1.092 , Expected less than: 12.566
Integrated gain phi: 3.186 , Expected less than: 12.566
```
The other antenna models have lower max gain, so they are definitely below the limit.

NuRadioReco/detector/antennapattern.py
```diff
@@ -1614,56 +1646,110 @@ def parametric_phase(self, freq, phase_type='theoretical'):
             # maximum frequency
             f = 1000. * units.MHz
             a = np.pi / np.log(tau) * np.log(freq / f) - 60
+        if phase_type == 'VPol_third_order':
+            a = -117.91689143632604 * freq + 74.56744619835776/2 * freq ** 2 -64.34287756833945/3 * freq ** 3 + np.pi/2
+        if phase_type == 'HPol_third_order':
+            a = -11.399516173113549 * freq + 39.58971631878798/2 * freq ** 2 -38.18091213162158/3 * freq ** 3 + np.pi/2
```
sjoerd-bouma: 3rd order seems like it might be slightly excessive, and I don't know that you need this many digits for something so approximate, but that's mostly nitpicking - by eye the match seems to have improved significantly (maybe there's still a slightly mysterious time shift for the Vpol for theta < 90 deg, but okay).

Out of curiosity, how did you arrive  at the constant term of np.pi/2 (I'm assuming it wasn't by fitting?).

NuRadioReco/detector/antennapattern.py
```diff
@@ -1581,20 +1582,51 @@ class AntennaPatternAnalytic(AntennaPatternBase):
...
+            Name of antenna model. Current implemented models are 'analytic_LPDA',
+            'analytic_VPol', and 'analytic_HPol'. By using the default values of cutoff_freq
+            and max_VEL, these models approximate the createLPDA_100MHz_InfFirn_n1.4,
+            RNOG_vpol_v3_5inch_center_n1.74, and RNOG_hpol_v4_8inch_center_n1.74
+            antenna models, respectively.
+        cutoff_freq: float
+            Sets the low frequency cutoff for the LPDA and VPol models, and the peak
+            frequency for the HPol model.
+        max_VEL: float
+            Sets the maximum value of the vector effective length for the antenna models.
```
sjoerd-bouma: This is great! I think adding the 'default' values for LPDA, Hpol and Vpol documented here would also potentially be helpful (e.g. if I want to slightly modify the antenna from the default response), right now a user has to read through the source code to find these values.

sjoerd-bouma (commented):
Thanks Martin! Just one minor addition to the docstring and I think this can be merged.

PR #1026, opened 2025-11-14
Use correct core to subract from positon

fschlueter:


tjibbe-g (dismissed):

fschlueter:
Could be, I do not think you get an error at this point.

fschlueter:
can you double check?

PhilippLaub (changes requested):
This seems to ignore the core completely

PhilippLaub:
> hmm, in the "before fix" version, has the core 2 or three coordinates?

the core that I get from `event.get_first_sim_shower().get_parameter(showerParameters.core)` has three coordinates with z being the detector altitude. I use a 2D core as input to the interpolator as suggested by the documentation.

fschlueter:
Hi all, I am trying to get back to this and digest all information. Looking at the code and the doc-string of the `coreasInterpolator.py` function `get_interp_efield_value` in is clear that the `position_on_ground` variable should be "absolue":
```
        Note that `position_on_ground` is in absolute coordinates, not relative to the core position.
        Extrapolation outside of the starshape is handled by the ``cr-pulse-interpolator`` package (see
        also the description of ``kwargs`` in `initialize_efield_interpolator`).
``` 
The old code `antenna_position[:len(core)] - core  # get the trace at the relative distance from the core` passed the relative antenna position (i.e., the antenna relative to the drawn shower core) if I understand everything correct. 

@PhilippLaub what exactly are you plotting (what positions, the in-file stored electric field positions should be relative). 

On a different (but maybe related note): The SKA detector class returns the antenna (channel) positions relative but with an almost absolute altitude while the station position `get_absolute_position` has a z-coordinate close to 0. Internally we are adding the parameter `detector_altitude` to the antenna positions but not the station positions, I think we should change that.

fschlueter:
I can reproduce what @PhilippLaub showed. Looking more into it, I expected the function `get_position_showerplane` to take the newly drawn core to transform the antenna position but it uses the original simulated core (typically at 0, 0, z).

fschlueter:
My last commit fixes the changes I made. @PhilippLaub would you mind testing it? 

In the end this PR does not fixes a bug but changes a bit how the module works. I think this is now in line with the comments while previously the comments were not reflecting what the code did. I think we have to make a choice here. Either we adopt this change or we revert them but correct the doc-strings. @PhilippLaub and @karenterveer what is your opinion on that (this module is most relevant for SKA so your opinion counts!) ? Do you want to discuss that in a call?

fschlueter:
PS: I somehow screed up the commit history - sorry for that.

fschlueter:
This PR will not be merged instead PR #1052 fixed the incorrect doc-strings

PR #1011, opened 2025-10-15
Patch rnog simulation example

fschlueter:
This  PR updates the RNO_G_trigger_simulation example to use the new PA trigger developed by @ryankrebs016, it updates the standard RNO-G config.yaml to enable focusing. It also adds a new script `NuRadioMC/examples/RNO_G_trigger_simulation/simulate_snr_trigger_curve.py`.

Bonus: Fixed Documentation (HDF5 structure)

PR #835, opened 2025-02-20
Add FAERIE support

fschlueter:
This PR replaces #723. #723 had the developed merged into it which broke the air-ice ray-tracing. We can only merge developed in here if that is fixed. This PR also fixes the name from FEARIE to FAERIE.

Issue #600, opened 2023-12-05
generator.py does not take energy/flavor dependence into account when generating ccnc, inelasticities


sjoerd-bouma:
In `generator.py`, when 'randomly' deciding whether a CC or NC interaction takes place, the code uses the `get_ccnc` function [here](https://nu-radio.github.io/NuRadioMC/_modules/NuRadioMC/utilities/inelasticities.html#get_ccnc). This corresponds to a constant ratio of CC : NC events, even though the ratio actually depends on the energy and the particle flavor. Weirdly, parameterizations for the cross section that do take this into account (as well as the slightly more significant dependence on neutrino vs antineutrino) are already available in the [NuRadioMC.utilities.cross_sections](https://nu-radio.github.io/NuRadioMC/NuRadioMC/apidoc/NuRadioMC.utilities.cross_sections.html#NuRadioMC.utilities.cross_sections.get_nu_cross_section) module. While the effect is relatively small, it seems strange not to take it into account, especially if the appropriate code is already there? 

On a similar note, the neutrino inelasticity is also generated independently of the neutrino energy - I'm guessing for similar reasons.
![image](https://github.com/nu-radio/NuRadioMC/assets/78698543/1b1d9a9e-5d0a-43b9-ab4a-ad2a01aed8fd)

sjoerd-bouma:
Potentially relevant in this context (?) - #396

cg-laser:
yes, this is still a todo for me to finally finish #396 ... the missing piece is to use the new BGR18 cross sections everywhere consistently

sjoerd-bouma:
Just as a gentle reminder, driven by the discussion started by https://arxiv.org/abs/2403.07984 - I think it's important to remember that until #396 gets merged, our inelasticities are off by somewhere between 10%, a factor of 2, or an order of magnitude, depending on the energy and how we define the comparison. This is not currently clearly documented (in my opinion), but should be kept in mind when running effective volume simulations.
![image](https://github.com/nu-radio/NuRadioMC/assets/78698543/3913fa5c-86db-41b5-a136-d963823b59b9)

cg-laser:
[...]

PR #1028, opened 2025-11-21
Raytracing air to ice + general analytic raytracer refactoring

alisanozdrina:
Attempt to add air to ice propagation back into analytic ray tracing. 

To do so added additional code into get_delta_y based on #558 and took all necessary internal helper functions of ray_tracing_2d and made them standalone numba compatible functions.  

Two potential issues:
1. Focusing calculation may need further validation. 
2. Fresnel Transmittance seems to be taken care of inside apply_propogation_effects().

All unit tests now passed. 

Example ray tracing map
516531688-4c1b5f3a-a090-4f9d-a24a-3282eca36e47


NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -430,12 +684,6 @@ def __init__(self, medium, attenuation_model=None,
                 self.__logger.warning('C++ raytracer is not available. Using Python raytracer.')
                 self.__logger.warning("check NuRadioMC/NuRadioMC/SignalProp/CPPAnalyticRayTracing for manual compilation")
 
-        if isinstance(medium, medium_util.uniform_ice):
-            msg = ('Analytic raytracer does not work with a uniform ice model. '
-                    'Abort.... ! Use direct raytracing or a non-uniform ice model instead.')
-            self.__logger.error(msg)
-            raise RuntimeError(msg)
-
```
sjoerd-bouma: Did this get fixed?

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1441,7 +1698,7 @@ def find_solutions(self, x1, x2, plot=False, reflection=0, reflection_case=1):
                 C_0_stop = self.get_C_0_from_angle(np.arcsin(1/self.medium.get_index_of_refraction([0, x1[0], x1[1]])), x1[1]).x[0]
                 logC0_stop = np.log(C_0_stop - 1/self.medium.n_ice)
                 delta_ys = [self.obj_delta_y(logC0, x1, x2, reflection, reflection_case) for logC0 in [logC0_start, logC0_stop]]
-                self.__logger.debug("Looking for ice-air solutions between log(C0) ({}, {}) with delta_y ({}, {})".format(logC0_start, logC0_stop, *delta_ys))
+                #print("Looking for ice-air solutions between log(C0) ({}, {}) with delta_y ({}, {})".format(logC0_start, logC0_stop, *delta_ys))
```
sjoerd-bouma: Can you revert this to a logger statement?

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -197,10 +197,242 @@ def get_y_turn( C_0, x1, n_ice, b, delta_n, z_0):
...
+    C_1 = x1[0] - get_y_with_z_mirror(x1[1], C_0, n_ice, b, delta_n, z_0)
+    y_turn = get_y(gamma_turn, C_0, C_1, n_ice, b, z_0)
     return y_turn
 
+def get_C_1(x1, C_0, n_ice, b, delta_n, z_0):
+    """
+    calculates constant C_1 for a given C_0 and start point x1
+    """
+    return x1[0] - get_y_with_z_mirror(x1[1], C_0, n_ice, b, delta_n, z_0)
+
```
sjoerd-bouma: This is not your fault so also not your responsibility, but the analytic raytracing module is full of 'helper' methods/functions which have missing/inaccurate docstrings and serve essentially no use. Maybe this would be a good moment to make them all private, i.e. `_get_C_1` etc?

sjoerd-bouma (commented):
Thanks very much @alisanozdrina and sorry for the late review - please don't consider this a request to work on this over Christmas!

I made a couple of minor fixes/changes that mostly affect the focusing, + a fix for air-to-ice trajectories where the emitter vertex has a non-zero horizontal coordinate - this was known (and I made a workaround for it) when it was working previously, but this time I managed to find the source. 

I think aside from some aesthetic comments which you may or may not address, there's two things that would be good to fix before merging this:
- Some methods now exist both as standalone functions (`get_C_1`) and as methods (`ray_tracing_2D.get_C_1`). To reduce code duplication and ensure always the faster method gets used, I would suggest either removing the method and replacing it everywhere, or (maybe less work) making `ray_tracing_2D.get_C_1` a simple wrapper for the standalone method, i.e. 
  ```
  def get_C_1(...):
      return get_C_1(..., self.x, self.y, ...)
  ```
- We should have an example as part of the CI, so this feature doesn't break again without us noticing! I think you can just add the example Christian made at `NuRadioMC/SignalProp/examples/E02ToAir.py` to the examples github action. If you're not sure how to do this I can also do it instead.

fschlueter:
@alisanozdrina, do you still consider working on this? #1077 will reimplement air-ice raytracing. I believe there is still merit in finalising this PR as the single-exponential raytracing is faster as the multiexponential raytracer (even if only a single layer ice model is configured) - is that correct @hanneswarnhofer. However, given that we are all short on time I am happy to close this PR if you think you wont get back to it.

alisanozdrina:
@fschlueter would it make sense to merge up to [66bf373](https://github.com/nu-radio/NuRadioMC/pull/1028/commits/66bf373c7eb5c2136a7fd61d1eeee47c96dc9614)? air-to-ice raytracing is implemented with Sjoerd's fixes to the focusing calculation, but before refactoring. since it's becoming a legacy code and exploration tool, I assume refactoring is not a major issue.

fschlueter:
@sjoerd-bouma what do you think?

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1359,8 +1614,10 @@ def obj_delta_y(self, logC_0, x1, x2, reflection=0, reflection_case=2):
         function to find solution for C0, returns distance in y between function and x2 position
         result is signed! (important to use a root finder)
         """
+        #print("logC_0", logC_0)
         C_0 = get_C0_from_log(logC_0,self.medium.n_ice)
-        return get_delta_y(C_0, np.array(x1), np.array(x2),self.medium.n_ice, self.__b, self.medium.delta_n, self.medium.z_0, self.reflection, (-1.0,-1.0), reflection, reflection_case)
+        x1_arr = np.array(x1, dtype=np.float64).copy()
```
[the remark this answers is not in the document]
fschlueter: @alisanozdrina beyond Sjoerds refractoring comments is that substantial for the physics?

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -149,11 +147,13 @@ def get_turning_point(c, b, z_0, delta_n):
...
     typle (gamma, z coordinate of turning point)
     """
-    gamma2 = np.array([b * 0.5 - (0.25 * b ** 2 - c) ** 0.5])  # first solution discarded
-    z2 = np.log(gamma2 / delta_n) * z_0
-    if(z2 > 0):
-        z2 = np.array([0], dtype=np.float64)  # a reflection is just a turning point at z = 0, i.e. cases 2) and 3) are the same
-        gamma2 = get_gamma(z2, delta_n, z_0)
+
+    gamma2 = float(b * 0.5 - np.sqrt(0.25 * b ** 2 - c))
+    z2 = float(np.log(gamma2 / delta_n) * z_0)
```
fschlueter: I think neither of them require to be floats, but I guess it does not harm?

fschlueter:
Running the E02ToAir.py example I get the following error (after my commit changing the error msg):
```
ERROR - 2026-07-02 17:10:38,255 - NuRadioMC.ray_tracing_2D - Error in compiling methods using jit: "A jit decorator was called on an already jitted function CPUDispatcher(<function get_z_mirrored at 0x128b7e160>).  If trying to access the original python function, use the CPUDispatcher(<function get_z_mirrored at 0x128b7e160>).py_func attribute." - proceeding without numba
```

fschlueter:
I believe the issue is that when calculating focusing we re-initialize the raytracer:
```
                self._r1 = ray_tracing(self._medium, self._attenuation_model, logging.WARNING,
                                self._n_frequencies_integration, self._n_reflections, use_cpp=self.use_cpp)
```
that throughs things off ...

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -464,9 +712,16 @@ def __init__(self, medium, attenuation_model=None,
...
                 try:
+                    get_z_mirrored = jit(get_z_mirrored, nopython=True, cache=True)
+                    get_C_1 = jit(get_C_1, nopython=True, cache=True)
+                    get_fresnel_angle = jit(get_fresnel_angle, nopython=True, cache=True)
+                    get_reflection_angle = jit(get_reflection_angle, nopython=True, cache=True)
+                    get_path_segments = jit(get_path_segments, nopython=True, cache=True)
+                    get_angle = jit(get_angle, nopython=True, cache=True)
                     get_reflection_point = jit(get_reflection_point, nopython=True, cache=True)
                     obj_delta_y_square = jit(obj_delta_y_square, nopython=True, cache=True)
                     get_delta_y = jit(get_delta_y, nopython=True, cache=True)
```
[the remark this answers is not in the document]
fschlueter: Ah sjoerd already spotted the issue. Its also means that for the focusing calculation the "second" raytracing is performed without numba?!

fschlueter:
@sjoerd-bouma and @alisanozdrina I took over this branch - I hope you are okay with that. I need that feature for a study of mine. Tried some AI-powered refactoring also solving a couple of issues along the way. That however means that this PR blows up quite a bit ...

fschlueter:
Okay, I compared the result of a simulation with 100000 events (190 triggers) for an ice-air raytracing scenario where the antenna is at 3km altitude between this branch and the version of the analyticraytracer from #835 (this PR uses a slightly modified version of the original raytracer from before #660) + develop branch. 

Beyond the fact that all `ray_tracing_solution_type` are wrong and thus also all `ray_tracing_C1` (the calculation of ray_tracing_C1` depends on the type). All other things are identical beyond numerical fluctuations `h5diff -d 0.0000001` passes.... 

This branch sets to the ray_tracing_solution_type to `1` (=direct) which I believe is correct (right @sjoerd-bouma ?). 

Next I will rerun this with more events (just to be sure) and second look at the calculated focusing factors as that was labelled "not much validated" also in the original code before #600. Does anyone know something about that?

fschlueter:
This PR now also contains #1117 - but I am confident that we can merge that one first :)

fschlueter:
@sjoerd-bouma shall I run validation on the regular ray-tracing solutions or do we think that the CI covers all that?

sjoerd-bouma:
Re jitting - I wrote this reply but apparently forgot to press send? 
> Running the E02ToAir.py example I get the following error (after my commit changing the error msg):
> 
> ```
> ERROR - 2026-07-02 17:10:38,255 - NuRadioMC.ray_tracing_2D - Error in compiling methods using jit: "A jit decorator was called on an already jitted function CPUDispatcher(<function get_z_mirrored at 0x128b7e160>).  If trying to access the original python function, use the CPUDispatcher(<function get_z_mirrored at 0x128b7e160>).py_func attribute." - proceeding without numba
> ```

This is an issue with the previous/current implementation of the jitting - in fact what happens is the python function is overwritten by the jitted version, which then (if the class is initialized again) can of course not be jitted again. The error message isn't correct though, what happens is that the new (jitted) function is always used. There is an ongoing discussion with @philippwindischhofer and @hanneswarnhofer about how to fix this going forward. I guess your solution is also fine.

sjoerd-bouma:
Without having had a look at this PR in detail yet, I am a bit worried about merge conflicts with #1087 and #1077, which my understanding was were/are also close to being ready to be merged. Wouldn't it simplify things to tackle these PRs serially rather than in parallel?

philippwindischhofer:
For added context, the "ongoing discussion" that @sjoerd-bouma points out proposes to embrace `numba` as a NuRadio core dependency and use `@njit` decorators on all to-be-jitted methods. If we want, we can have a fallback mechanism like

```
% cat maybenumba.py
try:
    from numba import njit
except:
    def njit(*args, **kwargs):
        def dec(func):
            return func
        return dec
```

in case `numba` fails to import  for some reason. If `numba` is available, but jitting is for some reason not desired, one can always run with `NUMBA_DISABLE_JIT=1` (https://numba.pydata.org/numba-doc/dev/reference/envvars.html#jit-flags).

fschlueter:
> Without having had a look at this PR in detail yet, I am a bit worried about merge conflicts with #1087 and #1077, which my understanding was were/are also close to being ready to be merged. Wouldn't it simplify things to tackle these PRs serially rather than in parallel?

So I agree that this will become messy. However, I think the best way forward would be to get this PR merged. Redo PR #1087 and than deal with smaller conflicts in #1077.

fschlueter:
> For added context, the "ongoing discussion" that @sjoerd-bouma points out proposes to embrace `numba` as a NuRadio core dependency and use `@njit` decorators on all to-be-jitted methods. If we want, we can have a fallback mechanism like
> 
> ```
> % cat maybenumba.py
> try:
>     from numba import njit
> except:
>     def njit(*args, **kwargs):
>         def dec(func):
>             return func
>         return dec
> ```
> 
> in case `numba` fails to import for some reason. If `numba` is available, but jitting is for some reason not desired, one can always run with `NUMBA_DISABLE_JIT=1` (https://numba.pydata.org/numba-doc/dev/reference/envvars.html#jit-flags).

Sounds very reasonable to me. I do not understand the code snippet though. The question is when does that need to happen? Can we do it once all of those 3 PR are merged?

philippwindischhofer:
> I do not understand the code snippet though. 

It just embellishes the `numba` import and replaces `@njit` with the identity decorator in case `numba` fails to import.

> The question is when does that need to happen? Can we do it once all of those 3 PR are merged?

In principle yes, although the jitting in #1077 has been broken by the refactoring and so `numba` does not currently work. So this should be fixed before #1077 can be merged (but we should continue that part of the discussion there).

fschlueter:
> > I do not understand the code snippet though.
> 
> It just embellishes the `numba` import and replaces `@njit` with the identity decorator in case `numba` fails to import.
> 
> > The question is when does that need to happen? Can we do it once all of those 3 PR are merged?
> 
> In principle yes, although the jitting in #1077 has been broken by the refactoring and so `numba` does not currently work. So this should be fixed before #1077 can be merged (but we should continue that part of the discussion there).

Ah now i get it. That’s a good suggestion (if you can still disable numba which it seems you can with the env variable). I would probably introduce our own decorator `njitfb` or something similar to make it clear to the developer that this may not actually always compile the function.

fschlueter:
I compared the ice-air raytracing with an adapted version of the pre #660 code. Same result. Do you want me to run more in depth comparison of in-ice raytracing with develop? Do you want me to run more tests on focusing factor calculations?

fschlueter:
I now used the new RNO-G example from #1011 that nothing changes for in-ice raytracing between this branch and develop (+ #1011). For 10000 events at 1e19 (32 triggered events) I get the same results (modulo some numerical differences at the level of `< 10^-14`. (checked with `h5diff`).

Do you want me to run it with more statistics.

NuRadioMC/examples/RNO_G_trigger_simulation/noise.py
```diff
@@ -60,8 +65,17 @@
...
 
+beamformedPowerIntegrationTrigger = BeamformedPowerIntegrationTrigger()
+beamformedPowerIntegrationTrigger.begin()
+
+
+
+# digitalBeamformedEnvelopeTrigger = DigitalBeamformedEnvelopeTrigger()
+# digitalBeamformedEnvelopeTrigger.begin()
+
+
```
sjoerd-bouma, line 76: delete duplicate commented out lines

NuRadioMC/examples/RNO_G_trigger_simulation/noise.py
```diff
@@ -241,6 +271,22 @@ def trigger_simulation(evt, station, det, vrms=None):
...
+            **power_trigger_kwargs
+        )
+
+        # pa_envelope_threshold = np.round(8 * np.sqrt(np.sum(np.array(list(threshold_high.values())))))
+        # digitalBeamformedEnvelopeTrigger.run(
+        #     evt, station, det,
+        #     trigger_name=f"pa_envelope_{thresh_key:.4f}_sigma",
+        #     threshold=pa_envelope_threshold,
+        #     **envelope_trigger_kwargs
+        # )
```
sjoerd-bouma, line 288: maybe add a comment stating why someoen might want to uncomment this, or otherwise remove it?

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -716,69 +1009,20 @@ def get_travel_time_analytic(self, x1, x2, C_0, reflection=0, reflection_case=1)
...
+            n_ice, z_0, alpha, gamma, l1, l2 = self._get_analytic_ray_params(x1, C_0, reflection, reflection_case)
 
             def get_ct(z):
-                ct = z_0 * (
+                return z_0 * (
                     np.sqrt(gamma(z)) - n_ice**2/np.sqrt(alpha) * np.log(l1(z))
                     + n_ice * np.log(l2(z))
                 ) + n_ice**2 * z / np.sqrt(alpha)
-                return ct
-
```
sjoerd-bouma, line 761: Surely stylistically this is not an improvement (multi-line return statement)? Not that it matters of course...

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1122,45 +1340,15 @@ def get_path_segments(self, x1, x2, C_0, reflection=0, reflection_case=1):
...
-            self.__logger.debug("setting x1 from %s to %s", x1, x2)
-            x1 = x2
-
-        return segments
+        rows = get_path_segments(
+            x1, x2, C_0,
+            self.medium.n_ice, self.__b, self.medium.delta_n, self.medium.z_0,
+            self.reflection, reflection, reflection_case)
+        return [[row[0:2], row[2:4], row[4:6], row[6:8], row[8], row[9]] for row in rows]
```
sjoerd-bouma, line 1347: :/

sjoerd-bouma: Is there any reason why the global (module-level) `get_path_segments` is public? Can you add a `_` so that at least it has a different name than the method?

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -2647,14 +2813,82 @@ def get_reflection_angle(self, iS):
...
+                    t_theta = geometryUtilities.get_fresnel_t_p(zenith_reflection, n_1=1., n_2=n_ice_surface)
+                    t_phi = geometryUtilities.get_fresnel_t_s(zenith_reflection, n_1=1., n_2=n_ice_surface)
+                    self.__logger.info(f"propagating from air to ice: transmission coefficient is {t_theta:.2f}, {t_phi:.2f}")
+
+                fresnel_coefficients.append(
+                    {'zenith': zenith_reflection, 'case': 'transmission', 'theta': t_theta, 'phi': t_phi})
+            else:
+                # in-ice propagation, reflection off the surface
+                r_theta = geometryUtilities.get_fresnel_r_p(zenith_reflection, n_2=1., n_1=n_ice_surface)
+                r_phi = geometryUtilities.get_fresnel_r_s(zenith_reflection, n_2=1., n_1=n_ice_surface)
```
sjoerd-bouma: Is it worth replacing the `1.` that stands for the refractive index of air by some module constant `_N_AIR = 1.0` (e.g.) in case we ever want to replace it by 1.0003 (or maybe we want to do this now, anyway?). Feels like it being hardcoded in all these different places as a float is slightly more error-prone as I think we use 1.0003 at least in one or two other places already.

NuRadioMC/SignalProp/analyticraytracing.py
fschlueter: Done - I guess we could make several more methods private but I did not want to add yet another change to this already big PR. We can decide to do that in a follow up PR?

sjoerd-bouma: We don't have to make pre-existing functions/methods private but I thought it might be sensible to at least make sure all the newly added ones that have no real use outside of the class are already private, rather than introducing public functions in this PR and then making them private later.

NuRadioMC/SignalProp/examples/E02ToAir.py
```diff
@@ -12,6 +13,14 @@
...
 # ray.cpp_available=False
 
+"""
+this example calculates ice-to-air and air-to-ice raytracing solutions for a
+few fixed geometries. Because the geometries (and hence the solutions) are
+fixed, the resulting C0, focusing factor and fresnel transmission
+coefficients (t_s, t_p) are validated against hardcoded reference values so
+that regressions in the ice/air raytracing code are caught in CI.
+"""
+
```
sjoerd-bouma, line 21: Module / script docstrings technically should go before import statements, but seeing as we don't document scripts anyway this doesn't really matter.

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -2647,14 +2815,82 @@ def get_reflection_angle(self, iS):
...
+                    # ice to air case
+                    t_theta = geometryUtilities.get_fresnel_t_p(zenith_reflection, n_2=N_AIR, n_1=n_ice_surface)
+                    t_phi = geometryUtilities.get_fresnel_t_s(zenith_reflection, n_2=N_AIR, n_1=n_ice_surface)
+                    self.__logger.info(f"propagating from ice to air: transmission coefficient is {t_theta:.2f}, {t_phi:.2f}")
+                else:
+                    # air to ice
+                    t_theta = geometryUtilities.get_fresnel_t_p(zenith_reflection, n_1=N_AIR, n_2=n_ice_surface)
+                    t_phi = geometryUtilities.get_fresnel_t_s(zenith_reflection, n_1=N_AIR, n_2=n_ice_surface)
+                    self.__logger.info(f"propagating from air to ice: transmission coefficient is {t_theta:.2f}, {t_phi:.2f}")
+
```
sjoerd-bouma: This is not the right angle, by definition the angle specified is the **incoming** angle which in this case should be the angle in air, you are using the in-ice angle here instead.

edit: I say you, probably it was me who implemented this? I'll fix it myself...

fschlueter:
Running some ice-air/air-ice raytracing and looking at the the focusing factor and fresnel coefficients. Looks reasonsable to me... ?!

ice2air_output
ice2air_output1

fschlueter:
After Sjoerds fix the air-to-ice focusing factor  and fresnel coefficients changed: 

ice2air_output_a2i_new

fschlueter:
I actually do not understand why they changed? Maybe its because of my change of `n_at_surface`? @sjoerd-bouma any comment on  this?

fschlueter:
@sjoerd-bouma and @philippwindischhofer what do you say?

sjoerd-bouma:
> I actually do not understand why they changed? Maybe its because of my change of `n_at_surface`? @sjoerd-bouma any comment on this?

I think my change should only affect the air-to-ice fresnel coefficients, not the focusing factor or anything ice-to-air or ice-to-ice. Looking at how the reference values in E02ToAir.py have changed this is (approximately) the case, so I'm not sure what's happened to your plots?

fschlueter:
I can imagine they are due to my change of `n_at_surface`. Originally I thought this strong raise in focusing factor for air-to-ice propagation close to the critical angle is sensible. Now I am not sure about it (could be confirmation bias since the new results now changed that feature ...). But thinking about it, to co located sources above the surface, will be seen in the ice under a smaller solid angle due to the refraction, correct? Currently the focusing factor is close to but always below 1 which would disagree with that statement? @sjoerd-bouma do I have a wrong picture in my head?

fschlueter:
@sjoerd-bouma (and maybe @philippwindischhofer) do you have suggestions how to move forward with this PR (this kind of holds up a lot of things - the other two raytracer PR and also the restructuring). 

I am less concerned about the changes in the focusing factor for the air-ice raytracing and believe that it is coming from the `n_at_surface` change (I can explicitly check if you want). Otherwise I think this is much better validate than it ever was.

sjoerd-bouma:
> I can imagine they are due to my change of `n_at_surface`. Originally I thought this strong raise in focusing factor for air-to-ice propagation close to the critical angle is sensible. Now I am not sure about it (could be confirmation bias since the new results now changed that feature ...). But thinking about it, to co located sources above the surface, will be seen in the ice under a smaller solid angle due to the refraction, correct? Currently the focusing factor is close to but always below 1 which would disagree with that statement? @sjoerd-bouma do I have a wrong picture in my head?

The change in `n_at_surface` should only have a very small effect, and for all the geometries in E02ToAir.py this is the case, as you can tell from the very small changes in the reference values. I suspect something went wrong in the plotting.

The focusing factor discussion is unfortunately complicated. The focusing factor is ~0.1 and actually symmetric between air-to-ice or ice-to-air, but the correction for the double-counting in the fresnel factor spoils that and makes things harder to interpret. I believe the results from the test script make sense at the moment, but your last plot looks wrong to me (for ice-to-air, the fresnel coefficients should be > 1, I believe).

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1590,13 +1764,11 @@ def get_angle_from_logC_0(self, logC_0, z_pos, angoff=0, in_air=False):
             angle corresponding to C_0, minus offset angoff
         '''
 
-        C_0 = get_C0_from_log(logC_0, self.medium.n_ice)
-
-        dydz = get_y_diff(z_pos, C_0, self.medium.n_ice, self.__b, self.medium.z_0, self.medium.delta_n, in_air=in_air)
-#        dydz = self.get_dydz_analytic(C_0, z_pos)
+        C_0 = _get_C0_from_log(logC_0, self.medium.n_ice)
+        C_0 = np.atleast_1d(C_0)[0]
```
sjoerd-bouma, line 1768: This line is weird but I guess necessary somehow?

sjoerd-bouma (approved):
CI is happy, I think I'm happy, too. Thanks @alisanozdrina and @fschlueter for reviving this

NuRadioMC/SignalProp/analyticraytracing.py
fschlueter: hmm

fschlueter:
> I believe the results from the test script make sense at the moment, but your last plot looks wrong to me (for ice-to-air, the fresnel coefficients should be > 1, I believe).

they are above 1?

PR #1030, opened 2025-12-11
set all antenna z positions to observational level

karenterveer:
The z positions of the antennas were previously set to 0 by default. To ensure they match the actual observational level of LOFAR, they were now all set to 7.6m.


sjoerd-bouma:
Closed in favour of #1032

PR #1031, opened 2025-12-12
update BGR18 cross sections to extend to 1e21 eV

### LINKED ISSUE #934: Tables for inelasticity distributions end at 1e19 eV
As discussed in PR #396 the tables do no extend far enough in energy. There is contact with the developers by @shallmann , @fschlueter and @sjoerd-bouma whether tables can be provided or whether we have to generate them ourselves. 

This issue should serve as reminder that the highest energies should not be trusted beyond 1e19.

sjoerd-bouma:
Updates and extends the 'new' BGR cross sections (and inelasticities) to 1e21 eV.

Fixes #934. The new cross sections live at https://rnog-data.zeuthen.desy.de/cross_sections/BGR18_dsigma_dy_H2O.npz; they are only 2 MB so could theoretically be included in the repository as previously, but having them downloaded on demand is probably also fine. As is to be expected, the Veff test fails, with a larger effective volume obtained with the new cross sections. This is probably okay, although the difference (17-18%) at 1e18 eV is larger than I would have expected from the plots in the next comment. If anyone can think of some good additional cross-checks to run I'm happy to run those.

Below is a quite extensive description of the new cross sections, for documentation purposes.

The cross sections were obtained with [GENIE](www.genie-mc.org) version 3.06.02, using the `gcalchedisdiffxsec` command:
```
gcalchedisdiffxsec -s -p $1 -t $2 -x bx.dat -y by.dat -e energies.dat --tune GHE19_00a_00_000 --event-generator-list CCHEDIS -o "./dx_nu_GHE19_00a_00_000_$1_$2.root"
```
where `$1` is the neutrino (-16,-14,-12,12,14,16), and `$2` is the target nucleus (1000010010 for hydrogen and 1000080160 for oxygen). GENIE returns the double differential cross section as multiple values for different struck quarks (?) at each value of `x` and `y`, so the total differential cross section was obtained by summing all contributions at each value of `x` and `y`. The single-differential cross section $d\sigma / dy$ was obtained by integrating over $x$; the integration was performed using the `integrate_pwpl` function also added in this PR. Both `x` and `y` span from $10^{-12}$ to 0.89*, with a logarithmic spacing with 20 bins per decade. Both spacing and range were motivated by some initial studies that showed that at both the lowest and highest energies going lower in `x`/`y` had no effect**, and finer binning changed the results by less than 0.1%.

After integration, the H20 cross section was obtained as $\sigma_\mathrm{H2O} = 2\sigma_\mathrm{H} + \sigma_\mathrm{O}$. This cross section agrees quite well with the previous BGR cross section up to $\sim 10^17$ eV; at the highest energies there is a significant difference. This difference is possibly due to the previous integration not going low enough in `x`, which results in a reduction of $d\sigma/dy$ at very high energies and large `y`. Comparison plots and a more extensive discussion of this disagreement in the next comment.

\* the values at `x=1` / `y=1` were either 0 or showed a sharp drop compared to the penultimate value, so these were excluded; the effect of extrapolating from 0.89 to 1 was estimated to be less than 0.1% in a couple of test cases
\** while $10^{-12}$ is probably slightly overkill, at the highest energies also the behaviour at $x << 10^{-6}$ was shown to be relevant.

### DIFF
--- a/NuRadioMC/test/Veff/1e18eV/T03check_output.py
+++ b/NuRadioMC/test/Veff/1e18eV/T03check_output.py
@@ -19,8 +19,8 @@ trigger_per_file = {
 }
 
 Veff_mean_per_file = {
-    "output.hdf5": 4.5246999,
-    "output_noise.hdf5": 7.1525037
+    "output.hdf5": 5.35558,
+    "output_noise.hdf5": 8.35748
 }
 Veff_sigma = 0.0001
 


sjoerd-bouma:
> Regarding the 17% increase in effective volume, from the top plot you have posted, you can see that the new 1e18 eV inelasticity distribution is pushed towards higher y than the old distribution (I assume "old" here refers to BGR). This means that more energy is deposited in the showers, and it should lead to higher effective volume. So I don't think the difference comes from any of our code, but rather what is stored in the files.

Yes, I'm not so much surprised the effective volume increases, but I wasn't expecting such a large effect. I did double-check this by running some larger sample sizes, where the difference is only 8%, so I'm putting it down to random fluctuations for now.

Below is the result of the Veff comparison at 4 different energies (data points slightly shifted for legibility). Within the (fairly large) error bars this seems reasonable, I think; the sample sizes were already fairly large (2e5 / 1e6 events at 10 PeV), but I guess we can go higher if we want to really rule out a random origin for the increase at the lower energies.
veff-comparison-all

sjoerd-bouma:
@MartinRavn I think from my side I have done all the checks that I could think of, so I've updated the reference Veff values. Let me know if you think there are any cross checks / changes that should be made, otherwise I think this can get merged?

PR #1032, opened 2025-12-17
Add detector site elevation

sjoerd-bouma:
Adds a new method `get_site_elevation` to the `Detector` class, which returns the elevation (altitude) of a given detector site. Elevations were obtained from the NOAA ETOPO1 dataset (which includes the Greenland and Antarctic ice sheets), with a claimed precision of about ~10 m, using the public API from www.opentopodata.org. 

The LOFAR elevation is manually set to 7.6 m instead of the NOAA value; if anyone else wants to supply more precise elevations for other sites, feel free.


fschlueter:
@sjoerd-bouma I was always bothered by the fact that we store the site coordinates (long, lat) in degree and not in the NuRadio base units of rad. I am not sure if we want to change that (and suffer the consequences...). Or is there a good reason we use deg (other than readability). However, shall we add that information to the online documentation?

sjoerd-bouma:
> @sjoerd-bouma I was always bothered by the fact that we store the site coordinates (long, lat) in degree and not in the NuRadio base units of rad. I am not sure if we want to change that (and suffer the consequences...). Or is there a good reason we use deg (other than readability). However, shall we add that information to the online documentation?

Hmm, I see. I _think_ passing this in degrees and not in rad probably causes less confusion than it would if we were more consistent with units in this case, so I'm okay with keeping it the way things are. Another (probably overengineered solution) would be to return an astropy [EarthLocation](https://docs.astropy.org/en/stable/api/astropy.coordinates.EarthLocation.html#astropy.coordinates.EarthLocation).

In terms of documentation, the docstring already specifies the longitude and latitude are returned in degrees; where else would you want to document this?

fschlueter:
I think we should overhaul the unit section here: https://nu-radio.github.io/NuRadioMC/Introduction/pages/conventions.html and list this as an exception. (Also see the discussion in SKA as an motivation for a better documentation).

sjoerd-bouma:
> I think we should overhaul the unit section here: https://nu-radio.github.io/NuRadioMC/Introduction/pages/conventions.html and list this as an exception. (Also see the discussion in SKA as an motivation for a better documentation).

It felt very weird to me to add this to the conventions, so I just added a big orange warning to `get_site_coordinates` which I think is the more natural place for it. Hope you can live with that. I did slightly rewrite the conventions bit about units, but I think the philosophy & correct usage of NuRadio units is described already in some detail in the `units` module which to me seems the most sensible place for it...

(I also snuck in an unrelated change to some docstrings and a deprecated dash method, but I'm hoping no one will notice!)

fschlueter:
@galaxyvita had complained to me that the information of units are at two places in the documentation. Maybe he can clarify what his original position was. I think we can still slightly improve the unit description but this is better done in a dedicated PR

fschlueter (approved):

galaxyvita:
Yes when reading the documentation you have 3 places that talk about the units:
https://nu-radio.github.io/NuRadioMC/NuRadioReco/pages/utilities.html#unit-system
https://nu-radio.github.io/NuRadioMC/Introduction/pages/conventions.html
https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.utilities.units.html#module-NuRadioReco.utilities.units

Which i found confusing since the 3th option is clearly the best one and explains everything very well but it is also the hardest one to find.

sjoerd-bouma:
> Yes when reading the documentation you have 3 places that talk about the units: https://nu-radio.github.io/NuRadioMC/NuRadioReco/pages/utilities.html#unit-system https://nu-radio.github.io/NuRadioMC/Introduction/pages/conventions.html https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.utilities.units.html#module-NuRadioReco.utilities.units
> 
> Which i found confusing since the 3th option is clearly the best one and explains everything very well but it is also the hardest one to find.

Good point, thanks - I think the first two should probably just link to the third one? (They already do, but should maybe be more explicit about it that this is where the unit system is explained?) I'll make a mental note and a separate PR.

PR #1033, opened 2026-01-06
Compatibility fixes

sjoerd-bouma:
A couple of things had broken on older Python / numpy versions (most of them my fault). This PR tries to fix these:
- `np.trapz` is deprecated in numpy 2.0, but replacing it with `np.trapezoid` (as I did in #1031) does not work in older numpy versions. I've replaced it with the (identical?) `trapezoid` method from `scipy.integrate` here
- The reference file for the SignalGen test was a pickle file with numpy 2.0 arrays. I've replaced it with a `.npy` file which is safer and compatible across numpy versions.
- The test nur file for the VoltageToEfieldConverter test contained some records in its `registered_modules_station` of dodgy things that happened to it. These included a pickled version of a scipy module that has moved in a semi-recent scipy release, making the file unreadable on Python 3.7. I have cleared the registered modules list (file updated on the DESY server, independent from this PR).

Finally, I have changed the integration used in the BGR18 cross section from `trapezoid` to the piecewise-powerlaw integration introduced in #1031. This was actually my intention already in that PR, but I seem to have missed this. The new differential cross sections only extend up to y=0.88 (instead of y=1; see #1031 for more details), which causes the trapezoid integration to slightly underestimate the integrated cross sections (by 2-4%). It looks like this does mean the reference file may also need to be adjusted (slightly) accordingly.

### DIFF
--- a/NuRadioReco/test/voltageToEfieldConverter/test-unfolding.py
+++ b/NuRadioReco/test/voltageToEfieldConverter/test-unfolding.py
@@ -64,7 +64,6 @@ if __name__ == "__main__":
 
 
     fig, axs = plt.subplots(3, 4, figsize=(8, 4.5), sharex='col', sharey='col', layout='constrained')
-    fig.get_layout_engine().set(h_pad=0, hspace=0)
     bandpass_kwargs = dict(passband=[80*units.MHz, 500*units.MHz], filter_type='butterabs')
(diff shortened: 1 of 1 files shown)


EnriqueHuesca (approved):
LGTM, as I said before.

PR #1037, opened 2026-01-21
Modify medium.py to include 3 part exponential ice model

avijai2000:
Added 3 part exponential ice model to medium.py


NuRadioMC/utilities/medium.py
```diff
@@ -153,6 +155,81 @@ def __init__(self):
...
+        z_shift_bubbly: float
+
+        refractive index profile in snow
+        n(z) = n_bubbly - delta_n_bubbly * exp(z_shift_bubbly * z)
+
+        z_firn: float, meters
+            transition between snow and firn
+
+        z_bubbly: float, meters
+            transition between firn and bubbly ice
```
sjoerd-bouma: I'm slightly confused; these are not actually parameters that the user can input to the `__init__`, right? It looks like these are actually hardcoded in line 208 and onwards and passed on to radiopropa. Maybe change this to a `Notes` section instead and provide the actual values for reference (+ for extra credit where the values come from, i.e. a link to the paper).

NuRadioMC/utilities/medium.py
```diff
@@ -341,7 +418,7 @@ class uniform_ice(medium_base.IceModelSimple):
     """
     uniform ice with refractive index of typical deep ice (1.78)
     """
-    def __init__(self, z_bottom=None):
+    def __init__(self, z_bottom=-10000*units.meter):
```
sjoerd-bouma: Can you comment on why this change was made?

sjoerd-bouma (commented):
Thanks @avijai2000, this is very helpful! Does this model work with radiopropa 'out of the box'?

NuRadioMC/utilities/medium.py
avijai2000: I just made the change, let me know if that works. I couldn't find the paper that the values were from (I pulled them from Philipp's github repo). I'll follow up with him on that

NuRadioMC/utilities/medium.py
avijai2000: I reverted that sorry, didn't mean to make that change

NuRadioMC/utilities/medium.py
philippwindischhofer: Ideally we would already use the 3-part exponential with continuous `dn/dz`. All the constants are defined here: https://github.com/philippwindischhofer/Reconal/blob/7204049c755a0678178821073fa73a476c49c491/defs.py#L72-L82

philippwindischhofer: (It looks like you're not actually defining the constants anywhere though, Aishwarya? They're only shown in the docstring as far as I could see.)

philippwindischhofer (changes requested):

philippwindischhofer:
> Thanks @avijai2000, this is very helpful! Does this model work with radiopropa 'out of the box'?

Aishwarya will know; my understanding is that it depends on this accompanying PR to be merged: https://github.com/nu-radio/RadioPropa/pull/23

NuRadioMC/utilities/medium.py
avijai2000: I thought that was what we wanted? I was previously hardcoding them in line 208, is there a specific place we want the constants to be?

avijai2000:
Yeah it will work with the changes made in the RadioPropa pull request

NuRadioMC/utilities/medium.py
philippwindischhofer: I think hardcoding is fine—but right now it looks like you put them in a comment and so the constants don't actually exist within the code? Perhaps I'm also just not properly awake yet ...

avijai2000: Should I define them in the init then like I did earlier?

philippwindischhofer: Yeah, I think that might be best! But let's first wait until we hear back from Sjoerd to make sure he is also on board with that.

avijai2000: Okay sounds good!

sjoerd-bouma: Yes, hardcoding is fine - my original comment was that the term 'Parameters' (in a docstring) is reserved for actual parameters you pass to the class/function, which you do not have here. So you can put the hardcoded parameters back, but just remove the 'Parameters' section in the docstring.

avijai2000: Oh I see okay I'll do that

sjoerd-bouma: Sorry for the repeated confusion but can we keep the `Notes` section from the previous commit? It was useful to have the parameter values documented and a link to where they come from! Just the `Parameters` section was out of place.

sjoerd-bouma:
Just for reference, the single event test failure is unrelated to this PR so shouldn't worry you (it maybe should and does worry me, but I'll make a separate issue for it).

NuRadioMC/utilities/medium.py
avijai2000: All good! I just added it back in

NuRadioMC/utilities/medium.py
```diff
@@ -153,6 +155,71 @@ def __init__(self):
...
+        return self._scalarfield.getValue(position)
+    def get_average_index_of_refraction(self, position1, position2):
+        position1 = RP.Vector3d(*(position1 / units.meter))
+        position2 = RP.Vector3d(*(position2 / units.meter))
+        return self._scalarfield.getAverageValue(position1, position2)
+    def get_gradient_of_index_of_refraction(self, position):
+        pos = RP.Vector3d(*(position / units.meter))
+        return self._scalarfield.getGradient(pos) / units.meter
+    def _compute_default_ice_model_radiopropa(self):
+        return medium_base.RadioPropaIceWrapper(self, self._scalarfield)
```
fschlueter, line 227: Can we add doc-strings and some separation between the functions

anelles:
Please make sure that the changes here and in #1077 remain coherent and that the two of you don't change the same thing twice.

fschlueter:
Can this PR make use of the icemodel implemented with #1077 (which looks to be more general)

fschlueter:
Is this PR ready?

sjoerd-bouma (approved):

fschlueter:
@philippwindischhofer I assume the changes you requested were implemented? Can you lift your stop?

philippwindischhofer:
I'm trying to think what the best way to proceed is for this ...

The n-layer exponential model defined by @hanneswarnhofer in #1077 includes the 3-layer exponential defined here as a special case (but does not yet have radiopropa support built in).

I think in an ideal world we would make the n-layer exponential model compatible with radiopropa rather than having two different and mutually incompatible 3-layer exponential models to choose from.

To understand better the broader context: @avijai2000 are you still relying on this PR (and radiopropa) for your diffuse neutrino analysis? I know at some point you were evaluating Alan Coleman's C8 ray tracer. Does it make sense to move the analysis pipeline over to the analytic raytracer, or is radiopropa support still a priority?

fschlueter:
@avijai2000 can you comment on Philipps question?

avijai2000:
Sorry for the delayed response, I'm not using the radiopropa anymore for the neutrino search. I've fully switched to the c8 ray tracer. But from what I understand, if people wanted to use the 3 part model in simulation, radiopropa support was still needed for that

fschlueter:
Okay, so would be a the best way forward to implement a RadioPropa Icemodel for the icemodels implemented in #1077? (Possibly with #1077)?

philippwindischhofer:
Yes, I think this would be ideal, provided somebody is willing to spend the time on implementation and validation.

If not, my personal opinion is that it may not be worth supporting these new ice models with RadioPropa. I think it is sufficient to have _one_ validated and understood raytracer, and it seems like @hanneswarnhofer's work has given us that (or is very close to it).

fschlueter:
but does that mean we have no means to validate the new multi-layer raytracer with a multi-layer ice model? Or we do since there are other multi-layer icemodels working right now with the nuradio radiopropa raytracer?

philippwindischhofer:
Closing here, since we will be replacing RadioPropa with the C8 raytracer. Thanks very much @avijai2000 for your continued work on this!

Issue #944, opened 2025-06-06
Add multi-station test to CI


sjoerd-bouma:
We currently do not have a test in the CI that uses multiple stations, which led to #938. It would be good to add a (small) simulation test that does this to ensure we don't break multi-station functionality in the future.

PR #1038, opened 2026-01-22
hdf5 error multi station

### LINKED ISSUE #1036: Simulation breaking in hdf5 writer for multi-station event tracks
On line 352 of output_writer_hdf5.py, the `triggered` array is initialized with the length of `shower_ids_stn`, whereas the similar `multiple_triggers` and `trigger_times` arrays are then initialized with the length of `shower_ids`. For mutli-station simulations with secondary tracks this can cause a simulation breaking error if `iSh`, line 356, becomes an index greater than the length of `triggered`. Which I think happens since `shower_ids_stn` can contain indices of a subset of the showers in a track instead of using the fuller list of indices of any shower in a track that triggered any station. 

https://github.com/nu-radio/NuRadioMC/blob/1ab1a3cbb0769ce0195e194418fd019a629c9bed/NuRadioMC/simulation/output_writer_hdf5.py#L351-L375

Initializing `triggered` with the length of `shower_ids_stn` looks like a typo to me. @cg-laser I think this was written by you so it would be helpful if you can double check the code that's written.

I'm running a multi-station simulations using primary and secondary events generated from my own code. Those events are converted so they're grouped by the parent neutrino's event group ids with each shower having its own shower id. This makes debugging a little harder since it's a custom event list. I have not run multi-station simulations with proposal generated secondary events yet to see if this error occurs for those as well.

sjoerd-bouma:
Fixes #1036 

Thanks @ryankrebs016. Having had a look I agree with your analysis: both line 366 and 379 clearly expect an array of length `shower_ids` rather than `shower_ids_stn`, and this also matches the expectation/specification of the HDF5 output format (https://nu-radio.github.io/NuRadioMC/NuRadioMC/pages/HDF5_structure.html#id3).

This does once again underline the importance of a multi-station test (#944 ) if someone has time to add one!


ryankrebs016 (approved):

cg-laser:
thanks for fixing this!

Issue #1040, opened 2026-02-09
Saving and Loading in Detector Description


avijai2000:
While trying and save and load in the detector description rather than get it from the hardware database, I'm getting an attribute error (see attached image). This happens while running simulate.py. The detector description was saved with the following script: https://github.com/nu-radio/NuRadioMC/blob/ee0ca42070dd666069851e02b264677200287ea1/NuRadioReco/detector/RNO_G/rnog_detector.py#L223 and loaded in by specifying it as the detector_file here: https://github.com/nu-radio/NuRadioMC/blob/ee0ca42070dd666069851e02b264677200287ea1/NuRadioReco/detector/RNO_G/rnog_detector.py#L66. 

<img width="2696" height="464" alt="Image" src="https://github.com/user-attachments/assets/b90e0a1b-29b6-424a-9a0d-d17c20162d40" />

fschlueter:
Thanks @avijai2000, after looking into it, I believe this can be avioded by setting `always_query_entire_description=True` when initialising the detector the first time before exporting. Can you test that? 

This is not super obvious, sorry that you run into this issue. I added an error msg with PR #1041 to avoid user having this issue in the future.

I am closing this issue. Feel free to reopen it if I have overseen anything and the issue persist for you.

PR #1041, opened 2026-02-10
Avoiding incomplete export of RNOG detector descriptions.

fschlueter:
In principle this is not strictly an error, i.e., it is possible to export a complete description without the flag set. So it could have been an warning but I thought a error is more prominent and the other case should not be necessary.

### DIFF
--- a/NuRadioReco/detector/RNO_G/rnog_detector.py
+++ b/NuRadioReco/detector/RNO_G/rnog_detector.py
@@ -242,6 +242,13 @@ class Detector():
             An optional comment describing this detector that will be added to the exported detector description.
         """
 
+        if not self._query_all:
+            self.logger.error(
+                "You are exporting the detector description but you have not queried the entire description at once. "
+                "This means that only the currently buffered station description is exported which is potentially "
+                "incomplete. If you want to export the entire description, set `always_query_entire_description=True` "
+                "in the constructor before exporting.")
+
         periods = {}
         for station_id in self.__buffered_stations:
 
--- a/NuRadioReco/detector/RNO_G/rnog_detector.py
+++ b/NuRadioReco/detector/RNO_G/rnog_detector.py
@@ -246,11 +246,11 @@ class Detector():
             # When not querying the entire description at once,
             # we need to make sure that the buffer is updated
             # for all stations before exporting.
-            self._query_all = True
+            self.logger.info("Query entire detector description at once before exporting (this might take a while) ...")
             for station_id in self.__buffered_stations:
                 # remove everything (could be handled smarter ...)
                 self.__buffered_stations[station_id] = {}
-                self._query_station_information(station_id)
+                self._query_station_information(station_id, all=True)
 
         periods = {}
         for station_id in self.__buffered_stations:
@@ -507,7 +507,7 @@ class Detector():
 
             for station_id, need_update in update_buffer_for_station.items():
                 if need_update and self.has_station(station_id):
-                    self._query_station_information(station_id)
+                    self._query_station_information(station_id, all=self._query_all)
 
         # Return when buffer is not empty. This has to come first ...
         for station_id in self.__buffered_stations:
@@ -578,7 +578,7 @@ class Detector():
         self.logger.debug(f"Station {station_id} not commissioned!")
         return False
 
-    def _query_station_information(self, station_id):
+    def _query_station_information(self, station_id, all):
         """
         Query information about a specific station from the database via the db_mongo_read interface.
         You can query only information from the station_list collection (all=False) or the complete
@@ -601,7 +601,8 @@ class Detector():
 
         self.logger.info(
             f"Query information for station {station_id} at {self.get_detector_time()}")
-        if self._query_all:
+
+        if all:
             station_information = self.__db.get_complete_station_information(
                 station_id, measurement_signal_chain=self.signal_chain_measurement_name)
         else:
--- a/NuRadioReco/detector/RNO_G/rnog_detector.py
+++ b/NuRadioReco/detector/RNO_G/rnog_detector.py
@@ -250,7 +250,7 @@ class Detector():
             for station_id in self.__buffered_stations:
                 # remove everything (could be handled smarter ...)
                 self.__buffered_stations[station_id] = {}
-                self._query_station_information(station_id, all=True)
+                self._query_station_information(station_id, query_all_information=True)
 
         periods = {}
         for station_id in self.__buffered_stations:
@@ -507,7 +507,7 @@ class Detector():
 
             for station_id, need_update in update_buffer_for_station.items():
                 if need_update and self.has_station(station_id):
-                    self._query_station_information(station_id, all=self._query_all)
+                    self._query_station_information(station_id, query_all_information=self._query_all)
 
         # Return when buffer is not empty. This has to come first ...
         for station_id in self.__buffered_stations:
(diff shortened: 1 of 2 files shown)


sjoerd-bouma:
Is it bad practice to log at error level when no error is actually raised? On the other hand, I can see why you want to escalate above warning, because most likely what the user wants hasn't happened. So maybe ignore these philosophical musings.

I guess in an _ideal_ world rather than logging an error message you download the missing bits of the description on demand (i.e. on exporting), but maybe this would be tricky to implement?

fschlueter:
> I guess in an ideal world rather than logging an error message you download the missing bits of the description on demand (i.e. on exporting), but maybe this would be tricky to implement?

hm - that could be implemented... Or we actually raise an error (or we log an error and skip the exporting)? 

What of the three options do you prefer?

sjoerd-bouma:
> > I guess in an ideal world rather than logging an error message you download the missing bits of the description on demand (i.e. on exporting), but maybe this would be tricky to implement?
> 
> hm - that could be implemented... Or we actually raise an error (or we log an error and skip the exporting)?
> 
> What of the three options do you prefer?

I'm guessing as a user what I would prefer is for things to work if I try to export the detector, if this is straightforward to implement? But I can live with the other two solutions as well.

fschlueter:
copy

fschlueter:
Have a look now @sjoerd-bouma

sjoerd-bouma:
@avijai2000 could you check if on this branch your issue (#1040) is fixed?

fschlueter:
@sjoerd-bouma lets merge this

fschlueter:
ping

sjoerd-bouma (dismissed):
If @avijai2000 has time to test this once that would be nice, but it looks like it should work (and maybe @fschlueter has tested it?), so we should probably be okay to merge this.

annanayjaitly:
i tried to do a simple test to check if its working an i get an error. this is what i tried

```python
from NuRadioReco.detector.RNO_G.rnog_detector import Detector
import datetime



det = Detector(select_stations=13)
det.update(datetime.datetime(2023, 6, 1))
det.export(f"data/station_{13}_detector.json")
print(det.export_as_string())

det_loaded = Detector(detector_file="data/station_13_detector.json.xz")

det_loaded.update(datetime.datetime(2023, 6, 1))

det_loaded.get_station_ids()
print(det_loaded.export_as_string())

det_loaded.update(datetime.datetime(2023, 6, 1))
det_loaded.get_signal_chain_components(13, 1)
```

it doesn't work, i get this error:
```
AttributeError                            Traceback (most recent call last)
Cell In[4], [line 19](vscode-notebook-cell:?execution_count=4&line=19)
     16 print(det_loaded.export_as_string())
     18 det_loaded.update(datetime.datetime(2023, 6, 1))
---> [19](vscode-notebook-cell:?execution_count=4&line=19) det_loaded.get_signal_chain_components(13, 1)

File ~/git/NuRadioMC/NuRadioReco/detector/RNO_G/rnog_detector.py:61, in _check_detector_time.<locals>._impl(self, *method_args, **method_kwargs)
     58 @wraps(method)
     59 def _impl(self, *method_args, **method_kwargs):
     60     self.get_detector_time()  # this will raise an error if time is not set
---> [61](https://vscode-remote+ssh-002dremote-002bice-002dwgs1.vscode-resource.vscode-cdn.net/cephfs/users/jaitlya/git/satelite-search/~/git/NuRadioMC/NuRadioReco/detector/RNO_G/rnog_detector.py:61)     return method(self, *method_args, **method_kwargs)

File ~/git/NuRadioMC/NuRadioReco/detector/RNO_G/rnog_detector.py:1037, in Detector.get_signal_chain_components(self, station_id, channel_id)
   1016 @_check_detector_time
   1017 def get_signal_chain_components(self, station_id, channel_id):
   1018     """
   1019     Returns the names of all components in the signal chain.
   1020 
   (...)   1035         the signal chain. (For an explanation of a weight see `detector.response.Response`)
   1036     """
-> [1037](https://vscode-remote+ssh-002dremote-002bice-002dwgs1.vscode-resource.vscode-cdn.net/cephfs/users/jaitlya/git/satelite-search/~/git/NuRadioMC/NuRadioReco/detector/RNO_G/rnog_detector.py:1037)     signal_chain_dict = self.get_channel_signal_chain(
   1038         station_id, channel_id)
   1039     signal_chain_components = {
   1040         ele["name"]: ele["weight"] for ele in
...
--> [662](https://vscode-remote+ssh-002dremote-002bice-002dwgs1.vscode-resource.vscode-cdn.net/cephfs/users/jaitlya/git/satelite-search/~/git/NuRadioMC/NuRadioReco/detector/RNO_G/rnog_detector.py:662) channel_sig_info = self.__db.get_channel_signal_chain(signal_id, measurement_name=self.__signal_chain_measurement_name)
    663 channel_sig_info.pop('channel_id', None)
    665 self.__buffered_stations[station_id]["channels"][channel_id]['signal_chain'] = channel_sig_info

AttributeError: 'NoneType' object has no attribute 'get_channel_signal_chain'
```
I am not sure if its the best way to test, let me know if i should try something else.

fschlueter:
Is that the same error which was fixed with #1058 ? Can you update your local branch and rerun the check?

fschlueter:
@annanayjaitly ?

annanayjaitly:
Hello, didn't get a chance to get back to this yet, thanks for reminding! I will check it ASAP

sjoerd-bouma:
...looks like Annanay's MWE still returns the same error as above also in the updated branch.

fschlueter:
Is that still the case if you comment out the `export_as_string`? Sorry I do not have time to take care of this in the next weeks. Maybe someone else can step in?

fschlueter:
@annanayjaitly, @sjoerd-bouma for me your MWE works!

fschlueter:
@sjoerd-bouma @annanayjaitly I can not reproduce the error with the MWE.

sjoerd-bouma:
> @sjoerd-bouma @annanayjaitly I can not reproduce the error with the MWE.

It does still raise the same error for me. I've modified the CI to also test this case.

FWIW, if you think it'll be too complicated to get this working I'm also still happy with just raising a helpful error if `always_query_entire_description` is `False`, as discussed previously. Your call.

fschlueter:
@sjoerd-bouma thanks! useful addition to our testing! It should be fixed now

NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -242,6 +242,16 @@ def export(self, filename, json_kwargs=None, additional_data=None, drop_response
             An optional comment describing this detector that will be added to the exported detector description.
         """
 
+        if not self._query_all:
+            # When not querying the entire description at once,
+            # we need to make sure that the buffer is updated
+            # for all stations before exporting.
+            self._query_all = True
```
sjoerd-bouma: this now silently changes the always_query_entire_description property, right? Maybe restore the old value of `_query_all` after exporting?

fschlueter: I do not quite see the point of this, since everything is downloaded by than. Ofcourse if you would require to update the description because you change the time, it than will again download everything but this is a very specific situation and weird usecase.

sjoerd-bouma: Then maybe just add a logging message to document that the setting has changed because the user called the export method. I'm not foreseeing concrete issues but it seems just unexpected behaviour which may give someone a headache in the future if they have to debug it. (e.g. let's say in the future we find a different issue that applies only if always_query_entire_description is True or False, and then whether or not that pops up or not will additionally start to depend on whether the detector has been exported or not).

fschlueter: Okay I think I fixed the behaviour by changing the interface of the function. Have a look

NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -578,7 +578,7 @@ def has_station(self, station_id):
         self.logger.debug(f"Station {station_id} not commissioned!")
         return False
 
-    def _query_station_information(self, station_id):
+    def _query_station_information(self, station_id, all):
```
sjoerd-bouma: I probably prefer the keyword name `query_all` over the built-in `all` but I guess this is not very likely to cause problems.

sjoerd-bouma (dismissed):
One comment about a parameter name that overwrites a built-in Python function, but up to you if you want to fix it.

NuRadioReco/detector/RNO_G/rnog_detector.py
annanayjaitly: +1 on not using the built in all, I think query_all is clearer for the user anyway. Renaming an in built function can cause new and mysterious problems in the future

fschlueter: done

fschlueter:
This can be merged?

sjoerd-bouma (approved):
Thanks Felix!

PR #1043, opened 2026-02-11
Time-dependent templates for impulse response

zsselcuk:
- In the new calibrated impulse responses from @CamphynR, the best-fitting template is determined seperately each year. Since we don't store time-dependent impulse responses in the DB, we (me and @jakobhenrichs ) decided to assign these templates to the `gain_calibration` collection under the correct station/channel and year. Since there is no `full_chain` entry (template) in the new signal chain, I adjusted the `db_mongo_read.py` script to obtain the impulse response template from the `gain_calibration` collection (also works for the previous impulse response, where the templates were not time-dependent):
**New `signal_chain` entry:**
Screenshot 2026-02-11 at 22 55 26

**New `gain_calibration` entry:**
Screenshot 2026-02-11 at 22 56 01
- I adjusted the `db_mongo_write.py` script to be able to update a primary measurement far in the future. The older version only updates the current primary measurement.
- The missing weight/attenuator for the collections `gain_calibration` and `time_delay` causes errors when "full_chain" is not the first `response_chain` component

### DIFF
--- a/NuRadioReco/detector/RNO_G/db_mongo_read.py
+++ b/NuRadioReco/detector/RNO_G/db_mongo_read.py
@@ -1083,6 +1083,7 @@ class Database(object):
             channel_signal_id=channel_signal_id, measurement_name=measurement_name, verbose=verbose)
 
         for chain_key in ['response_chain', 'trigger_response_chain']:
+            templates_to_append = []
 
             # Not every channel has a trigger response chain
             if chain_key not in channel_sig_info:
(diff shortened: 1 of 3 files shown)


NuRadioReco/detector/RNO_G/db_mongo_write.py
```diff
@@ -118,7 +118,7 @@ def add_entry_to_database(self, collection, identification_key, identification_v
             If this quantity is given, the start time of the primary measurement is set to this value. Otherwise, the primary start time will be set to the current time
         """
 
-        self.set_database_time(datetime.datetime.now(tz=datetime.timezone.utc))
+        #self.set_database_time(datetime.datetime.now(tz=datetime.timezone.utc))
```
fschlueter: So the database time is now set elsewhere? Would it be saver to set the database time in cases where it is not yet set (to maintain the old behaviour) or raise an error if it is not set at this point (to force the user to set it)? In any case please do not just comment it but remove it if not needed anymore

fschlueter:
Thanks for doing this! How was the template defined before? Via the signal chain? In my mind it would be somewhat cleaner to just make a new signal chain rather than adding this to the gain calibration

fschlueter:
@zsselcuk that looks good. A nit-picking comment: Why do we call it `time_dependent_template` when the `gain_factor` is simply the `gain_factor`. I think the new structure implies time dependency and I would just call it `template` or something like `template_response`

zsselcuk:
@fschlueter, I named it so to emphasize the time-dependendency. But you are right, it is a bit unnecessary. But I have already uploaded them to the DB under the name `time_dependent_template` and would rather not change it. Is it fine?

fschlueter:
Is it a big effort to change it?

zsselcuk:
Changing it would take time but I can actually just delete the entries and reupload them with the better naming. This should be easy.

zsselcuk:
@fschlueter I changed the naming to `template_response`.

NuRadioReco/detector/RNO_G/db_mongo_read.py
```diff
@@ -1083,30 +1090,53 @@ def get_channel_signal_chain(self, channel_signal_id, measurement_name=None, ver
...
                         verbose=verbose,
                         sparameter='S21')
-
                 # add the component data to the channel_sig_info dict
                 channel_sig_info[chain_key][ice].update(component_data)
+                
+                if component_data_template is not None: 
+                    templates_to_append.append(component_data_template)
+            
+            channel_sig_info[chain_key].extend(templates_to_append)
```
fschlueter: Why are we not directly extending `channel_sig_info[chain_key]`? Why do we need `templates_to_append`?

NuRadioReco/detector/RNO_G/db_mongo_read.py
```diff
@@ -1083,30 +1090,53 @@ def get_channel_signal_chain(self, channel_signal_id, measurement_name=None, ver
...
+                        template_name = component_data["template_response"]["name"]
+                        component_data_template = {"collection": "full_chain"}
+                        component_data_template["name"] = template_name
+                        component_data_template.update(self.get_component_data(
+                            collection_name="full_chain",
+                            component_name=component_data["template_response"]["name"],
+                            sparameter="S21",
+                            verbose=verbose))
+                    else:
+                        component_data_template = None
```
fschlueter, line 1118: that seems not necessary?

NuRadioReco/detector/RNO_G/db_mongo_read.py
```diff
@@ -1083,30 +1090,53 @@ def get_channel_signal_chain(self, channel_signal_id, measurement_name=None, ver
...
-                    # only load a single calibration value
+                component_data_template = None
+                if component_entry["collection"] == "gain_calibration":
+                    component_data = self.get_time_dependent_factor(
+                        collection=component_entry["collection"], search_id=component_entry["id"])
+                    
+                    if "template_response" in component_data.keys():
+                        template_name = component_data["template_response"]["name"]
+                        component_data_template = {"collection": "full_chain"}
+                        component_data_template["name"] = template_name
```
fschlueter, line 1111: Is `"name"` not already set inside `self.get_component_data`?

fschlueter (approved):

PR #1044, opened 2026-02-12
Add custom (user-specified) amplifiers

sjoerd-bouma:
This PR makes two changes:
- It allows to provide custom amplifiers as .csv files under `NuRadioReco/detector/amps/custom`. While it is already possible to do all kinds of custom processing during the simulation (by implementing custom logic in `_detector_simulation_filter_amp`), it is not so easy to retrieve this afterwards, e.g. for analysis/reconstruction. This PR allows users to specify custom amplifier responses in  a way that makes this (hopefully) easier to generalize; I used it for example in the reconstruction of simulated IceCube-Gen2 events.
- It also no longer raises an error if no amplifier is specified in the detector (though it will very noisily emit error messages), but instead default to a unit response. This makes it possible to simulate no amplifier while still reminding the user that this is very naughty and should not be done unintentionally (I implemented this a fairly long time ago and am no longer so sure why).

For context, this is part of #1010 but I figured this is an isolated feature which would be easier to review separately.

### DIFF
--- a/NuRadioReco/detector/RNO_G/analog_components.py
+++ b/NuRadioReco/detector/RNO_G/analog_components.py
@@ -1,12 +1,13 @@
 import numpy as np
+import logging
 from scipy.interpolate import interp1d
 import os
 from NuRadioReco.utilities import units
+from functools import lru_cache
 from radiotools import helper as hp
-import logging
 logger = logging.getLogger('NuRadioReco.analog_components')
 
-
+@lru_cache(maxsize=128)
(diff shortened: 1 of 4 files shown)


NuRadioReco/detector/RNO_G/analog_components.py
fschlueter: Can you update this doc-string. it seems outdated to me

fschlueter:
Looks fine to me - you can ignore the one comment I made but you will need to resolve the merge conflict.

fschlueter (approved):

PR #1045, opened 2026-02-14
adding jittering functions to examples

Skoishi:
update the Gap time(t_trigger_condition - t_start) in simulation to discribe the behavior of the jittering of gap time in actual data:

in data readout the radiant board use LAB4d which readout the input waveform in 128-sample blocks (https://arxiv.org/pdf/1803.04600). Whenever the trigger arrives, sampling will continue until the current 128-sample block is complete, and only then the readout will start. For randomly-arriving trigger signals, this would lead to a uniformly-jittering trace start time +-64 samples/radiant sampling frequency.
This update add this function to current simulations and give an option for most recent trigger set-up the phased array in the example folder


Skoishi:
this file is also added to plan for future integration for Tong's work regarding to add various t_delay onto simulation gap time to better represent the real data behavior.

tongnpdl:
Thank you! I think we should jitter the same amount for all channel ie
pre_trigger_time = 250*units.ns + uniform(-64,64) * 1./ sample_rate

NuRadioMC/examples/RNO_G_trigger_simulation/simulate_array_PA.py
```diff
@@ -0,0 +1,564 @@
...
+                in zip(trigger_channels, vrms_after_gain)}
+        else:
+            # We round here. This is not how an ADC works but I think this is not needed here.
+            threshold_high = {channel_id: int(round(threshold * vrms)) for channel_id, vrms
+                in zip(trigger_channels, vrms_after_gain)}
+            threshold_low = {channel_id: int(round(-1 * threshold * vrms)) for channel_id, vrms
+                in zip(trigger_channels, vrms_after_gain)}
+
+        # Use RADIANT sampling rate (2.4 GHz) for jitter calculation
+        pre_trigger_dict = {channel_id: 250 * units.ns + _jitter_adder_internal(radiant_sampling_rate, rng) for channel_id in trigger_channels}
```
fschlueter: Does that mean that each channel has its own trace start time? I thought all channels should be in sync?

NuRadioMC/examples/RNO_G_trigger_simulation/simulate_array_PA.py
```diff
@@ -0,0 +1,564 @@
...
+    
+    # Create pre_trigger_times for ALL channels (ch0-ch23) to avoid KeyError during readout
+    # Use RADIANT sampling rate (2.4 GHz) for jitter calculation
+    phased_array_pre_trigger_dict = {
+        channel_id: 250 * units.ns + _jitter_adder_internal(radiant_sampling_rate, rng) 
+        for channel_id in range(24)
+    }
+
+    # Set pre_trigger_times on the phased_array trigger (used for readout window calculation)
+    station.get_trigger("phased_array").set_pre_trigger_times(phased_array_pre_trigger_dict)
```
fschlueter: Does that mean that the phased array trigger is simulated with the same jitter for every event? (Same comment as for the other trigger applies here too).

fschlueter:
Thank you @Skoishi and @tongnpdl for starting this! We will be talking about this during the call tomorrow, so no need to reply right away. However, I wanted to already post some comments / questions now. I think this is an important detector effect which we should build into NuRadioMC's simulation.py. And I agree, I think varying the pre_trigger_time is the correct approach.

tongnpdl:
I agree with both of @fschlueter  comments. I haven't thought about the effects on event-to-event basic but I like how @Skoishi implemented it as a separate function that you can add on top giving the flexibility. 

Another approach I have in mind was changing how "channelReadoutWindowCutter" is implemented, like adding the the uniform random 64 samples jitter on top of the pre_trigger_time instead of trying to control pre_trigger_time prior the simulation. And one have to pass kwarg somehow to the module to include the jitter (or not). It's not obvious to me where would be a good place to do that .. maybe simulation class argument but I haven't thought it through. What do you think?

fschlueter:
> Another approach I have in mind was changing how "channelReadoutWindowCutter" is implemented, like adding the the uniform random 64 samples jitter on top of the pre_trigger_time instead of trying to control pre_trigger_time prior the simulation. And one have to pass kwarg somehow to the module to include the jitter (or not). It's not obvious to me where would be a good place to do that .. maybe simulation class argument but I haven't thought it through. What do you think?

I agree the `channelReadoutWindowCutter` module seems to be the (one) logical place to add this functionality. The module `analogToDigitalConverter` already offers to add a "random time jitter" to the digitized trace. I am not quite sure yet what is the right way. The `channelReadoutWindowCutter` pretends that we know the time where the trace starts exactly, which is, afaiu, not the situation. I think the `analogToDigitalConverter` but there it is kind of awkward to roll the traces without cutting them.

Another though was that we add this to an RNO-G specific module if this is to RNO-G specific.

Maybe @cg-laser or @sjoerd-bouma have thoughts on that.

Skoishi:
Ok, I changed the push request. It is now on window cutter and allow user defined random seed so that the result will be trackable.

fschlueter:
Hey @Skoishi, thanks! I have three requests: 
- Can you delete the script you are adding? It is not needed anymore if I understand correctly.
- Can you make the jitter in the channelReadoutWindowCutter configurable? Right now it hardcoded and tuned for RNO-G (keep in mind other people might use this module in the future). So I would argue we need to 1) enable/disable the jitter completely, 2) be able to change the values with which the pre_trigger_time is jittered. Maybe also whether we want to have a uniform and gaussian part.

Skoishi:
Will do. In that case I'll add it as an seperate function in window cutter. so people can comment out or change the function input. or do you think it will be better to add the function input it onto run flag?

Skoishi (commented):
changed the jitter adder to a user controlled function as requested. and adding comments for readability

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -233,6 +260,48 @@ def get_empty_channel(station_id, channel_id, detector, trigger, sampling_rate):
...
+    Returns
+    -------
+    sample_jitter_time : float
+        Time jitter from the discrete sample offset
+        (``jitter_sample / sampling_rate``).
+    norm_smear : float
+        Gaussian time smear drawn from ``N(0, gaussian_spread)``.
+    """
+    if rng is None:
+        rng = Generator(Philox(secrets.randbits(128)))
```
fschlueter: not 42?

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -66,6 +76,22 @@ def run(self, event, station, detector):
...
+        # unknown trigger time in data(gaussian spread by empirical data is 10.5ns)
+        first_channel = next(station.iter_channels())
+        first_detector_sampling_rate = detector.get_sampling_frequency(
+            station.get_id(), first_channel.get_id())
+
+        jitter_time = jitter_adder(
+            gaussian_spread=10.5 * units.ns,
+            sample_block_size=64,
+            sampling_rate=first_detector_sampling_rate * units.GHz,
+            rng=self._rng,
```
fschlueter: That is a really good idea! Can you make these hard-coded values argument in the run method?

fschlueter: Also, we want to be able to simulate no time_jitter. Is that possible when `sample_block_size` and `gaussian_spread` is 0? Maybe it is worth making it explicit with a if condition?

NuRadioReco/modules/channelReadoutWindowCutter.py
Skoishi: I took 128 from the trigger simulation examples. will 42 works better?

NuRadioReco/modules/channelReadoutWindowCutter.py
Skoishi: Now I changed the sample block exception case, for gaussian it allow 0 as input and will just be a delta function based on my test

NuRadioReco/modules/channelReadoutWindowCutter.py
fschlueter: Nope - sorry is an nerd joke

fschlueter:
@Skoishi, I made some changes which show what I was after. However, I have another comment. For a digital readout, the gaussian spread needs to be discretised in samples as well, correct?

Skoishi:
> @Skoishi, I made some changes which show what I was after. However, I have another comment. For a digital readout, the gaussian spread needs to be discretised in samples as well, correct?

Actually, after think about this, you are completely right. I do want to run one more test with this though, I'd like to check if this gaussian time varies or not after the sampling rate change. If yes, we can check if it's actually vary by sample amount.

fschlueter:
What this PR also needs is a way to configure the time jitter when the module is used in `simulation.py`

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -2,27 +2,53 @@
...
+        --------
+        time_jitter
+        """
+        
         self.__sampling_rate_error_issued = False
-        pass
+        self._rng = Generator(Philox(random_seed or secrets.randbits(128)))
+        self._gaussian_jitter = gaussian_jitter
+        self._sample_block_size = sample_block_size
+
```
sjoerd-bouma: The seed should be passed from `simulation.py`, otherwise simulations are no longer deterministic.

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -66,6 +92,25 @@ def run(self, event, station, detector):
             return
 
         trigger_time = trigger.get_trigger_time()
+
+        # Compute jitter time caused by multiple experimental sources
+        # like readout sampling block(L4D board designed to be 128 samples 
+        # empirically data gives 64sample block) and 
+        # unknown trigger time in data(gaussian spread by empirical data is 10.5ns)
```
sjoerd-bouma: Remove/adjust comment (this module is not just for RNO-G)

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -66,6 +92,25 @@ def run(self, event, station, detector):
...
+        first_channel = next(station.iter_channels())
+        first_detector_sampling_rate = detector.get_sampling_frequency(
+            station.get_id(), first_channel.get_id())
+
+        gaussian_spread=10.5 * units.ns
+        sample_block_size=64
```
sjoerd-bouma: remove

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -2,27 +2,53 @@
...
+        random_seed : int or None, optional
+            Set a random seed, used if simulating a random jitter for the readout window.
+            (Default: secrets.randbits(128))
+
+        gaussian_jitter : float, optional
+            Gaussian time jitter in ns (default unit) to smear the position of the trigger in the trace (nominal
+            value defined by ``trigger.pre_trigger_time``). (Default: 0)
+
+        sample_block_size : int, optional
+            Defines "block" of samples within which the trigger is uniformly smeared. (Default: 0)
```
sjoerd-bouma: Typically, the user does not explicitly pass anything to the `channelReadoutWindowCutter` in simulations; instead, you can define triggers. I would suggest 
1. storing the jitter_time in the `Trigger` class ([NuRadioReco.framework.trigger](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.framework.trigger.html#module-NuRadioReco.framework.trigger)) in the same way that the `pre_trigger_times` are. The default value should be 0 (no jitter added).
2.  and then make this module read the jitter_time in the same way as the pre_trigger_times in the run function

sjoerd-bouma (commented):
Thanks all for the good work! I've left some comments. I think the definition of the jitter time should move to the Trigger object and added as an optional parameter to the trigger running modules in `NuRadioReco/modules/trigger`, similar to how the pre_trigger_time can be defined, because there is currently no way to pass arguments to the `channelReadoutWindowCutter` when running a simulation.

I'm also not entirely convinced by the introduction of the gaussian jitter. I'm happy to merge this with only uniform jitter (which we know should definitely be there), but I would like to see a real 1:1 comparison with data, as we discussed, before we add an additional gaussian component, I think. So my suggestion would be to introduce only the uniform component in this PR, and then once that study is done we can decide if we want to introduce additional components, which should be relatively straightforward to implement.

NuRadioReco/modules/channelReadoutWindowCutter.py
fschlueter: I agree that this is probably the cleanest solution!

Skoishi: isn't adding a setter for the seeds works easier? since default is 0, so people do not need to set the seed if they do not need the jitter simulation, but if they need one they can reproduce it by passing the seed like in the trigger simulation example to device incorporate. and in that way for the simulation module it can use 1 global seed then.

NuRadioReco/modules/channelReadoutWindowCutter.py
fschlueter: But that is possible right?

sjoerd-bouma: Also this is a case where the `or` shortcut does not do exactly what you expect - if I want to use the random seed `0` this will not work. Better to write an explicit `None` check here.

On second thought - `Philox(random_seed)` should really cover all cases already, if `random_seed==None` this will generate a random seed from the OS automatically anyway.

NuRadioReco/framework/trigger.py
```diff
@@ -59,6 +60,16 @@ def __init__(self, name, channels=None, trigger_type='default', pre_trigger_time
...
             if only a float is given, the same pre_trigger_time is used for all channels
+        sample_block_size: int
+            range of the uniform integer jitter (in samples) caused by the
+            readout window being collected in sample blocks.
+            An integer is drawn from ``[-sample_block_size / 2, sample_block_size / 2)``.
+            Default: 0 samples(no block jitter)
+        gaussian_jitter: float
+            standard deviation of a Gaussian time smear applied to the
+            readout window position (in NuRadioReco time units).
+            by testing, before sampling rate change this value is about 10.5ns
```
sjoerd-bouma: remove line 71, or specify that this is only true for RNO-G (my preference is to remove this altogether - I'm still not completely convinced we need gaussian jitter!)

NuRadioMC/simulation/simulation.py
```diff
@@ -1702,6 +1729,9 @@ def run(self):
         self._output_writer_hdf5.calculate_Veff()
         if not self._output_writer_hdf5.write_output_file():
             logger.warning("No events were triggered. Writing empty HDF5 output file.")
+            #Youwei"s edit
+            print("No events were triggered. Writing empty HDF5 output file.")
+
```
sjoerd-bouma: The message is already logged, can you remove the comment and print statement?

Skoishi: I've updated my commit. It should be fine now.

NuRadioReco/framework/trigger.py
```diff
@@ -59,6 +60,15 @@ def __init__(self, name, channels=None, trigger_type='default', pre_trigger_time
...
             if only a float is given, the same pre_trigger_time is used for all channels
+        sample_block_size: int
+            range of the uniform integer jitter (in samples) caused by the
+            readout window being collected in sample blocks.
+            An integer is drawn from ``[-sample_block_size / 2, sample_block_size / 2)``.
+            Default: 0 samples(no block jitter)
+        gaussian_jitter: float
+            standard deviation of a Gaussian time smear applied to the
+            readout window position (in NuRadioReco time units).
+            Default: 0 ns (no Gaussian jitter)
```
fschlueter: is it not awkward to define the `sample_block_size` in samples (int) and the `gaussian_fitter` in time/

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -2,27 +2,51 @@
...
+
+        Also see
+        --------
+        time_jitter
```
fschlueter: remove

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -66,6 +90,22 @@ def run(self, event, station, detector):
...
+
+        # Compute jitter time caused by multiple experimental sources
+        # like readout sampling block(L4D board designed to be 128 samples 
+        # empirically data gives 64sample block) and 
+        # unknown trigger time in data(gaussian spread by empirical data is 10.5ns)
+        first_channel = next(station.iter_channels())
+        first_detector_sampling_rate = detector.get_sampling_frequency(
+            station.get_id(), first_channel.get_id())
+
+        jitter_time = time_jitter(
```
fschlueter, line 86: should we store this time jitter back to the trigger object to keep the information on simulations? In data we do not have the information (so I we might decide to not store it)

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -233,6 +274,54 @@ def get_empty_channel(station_id, channel_id, detector, trigger, sampling_rate):
     return channel
 
 
+def time_jitter(gaussian_spread=0*units.ns, sample_block_size=64, sampling_rate=2.4*units.GHz, rng=None):
```
fschlueter: should we store this in the trigger module?

NuRadioMC/simulation/simulation.py
```diff
@@ -1153,6 +1153,9 @@ def __init__(
             allows to specify a custom ice model. This model is used if the config file specifies the ice model as "custom".
         trigger_channels: list of ints or dict of list of ints
             list of channel ids that are used for the trigger (per station_id). If None, all channels are used.
+        jitter_random_seed: int or None
+            Random seed for the readout window jitter RNG in
+            channelReadoutWindowCutter. If None, a random seed is generated.
```
fschlueter: please use `self._config['seed']` instead of an additional seed

NuRadioReco/modules/channelReadoutWindowCutter.py
Skoishi: Sorry, what do you mean in data we do not have the information？I believe at least the sample_block is understood both in board design and have the data evidence?

NuRadioReco/framework/trigger.py
Skoishi: Sorry to reply late, busy with diffuse search recently. Gaussian_jitter in time was proposed by Tong. Recently I haven't get enough time to do a more through study to confirm this is independent respect to frequency(not that much high SNR pulser runs too) But so far, gaussian_jitter is an extra method provided for people to put in their empirical observations.

NuRadioReco/modules/channelReadoutWindowCutter.py
fschlueter: In simulation we can trac the result of `time_jitter(...)` this might be useful to understand some results better. In data although we now how `time_jitter(..)` works, we do not know the random value per event.

NuRadioReco/framework/trigger.py
fschlueter: Regardless of whether there is a gaussian component present in the "trigger time uncertainty" in RNO-G, I can see that it make sense to simulate a gaussian jitter. However, I do not know if it use confusing or not to define one part of the uncertainty in samples and the other one in time.

NuRadioReco/modules/channelReadoutWindowCutter.py
fschlueter: @Skoishi

NuRadioMC/examples/RNO_G_trigger_simulation/simulate.py
```diff
@@ -143,6 +143,10 @@ def rnog_flower_board_high_low_trigger_simulations(evt, station, det, trigger_ch
             trigger_name=f"deep_high_low_{thresh_key}",
             pre_trigger_time=250 * units.ns,
         )
+        station.get_trigger(f"deep_high_low_{thresh_key}").set_jitter_params(
+            sample_block_size=64,
```
fschlueter: Is the `64` that actually accurate if we randomly sample in `rng.integers(-sample_block_size / 2, sample_block_size / 2)`? I thought the sampling windows are 128 sample long.

NuRadioReco/modules/channelReadoutWindowCutter.py
```diff
@@ -233,6 +261,54 @@ def get_empty_channel(station_id, channel_id, detector, trigger, sampling_rate):
...
+        ``Generator(Philox(random_seed or secrets.randbits(128)))``.
+
+    Returns
+    -------
+    sample_jitter_time : float
+        Time jitter from the discrete sample offset
+        (``jitter_sample / sampling_rate``).
+    norm_smear : float
+        Gaussian time smear drawn from ``N(0, gaussian_spread)``.
+    """
```
fschlueter: This is not returned

NuRadioReco/modules/channelReadoutWindowCutter.py
fschlueter: fine like this?

NuRadioReco/framework/trigger.py
Skoishi: Ok, I see what you are saying. I'll convert the gaussian to samples based jitter too.

NuRadioReco/modules/channelReadoutWindowCutter.py
Skoishi: I think most of this jitter is used in ML to compare with data? Only the distribution matters in this use case. Do you have any other use case in mind which the individual jitter info might be helpful?

NuRadioMC/examples/RNO_G_trigger_simulation/simulate.py
Skoishi: Thanks for catching it should be 128 based on current study

For documentation purpose:

by Tong:
In [glitch detection](https://github.com/nu-radio/NuRadioMC/blob/develop/NuRadioReco/modules/RNO_G/channelGlitchDetector.py#L52), lab4d sampling block size is 64. But in [block offset](https://github.com/nu-radio/NuRadioMC/blob/develop/NuRadioReco/modules/RNO_G/channelBlockOffsetFitter.py#L35), the offset block size is 128.

By Ryan Krebs:
The trigger processing and the 128 sample block are linked, so it should just reduce to where triggering signals can appear in the 128 sample block.

Skoishi:
I believe I addressed all the questions from this commit, but if there is any thing I missed. Let me know.
Thank you Felix and Sjored for patiently bear with me to adjust the detail of this new function to NuRadioMC

NuRadioReco/modules/channelReadoutWindowCutter.py
fschlueter: In detailed studies of the trigger (to know where the trigger fired). But I guess you could just disable this. If someone wants this feature, they can implement it

fschlueter (approved):

fschlueter:
Thanks for your patient and getting this done @Skoishi

fschlueter:
Feel free to merge it yourself!

Skoishi:
Thanks! It is merged now!

Issue #462, opened 2022-11-04
default argument in voltageToEfield converter modules


shallmann:
Is there a reason, why the default value from `use_channels=None` is silently overwritten to [0,1,2,3] (I guess this is probably ARIANNA heritage?)

https://github.com/nu-radio/NuRadioMC/blob/3550584f62769a91afe62d645ca71718d47b1e4e/NuRadioReco/modules/voltageToAnalyticEfieldConverter.py#L297
https://github.com/nu-radio/NuRadioMC/blob/3550584f62769a91afe62d645ca71718d47b1e4e/NuRadioReco/modules/voltageToEfieldConverter.py#L148

1. This behaviour looks potentially dangerous
2. (same is maybe true for the set bandpass in the AnalyticEfieldConverter)?
3.  Also the voltageToEfieldConverter fails with a hard-to-track error message, if only called with single antenna in `use_channels`. Since I don't necessarily expect the module does return anything useful in that case, we should either add some logging hint, or require > 1? Any preference?

christophwelling:
Yes, I'm pretty sure the default `[0, 1, 2, 3]` is a leftover from ARIANNA.
If you have a default that is an object (like an array), the recommended way to handle it is actually to have `None` in the function definition and then overwrite it with the actual default. If you specify the array right in the function definition, this will create an array object, which is then used as the default. The problem is that on subsequent function calls, this array object will not be recreated, so any changes that are done to the array will change the default the next time the function is called.
In this case, I would suggest to just not have a default for `use_channels`. For the efieldToAnalyticVoltageConverter, it is more complicated, because the `use_channels` parameter has other parameters with defaults before it. So we would either have to remove the defaults for those too or change the order of the parameters. Both will probably break things.
I can check why exactly the module fails if you only have one channel and put in a warning.

fschlueter:
I stumbled across that a couple of time too and do not like it neither. My suggestion would be to keep the default to `None` but instead of silently set it to `[0, 1, 2, 3]` raise an error with an appropiate error message that this argument now has to be set. This makes the argument defacto positional but keeps backwards compatibility

sjoerd-bouma:
> I stumbled across that a couple of time too and do not like it neither. My suggestion would be to keep the default to `None` but instead of silently set it to `[0, 1, 2, 3]` raise an error with an appropiate error message that this argument now has to be set. This makes the argument defacto positional but keeps backwards compatibility

[...]

PR #1047, opened 2026-02-19
remove arianna heritage default arguments - make them required argume…

fschlueter:
…nts - added appropiate error message

Addresses #462

### DIFF
--- a/NuRadioReco/modules/voltageToAnalyticEfieldConverter.py
+++ b/NuRadioReco/modules/voltageToAnalyticEfieldConverter.py
@@ -284,20 +284,26 @@ class voltageToAnalyticEfieldConverter:
         debug_plotpath: string or None
             if not None plots will be saved to a file rather then shown. Plots will
             be save into the `debug_plotpath` directory
-        use_channels: array of ints (default: [0, 1, 2, 3])
-            the channel ids to use for the electric field reconstruction
-            default: 0 - 3
-        bandpass: [float, float] (default: [100 * units.MHz, 500 * units.MHz])
-            the lower and upper frequecy for which the analytic pulse is calculated.
+        use_channels: array of ints
+            The channel ids to use for the electric field reconstruction
+        bandpass: [float, float]
+            The lower and upper frequecy for which the analytic pulse is calculated.
             A butterworth filter of 10th order and a rectangular filter is applied.
-            default 100 - 500 MHz
         use_MC_direction: bool
             use simulated direction instead of reconstructed direction
         """
         if use_channels is None:
-            use_channels = [0, 1, 2, 3]
+            msg = ("No channels specified for electric field reconstruction. "
+                "Please provide `use_channels` argument to the voltageToAnalyticEfieldConverter.")
+            logger.error(msg)
+            raise ValueError(msg)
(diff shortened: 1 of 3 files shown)


NuRadioReco/modules/voltageToAnalyticEfieldConverter.py
```diff
@@ -284,20 +284,26 @@ def run(self, evt, station, det, debug=False, debug_plotpath=None,
...
+                "Please provide `use_channels` argument to the voltageToAnalyticEfieldConverter.")
+            logger.error(msg)
+            raise ValueError(msg)
+
         if bandpass is None:
-            bandpass = [100 * units.MHz, 500 * units.MHz]
+            msg = ("No bandpass specified for electric field reconstruction. "
+                "Please provide `bandpass` argument to the voltageToAnalyticEfieldConverter.")
+            logger.error(msg)
+            raise ValueError(msg)
```
sjoerd-bouma, line 305: Is there any reason why you log the error and then raise it? (There may be a good reason, just curious to learn it!) Naively, I would have thought this just results in the user seeing the error message twice.

sjoerd-bouma (commented):
I think you're right and this needs to be deprecated, even if it may theoretically break some existing ARIANNA scripts. I would suggest documenting this explicitly in the changelog (under 'deprecations / breaking changes / some header that's not features/bug fixes).

anelles:
The ARIANNA experiment has been decommissioned and likely all analyses have been published, I think I am fine with breaking backwards compatibility here.

fschlueter:
@anelles does that mean we can remove all already depreciated because unused and untested Arianna code? We could create an archival branch before doing that!

NuRadioReco/modules/voltageToAnalyticEfieldConverter.py
fschlueter: Honestly I have been asking myself this question in the past. It seemed to me to be some kind of standard. I guess it comes from the fact that logger.error doesn’t actually abort the program while raise ValueError might not nicely format the error message? I am open to alternatives! Just keep in mind that we have this structure in dozen places in nuradio

sjoerd-bouma: > It seemed to me to be some kind of standard.

[citation needed] ;) I wouldn't do it myself, but it only adds one line of output to the usually fairly long traceback from raising `ValueError` so I don't think it's a dealbreaker for me. I guess if you're logging the NuRadio logger to a file that wouldn't record the ValueError, so in that case there's even some benefit.

fschlueter: standard in NuRadio - see for yourself :) Is there a better way to abort the Program and log the error?

sjoerd-bouma (approved):

anelles:
> @anelles does that mean we can remove all already depreciated because unused and untested Arianna code? We could create an archival branch before doing that!

Maybe that is the way to go. Create an archival branch and then clean up, because it is indeed very confusing and I am sure that we inadvertently already broke a couple of things, because it is not tested.

PR #1048, opened 2026-02-19
Store shower Xmax in simulated shower when simulating with ARZ

### LINKED ISSUE #1042: shower_maximum currently not stored in simulated shower
The parameter `shower_maximum` ([link](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.framework.parameters.html#NuRadioReco.framework.parameters.showerParameters.shower_maximum)) is currently not stored in the simulated shower by default. For neutrino events simulated with ARZ this is relevant in reconstruction, because the viewing angle and travel times are adjusted based on the simulated xmax.

While in principle the xmax can be retrieved indirectly via the [charge_excess_profile_id](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.framework.parameters.html#NuRadioReco.framework.parameters.showerParameters.charge_excess_profile_id), this is rather cumbersome and unintuitive. It would be better to just store xmax directly during `simulation.py`.

This should hopefully be a relatively easy change. We could consider also storing xmax in the HDF5 files.

sjoerd-bouma:
Closes #1042 

Store the shower Xmax in the `shower_maximum` parameter when simulating with ARZ, similar to how the shower realization is already stored.

I guess there is some slight ambiguity in how we store Xmax (occasionally we refer to it as a length); I think probably the most sensible/consistent thing to do is to store it in units of weight / length^2, which is what I've done here.

I've also fixed a bug in the ARZ module where the signature expected from `output_mode` was not respected if an empty trace was returned (because we're too far from the Cherenkov angle).

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -236,6 +236,9 @@ def calculate_sim_efield(
                 if not shower.has_parameter(shp.charge_excess_profile_id):
                     shower.set_parameter(shp.charge_excess_profile_id, additional_output['iN'])
                     logger.debug(f"setting shower profile for ARZ shower library to i = {additional_output['iN']}")
+                if not shower.has_parameter(shp.shower_maximum):
+                    shower_xmax = additional_output['profile_depth'][np.argmax(additional_output['profile_ce'])]
+                    shower.set_parameter(shp.shower_maximum, shower_xmax)
 
             if config['signal']['model'] == "Alvarez2009":
                 if not shower.has_parameter(shp.k_L):
--- a/NuRadioMC/SignalGen/ARZ/ARZ.py
+++ b/NuRadioMC/SignalGen/ARZ/ARZ.py
@@ -607,63 +607,62 @@ class ARZ(object):
         # to the Cherenkov angle, but NuRadioMC will reuse the shower realization of the first ray tracing solution.
         if np.abs(theta - cherenkov_angle) > maximum_angle:
             logger.info(f"viewing angle {theta/units.deg:.1f}deg is more than {maximum_angle/units.deg:.1f}deg away from the Cherenkov cone. Returning zero trace.")
-            empty_trace = np.zeros((3, N))
-            return empty_trace
+            trace_onsky = np.zeros((3, N))
+        else:
+            # get the appropriate model parameters
+            if shower_type == "HAD":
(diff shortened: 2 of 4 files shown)


NuRadioMC/simulation/simulation.py
```diff
@@ -236,6 +236,9 @@ def calculate_sim_efield(
                 if not shower.has_parameter(shp.charge_excess_profile_id):
                     shower.set_parameter(shp.charge_excess_profile_id, additional_output['iN'])
                     logger.debug(f"setting shower profile for ARZ shower library to i = {additional_output['iN']}")
+                if not shower.has_parameter(shp.shower_maximum):
+                    shower_xmax = additional_output['profile_depth'][np.argmax(additional_output['profile_ce'])]
```
fschlueter, line 240: we do not need a more robust calculation? I guess for LPM showers, there is often anyway no defined Xmax?

fschlueter:
With this implementation, is the xmax recalculated for every event? Can be made more efficient if we dont do that? It might not matter (sorry have not checked it myself).

NuRadioMC/simulation/simulation.py
sjoerd-bouma: This is how xmax is defined 'internally' in ARZ anyway, also for LPM showers. It may not be particularly reliable/meaningful for those showers, but I think I'm happy to leave that for the user to deal with.

sjoerd-bouma:
> With this implementation, is the xmax recalculated for every event? Can be made more efficient if we dont do that? It might not matter (sorry have not checked it myself).

Yes, it's calculated for each event (but not for each channel). Technically one could cache this (it is calculated internally when computing the askaryan emission in ARZ also) and take us from 2 to just 1 call of `np.argmax`. But seeing as the rest of the ARZ computation involves integrals, FFTs, etc., my estimate is this would speed things up by less than 0.01%.

NuRadioMC/simulation/simulation.py
fschlueter: sure

fschlueter:
Okay, I agree probably not necessary. I am afraid you have to solve a merge conflict again

sjoerd-bouma:
ping @fschlueter

fschlueter (approved):

PR #1050, opened 2026-02-20
replace deprecated numpy.trapz with scipy.integrate.trapezoid

sjoerd-bouma:
`np.trapz` is being deprecated in favour of `np.trapezoid`, and seems to be fully removed in the latest versions of numpy. Because `np.trapezoid` is not available in some older versions of numpy, I've instead replaced it with `scipy.integrate.trapezoid` which should work on all Python versions that we claim to support.

### DIFF
--- a/NuRadioMC/SignalGen/ARZ/ARZ.py
+++ b/NuRadioMC/SignalGen/ARZ/ARZ.py
@@ -15,8 +15,11 @@ try:
 except ImportError:
     numba_available = False
 
+if not hasattr(np, 'trapezoid'):
+    # needed for compatibility with old numpy versions
+    np.trapezoid = np.trapz
+
 logger = logging.getLogger("NuRadioMC.SignalGen.ARZ")
-# logger.setLevel(logging.INFO)
 
 ######################
 ######################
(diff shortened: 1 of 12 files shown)


NuRadioMC/SignalGen/ARZ/ARZ.py
```diff
@@ -15,8 +15,11 @@
 except ImportError:
     numba_available = False
 
+if not hasattr(np, 'trapezoid'):
+    # needed for compatibility with old numpy versions
+    np.trapezoid = np.trapz
+
```
sjoerd-bouma, line 21: I thought `scipy.integrate.trapezoid` would be the same routine as `np.trapezoid` but it seems numba doesn't like it, so at least for ARZ we have to use the numpy version. In principle we could use this same workaround everywhere but it doesn't seem particularly elegant...

fschlueter:
What old numpy version don't you want to sacrifice?

sjoerd-bouma:
> What old numpy version don't you want to sacrifice?

1.x, trapezoid was only added with numpy 2.0 https://numpy.org/devdocs/reference/generated/numpy.trapezoid.html. If we deprecate it we have to deprecate one or two Python versions as well.

fschlueter:
> > What old numpy version don't you want to sacrifice?
> 
> 1.x, trapezoid was only added with numpy 2.0 https://numpy.org/devdocs/reference/generated/numpy.trapezoid.html. If we deprecate it we have to deprecate one or two Python versions as well.

are those already end of live?

fschlueter (approved):

fschlueter:
ping

fschlueter:
@sjoerd-bouma ?

PR #1051, opened 2026-02-23
Calculating focusing factor without setting shower axis

fschlueter:
My understanding is that the shower axis is not needed for the calculation of the focusing factor... might be wrong

### DIFF
--- a/NuRadioMC/SignalProp/propagation.py
+++ b/NuRadioMC/SignalProp/propagation.py
@@ -6,7 +6,7 @@ solution_types = {
     3: 'reflected'
 }
 
-solution_types_revert = {v:k for k, v in solution_types.items()}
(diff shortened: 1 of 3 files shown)


fschlueter:
I am miss using this PR to sneak in a few other unrelated changes (should be uncontroversal) - sorry.

sjoerd-bouma:
I'm assuming you've tested this? (Radiopropa is unfortunately still not part of the CI). If it works, it's fine by me - from Bob's comment on line 114, it seems to be just a performance feature (?) to avoid raytracing for geometries where we expect to be very far from the Cherenkov angle.

fschlueter:
Yes I tested it - seems fine!

sjoerd-bouma (approved):

Issue #1054, opened 2026-02-24
Error in retrieving device positions through rnog_detector class


sanyukta13:
Using `RNO_G.rnog_detector.Detector` class, extracting device positions using `Detector.get_relative_position_device(station_id, device_id)` leads to error as follows:

```jupyter-notebook
----> [7](vscode-notebook-cell:?execution_count=3&line=7) det.get_relative_position_device(11,0)

File ~/software/source/NuRadioMC/NuRadioReco/detector/RNO_G/rnog_detector.py:61, in _check_detector_time.<locals>._impl(self, *method_args, **method_kwargs)
     58 @wraps(method)
     59 def _impl(self, *method_args, **method_kwargs):
     60     self.get_detector_time()  # this will raise an error if time is not set
---> [61](https://vscode-remote+ssh-002dremote-002bcobalt.vscode-resource.vscode-cdn.net/data/user/sanyukta/rno_code/triboelectric/~/software/source/NuRadioMC/NuRadioReco/detector/RNO_G/rnog_detector.py:61)     return method(self, *method_args, **method_kwargs)

File ~/software/source/NuRadioMC/NuRadioReco/detector/RNO_G/rnog_detector.py:1091, in Detector.get_relative_position_device(self, station_id, device_id)
   1086         raise KeyError(
   1087             f"\"id_position\" not in buffer for device {device_id}.")
   1089     position_id = self.__buffered_stations[station_id]["devices"][device_id]["id_position"]
-> [1091](https://vscode-remote+ssh-002dremote-002bcobalt.vscode-resource.vscode-cdn.net/data/user/sanyukta/rno_code/triboelectric/~/software/source/NuRadioMC/NuRadioReco/detector/RNO_G/rnog_detector.py:1091)     device_pos_info = self.__db.get_device_position(
   1092         device_position_id=position_id)
   1093     self.__buffered_stations[station_id]["devices"][device_id]['device_position'] = device_pos_info
   1095 return np.array(self.__buffered_stations[station_id]["devices"][device_id]["device_position"]["position"])

AttributeError: 'Database' object has no attribute 'get_device_position'
``` 
`get_device_position` is not defined in `RNO_G.db_mongo_read.Database`, can use `get_complete_device_information().['device_position']` instead

fschlueter:
Thanks for the report, I will try to take a look at it.

fschlueter:
@sanyukta13 can you checkout #1056

PR #1056, opened 2026-02-25
Fix function call in rnog detector

fschlueter:
Adressing #1054


sanyukta13:
@fschlueter works now, thanks!

sanyukta13 (approved):
fixed #1054

PR #1060, opened 2026-03-02
move unscramble and diff_sq

tongnpdl:
Move 2 methods outside of channelGlitchDetector class to be used for monitoring.
Expecting no functional changes.

Key changes:
- `unscramble` now live outside of class definition and can be called without class object
- `diff_sq` now live outside of class definition and can be called without class object
- LAB4D readout size and sampling block size are now global variables that are also default values for the two methods and the class when call with or without the class.

TODOs

- [x] update changelog.txt
- [x] update docstrings


fschlueter (dismissed):
Thanks Tong

fschlueter (dismissed):

fschlueter (approved):

PR #1062, opened 2026-03-03
Fix bug in SimpleIceModel.get_gradient_of_index_of_refraction - bug o…

### LINKED ISSUE #1061: Gradient error from bad dtype
In [this line](https://github.com/nu-radio/NuRadioMC/blob/0539f545bf5cc922a3f25d7ce323b09b9da86a2f/NuRadioMC/utilities/medium_base.py#L340C13-L340C41) the code creates a zero vector and then fills the z-value with the gradient. However, it initializes a numpy array of `dtype=int` so any gradients with magnitude less than 1 will return a gradient of zero since they will be cast to an int as well.

```python
        if (isinstance(position, list) or position.ndim == 1):
            gradient = np.array([0,0,0])   # <------ this had dtype=int
            if (position[2] - self.z_air_boundary) <= 0:
                gradient[2] = gradient_z(position[2])  # <------ this gets cast to an int
```

fschlueter:
…nly occurred when passing single position (1-dim array)

Addressing #1061 

The bug only occurs when passing a 1-d array describing a single 3-coordinate position to the function: 
```
n_grad = ice_simple.get_gradient_of_index_of_refraction(pos)  # <--- This was always fine
n_grad2 = np.array([ice_simple.get_gradient_of_index_of_refraction(np.array([0, 0, z_i])) for z_i in z])  # <--- here was the bug
```
 
This is the plots before and after fixing the issue:
image
image

Thanks for the bug report @colemanalan.

I do not see any call of `get_gradient_of_index_of_refraction` inside any raytracing code in NuRadio. Is it safe to assume that this bug had no consequences on physics results? 

In any case I am tagging a few people so that they are aware of this bug: @sjoerd-bouma, @philippwindischhofer, @avijai2000, @mcb28

### DIFF
--- a/NuRadioMC/utilities/medium_base.py
+++ b/NuRadioMC/utilities/medium_base.py
@@ -230,10 +230,6 @@ class IceModelSimple(IceModel):
 
         Parameters
         ----------
-        z_air_boundary: float, NuRadio length units
-            z coordinate of the surface of the glacier
-        z_bottom: float, NuRadio length units
-            z coordinate of the bedrock/bottom of the glacier.
         n_ice: float, dimensionless
             refractive index of the deep bulk ice
         delta_n: float, NuRadio length units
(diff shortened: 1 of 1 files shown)


sjoerd-bouma (approved):
`np.zeros(3)` (with optional type specification) would have been my choice but this is fine. Thanks @colemanalan and @fschlueter!

I'm not aware of any code that would be impacted by this. Might be worth putting it in the changelog as a bug fix anyway?

PR #1066, opened 2026-03-10
Remove (almost) redundant and less precise distance cut

fschlueter:
Apparently the distance cut is applied twice. The first time between every shower and the stations barycenter (including some padding) and a second time between each shower and channel. The first time this cut is calculated it is commented out. This PR reenables this cut. If not need because to inaccurate and we have the second stage in which we apply the cut, we should remove the whole first part. 

@cg-laser you commented the line in question out in this [commit](https://github.com/nu-radio/NuRadioMC/commit/6293e6eb0b4c895cdc8960c8dea94e73acb1f50e). Was this intentional? 

Issue spotted by @Gursharan2803

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -1509,6 +1509,7 @@ class simulation:
                             emitters=event_group.get_sim_emitters(),
                             station_id=station_id, channel_id=channel_id,
                             min_efield_amplitude=min_amplitude,
+                            rnd=self._rnd, antenna_pattern_provider=self._antenna_pattern_provider,
                             **sim_efield_kwargs)
 
                     if sim_station.is_candidate():
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -1448,9 +1448,8 @@ class simulation:
         # calculate bary centers of station
         station_barycenter = np.zeros((len(self._station_ids), 3))
         for iSt, station_id in enumerate(self._station_ids):
-            pos = []
-            for channel_id in self._det.get_channel_ids(station_id):
-                pos.append(self._det.get_relative_position(station_id, channel_id))
+            pos = [self._det.get_relative_position(station_id, channel_id)
+                for channel_id in self._det.get_channel_ids(station_id)]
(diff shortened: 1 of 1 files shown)


cg-laser:
don't we have the potential problem for secondary simulations: Here we take the position of the initial interaction (if I recall correctly). So me might exclude far away secondary showers that might produce a visible signal?

fschlueter:
If I read the code correctly:
```
                    vertex_distances_to_station = np.linalg.norm(vertex_positions - station_barycenter[iSt], axis=1)
                    distance_cut = self._get_distance_cut(np.sum(shower_energies)) + 100 * units.m  # 100m safety margin is added to account for extent of station around bary center.
                    if vertex_distances_to_station.min() > distance_cut:
```
it calculates the distance between station and each `vertex_positions` and checks if the closest is close enough. So I think its correct but I have not tested it extensively. I am also fine with removing this stage all together.

cg-laser:
to be honest, I don't remember why I removed it. It is probably best, to just remove it. I probably had a good reason for not implementing it again. It is also unclear how much computing time can actually be saved. 
So I would vote for removing this option.

fschlueter:
Okay, deleted it. This actually saved quite some lines!

sjoerd-bouma:
- maybe a simple benchmark would be nice, just to verify that in a typical case this does not make a big difference (I would be surprised if it does); I'm also fine to just take the run time of the CI as a rough indicator for this (broken CI is fixed by #1070 ).
- Please remove the `distance_cut` related options from the `config_default.yaml` file in `NuRadioMC/simulation` also.

cg-laser:
This feature is only really useful for secondary and multi-station simulations where you often have many many stations that are far away from the shower. So benchmarking this is not so easy. We should definitively keep the "shower to atnenna" distance cut.

sjoerd-bouma:
> This feature is only really useful for secondary and multi-station simulations where you often have many many stations that are far away from the shower. So benchmarking this is not so easy. We should definitively keep the "shower to atnenna" distance cut.

Got it, thanks, in that case please ignore my previous comment, I had somehow missed that the shower to antenna distance cut is still in.

fschlueter:
@cg-laser @sjoerd-bouma someone please approve.

sjoerd-bouma (approved):
Thanks Felix!

PR #1067, opened 2026-03-10
Fix simulation of non-trigger channels when simulating secondaries

fschlueter:
Before the signals of non-trigger channels were only added to the station of the last event (when `group_into_events` created several events).

Thanks @Gursharan2803 for spotting it!

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -1642,16 +1642,18 @@ class simulation:
                             event_time=self._evt_time,
                             detector_simulation_part1=self.detector_simulation_part1)
 
-                        logger.debug(f"Adding sim_station to station {station_id} for event group "
-                                     f"{event_group.get_run_number()}, channel {channel_id}")
-                        station.add_sim_station(sim_station)  # this will add the channels and efields to the existing sim_station object
-
                         # The non-triggered channels were simulated using the efieldToVoltageConverterPerEfield
                         # (notice the "PerEfield" in the name). This means that each electric field was converted to
                         # a sim channel. Now we still have to add together all sim channels associated with one "physical"
                         # channel. Furthermore we have to cut out the correct readout window. For the trigger channels
                         # this is done with the channelReadoutWindowCutter, here we have to do it manually.
                         for evt in output_buffer[station_id].values():
+                            station = evt.get_station()
+
+                            logger.debug(f"Adding sim_station to station {station_id} for event group "
+                                        f"{event_group.get_run_number()}, channel {channel_id}")
+                            station.add_sim_station(sim_station)  # this will add the channels and efields to the existing sim_station object
+
                             for sim_channel in sim_station.get_channels_by_channel_id(channel_id):
                                 if not station.has_channel(sim_channel.get_id()):
                                     # For each physical channel we first create a "empty" trace (all zeros)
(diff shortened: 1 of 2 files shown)


anelles (approved):
Good catch! It is hard to test whether this breaks our logic anywhere else, but at least it reads like it shouldn't. So I will trust you on this one!

NuRadioMC/simulation/simulation.py
```diff
@@ -1642,16 +1642,18 @@ def run(self):
...
                         # channel. Furthermore we have to cut out the correct readout window. For the trigger channels
                         # this is done with the channelReadoutWindowCutter, here we have to do it manually.
                         for evt in output_buffer[station_id].values():
+                            station = evt.get_station()
+
+                            logger.debug(f"Adding sim_station to station {station_id} for event group "
+                                        f"{event_group.get_run_number()}, channel {channel_id}")
+                            station.add_sim_station(sim_station)  # this will add the channels and efields to the existing sim_station object
+
                             for sim_channel in sim_station.get_channels_by_channel_id(channel_id):
```
sjoerd-bouma, line 1657: If I understand correctly, this means that the sim_station non-trigger channels will contain traces from all showers for each event, not just the showers which are contained in this particular 'sub-event'. This is different from the trigger channels (`group_into_events` only stores sim_channels / showers associated with the sub-event in each sim_station) and also results in the duplication of some data.

I think we should first check which sim_showers are stored in the existing event, and only add sim_channels which originate from those sim_showers. I'm wondering if there are edge cases where this goes wrong (i.e. if there is a scenario in which the trigger sim_channel from shower 1 belongs to sub-event 1 but a non-trigger channel from the same shower instead would be seen in sub-event 2), but I think this would require very large physical / temporal separation between the trigger- and non-trigger channel. Alternatively, you could try to factorize the required logic from the `group_into_events` function and use that also to determine which bits of the non-triggered channels need to go into which sub-event sim_station.

fschlueter: Keep in mind that you could trigger on a direct ray and record from the same shower a reflected ray in a non-trigger channel, I think that could easily create a time difference of 1 micro second which is the value used for the Gen2 simulations.

Hence, we either implement a proper logic also for the non-trigger channels or we keep it as it?

Do you think there is any potential bug in how it is currently implemented?

cg-laser: If I see it correctly, we really had a bug before, because in the line
`station.add_sim_station(sim_station)` 
the `station` object was the station of the last event (as it was defined in line 1580. So Felix's change should restore the intended behaviour.
Shall we discuss it in the call tomorrow?

sjoerd-bouma: No, I agree that the previous version had a bug, which should be fixed by this change.

My only issue is that the logic it uses to determine which sim_channels to include is different from that for the trigger channels, which will likely result in some data duplication (the same efields/sim_channels will be stored in multiple events). This is not a 'bug' per se, and we can decide to live with that for now, but I think it probably makes more sense to try to do something similar to the group_into_events function also with the non-trigger channels?

fschlueter:
Sjoerd and I discussed this PR. While this PR fixes the obvious bug that non-trigger channels were not added to the appropriate event in case the event_group contained more than one event, it handles the `sim_channels` in an un-ideal way: adding all `sim_channel` objects to all events, also those where the signal of the `sim_channel` is out of the time window of the particular event. This can cause confusion when inspecting the `sim_channels` but is not really a bug. Furthermore, implementing a smarter splitting of the `sim_channels` is not an easy task.

PR #1069, opened 2026-03-12
Skip Git version check in release versions

### LINKED ISSUE #1068: Missing git hash for release versions prints warnings
I am not entirely sure if this is an error on my side, but when working with a release version of NuRadioMC, I always get a warning due to Git not finding a repository to read a commit hash from. 
```
$ python3
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import NuRadioReco.utilities.version
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
>>>
```

The relevant call stack in my specific case is

```python
from NuRadioReco.modules.channelBandPassFilter import channelBandPassFilter
↓
import NuRadioReco.framework.event
↓
from NuRadioReco.utilities import io_utilities, version
```

As per #924, the herein-imported `version.py` tries to read the git hash upon loading.

But if I'm not mistaken, this will never work with a release version – the pip-installed package does not come with a `.git` directory.

Upon researching this, I also found that some code tries to warn the user if no hash is specified but will never actually do so, as all exceptions are caught before in `version.py`:
https://github.com/nu-radio/NuRadioMC/blob/5d309bfdf052dec00413d2657aebad10e3a3fe36/NuRadioReco/framework/event.py#L691-L696

Here are my suggestions of what we could do:
- check if we are in a release version before trying to access the hash (speedup!)
- bake in the hash somewhere in the release pipeline
- silence the error by redirecting Git's stderr

And we _should_ do the following:
- make the hash parsing more reliable by using the porcelain `git rev-parse HEAD` command instead of `git status`
- fix the defunct warning mentioned above, either by removing or rewriting it

I'll gladly create a PR, but we first need to decide on how we intend to handle the hashes.

NicoWeio:
Fixes #1068.

I have yet to test this and will update this PR once I do.

### DIFF
--- a/NuRadioReco/utilities/version.py
+++ b/NuRadioReco/utilities/version.py
@@ -10,7 +10,7 @@ logger = logging.getLogger("NuRadioReco.utilities.version")
 
 def get_git_commit_hash(module) -> str:
     try:
-        if not module.__version__.endswith('-dev'):
+        if 'dev' not in module.__version__:
             # distribution versions won't have a Git repository;
             # eagerly return to not waste time
             return "none"
--- a/NuRadioReco/utilities/version.py
+++ b/NuRadioReco/utilities/version.py
@@ -10,7 +10,7 @@ logger = logging.getLogger("NuRadioReco.utilities.version")
 
 def get_git_commit_hash(module) -> str:
     try:
-        if 'dev' not in module.__version__:
+        if not module._IS_DEV_VERSION:
             # distribution versions won't have a Git repository;
             # eagerly return to not waste time
             return "none"
--- a/NuRadioReco/utilities/version.py
+++ b/NuRadioReco/utilities/version.py
@@ -8,34 +8,40 @@ import logging
 logger = logging.getLogger("NuRadioReco.utilities.version")
 
 
-def get_git_commit_hash(path):
+def get_git_commit_hash(module) -> str:
     try:
-        gitproc = Popen(['git', 'show'], stdout=PIPE, cwd=path)
-        (stdout, stderr) = gitproc.communicate()
-        h = stdout.decode('utf-8').split('\n')[0].split()[1]
-        check = re.compile(r"^[a-f0-9]{40}(:.+)?$", re.IGNORECASE)
-        if(not check.match(h)):
-            logger.error("NuRadioMC version could not be determined, returning None")
+        if not module.__version__.endswith('-dev'):
+            # distribution versions won't have a Git repository;
+            # eagerly return to not waste time
             return "none"
+
+        path = os.path.dirname(module.__file__)
+
+        gitproc = Popen(['git', 'rev-parse', 'HEAD'], stdout=PIPE, cwd=path)
+        (stdout, stderr) = gitproc.communicate()
+        h = stdout.decode('utf-8').strip()
+        check = re.compile(r"^[a-f0-9]{40}(:.+)?$", re.IGNORECASE)
+        assert check.match(h)
     except:
+        logger.error("NuRadioMC/Reco version could not be determined, returning None")
         return "none"
     return h
 
-path = os.path.dirname(NuRadioMC.__file__)
-NuRadioMC_hash = get_git_commit_hash(path)
+
+# preload in case the source changes during execution
+_NuRadioMC_hash = get_git_commit_hash(NuRadioMC)
+_NuRadioReco_hash = get_git_commit_hash(NuRadioReco)
 
 
-def get_NuRadioMC_commit_hash():
+def get_NuRadioMC_commit_hash() -> str:
     """
     returns the hash of the current commit of the NuRadioMC git repository
     """
-    return NuRadioMC_hash
+    return _NuRadioMC_hash
 
-path = os.path.dirname(NuRadioReco.__file__)
-NuRadioReco_hash = get_git_commit_hash(path)
 
-def get_NuRadioReco_commit_hash():
+def get_NuRadioReco_commit_hash() -> str:
     """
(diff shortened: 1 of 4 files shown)


NuRadioReco/utilities/version.py
```diff
@@ -8,34 +8,40 @@
...
 
 
-def get_git_commit_hash(path):
+def get_git_commit_hash(module) -> str:
     try:
-        gitproc = Popen(['git', 'show'], stdout=PIPE, cwd=path)
+        if not module.__version__.endswith('-dev'):
+            # distribution versions won't have a Git repository;
+            # eagerly return to not waste time
+            return "none"
```
fschlueter, line 16: Why not returning `None`

fschlueter:
I thought the idea is to return something like `NuRadioMC.__version__` when no hash can be inferred? @sjoerd-bouma do you know why we do it separately for Reco and MC? Is that from the days they were independent repositories? 
should we ditch one of them?

NuRadioReco/utilities/version.py
NicoWeio: We'd first need to verify that all consumers can handle a `None` value. I can do that, of course, but assumed that this shouldn't be touched.

sjoerd-bouma:
The fix looks good to me, but you seem to have discovered an (independent) issue in how our CI works? I will try to have a look at that today.

> I thought the idea is to return something like `NuRadioMC.__version__` when no hash can be inferred? @sjoerd-bouma do you know why we do it separately for Reco and MC? Is that from the days they were independent repositories? should we ditch one of them?

The [NuRadioReco hash](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.framework.parameters.html#NuRadioReco.framework.parameters.eventParameters.hash_NuRadioReco) has indeed been deprecated for some time now, and I think we don't fill it anymore; we can decide to deprecate the function in `version.py` also, but I don't think it's a problem it returns the same hash as for NuRadioMC.

I'm not sure how I feel about storing the version in a parameter called `commit_hash`. For the [generatorAttributes](https://nu-radio.github.io/NuRadioMC/NuRadioReco/apidoc/NuRadioReco.framework.parameters.html#NuRadioReco.framework.parameters.generatorAttributes.NuRadioMC_version) we store both the version and the hash separately - maybe we want to do this in general also in eventParameters (i.e. add an `eventParameter.NuRadioMC_version` parameter)?

NicoWeio:
Thanks for your feedback! I'll leave this as is until we (i.e., you) reach a decision and debug the flaky CI. :)

NuRadioReco/utilities/version.py
```diff
@@ -8,34 +8,40 @@
 logger = logging.getLogger("NuRadioReco.utilities.version")
 
 
-def get_git_commit_hash(path):
+def get_git_commit_hash(module) -> str:
     try:
-        gitproc = Popen(['git', 'show'], stdout=PIPE, cwd=path)
+        if 'dev' not in module.__version__:
```
sjoerd-bouma: This puts an implicit requirement on future versions to use this naming convention for development version (which we have stuck to so far, but we should probably document this explicitly).

It will also fail if I happen to have checked out a release version. I don't know if that's a problem, as release versions should by definition be identical between the PyPI-released and Github versions. But seeing as this function should only get called once anyway, it shouldn't waste that much time typically?

NicoWeio: Good point. Unfortunately, I cannot think of a better way (except for toying with the release pipeline), so documenting the naming scheme might indeed be the best course of action. It comes down to how much we want to have hashes _all the time_.

I didn't benchmark it, but the delay feels like 1-2 seconds for me, likely because a) my Python environment is network-mounted and b) Git tries to search up to file system boundaries to find a repository.

sjoerd-bouma:
Hi Nico, sorry for the slow feedback. I've since realised there is essentially already another place that checks for git / pip source in the top-level `__init__.py` of NuRadioReco and NuRadioMC, where the version number is attempted to be read from the `pyproject.toml` file, with fallback to importlib_metadata (= the version reported by `pip`). Maybe you can just introduce a constant `_IS_DEV_VERSION` (or similar) there, and use this instead of the `module.__version__.endswith('-dev')` check above?

fschlueter:
hey the two of you are more deep into this. It seems Sjoerd had another suggestions. @NicoWeio Do you want to implement that or are both of you fine to merge it as it?

NicoWeio:
I'm fine with @sjoerd-bouma's suggestion but unsure about the details:
- How do we make sure that changing the `_IS_DEV_VERSION` constant for a release is not forgotten? Do we add a step to CI that changes/checks it? I'm not too familiar with publishing on PyPA; maybe this needs to be committed as well? Or is there some release checklist?
- Do we then duplicate the constant for NuRadioMC/Reco?
- In `version.py`, everything but the `if 'dev' not in module.__version__` is fine as it is?

sjoerd-bouma (approved):
Thanks @NicoWeio, I've for now decided to implement it in the same way as we do it for the version number; there are also some arguments to do it differently but I think this should hopefully be good enough in almost all cases. Thanks for pointing out the issue and fixing it, and sorry about the slow responses, feel free to merge this!

sjoerd-bouma:
It's been a month - unless there are any objections by @NicoWeio or @fschlueter I think we can just merge this?

NicoWeio:
Sorry, I was meaning to have another thorough look ever since ARENA. But I think it's fine as it is. 👍🏻

PR #1070, opened 2026-03-13
Update GitHub actions & fix Python environment caching

sjoerd-bouma:
Three minor fixes/updates:
- The caching of the Python environment would occasionally fail. It was possible for the `setup` action to set up a Python environment on a different Ubuntu version than used in subsequent actions, because `ubuntu-latest` can correspond to different Ubuntu versions. But caches (by default) are different across different OS versions (even if only differing by a minor update), which would result in a cache miss on some of the unit tests, which then would not have NuRadioMC installed. I fixed this by including a step to install NuRadioMC (`pip install -e .`) in case of a cache miss.
- I updated the marketplace github actions to the latest versions, as the versions we were using will be deprecated soon and stop working;
- I included the antenna models in the CI cache. If we're going to constantly request the same files from a server, it might as well be Microsoft's rather than DESY's.

### DIFF
--- a/.github/actions/cache-ci-data/action.yml
+++ b/.github/actions/cache-ci-data/action.yml
@@ -10,7 +10,7 @@ runs:
   using: 'composite'
   steps:
     - name: Cache CI data
-      uses: actions/cache@v4
+      uses: actions/cache@v5
       with:
         path: |
           NuRadioReco/examples/Interactive/data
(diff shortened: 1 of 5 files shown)


anelles (approved):
I appreciate that DESY is not becoming Microsoft. :)
I guess the workflows basically review themselves, still all changes look good to me.

PR #1071, opened 2026-03-16
[RNO-G glitch detector] ignore first block

philippwindischhofer:
Small change to also ignore the first block, as discussed with @fschlueter. The last block was already ignored.

Verified that this leads to insignificant changes in the glitch detector output, shown below for station 11, run 683.

Before this change:

station11_run683_original

After this change:

station11_run683_modified


anelles (approved):
Simple changes, looks good to me.
(Physics wise, it make me more skeptical about the boards, but that is a different story).

fschlueter (approved):

PR #1072, opened 2026-03-23
[LOFAR] Setting array element with a sequence

sjoerd-bouma:
Setting an array element with a sequence has been deprecated in numpy for a while, and has started raising an error with numpy 2.4. For some reason, it seems the "DIPOLE_CALIBRATION_DELAY_VALUE" in the raw TBB files is an array with shape (1,) instead of a single float, so this raises an error (see https://github.com/nu-radio/NuRadioMC/actions/runs/23240232748/job/67555126162). This simply reads out the first entry to fix that.

Maybe someone who's looked at more LOFAR data can confirm if this works consistently for older/newer lofar files also (i.e. that this "DIPOLE_CALIBRATION_VALUE" is always an array and never a float). But on the example file this seems to fix the issue (see https://github.com/nu-radio/NuRadioMC/actions/runs/23447696208/job/68216002463).


karenterveer:
The fix works fine both for older and newer LOFAR files.

karenterveer (approved):

galaxyvita:
I have never use LOFAR data, so i will just approve if the others agree.

PR #1073, opened 2026-03-24
Fix ModDetector - remnant from replacing signal_chain dict to list

fschlueter:
This fix enables (again) to change the cable delay of our detector. Remnant from when we changed the data type of the signal chain "collection" from a dict to a list.


Gursharan2803 (approved):
It's working fine now.

PR #1075, opened 2026-03-25
Only skip direction/pixel for in-ice channels in case one simulates b…

fschlueter:
…oth, in-ice and in-air channels

The previous if condition: `if n_ice != n_air: # consider signal reflection at ice surface` was always true.

### DIFF
--- a/NuRadioReco/modules/channelGalacticNoiseAdder.py
+++ b/NuRadioReco/modules/channelGalacticNoiseAdder.py
@@ -115,8 +115,8 @@ class channelGalacticNoiseAdder:
             by a lot. If the frequencies of the channels change, the cache is cleared.
         scaling: float, default: 1.0
             Scaling factor for the noise. This is useful when doing interferometry with extremely large arrays
-            such as SKA-low. For such an array it is very expensive to simulate/interpolate/process all antennas. 
-            Instead, one can use every nth antenna and scale the noise by a factor of 1/\sqrt{n} (since the SNR 
+            such as SKA-low. For such an array it is very expensive to simulate/interpolate/process all antennas.
+            Instead, one can use every nth antenna and scale the noise by a factor of 1/\sqrt{n} (since the SNR
             is expected to scale with the square root of the number of antennas when using interferomtery/beamforming).
         """
         if debug:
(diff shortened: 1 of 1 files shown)


sjoerd-bouma:
So here you're being bitten by the badly named `ice` module - for land-based detectors it always returns the refractive index of air, so `n_ice != n_air` is not always true. In fact, your new condition will cause problems for LOFAR, because the z-coordinate of some channels is slightly negative, and they're definitely not in ice!

fschlueter:
> So here you're being bitten by the badly named `ice` module - for land-based detectors it always returns the refractive index of air, so `n_ice != n_air` is not always true. In fact, your new condition will cause problems for LOFAR, because the z-coordinate of some channels is slightly negative, and they're definitely not in ice!

Yes you are right, I noticed that while working on #1076 too. Okay but the code would still produce a bug for an experiment like RNO-G with in-air antennas (for those certain zenith angle would not be simulated). The fact that some lofar antennas have a negative z-coordinate is a problem. What do you suggest?

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -315,17 +319,11 @@ def run(
             if zenith > 90. * units.deg:
                 continue
 
-            if n_ice != n_air: # consider signal reflection at ice surface
+            if any_in_ice_channel:
```
sjoerd-bouma, line 322: how about `if n_ice != n_air and any_in_ice_channel`?

sjoerd-bouma: though I guess this is actually fine, just slightly inefficient, because we will just get the same result as in the (now removed) `else` case below.

sjoerd-bouma (approved):
Took me a while to see which bug you had fixed, and I think the change from `n_ice != n_air` to `any_in_ice_channel` does not fix any issues, but it actually shouldn't introduce any new ones either, so your call if you want to revert that or not.

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -358,6 +356,9 @@ def run(
                     curr_fresnel_zenith = zenith
                     curr_n = n_air
 
+                if curr_fresnel_zenith is None:
+                    continue
+
                 antenna_pattern = self.__antenna_pattern_provider.load_antenna_pattern(
```
sjoerd-bouma, line 362: So I thought this was the 'fix', in case a detector has both in-ice and in-air channels, which previously would have resulted in in-air channels incorrectly being skipped if there is no ray path to the in-ice channels. However, I've just realized that this condition will never be true anyway, because n_ice >= n_air.

fschlueter: Okay, I agree. Why was the check 
```
                if curr_fresnel_zenith is None:
                    continue
```
there is the first place?

sjoerd-bouma:
Okay, so it looks like in the end I don't think this PR fixes anything, but it also doesn't break anything. :upside_down_face:

NuRadioReco/modules/channelGalacticNoiseAdder.py
sjoerd-bouma: TIR is scary and you can never be too careful? Maybe in the future we will want to use this module to simulate galactic emission coming from inside the ice to a detector built in the sky? Those are my leading theories.

fschlueter:
Lets close it and see what we adopt in #1076?

PR #1076, opened 2026-03-25
Galactic noise adder improve sim for deep antennas

fschlueter:
Based on #1075. This PR improves the simulation of the galactic noise in deep in-ice antennas. It takes a depth dependent refractive index to calculate the change in signal direction and implements a rough simulation of the signal attenuation.

### DIFF
--- a/NuRadioReco/modules/channelGalacticNoiseAdder.py
+++ b/NuRadioReco/modules/channelGalacticNoiseAdder.py
@@ -115,8 +115,8 @@ class channelGalacticNoiseAdder:
             by a lot. If the frequencies of the channels change, the cache is cleared.
         scaling: float, default: 1.0
             Scaling factor for the noise. This is useful when doing interferometry with extremely large arrays
-            such as SKA-low. For such an array it is very expensive to simulate/interpolate/process all antennas. 
-            Instead, one can use every nth antenna and scale the noise by a factor of 1/\sqrt{n} (since the SNR 
+            such as SKA-low. For such an array it is very expensive to simulate/interpolate/process all antennas.
+            Instead, one can use every nth antenna and scale the noise by a factor of 1/\sqrt{n} (since the SNR
             is expected to scale with the square root of the number of antennas when using interferomtery/beamforming).
         """
         if debug:
@@ -305,8 +305,12 @@ class channelGalacticNoiseAdder:
         n_air = ice.get_refractive_index(depth=1, site=detector.get_site(station.get_id()))
 
         channel_spectra = {}
+        channel_depths = []
         for channel in station.iter_channels(use_channels=selected_channel_ids):
             channel_spectra[channel.get_id()] = channel.get_frequency_spectrum()
+            channel_depths.append(detector.get_relative_position(station.get_id(), channel.get_id())[2])
+
+        any_in_ice_channel = np.any(np.array(channel_depths) < 0)
 
         for i_pixel in range(healpy.pixelfunc.nside2npix(self.__n_side)):
             azimuth = local_coordinates[i_pixel].az.rad
@@ -315,17 +319,11 @@ class channelGalacticNoiseAdder:
             if zenith > 90. * units.deg:
                 continue
 
-            if n_ice != n_air: # consider signal reflection at ice surface
+            if any_in_ice_channel:
                 t_theta = geometryUtilities.get_fresnel_t_p(zenith, n_ice, n_air)
                 t_phi = geometryUtilities.get_fresnel_t_s(zenith, n_ice, n_air)
                 fresnel_zenith = geometryUtilities.get_fresnel_angle(zenith, n_ice, n_air)
-            else: # we are at an in-air site; no refraction
-                t_theta = 1
-                t_phi = 1
-                fresnel_zenith = zenith
 
-            if fresnel_zenith is None:
-                continue
 
             if self.__caching:
                 noise_temperature = self._get_cached_noise_temperature_for_pixel(i_pixel)
@@ -357,6 +355,9 @@ class channelGalacticNoiseAdder:
                     curr_t_phi = 1
                     curr_fresnel_zenith = zenith
                     curr_n = n_air
+
+                if curr_fresnel_zenith is None:
+                    continue
 
                 antenna_pattern = self.__antenna_pattern_provider.load_antenna_pattern(
                     detector.get_antenna_model(station.get_id(), channel.get_id()))
--- a/NuRadioReco/modules/channelGalacticNoiseAdder.py
+++ b/NuRadioReco/modules/channelGalacticNoiseAdder.py
@@ -98,6 +98,7 @@ class channelGalacticNoiseAdder:
             scaling=1.0,
             ice_model=None,
             attenuation_model=None,
+            simulate_sun=False,
     ):
         """
         Set up important parameters for the module
@@ -138,6 +139,8 @@ class channelGalacticNoiseAdder:
             The ice model to use for the simulation of in-ice antennas.
         attenuation_model: str, default: None
             The attenuation model to use for the simulation of in-ice antennas.
+        simulate_sun: bool, default: False
+            If True, the sun's contribution to the galactic noise is simulated. This is experimental!
         """
         if debug:
             warnings.warn("This argument is deprecated and will be removed in future versions.", DeprecationWarning)
@@ -145,6 +148,11 @@ class channelGalacticNoiseAdder:
         self.__random_generator = Generator(Philox(seed))
         self.__n_side = n_side
         self.solid_angle = healpy.pixelfunc.nside2pixarea(self.__n_side, degrees=False)
+
+        self._simulate_sun = simulate_sun
+        if self._simulate_sun:
+            logger.warning("You are simulating the sun's contribution to the galactic noise. "
+                "This is experimental and might not be very accurate! See in-code comments for details.")
 
         self.__caching = caching
         self.scaling = scaling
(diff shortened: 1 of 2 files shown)


fschlueter:
This plot shows the effect of the two new effects simulated (changed signal direction "better refractive index" and, in addition, attenuation effects) for a 100m antenna (channel "0") and 40m deep antenna (channel "7"). In the current situation "old" no difference is made between antennas at different dephs. Both antennas use the same antenna pattern (RNO-G Vpol). Note while the effect are sizable, the overall strength of the signal is relatively small compared to other noise sources such as thermal and electronic noise.

image

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -367,20 +390,40 @@ def run(
...
+                    distance_in_ice = abs(channel_pos[2]) / np.cos(curr_fresnel_zenith)
+
+                    # Since channel_pos[2] < -10, this ensures at least num >= 2
+                    dist_steps = np.linspace(0, distance_in_ice, num=int(distance_in_ice / 10) + 1 )
+                    d_step = dist_steps[1] - dist_steps[0]
+                    depths = channel_pos[2] + np.cos(curr_fresnel_zenith) * (dist_steps[:-1] + d_step / 2)
+
+                    exponent = -d_step / np.array(
+                        [attenuation.get_attenuation_length(d, freqs[passband_filter], self._attenuation_model) for d in depths])
+                    attenuation_factor = np.product(np.exp(exponent), axis=0)
```
sjoerd-bouma: depths is just linspace(channel_depth, 0), no? It feels like you're slightly overcomplicating things here. I suggest calculating the average attenuation length once per channel, and then just using `attenuation_factor = np.exp(-distance_in_ice/average_attenuation_length)`. The average/effective attenuation length is just `np.mean(1/attenuation.get_attenuation_length(np.linspace(channel_depth,-0.01), frequencies, model))`. This is correct in the straight-line approximation you're using (but feel free to double check my maths ;)

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -367,20 +390,40 @@ def run(
...
+
                 if channel_pos[2] < -5:
-                    logger.warning(
-                        "Galactic noise cannot be simulated accurately for deep in-ice channels. "
-                        "Coherence and arrival direction of noise are probably inaccurate.")
+                    if not self.__you_have_been_warned:
+                        logger.warning(
+                            "Galactic noise is not yet simulated most accurately for deep in-ice channels. "
+                            "The phase shift between different antennas are probably inaccurate.")
+                        self.__you_have_been_warned = True
```
sjoerd-bouma: I think with this improvement I'm also happy to move this warning to the docstring (maybe a `Notes` section), which also gives you some more space to describe the current caveats and cleans up the code a little bit.

sjoerd-bouma (commented):
This is a really nice improvement, I think! I suggested an improvement to the attenuation calculation, but other than that (and pending discussion in #1075) I think this can be merged.

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: yes I think you are correct

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: do you actually think using n(z=depth of antenna) to calculate the phase shift makes the calculation more accurate? we could also try to calculate an average (going hand-in-hand with the straight-line approximation in the attenuation calculation)

sjoerd-bouma: No, for coherence you need to be better than half a wavelength, and the straight-line approximation is definitely not good enough for that. For attenuation being off by 10s of meters is fine.

fschlueter:
The last commits did not change the results. Here is another view on it, the averaged and normalized frequency spectra:

image

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -324,12 +336,25 @@ def run(
...
+            # since it assumes a straight line path and not a bend ray.
+            if self._attenuation_model is not None and channel_depth < -10:
+                # Approximate attenuation along a straight line and 10 steps.
+                n = int(abs(channel_depth) // 10) + 1
+                depth_bins = np.linspace(0, channel_depth, num=n)
+                depths = depth_bins[:-1] + np.diff(depth_bins) / 2
+                average_attenuation_length[channel.get_id()] = np.mean(
+                    [attenuation.get_attenuation_length(d, freqs[passband_filter], self._attenuation_model) for d in depths],
+                    axis=0,
+                )
```
sjoerd-bouma: please see my previous comment - for a straight line, you should just take the average over 1/attenuation length, no need to vary the number of bins as the result is exact.

fschlueter: I am not following you, we have to calculate the mean attenuation length for a given antenna deph, averaging over a few steps in depth (independent of the actual distance travelled/zenith) - exactly this is done here.

sjoerd-bouma: the attenuation term is exp - sum_i (distance_i / attenuation_length_i), in the straight-line approximation distance_i is constant over depth and can be taken out of the sum, and the attenuation term can be replaced with its average = 1/N sum_i (1 / attenuation_length_i). The average should therefore be over 1/attenuation_length not over attenuation_length. Difference will be small in most cases.

sjoerd-bouma: I would genuinely just take the expression I suggested in the previous comment and leave the number of steps at the linspace default, I don't think there's anything to be gained with your different step sizes and number of steps etc

fschlueter: the issue with my code has nothing to do with number of steps?

fschlueter: > please see my previous comment - for a straight line, you should just take the average over 1/attenuation length, no need to vary the number of bins as the result is exact.

dont quite understand this statement, with a depth dependent attenuation length it somewhat matters to sample the depth?

fschlueter: I fixed the equation and I changed to a constant n_steps in linespace (however, now the accuracy with which the attenuation length is averaged over is depth dependent.

fschlueter:
@sjoerd-bouma does the latest fix resolve your issue with the calculation?

NuRadioReco/modules/channelGalacticNoiseAdder.py
sjoerd-bouma: Yeah, you could make the number of steps depth-dependent if you want, I just thought it was unnecessary complexity.

fschlueter: fair

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -63,13 +65,25 @@ class channelGalacticNoiseAdder:
...
+    the galactic noise is expected to be very low and will not be detectable without a large number of
+    antennas.
     """
 
     def __init__(self):
         self.__n_side = None
         self.__interpolation_frequencies = None
         self.__radio_sky = None
         self.__noise_temperatures = None
+        self.__you_have_been_warned = False
```
sjoerd-bouma, line 86: This line can go now that we've gotten rid of the warning. Or should we have a warning if we have deep in-ice antennas but no specified ice model? I'm also happy to just make this the user's responsibility to be honest.

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: If you are happy please approve

sjoerd-bouma (dismissed):
Looks good to me - there is one more question mark about focussing (if I believe my own thesis I think there is a missing factor of `sqrt[(n_ice_surface cos(theta_ice_surface) / n cos(theta)]` now, which I think has decent chances of being correct. But I don't have time to think about this too much today.

(edited - equation was wrong)

fschlueter:
> sqrt[n cos(theta) / (n_at_surface cos(theta_at_surface)]

This is the difference including this focusing factor (simple multiplication with the spectrum):
image

fschlueter:
With the latest commit/fix of the calculation of the focusing factor the effect is shown in this plot:
image

fschlueter:
To summarise the (optional) changes introduced with this PR (ignoring the sun): A change in the signal direction at the antenna due to sub-surface refraction, de-focusing due to sub-surface refraction, and attenuation, I am getting the following spectra for RNO-G channels 0 (100m deep Vpol), 7 (40k deep Vpol), 12 (sub-surface downward facing LPDA), and 13 (sub-surface upward facing LPDA) 
image

fschlueter:
@sjoerd-bouma happy with it?

fschlueter:
@sjoerd-bouma what do you think?

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -506,3 +604,52 @@ def get_local_coordinates(coordinates, obs_time, n_side):
...
+
+@functools.lru_cache(maxsize=2)
+def Tb_quiet_sun_paper(kind='linear'):
+    """ Returns function of Tb(f) """
+    
+    # This is data of the quiet sun brightness temperature digitized from Fig. 5, of the paper:
+    # https://iopscience.iop.org/article/10.3847/1538-4357/ac6b37/pdf
+    # First column (i.e., entires data[::2]) is the frequency in MHz,
+    # second column (i.e., entries data[1::2]) is the brightness temperature in K.
+
```
sjoerd-bouma: Can you either change the function names to something sensible (e.g. `brightness_temp_sun`?) and add parameter docstrings, or make one / both of them private? In any case the reference for the data should probably be in a public docstring (either of the public function, or in the `Notes` section of the class above)..

fschlueter:
@sjoerd-bouma last notice before I merge it myself

NuRadioReco/modules/channelGalacticNoiseAdder.py
```diff
@@ -298,40 +329,94 @@ def run(
...
+        one_over_average_attenuation_length = {}
         for channel in station.iter_channels(use_channels=selected_channel_ids):
             channel_spectra[channel.get_id()] = channel.get_frequency_spectrum()
+            channel_depth = detector.get_relative_position(station.get_id(), channel.get_id())[2]
+            channel_depths.append(channel_depth)
+
+            # calculate the signal attenuation for in-ice channels. This is a very rough approximation,
+            # since it assumes a straight line path and not a bend ray.
+            if self._attenuation_model is not None and channel_depth < -10:
+                # Approximate attenuation along a straight line and 10 steps.
```
sjoerd-bouma: This comment is no longer accurate (my fault, sorry!)

sjoerd-bouma (commented):
The addition of the sun is nice! Would be nice if you could tidy up the function names / docstrings for that still a little bit (see suggestion). Otherwise, this is good to merge - sorry for the long wait!

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: it is? its only a good approximation now?

sjoerd-bouma: Sorry, I figured this would show up as a comment for line 379 specifically - I meant the 10 steps thing only.

NuRadioReco/modules/channelGalacticNoiseAdder.py
sjoerd-bouma: I'm happy enough with this. As you are probably aware, functions in Python conventionally start with a lower case character (in contrast to classes), but then as you are definitely aware this is not a convention we adhere to at all in NuRadio, so the capital T can stay.

sjoerd-bouma (dismissed):
Thanks Felix, this is a really nice upgrade. In fact, it's so nice we should probably add a note to the changelog... otherwise let's get this merged!

fschlueter:
done

sjoerd-bouma (approved):

NuRadioReco/modules/channelGalacticNoiseAdder.py
fschlueter: ubs, never read this. I should have obeyed this convention - will fix it in the future.

PR #1077, opened 2026-03-26
Multilayeranalyticraytracing

hanneswarnhofer:
**Overview:** 

The new `multilayeranalyticraytracing.py `module contains functionality for the multi layer expansion of the currently used analyticraytracing module. We change the medium definitions to a more general layer structure that allows to find the analytical solutions for the light path in a medium that can be described as layers of single exponential models, also with discontinuities at the layer boundaries. Air to ice tracing is replaced by modelling the air as one of those exponential layers, giving the possibility to model it in a much more realistic way than just n=1.0. The module follows the same solution finding and minimization steps as the analyticraytracing module. 

Additions to the `SignalProp.propagation` to allow for correct handling of the module are in planed, as well as additions to `utlilities.medium` to include the layer definitions in a central place. 

**Work in progress:** Currently adding the functions to evaluate path specific parameters, such as get_travel_time, get_attenuation, get_launch_angle, etc.

**ToDo:** Adding unit tests, what is a changelog file?


NuRadioMC/SignalProp/propagation.py
```diff
@@ -3,7 +3,8 @@
 solution_types = {
     1: 'direct',
     2: 'refracted',
-    3: 'reflected'
+    3: 'reflected',
+    4: 'from_air'
```
fschlueter: why do we need a 4th solution? In case of an ice-to air raytracing solution we would use `direct`?

fschlueter:
@hanneswarnhofer let me start with saying (again) that is impressive work! I have only superficially scrolled through your changes. As far as I can see you have not implemented a ray-tracing class but only helper functions. Is that still coming or is that using the existing analytic raytracing class? 

Also let me say early on: If it is possible to combine as much code as possible between the existing analytic raytracer and your multi-layer implementation in terms of helper functions, i.e., generalizing existing functions to more layers, without a noticeable loss of efficiency, that would be very nice!

NuRadioMC/SignalProp/propagation.py
hanneswarnhofer: Ahh right, I used this for plotting as a sanity check. In the air-to-ice case we have to skip the part where the ray gets reflected on the surface and this distinctive solution type made it easier to investigate those solutions seperated from the rest. But we can reduce this again to the three solution types to be consistent with the other raytracers.

hanneswarnhofer:
> @hanneswarnhofer let me start with saying (again) that is impressive work! I have only superficially scrolled through your changes. As far as I can see you have not implemented a ray-tracing class but only helper functions. Is that still coming or is that using the existing analytic raytracing class?

@fschlueter Thanks a lot! And yes so far it is only a set of seperated (helper) functions, but I want to rewrite it so it follows a class structure using the `ray_tracing_base` class, so that we can use it as far as it is possible in the same way as the current analytic raytracer. 
 
> Also let me say early on: If it is possible to combine as much code as possible between the existing analytic raytracer and your multi-layer implementation in terms of helper functions, i.e., generalizing existing functions to more layers, without a noticeable loss of efficiency, that would be very nice!

I am trying to do this as much as possible where it makes sense and where the multilayer structure still allows to follow the same concept but generalized to more layers. However, it seems that in most cases it is not so simple to expand the single layer case to the multi layer logic. I am currently working on the `path_length_analytic` function for example, where I need to calculate the integral within each layer segment starting from the emitter and going layer by layer up to the turning point and then down again layer by layer to the reciever. So I will follow the same logic as done in the old version (finding segments -> turning downwards going segments into upwards one, so the integral can be solved -> calculate path length integral in each segment -> get overall path length) but with more segments and evaluating the integral with the parameters of the layers we are in...  soo I am using the same concepts and am trying to reuse code where I can, but often I still have to rewrite the functions to the largest parts I guess.

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,1355 @@
...
+* classify solutions (direct, refracted, reflected)
+
+...if they are in our nice layer format that expects smooth boundaries of the n(z) definition that we apply
+
+The implementation is an expansion of the previous analytic ray
+tracing solver used in NuRadioMC . In order to enable compilation with
+``numba.njit(nopython=True)``, the layer definitions are internally
+converted from dictionary-based objects to arrays. 
+For more insight into the physics behind this approach and remarks on the solving strategy it is recommended 
+to have a look at the companion note to this module ``MultilayerAnalyticRayTracting.md`` and the appendix C of
```
sjoerd-bouma: You didn't add this note; I also think it would make more sense to stick to rst, either as part of this docstring, or if it's very long, we can add it to one of the 'manuals' in the documentation (https://nu-radio.github.io/NuRadioMC/NuRadioMC/pages/manuals.html). But we can discuss this in more detail later.

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,1355 @@
...
+
+region_name : str
+    Human-readable name of the region.
+
+Internally these definitions are converted to arrays using
+:func:`layers_to_arrays` in order to support Numba compilation.
+
+Contact: hannes.warnhofer@desy.de
+Created 2026
+Developed for use in NuRadioMC ray tracing.
```
sjoerd-bouma: Feel free to keep this, but git already does keep track of all of this for us : )

NuRadioMC/utilities/medium.py
```diff
@@ -369,3 +369,64 @@ def get_ice_model(name):
...
+        return self._scalarfield.getValue(position)
+    def get_average_index_of_refraction(self, position1, position2):
+        position1 = RP.Vector3d(*(position1 / units.meter))
+        position2 = RP.Vector3d(*(position2 / units.meter))
+        return self._scalarfield.getAverageValue(position1, position2)
+    def get_gradient_of_index_of_refraction(self, position):
+        pos = RP.Vector3d(*(position / units.meter))
+        return self._scalarfield.getGradient(pos) / units.meter
+    def _compute_default_ice_model_radiopropa(self):
+        return medium_base.RadioPropaIceWrapper(self, self._scalarfield)
```
sjoerd-bouma: see comments on #1037, I guess, though you can fight it out with @avijai2000 who actually gets to address them :)

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,1355 @@
...
+        z1, z2 = z2, z1
+        downgoing = True
+
+    n_deep = n_ice[-1]
+
+    ## Here something is still wrong
+    ## theta skim goes to inf for too horizontal geometries when z1 is on the same height as z2.
+
+    theta_straight = np.arctan((z2-z1)/(y2-y1))
+    #print(f"theta_straight: {theta_straight}")
```
sjoerd-bouma: once this is in a more finished state, it would be good to remove these commented out lines

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,1355 @@
...
+        layers
+    )
+    #print(f"C0straight: {C0straight}")
+    
+    logC0straight = np.log(max(C0straight - 1./n_deep, 1e-12))
+    logC0skim = np.log(max(C0skim - 1./n_deep, 1e-12))
+    #print(f"logC0skim: {logC0skim}")
+    print(f"-------------------------------------------------------")
+    print(f"Original x1 and x2: {x1} and {x2}. With air: {with_air}. Downgoing: {downgoing}")
+    print(f"Searching for Ray-Tracing Solutions from x1 ({y1},{z1}) to x2 ({y2},{z2})...")
```
sjoerd-bouma: debug information is useful, but please use a logger rather than print statements, i.e.
```
import logging
logger = logging.getLogger("NuRadioMC.SignalProp")

logger.info("logging message") # or logger.status / logger.debug / logger.warning, as appropriate
```

sjoerd-bouma (commented):
This is great! To figure out what's still missing, have a look at the 'propagation_base_class.py` in `SignalProp` - this is the interface that your raytracer will also need to have. As Felix mentioned, the more code you can re-use from `analyticraytracing`, the better, but I trust you're deep enough in the code by now you can make your own calls there. Once you have the same interface, it will be very straightforward to use this code (by just changing one line in the simulation config) & add it to the automatic tests.

I'm marking the PR as draft for now, feel free to ping us / mark as ready for review once you want some more input.

NuRadioMC/utilities/medium.py
hanneswarnhofer: I removed it again. I only copied and slightly adapted this from Aishwarya's PR to be able to compare my stuff to radipropa.

hanneswarnhofer:
Hello again :) 

I now implemented all the functionality needed for using the new ray tracer in full scale, including all the (important) functionality that is provided by the `analyticraytracing.ray_tracing_2D` class, except the birefringence functions. The idea is to provide all the functions so that they are named and can be called in the same way as for the currently used raytracer, allowing to potentially reuse the whole `ray_tracing` class (so the 3D wrapper around the 2D analytic raytracer) here. For this we would still need to change this class, so that we can also specify `self._r2d = mulan_ray_tracing_2D`. However, this would be something where I would like your opinion on how we should do that exactly. 

I also modified the `medium_base` and added a correpsonding layer oriented medium class, and defined the corresponding ice models in `medium`. Here I would also be happy about some input from your side on how we should do this.

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
+"""
+Ray tracing module for layered exponential refractive index profiles.
+
+This module implements analytic ray tracing in glacial ice or another medium where the
+refractive index can be described as layers of exponentials where each layer follows 
+
+n(z) = n_ice - delta_n * exp(z / z_0)
+
+This also suppports propagation across layer boundaries where n(z) is not continuous, so also air-to-ice tracing.
+We can also model the refractive index of air as an exponential layer of course.
```
fschlueter: we can or we have to (i.e. is the combination of exponential layers + constant layer allowed)? Dont get me wrong, I am happy to model the atmosphere with an exponential layer (or actually 5 as common in air-shower physics) but its good to be precise.

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
...
+To find the ray solutions we needed to implement a number of functions
+that you can use to :
+
+* evaluate refractive index profiles
+* compute analytic ray trajectories
+* determine turning points
+* solve for ray parameters connecting two points
+* classify solutions (direct, refracted, reflected)
+
+...if they are in our nice layer format that expects smooth boundaries of the n(z) definition that we apply
```
fschlueter: what does that mean?

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
...
+
+
+Examples
+--------
+Typical workflow:
+
+1. Define the ice layers (e.g. ``LAYERS``).
+2. Call :func:`find_solutions` to determine valid ray parameters between
+   two points.
+3. For each solution, compute the ray path using :func:`get_path`.
```
fschlueter: Great that you thought about this already but it would be important that you include a dedicated example script which defines the layers and runs the raytracing for a set of source/receiver pairs

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
...
+
+Notes
+-----
+Coordinates are given as (y, z) with units of meters:
+
+* y : horizontal distance
+* z : depth (negative downward)
```
fschlueter: strictly speaking if z is pointing downwards for negative values its not the depth but the z-coordinate. Maybe we should try implementing this convention so everyone knows exactly how a parameter is to interpret by its name (depth vs z)

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
...
+
+z = 0 corresponds to the surface.
+
+The ray parameter ``C0`` represents the **inverse horizontal slowness**
+of the ray and determines the curvature of the trajectory.
+It can also be defined as C0 = 1/(n(z)*sin(theta)) where theta is the angle relative to the horizontal.
```
fschlueter: elevation angle?

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
...
+    Lower depth boundary of the layer (m).
+
+z_max : float
+    Upper depth boundary of the layer.
+
+n_ice : float
+    Asymptotic refractive index of deep ice.
+
+delta_n : float
+    Surface-to-deep refractive index contrast.
```
fschlueter: can you explain that better?

NuRadioMC/utilities/medium_base.py
```diff
@@ -830,3 +831,97 @@ def getGradient(self, position): #name may not be changed because linked to c++
             pos = np.array([position.x, position.y, position.z]) / nu2rp_meter
             gradient = self.__ice_model_nuradio.get_gradient_of_index_of_refraction(pos) / nu2rp_meter
             return RP.Vector3d(*gradient)
+
+class IceModelExpLayers(IceModel):
+    """
+    Medium class for defining multilayered media
+    Dedicated to media where the refractive index within each layer can be described by an exponential model. Needed for the usage of the multilayer analytic raytracer.
+    """
+    def __init__(self, layers, z_air_boundary=0.0, z_bottom=None):
```
fschlueter: Please add a doc-string for this function

NuRadioMC/utilities/medium_base.py
```diff
@@ -830,3 +831,97 @@ def getGradient(self, position): #name may not be changed because linked to c++
...
+        super().__init__(z_air_boundary, z_bottom)
+
+        self.layers = layers
```
fschlueter: remove

NuRadioMC/utilities/medium_base.py
```diff
@@ -830,3 +831,97 @@ def getGradient(self, position): #name may not be changed because linked to c++
...
+
+        self.layers = sorted(layers, key=lambda L: L["z_min"])
+        self._validate_layers()
+
+        self._layers_arr = self._layers_to_arrays()
+
+    def _validate_layers(self):
+        for i in range(len(self.layers) - 1):
+            assert np.isclose(
```
fschlueter: I was told (by the internet) to not use `assert` to verify user input. Its only meant to "spot invariants in your code that should never, ever happen once the code is stable and deployed"

fschlueter: so you could use `raise ValueError(...)`

NuRadioMC/utilities/medium_base.py
```diff
@@ -830,3 +831,97 @@ def getGradient(self, position): #name may not be changed because linked to c++
...
+                Exponential scale depth. 
+        """
+        n = len(self.layers)
+        z_min = np.zeros(n)
+        z_max = np.zeros(n)
+        n_ice = np.zeros(n)
+        delta_n = np.zeros(n)
+        z0 = np.zeros(n)
+
+        for i,L in enumerate(self.layers):
```
fschlueter: space (sorry for not controlling my ocd)

NuRadioMC/utilities/medium_base.py
```diff
@@ -830,3 +831,97 @@ def getGradient(self, position): #name may not be changed because linked to c++
...
+            z_min[i] = L["z_min"]
+            z_max[i] = L["z_max"]
+            n_ice[i] = L["n_ice"]
+            delta_n[i] = L["delta_n"]
+            z0[i] = L["z_0"]
+
+        return z_min, z_max, n_ice, delta_n, z0
+
+    def get_index_of_refraction(self, position):
```
fschlueter, line 943: Can you add a one-line doc-string. I know that this is a function which appears in many different version across multiple inheritances

NuRadioMC/utilities/medium_base.py
```diff
@@ -830,3 +831,97 @@ def getGradient(self, position): #name may not be changed because linked to c++
...
+
+        def n_of_z(z):
+            for L in self.layers:
+                if L["z_min"] <= z < L["z_max"]:
+                    return L["n_ice"] - L["delta_n"] * np.exp(z / L["z_0"])
+            return 1.0 # fallback value
```
fschlueter: Would it make sense computationally to use internally arrays also here, not only for functions related to numba? This could make this code much faster. 

Is the fallback justified here or shall we better raise an error?

NuRadioMC/utilities/medium.py
```diff
@@ -369,3 +370,98 @@ def get_ice_model(name):
         raise NotImplementedError('The ice model you are trying to use is not implemented. Please choose another ice model or implement a new one.')
     else:
         return globals()[name]()
+
+
+class greenland_simple_layered(medium_base.IceModelExpLayers):
```
fschlueter, line 462: maybe add a note that this icemodel is only meant as a cross check?

NuRadioMC/utilities/medium.py
```diff
@@ -369,3 +370,98 @@ def get_ice_model(name):
...
+            "region_name": "SingleModel"
+        }]
+
+        super().__init__(
+            layers=layers,
+            z_bottom=-3000.0,
+        )
+
+class greenland_firn_layered(medium_base.IceModelExpLayers):
+    def __init__(self):
```
fschlueter, line 545: Please add a short doc-string describing what the model does: 2 layer model firn + deep ice

NuRadioMC/utilities/medium.py
```diff
@@ -369,3 +370,98 @@ def get_ice_model(name):
...
+            }
+        ]
+
+        super().__init__(
+            layers=layers,
+            z_bottom=-3000.0,
+        )
+
+class greenland_3exp_layered(medium_base.IceModelExpLayers):
+    def __init__(self):
```
fschlueter, line 579: same

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
...
+
+from NuRadioReco.utilities import units, geometryUtilities
+from NuRadioMC.utilities import attenuation as attenuation_util, medium as medium_util
+from NuRadioMC.SignalProp.propagation_base_class import ray_tracing_base
+
+from math import sqrt, log, sin
+
+import logging
+logger = logging.getLogger("NuRadioMC.analytic_ray_tracing")
+
```
fschlueter: the following later definitions are redundant with the definitions in medium.py and can be deleted?

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
...
+        z_min[i] = L["z_min"]
+        z_max[i] = L["z_max"]
+        n_ice[i] = L["n_ice"]
+        delta_n[i] = L["delta_n"]
+        z0[i] = L["z_0"]
+
+    return z_min, z_max, n_ice, delta_n, z0
+
+
+def init_layer_arrays(z_min, z_max, n_ice, delta_n, z0):
```
fschlueter: do you want to add this to the validation function of the icemodel class?

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
...
+    C0 = get_C0_from_log(logC0, n_deep)
+    dy = get_delta_y(C0, y1, z1, y2, z2, layers, (-1., -1.),downgoing,with_air)
+    if not np.isfinite(dy):
+            return 1e30
+    return dy
+
+DIRECT = 1
+REFRACTED = 2
+REFLECTED = 3
+FROM_AIR = 1
```
fschlueter: Please do not redefine them here but use `solution_types_revert` from the other module. Do we need `FROM_AIR`?

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
...
+        if get_segments is True:
+            return y_path, z_path, yb, zb
+        else:
+            return y_path, z_path
+
+
+
+
+UPGOING = 1
+DOWNGOING = 2
```
fschlueter: is that necessary?

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
...
+
+        downgoing = False
+        if z1 > z2:
+            z1, z2 = z2, z1
+            downgoing = True
+
+        return determine_solution_type(y1, z1, y2, z2, C0, layers_arr, downgoing, with_air)
+    
+    def find_solutions(self, x1, x2, plot=False, *_, **__):
+        layers_arr = layers_to_arrays(self.medium)
```
fschlueter: seems like we can store `self._layers_arr` in the class instead of running this function for each function?

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2764 @@
...
+        n1 * n2
+        * abs(np.cos(launch_angle) * np.cos(receive_angle))
+        * (w_theta * w_phi / (s**2))
+        )
+    
+    return 
+
+
+
+class mulan_ray_tracing_2D(ray_tracing_base):
```
fschlueter: I would choose a more descriptive (longer) name, e.g.,: `multi_layer_ray_tracing_2D`

NuRadioMC/SignalProp/propagation.py
```diff
@@ -3,7 +3,8 @@
 solution_types = {
     1: 'direct',
     2: 'refracted',
-    3: 'reflected'
+    3: 'reflected',
+    0: 'from_air'
```
fschlueter: do need that new type?

fschlueter (commented):
Hi @hanneswarnhofer, first of all: Impressive work! Thank you very much. I left a bunch of inline comments but I will probably need some more time to digest everything better. 
Regarding your questions (at least one of them): I think your multi-layer-2d-ray-tracer should be seemingly replaceable with the other 2d ray-tracer to be used with the `ray_tracing` class (and I would chose either of the two 2d classes base on the medium class/type). Does that sound sensible? And I think if you do that you do not have to reimplement the birefringence functions at all.

fschlueter:
> I also modified the medium_base and added a correpsonding layer oriented medium class, and defined the corresponding ice models in medium. Here I would also be happy about some input from your side on how we should do this.

Yes I think that is exactly the correct approach.

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
hanneswarnhofer: The whole description in the beginning is a bit outdated and partly a bit confusing how it is written I gotta admit, I understad all of this better now and will try to describe everything in a better way and be more concise with all the definitions and so.

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -1093,10 +1090,9 @@ def determine_solution_type(y1, z1, y2, z2, C0, layers, downgoing, with_air):
     )
 
     if with_air and downgoing:
-        #return solution_types_revert['from_air']
-        return FROM_AIR
-        #return 0
-
+        #return solution_types_revert['direct']
```
fschlueter: remove those comments? Why are you defining `DIRECT` instead of using `solution_types_revert['direct']` directly?

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -1415,8 +1411,8 @@ def find_solutions(x1, x2, layers,tol=1e-6):
                                 'D' : result2,
                                 'x1': x1,
                                 'flag' : 3})
-        else:
-            print("no solution with logC0 > {:.3f} exists".format(result.x[0]))
+        #else:
+        #    print("no solution with logC0 > {:.3f} exists".format(result.x[0]))
```
fschlueter: In general please only keep comments like this when really necessary

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
hanneswarnhofer: This flag makes it nicer to use the segments in the following functions so I would prefer to leave it. I changed it to 1 for UPGOING and 0 for DOWNGOING and we can avoid defining this global variable here and just hide this assignment and check inside the functions if that's better.

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
hanneswarnhofer: Yes, did that now!

NuRadioMC/SignalProp/propagation.py
hanneswarnhofer: No that was only for me to debug and plot, forgot to remove it

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
hanneswarnhofer: Numba has a problem with the dict style of solution_types_revert if I use it directly, so I am doing this as a workaround. I wasn't sure yet if we can go on like this so I left the comments. Will remove eventually

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
hanneswarnhofer: Will do! Still gotta include some logging stuff and then I remove those lines

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
fschlueter: If there is a purpose for it keep it. Whether to implement it as local or global variable I can not say (would need to better understand the code), but typically local is better (there are few reasons to have global variable). Can you just use one? Downgoing=1 or 0 (0 means its upgoing) or the other way around.

hanneswarnhofer: Ok 👍  I changed it like this now so that we just have a 1 or 0 flag for upgoing or not upgoing (aka downgoing)

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
hanneswarnhofer: Good idea, will do

NuRadioMC/utilities/medium_base.py
hanneswarnhofer: I added some better docstrings to all of that IceModelExpLayers class 👍

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
hanneswarnhofer: In principle yes. I was thinking we could still leave it for people that already want to use it for evaluating path parameters such as Jethro currently and Nils for his refractive index fit without the need to have all dependencies working. Like this the stuff can be used right away by downloading this file without any possibly complicated depency issues. But yes at some point we should remove it to avoid slipping errors into somewhere with the redundant double definition 👍

fschlueter:
@hanneswarnhofer give us a hint one you think this is ready to be further reviewed.

hanneswarnhofer:
Hi @fschlueter and @sjoerd-bouma, I tested now a bit if it works when I define the `ray_tracing` class with my `multi_layer_ray_tracing_2D` instead of the `ray_tracing_2D` from the single layer analytic module and it looks good. I still gotta test it a bit more systematically, but I guess we can already think about how we do it the least messy way to include both the changes from here, as well as the ones from #1087. 

We might also want to add this to the `SignalProp.propagation` module, that it takes care of loading `ray_tracing(r2d=ray_tracing_2D)` when `analytic` is provided and `ray_tracing(r2d=_multi_layer_ray_tracing_2D)` when `multilayer` is used. 

Also I was wondering if I should leave it in one big single file as it is or maybe split it up a bit into the solver functions, path evaluations (travel time, attenuation, ...) and so on and structure it better?

fschlueter:
> Hi @fschlueter and @sjoerd-bouma, I tested now a bit if it works when I define the `ray_tracing` class with my `multi_layer_ray_tracing_2D` instead of the `ray_tracing_2D` from the single layer analytic module and it looks good. I still gotta test it a bit more systematically, but I guess we can already think about how we do it the least messy way to include both the changes from here, as well as the ones from #1087.
> 
> We might also want to add this to the `SignalProp.propagation` module, that it takes care of loading `ray_tracing(r2d=ray_tracing_2D)` when `analytic` is provided and `ray_tracing(r2d=_multi_layer_ray_tracing_2D)` when `multilayer` is used.
> 
> Also I was wondering if I should leave it in one big single file as it is or maybe split it up a bit into the solver functions, path evaluations (travel time, attenuation, ...) and so on and structure it better?

Hi Hannes, I think the logical way would be to first merge #1087 and than this class. You could already merge the current version of 1087 into this branch/PR and make the necessary adjustments. Of course this is only meaningful if 1087 gets merged (which I hope it does). 

Regarding the refactoring of the code into multiple files, I am all for it if it increases readability!

fschlueter:
Feel free to share the tests/comparison you have done here (its a good way to archive). Also please think about adding unit tests for your new raytracer (maybe even comparing the output of the two raytacer implementations)

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,2992 @@
...
+    x2 = (0, 0)
+    travel_time = get_travel_time_analytic(C0, x1, x2, layers)
+
+    C1, _, _, _ = compute_offsets(C0, pos[1], pos[2], layers)
+
+    horizontal_offset =  evaluate_y(C0, C1, 0, layers) - evaluate_y(C0, C1, pos[2], layers)
+
+    return horizontal_offset, travel_time
+
+def get_time_difference_plain_wave_analytic(pos1, pos2, theta_air, phi_air, layers):
```
philippwindischhofer: Replace with `get_time_difference_plane_wave_analytic`

philippwindischhofer (changes requested):

NuRadioMC/utilities/medium_base.py
```diff
@@ -830,3 +827,202 @@ def getGradient(self, position): #name may not be changed because linked to c++
...
+
+        Parameters
+        ----------
+        layers : list of dict
+            Layer definitions (see class docstring).
+        """
+        self.layers = sorted(layers, key=lambda L: L["z_min"],reverse = True)
+        self._validate_layers()
+
+        self._layers_arr = self._layers_to_arrays()
```
philippwindischhofer: Could I suggest to factor these three lines into a simple function `_set_layers(self, layers)` that is then called from the constructor? This will make it possible to also call `_set_layers` at runtime to update the ice model, which will be useful e.g. in the geometry calibration.

philippwindischhofer (changes requested):

NuRadioMC/utilities/medium.py
```diff
@@ -369,3 +376,123 @@ def get_ice_model(name):
...
+                "region_name": "Firn"
+            },
+            {
+                "z_min": -3000.0,
+                "z_max": -80.5,
+                "n_ice": 1.77468,
+                "delta_n": 1.41573,
+                "z_0": 1/0.0387882,
+                "region": "bubbly_ice",
+                "region_name": "Ice"
```
philippwindischhofer, line 631: One thing that could be useful here is introduce a dedicated ice model class for cases where the ice `n(z)` is continuous and continuously differentiable. This significantly reduces the number of free parameters. (Secretly, the 3-part model here also has continuous gradient, but that's not obvious at all from the parameters.)

I already have an implementation of this in the geometry calibration code [here](https://github.com/RNO-G/antenna-positioning/blob/d3a7f11ba434b1b69d45c47e46eb83db0fed9185/AntPosCal/ice_model/icemodel.py#L11-L86), but I think it might fit better in NuRadio. If you want, I can add it.

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,3018 @@
...
+            np.abs(theta_straight),
+            layers
+        )
+
+        n_z = get_refractive_index(z1,layers)
+        logC0straight = np.log(max(C0straight - 1./n_deep, 1e-12))
+        logC0skim_nz = np.log(max(1/n_z - 1./n_deep, 1e-12))
+        logC0skim = np.log(max(C0skim- 1./n_deep, 1e-12))
+
+        result = optimize.root(obj_delta_y_sqr, x0=logC0straight, args=(y1,z1,y2,z2,layers, n_deep,downgoing,with_air), tol=tol)
```
philippwindischhofer: With `logC0straight` as initial condition, the root finding fails for me when I try to calculate rays for which start and end are at similar depths.

I think it's important that we make it so that the raytracer "never" fails to find an existing solution between two points in the bulk. (Of course this gets arbitrarily complicated near the caustic, but let's leave that case for now.) What do you think about making this section more robust by looping over the different starting values for the initial conditions until a solution is found?

philippwindischhofer (changes requested):

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,3018 @@
...
+    # same as multiplying by sin(theta)
+
+    surface_shift = np.inner(src_hvec, dP)
+    delta_L_air = surface_shift * np.sin(theta_air)
+
+    # convert to travel time
+    delta_t_air = n_air * delta_L_air*units.m / constants.c
+
+    # total relative arrival time
+    delta_t = (t1 - t2) + delta_t_air
```
philippwindischhofer: Please double-check if I'm missing something here, but I think the relative sign between the in-ice and in-air contributions should be inverted.

philippwindischhofer (changes requested):

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
hanneswarnhofer: Thanks for catching! Changed dP=P1-P2 to dP=P2-P1, now it should be fine

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
hanneswarnhofer: I gotta admit that there is something going on in the minimzation process that I don't really understand. 

For example in the new function `find_solutions_bulk`: If I try three different starting parameters, for example `logC0guess_1, `logC0guess_2` and `logC0guess_2`  which correspond to different launch angles so to say, then I get worse performance of the solver then when only trying the first two starting parmeters. 

However, this should not make any difference if we just try again afterwards, because we should have stored already the found solutions for the given start and end point. In my understanding this is what should happen: If for a new minimization with a new starting parameter no solutions are found, no problem, if we find one and it has the same C0, we don't append the new result, if it has a solution with different C0 we would still see that we found a wrong solution or something later. However we end up with areas where no solutions are found when we include the other try with another starting value. Anyone an idea/experience on what might be the underlying issue here?

hanneswarnhofer: I bypassed this weird behavior now like this: 

The "new" function `find_solutions_bulk` finds (nearly) all the solutions for starting points below ice, but fails when finding solutions that lie in the air, so across a layer boundary where n(z) makes a quite large jump. 

The "old" approach, which is the same as used in the old analytic raytracer, used in function `find_solutions` however was also stable across the ice surface and finds solutions for air-to-ice rays (except for some that are far away horizontally and very close above the surface, but we anyhow don't really understand how a signal like this is expected to behave I guess, so it's not as important). 

I could neither of them get to work universally for all points, so I have them as separate functions that are called depending on the geometry of the problem (x1 in air or vertically below x2) and if no solution is found we try the other one. Ugly and makes the solver unnecessarily try two solution finding functions also in the shadow zone, but it works. This happens in the `find_solutions` method of the `multi_layer_ray_tracing_2D` class. Since in some special geometrical cases the `find_solution_bulk` function tends to return three solutions with one bein wrong , I included the `reduce_solutions` function, which checks again if we have duplicate C0s and reduces those and if after that we still have three or more solutions we omit the lowest C0 solutions (because the wrong minimization(s) result in negative or very low C0) and only keep 2. Also inefficient but seems to work.

NuRadioMC/utilities/medium.py
philippwindischhofer: @fschlueter @sjoerd-bouma does one of you have any opinion on this?

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,3664 @@
...
+        solution_type = self.determine_solution_type(x1, x2, C0)
+
+        # Only sanity-check DIRECT solutions
+        if solution_type == solution_types_revert['direct']:
+
+            direct_len = np.sqrt((x2[0]-x1[0])**2 + (x2[1]-x1[1])**2) * units.m
+            direct_time_lightspeed = direct_len / constants.c
+
+            if travel_time > direct_time_lightspeed * 2: # Break for obviously unphysical solutions (from wrong solutions, makes plotting easier) 
+                return None
```
philippwindischhofer: This test fires for me in cases when (I think) it shouldn't: I have a vertical ray in a medium with `n > 2`, mulan finds the correct solution but then gets hit by this.

philippwindischhofer (changes requested):

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
hanneswarnhofer: True, in media with refractive index above 2 this always fails! To be on the safer side here, I included n(z1) as a factor in the threshold calculation

fschlueter:
FYI, I changed (temporarily) the base for this PR to simplify the diff. (That seems not to have worked so well because I see now unrelated changes in the diff... maybe they vanish when github realizes that I just updated the base branch?

.github/workflows/run_tests.yaml
```diff
@@ -78,6 +78,7 @@ jobs:
     - uses: actions/checkout@v6
 
     - name: Setup GSL
+      timeout-minutes: 5
```
fschlueter: Can't we add this reoccurring timeout to the`setup-gsl` action?

NuRadioMC/SignalProp/analyticraytracing.py
fschlueter, line 10: @hanneswarnhofer could you briefly mention here the analytic signal and multi later ray tracers. Since we now have two different implementations?

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,3437 @@
...
+    
+    return np.sqrt(1 / f_inv_sq)
+
+
+import time
+import functools
+import logging
+
+
+def log_timing(level=logging.DEBUG):
```
fschlueter: That could be moved into a utilities/ folder

NuRadioMC/SignalProp/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,3437 @@
...
+            return result
+
+        return wrapper
+
+    return decorator
+
+
+
+
+class multi_layer_ray_tracing_2D(ray_tracing_base):
```
fschlueter: The file is quite monolithic. Can you break it up? One file containing the class and one containing all "helper" functions. You can keep them in the same folder.

NuRadioMC/SignalProp/analyticraytracing.py
fschlueter: If we use the multi layer raytracer we should set this actively to false here?

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -186,6 +199,34 @@ def set_start_and_end_point(self, x1, x2):
         self._x2 = np.array([X2r[0], X2r[2]])
         self.__logger.debug("2D points {} {}".format(self._x1, self._x2))
 
+    def set_start_and_end_point_no_swap(self, x1, x2):
```
fschlueter, line 236: cant you use the function `set_start_end_point` with a optional argument `swap=True` which in cases you need it set to False?

NuRadioMC/SignalProp/propagation.py
```diff
@@ -43,7 +45,10 @@ def get_propagation_module(name=None):
         return ray_tracing_base
     elif name == available_modules[0]:
         from NuRadioMC.SignalProp.analyticraytracing import ray_tracing
-        return ray_tracing
+        return partial(ray_tracing, ray_tracing_module = 'analytic')
+    elif name == available_modules[3]:
+        from NuRadioMC.SignalProp.analyticraytracing import ray_tracing
+        return partial(ray_tracing, ray_tracing_module = 'multilayer')
```
fschlueter: by default I would chose the implementation purely based on icemodel?

hanneswarnhofer: Sounds good to me!

NuRadioMC/utilities/medium.py
fschlueter: I think that sounds reasonable. As far as I understand @hanneswarnhofer uses modules which are in fact continuous. Do we want to keep the "generic" ExpLayer model too? I am getting worried that we have to many different icemodel classes around (but might be wrong on that)

NuRadioMC/SignalProp/analyticraytracing.py
hanneswarnhofer: For the multilayer one I made it so we can pass this use_cpp variable but it gets always set to False in the init of mulit_layer_ray_Tracing_2D no matter what was specified, so that it still has the same interface as the analytic one and doesn't break when this is given (e.g. when switching the ice model from simple to multilayer in an existing script). Might be a bit misleading though

NuRadioMC/SignalProp/analyticraytracing.py
hanneswarnhofer: Can also do it this way 👍 I would maybe call it `autoswap=True` or something, so it doesn't cause confusion with the `self._swap` variable that is set to True or False depending on whether the swapping was done or not

NuRadioMC/utilities/medium.py
```diff
@@ -375,3 +387,438 @@ def get_ice_model(name):
...
+                "delta_n": 0.446,
+                "z_0": 40.9,
+                "region": "ice",
+                "region_name": "Ice"
+            }
+        ]
+
+        super().__init__(
+            layers=layers
+            #z_bottom=-3000.0,
```
fschlueter: remove

NuRadioMC/utilities/medium.py
```diff
@@ -375,3 +387,438 @@ def get_ice_model(name):
...
+        )
+
+class greenland_3exp_layered(medium_base.IceModelExpLayers):
+    """
+    Four layer refractive index model.
```
fschlueter: 3 + air

NuRadioMC/utilities/medium_base.py
```diff
@@ -826,3 +827,355 @@ def getGradient(self, position): #name may not be changed because linked to c++
...
+        Raises
+        ------
+        ValueError
+            If a position lies outside all defined layers.
+        """
+        def n_of_z(z):
+            for L in self.layers:
+                if L["z_min"] <= z < L["z_max"]:
+                    return L["n_ice"] - L["delta_n"] * np.exp(z / L["z_0"])
+            raise ValueError(f"Position z={z} is not covered by any layer!")
```
fschlueter, line 980: feels like low hanging fruits to increase speed?

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -83,6 +88,9 @@ class describing the index-of-refraction profile
 
         detector: detector object
 
+        ray_tracing_module: string
+            'analytic' or 'multilayer'. Specifies which 2D raytracer to use. Default: 'analytic'
+
```
fschlueter: remove

NuRadioMC/SignalProp/analyticraytracing.py
fschlueter: I think you have to add a check for the ice model

NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -127,9 +130,24 @@ class describing the index-of-refraction profile
...
+            
+        elif isinstance(medium, IceModelExpLayers):
+
+            self.__logger.status("IceModelExpLayers was provided: Using 'new' multilayer analytic version as the 2D raytracing module.")
+            self._r2d = multi_layer_ray_tracing_2D(self._medium, self._attenuation_model, log_level=log_level,
+                            n_frequencies_integration=self._n_frequencies_integration,
+                            **ray_tracing_2D_kwards, use_cpp=use_cpp, compile_numba=compile_numba)
+            
+        else:
+            self.__logger.error("The analytic raytracer can only handle ice model of the type 'IceModelSimple' (single layer) or 'IceModelExpLayers' (multilayer) (see NuRadioMC.utilities.medium), might break...")
```
fschlueter: might break? remove trailing comment

NuRadioMC/SignalProp/analyticraytracing.py
fschlueter: do we still need the function?

NuRadioMC/SignalProp/AnalyticRayTracing/MultilayerAnalyticRayTracing/corefunctions.py
```diff
@@ -0,0 +1,923 @@
...
+from NuRadioMC.SignalProp.propagation_base_class import ray_tracing_base
+
+from math import sqrt, log, sin
+
+
+NumbaList = list # fallback for get_path_segments function
+
+# ------------------------------
+# Layer definitions
+# ------------------------------
```
fschlueter: why do we need these layer definitions here?

NuRadioMC/SignalProp/AnalyticRayTracingImpl/MultilayerAnalyticRayTracing/corefunctions.py
```diff
@@ -0,0 +1,923 @@
...
+        "z_min": -3000.0,
+        "z_max": -300.0,
+        "n_ice": 1.9468,
+        "delta_n": 1.41573,
+        "z_0": 1/0.0387882,
+        "region": "lower_ice",
+        "region_name": "Lower Ice"
+    }]
+
+def layers_to_arrays(layers):
```
fschlueter, line 12: functions are already defined in medium_base. remove one.

NuRadioMC/SignalProp/AnalyticRayTracingImpl/MultilayerAnalyticRayTracing/multilayeranalyticraytracing.py
```diff
@@ -0,0 +1,541 @@
...
+* z : vertical coordinate
+
+z = 0 corresponds to the ice surface.
+
+The ray parameter ``C0`` determines the curvature of the trajectory.
+It can be seen as C0 = 1/(n(z)*sin(theta)) where n(z) is the refractive index and theta is the angle relative to the horizontal at the current depth z.
+
+Layer definitions
+-----------------
+Layers are initially defined as dictionaries with the following keys:
```
fschlueter, line 55: should we add this to medium_base or medium? (or at least reference it there to here?)

NuRadioMC/utilities/medium_base.py
hanneswarnhofer: Changed it to directly using the layers array instead 👍

NuRadioMC/SignalProp/AnalyticRayTracing/MultilayerAnalyticRayTracing/corefunctions.py
hanneswarnhofer: We don't, this is more of a leftover code so some old plotting and testing functions were still working, but now after the refactoring I guess we can remove this

NuRadioMC/SignalProp/AnalyticRayTracingImpl/MultilayerAnalyticRayTracing/corefunctions.py
hanneswarnhofer: I removed the methods in `medium_base` and instead use the one defined here to get `self._layers_arr` in the icemodel

NuRadioMC/SignalProp/AnalyticRayTracingImpl/MultilayerAnalyticRayTracing/multilayeranalyticraytracing.py
hanneswarnhofer: This is also shown in the docstring of the `IceModelExpLayers` in `medium_base`. Maybe I would keep it in both places and here reference the ice model class and where to find it 🤔

philippwindischhofer:
@hanneswarnhofer and others: I took the liberty of adding the continuously-differentiable version of the n-layer exponential ice model, as recently discussed on the NuRadio call. More documentation to come.

NuRadioMC/SignalProp/analyticraytracing.py
fschlueter: can we remove this function or replace it implementation by calling `set_start_and_end_point(x1, x2, False)`

NuRadioMC/SignalProp/AnalyticRayTracingImpl/MultilayerAnalyticRayTracing/corefunctions.py
```diff
@@ -0,0 +1,774 @@
+
+import numpy as np
+from scipy import optimize
+from operator import itemgetter
+#from numba import njit
+#from numba.typed import List
```
fschlueter: remove stale imports

NuRadioMC/SignalProp/AnalyticRayTracingImpl/MultilayerAnalyticRayTracing/corefunctions.py
```diff
@@ -0,0 +1,774 @@
...
+    if (z_turn >= z_surface) and not with_air:
+            z_turn = z_surface
+    
+    # Evaluate horizontal coordinate at turning point
+    y_turn = evaluate_y(C0, C1, z_turn, layers)
+    
+    return y_turn, z_turn
+
+@njit(cache = True)
+def get_delta_y(C0, y1, z1, y2, z2, layers, C0range,
```
fschlueter: not your doing buy why not renaming all normal variables to start with a non-capital character?

fschlueter:
> ## Coverage report
>  
> 
> Click to see where and how coverage changed

@hanneswarnhofer this is a new feature we just implemented. I think we will have to understand how to "use" it / interpret these reports. However, maybe you can have a look how to increase the test coverage for the code added in this PR?

NuRadioMC/SignalProp/AnalyticRayTracing/CPPAnalyticRayTracing/wrapper.cpp
```diff
@@ -0,0 +1,8806 @@
+/* Generated by Cython 3.1.3 */
```
fschlueter: I guess this was accidentally committed?

NuRadioMC/SignalProp/AnalyticRayTracingImpl/MultilayerAnalyticRayTracing/multilayeranalyticraytracing.py
```diff
@@ -103,10 +103,10 @@
...
-from NuRadioMC.SignalProp.AnalyticRayTracingImpl.MultilayerAnalyticRayTracing.planewave import get_inice_quantities, get_time_difference_plane_wave_analytic
+from NuRadioMC.SignalProp.AnalyticRayTracingImplImpl.MultilayerAnalyticRayTracing.corefunctions import get_layer_index, get_n_1D, analytic_F, compute_offsets, build_y_field, evaluate_y, find_z_turn, get_turning_point, get_C0_from_theta, get_delta_y, get_skim_angle, determine_solution_type
+from NuRadioMC.SignalProp.AnalyticRayTracingImplImpl.MultilayerAnalyticRayTracing.solver import find_solutions, find_solutions_bulk, reduce_solutions
+from NuRadioMC.SignalProp.AnalyticRayTracingImplImpl.MultilayerAnalyticRayTracing.getrayparameters import get_path, get_path_segments, get_path_length_analytic, get_travel_time_analytic, get_launch_angle, get_receiving_angle, get_reflection_angle, get_attenuation_along_path, get_focusing_factor, get_launch_vector, get_receiving_vector, ds_dz_layer, get_path_length_numerical, get_travel_time_numerical
+from NuRadioMC.SignalProp.AnalyticRayTracingImplImpl.MultilayerAnalyticRayTracing.planewave import get_inice_quantities, get_time_difference_plane_wave_analytic
```
fschlueter: Double `Impl`

NuRadioMC/test/SignalProp/G02generate_data_set_for_unit_test_mooresbay.py
```diff
@@ -35,7 +35,7 @@
 # tt = 0
 for iX, x in enumerate(points):
 #     t_start2 = time.time()
-    r = ray.ray_tracing(x, x_receiver, ice, n_reflections=2)
+    r = ray.ray_tracing(x, x_receiver, ice)
 #     tt += (time.time() - t_start2)
```
fschlueter, line 43: where these changes intended?

NuRadioMC/SignalProp/AnalyticRayTracing/CPPAnalyticRayTracing/wrapper.cpp
hanneswarnhofer: Yes I messed something up in the process of commiting my last changes and rebasing

NuRadioMC/test/SignalProp/G02generate_data_set_for_unit_test_mooresbay.py
hanneswarnhofer: Oh no this one was also not planned to be kept

NuRadioMC/SignalProp/AnalyticRayTracingImpl/MultilayerAnalyticRayTracing/corefunctions.py
fschlueter: why would we need to reimplement those?

NuRadioMC/SignalProp/AnalyticRayTracing/CPPAnalyticRayTracing/wrapper.cpp
fschlueter: @hanneswarnhofer why did you resolve this? First this file is now under the wrong path and second (more importantly) this is an autogenerated file which should not be commited! I would actually suggest you try to remove it from the entire git history with some forced pushes since its almost 9k lines!

NuRadioMC/utilities/medium_base.py
```diff
@@ -1,9 +1,12 @@
 from NuRadioReco.utilities import units
+from NuRadioMC.SignalProp.AnalyticRayTracingImpl.MultilayerAnalyticRayTracing.corefunctions import layers_to_arrays
+from NuRadioMC.SignalProp.AnalyticRayTracingImpl.MultilayerAnalyticRayTracing.corefunctions import layers_to_arrays
```
fschlueter: Remove one

NuRadioMC/utilities/medium_base.py
```diff
@@ -827,3 +830,429 @@ def getGradient(self, position): #name may not be changed because linked to c++
...
+        Parameters
+        ----------
+        layers : list of dict
+            Layer definitions (see class docstring).
+        """
+        
+        #self.z_air_boundary = z_air_boundary
+        #self.z_shift = z_shift
+        self.layers = sorted(layers, key=lambda L: L["z_min"],reverse = True)
+        self._validate_layers()
```
fschlueter, line 910: what about the two unused arguments?

hanneswarnhofer: I intermediately disabled them in the process of a lazy error fix but I will change it so that z_air_boundary and z_shift are both set to 0 meters here to be backwards compatible with the single layer raytracer.

NuRadioMC/SignalProp/AnalyticRayTracing/CPPAnalyticRayTracing/wrapper.cpp
hanneswarnhofer: I removed it now, Sorry!

fschlueter:
Believe it or not - this PR is now first in the row, nothing else "to merge before" :)

NuRadioMC/SignalProp/AnalyticRayTracingImpl/MultilayerAnalyticRayTracing/corefunctions.py
hanneswarnhofer: Changed it

hanneswarnhofer (commented):
Soo, hello again :) I went through all the comments and it seems to me that we are close to being able to merge this. What do you all think?

It shows me that @philippwindischhofer is still requesting changes, but I can't really see any specific parts where this refers to. I am not super familiar with this GitHub Review thingy though, so might be that I am just missing something

philippwindischhofer (dismissed):

philippwindischhofer:
Hi @hanneswarnhofer I also don't know how to make the "requested changes" thing go away ... I tried re-submitting the review and not ticking that box, which I think helped?

NuRadioMC/test/SignalProp/T10unit_test_C0_greenland3exp.py
```diff
@@ -0,0 +1,58 @@
...
+for iX, x in enumerate(points):
+    r.set_start_and_end_point(x, x_receiver)
+    r.find_solutions()
+    if(r.has_solution()):
+        for iS in range(r.get_number_of_solutions()):
+            results_C0s_cpp[iX, iS] = r.get_results()[iS]['C0']
+
+# with open("reference_C0_MooresBay.pkl", "wb") as fout:
+#     pickle.dump(results_C0s_cpp, fout)
+results_C0s_cpp_ref = io_utilities.read_pickle("reference_C0_greenland3exp.pkl", encoding='latin1')
```
fschlueter: do we need to use pickle here? can we use json?

hanneswarnhofer: I guess we can also switch to json here, we are just comparing simple values for 1000 events, so we don't really need to worry about file size or speed here I think. I took basically all of this from the existing tests T05 and T06, where pickle is used, so I just copied it. Should I then also change those to using json?

hanneswarnhofer: I changed it now so that also the other tests write and load a json file instead of the pickle one

hanneswarnhofer:
So where are we in terms of merging now? @fschlueter @sjoerd-bouma 
The tests all run and work, the failing one is just this network connection timeout that occurs once in a while. 

> Hi @hanneswarnhofer I also don't know how to make the "requested changes" thing go away ... I tried re-submitting the review and not ticking that box, which I think helped?

@philippwindischhofer This worked, but then after I added another commit it changed again to "Review required", so I guess I would just have to ask again for a review and you do the same thing you did it 2 days ago.

anelles (approved):
Let's go merge this!

PR #1080, opened 2026-04-03
Add KM3-20213A to sensitivities plot

cozzyd:
Add KM3-230213A to E2_fluxes3.py 

This also slightly shifts the Auger label to avoid an overlap. 

Note that I'm not sure the current default when running the script as main makes too much sense, but this PR doesn't address that. (Maybe it's supposed to imply that you should write your own driver script).


fschlueter (approved):
Havnt checked if myself but trust Cosmin

fschlueter:
And yes this script should be cleaned up and probably merged with an other versions of it

anelles:
@fschlueter this is on my eternal todo ... but I guess you are in spring cleaning mode

fschlueter:
nope, sorry - will not touch it.

PR #1025, opened 2025-11-10
Likelihood reconstruction

MartinRavn:
This PR contains the main modules used for the results presented in our recent paper: https://arxiv.org/pdf/2510.21925

The code consists of 4 core modules that handle different parts of the reconstruction:
  - `NuRadioReco.modules.likelihood_reconstruction.likelihood_calculator.LikelihoodCalculator`: Calculates the likelihood for a simulated signal trace given a measured data trace.
  - `NuRadioReco.modules.likelihood_reconstruction.shower_simulator.ShowerSimulator`: Simulates the signal traces from a neutrino-induced shower for user-defined shower parameters, a given detector, and a simulation config file. This acts as a pure neutrino signal model and is used in forward-folding reconstruction of neutrino
  signals.
  - `NuRadioReco.utilities.minimization.Minimization`: Class for minimization of, e.g., a -2 log likelihood that unifies
  the interfaces of `scipy.optimize` and `iminuit`. The class adds additional functionality like normalization
  of the fitted parameters.
  - `NuRadioReco.utilities.matched_filter.MatchedFilter`: Class for performing a matched filter search of simulated signal templates in noisy data.

These modules are then combined in modules for specific reconstruction tasks with easy-to-use interfaces:
  - `NuRadioReco.modules.likelihood_reconstruction.neutrinoLikelihoodReconstructor`: Reconstructs a neutrino
  signal using the data traces in a `station` object, the spectra of the noise for each antenna, a detector
  description, and a simualtion config file.
  - `NuRadioReco.modules.likelihood_reconstruction.electricFieldLikelihoodReconstructor`: Reconstructs a pulsed
  electric field in an ensemble of antennas in close proximity to each other using the data traces in a `station`
  object.

I also implemented 5 examples demonstrating these classes:
- `LikelihoodCalculator`, `Minimization`, and `MatchedFilter` in `NuRadioReco/examples/likelihood_reconstruction/run_shower_simulator.py`
- `ShowerSimulator` in `NuRadioReco/examples/likelihood_reconstruction/run_shower_simulator.py`
- `neutrinoLikelihoodReconstructor` in `NuRadioReco/examples/likelihood_reconstruction/neutrino_signal_reconstruction.py`
- `A more advanced neutrino reconstruction in `NuRadioReco/examples/likelihood_reconstruction/neutrino_signal_reconstruction_advanced.py`
[...]

PR #1081, opened 2026-04-13
ShowerSimulator (neutrino signal model)

MartinRavn:
This PR implements a class that simulates a neutrino signal with a given energy, direction, vertex position, interaction time, and shower type/profile. The simulation is equivalent to the main simulation class in NuRadioMC, but doesn't add and doesn't run a trigger simulation. It is hence a pure neutrino signal model, which is useful for studies where only the true signal is of interrestm, and it is useful for forward-folding reconstruction methods.


MartinRavn:
I renamed and refactored this class to `ShowerSimulator` since it doesn't do anything specifically neutrino-related, like calculating fluxes, inelasticities, secondaries, etc. It now also has the option to take a list of initialized shower objects as input to be able to simulate several showers.

MartinRavn:
Here are the run times for initializing, first simulation, and subsequent simulations (24 antennas, ARZ2020, electromagnetic shower, analytic ray-tracing):
image

MartinRavn:
Here is the output of an example simulation using the class:
image

MartinRavn:
Here are some bullet points comparing the `ShowerSimulator` class and the example script in `NuRadioMC/simulation/examples/A02calculate_channel.py:`
image

MartinRavn:
@fschlueter, @cg-laser, @sjoerd-bouma, I now improved the example script with some of the things we discussed yesterday:
- The ice model and propagator are now initialized using the config file
- I removed the part with `apply_det_response_sim`, since the detector response was already applied using `apply_det_response`. There is no need to do this twice in an example.
- I added resampling to the detector sampling rate
- I cut the trace to a user-defined `trace_start_time` using `channelReadoutWindowCutter`

To cut the trace to the readout window, I had to add a new behaviour in `channelReadoutWindowCutter`. Now, if `trace_start_times` are provided, the trigger times are ignored (they are not there in this case), and the channels will be cut using the `trace_start_times` and the trace lengths (`n_samples / sampling_rate`) defined in the detector. There may be a more elegant way to implement this, but it shows the concept.

MartinRavn:
Here is a new comparison slide between the `ShowerSimulator` and the example script with highlighted changes:
image

Now, the main downsides of the example script are that it only runs for a single channel, it is not importable, it will initialize everything every time it is run, and we have no way to implement additional functionality (e.g, automatic trace start times, Hann windows to the readout traces).

MartinRavn:
I still think there is a lot of value in providing a class that can be imported by users. I know that students find NuRadioMC difficult, and this would make it easier for someone who just wants to simulate a neutrino signal in a detector. The only risk is that the class and example scripts serve a similar purpose, but I think it can be argued that the scope of the class is already more comprehensive and general than the example script. I also definitely need it for the likelihood-based forward folding reconstruction of neutrino signals that I am working on.

MartinRavn:
The example script also does not appear in the documentation. @sjoerd-bouma , do you know if this can be fixed?

MartinRavn:
I am closing this PR since everything discussed here was implemented through #1025 and #1098

PR #1083, opened 2026-04-23
Hotfix: pin pygdsm on Python versions <3.9

sjoerd-bouma:
The most recent pygdsm release (1.7.0) no longer works on Python 3.7/3.8, but did not update the minimum Python version required (yet). This PR just pins it to the older version which still works for the older Python versions, which doesn't change anything for the newer Python versions.

I also noticed a couple of `setup GSL` jobs timing out on the most recent CI run, taking 6 hours before giving up. That seems an excessive amount of compute time to spend waiting for a confused server; I've changed the timeout for this step to 5 minutes instead (when it completes successfully it usually takes O(8 s) so this should be plenty).

### DIFF
--- a/.github/workflows/deploy_documentation.yaml
+++ b/.github/workflows/deploy_documentation.yaml
@@ -22,6 +22,7 @@ jobs:
           cache: 'pip'
 
       - name: Install dependencies
+        timeout-minutes: 20
         run: |
           sudo apt-get install libgsl-dev
           python -m pip install --upgrade pip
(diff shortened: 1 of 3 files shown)


fschlueter (approved):
I home pygdsm did not introduce physics changing changes. Why are we not killing <py3.10?

sjoerd-bouma:
Closing this PR, as pygdsm 1.7.0 was yanked and the correct Python version specifications included for 1.7.1, so pip can correctly resolve the right version on its own again. I've made a separate PR for the GSL timeout issue: #1089  (which hasn't appeared since, but it doesn't hurt to keep the fix, I guess).

PR #1088, opened 2026-05-05
Update simulation.py

avijai2000:
Call group_into_events with zerosignal boolean specified from config in simulation.py.


fschlueter (approved):
I assume you tested the code?

avijai2000:
Yeah I tested it

PR #1091, opened 2026-05-10
optional detector object for dataProviderRNOG

zsselcuk:
If the detector object is not set, `channelCableDelayAdder` throws errors; it needs to be optional.

### DIFF
--- a/NuRadioReco/modules/RNO_G/dataProviderRNOG.py
+++ b/NuRadioReco/modules/RNO_G/dataProviderRNOG.py
@@ -61,12 +61,17 @@ class dataProviderRNOG:
         files: list of str
             List of files to read (are passed to the readRNOGDataMattak module).
         det: Detector
-            Detector object.
+            Detector object. The detector object is used to retrieve the cable delays for the channelAddCableDelay module. 
+            If set to None, channelAddCableDelay module will be skipped and no cable delays will be added.
+            channelGlitchDetector and channelBlockOffsetFitter do not use the detector object, so they will run regardless of whether the detector object is set or not.
         reader_kwargs: dict (default: {})
             Keyword arguments passed to the reader module `NuRadioReco.modules.io.RNO_G.readRNOGDataMattak`.
         """
         self.files = files
         self.detector = det
+        
+        if det is None:
+            logger.warning("The detector object is not set. Cable delays will not be added.")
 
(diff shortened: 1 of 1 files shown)


NuRadioReco/modules/RNO_G/dataProviderRNOG.py
```diff
@@ -95,15 +95,19 @@ def run(self):
         event: Event
             The processed event
         """
-
+        if self.detector is None:
+            logger.warning("The detector object is not set. Cable delays will not be added.")
+
```
fschlueter: Do you want to move this warning to the begin method. Than it does not spam the cmdline

NuRadioReco/modules/RNO_G/dataProviderRNOG.py
fschlueter: Can you extend the doc-string. For what is the detector class used and elaborate that passing `None` is now an option and what it means.

NuRadioReco/modules/RNO_G/dataProviderRNOG.py
```diff
@@ -95,15 +95,19 @@ def run(self):
...
             station = event.get_station()
-            self.detector.update(station.get_station_time())
 
+            if self.detector is not None:
+                self.detector.update(station.get_station_time())
+                self.channelCableDelayAdder.run(event, station, self.detector, mode='subtract')
+                
             self.channelBlockOffsetFitter.run(event, station, self.detector)
             self.channelGlitchDetector.run(event, station, self.detector)
-            self.channelCableDelayAdder.run(event, station, self.detector, mode='subtract')
```
fschlueter, line 107: you changed the order of the modules being applied. I am not sure if that matters, if not I would keep the old order

fschlueter: Are the other modules using the detector class at all? Apparently not? In that case we might want to explicitly pass `None` always to make it clear?

fschlueter (commented):
Thanks @zsselcuk, I added some comments. See below

fschlueter (approved):
Thanks @zsselcuk

fschlueter:
@sjoerd-bouma I think you noticed already that the South Pole test sporadically fails. Do you understand why?

PR #1092, opened 2026-05-17
Add 'header' to string type attributes list

clark2668:
If no events trigger, the file can also end up with a copy of the `header` object from the input file generator. In which case, we want to allow that to go in the merge, or else things crash with an ugly error, which looks like this

```
numpy.exceptions.DTypePromotionError: The DType <class 'numpy.dtypes.StrDType'> could not be promoted by <class 'numpy.dtypes._PyFloatDType'>. This means that no common DType exists for the given inputs. For example they cannot be stored in a single array unless the dtype is `object`. The full list of DTypes is: (<class 'numpy.dtypes.StrDType'>, <class 'numpy.dtypes._PyFloatDType'>)

```

### DIFF
--- a/NuRadioMC/utilities/merge_hdf5.py
+++ b/NuRadioMC/utilities/merge_hdf5.py
@@ -10,7 +10,7 @@ import logging
 import math
 logger = logging.getLogger("NuRadioMC.HDF5-merger")
 
-str_type_attrs = ['trigger_names', 'NuRadioMC_EvtGen_version', 'NuRadioMC_EvtGen_version_hash', 'NuRadioMC_version', 'NuRadioMC_version_hash', 'config']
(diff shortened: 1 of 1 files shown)


fschlueter:
@clark2668 can you provide some more informations? `header` is not an nuradio hdf file attribute? I am not quite understanding what issue this fixes.

fschlueter:
ping

clark2668:
Hi, sorry for my delay. What I found was that when I ran simulations, if there was no events triggering, the output hdf5 file seems to contain a copy of the header object. So it gets information about the cylinder radius, etc. So if that header information isn't in the list of things the merger expects, then the merge seems to fail. I can try to upload a MWE, if that is needed.

fschlueter:
please, I am not sure what you mean with header. We have headers in our `nur` files but not in the hdf5 files afair. Maybe I am overlooking something

fschlueter:
Closing this PR for now. Feel free to reopen it if the problem still persist and the questions can be clarified.

PR #966, opened 2025-08-14
Exporting simulation to CoREAS-style HDF5 files.

galaxyvita:
wrote a function write_CORSIKA7() that is the inverse of the read_CORSIKA7() function. Also wrote all the needed helper functions. Updated the conversion_fieldstrength_cgs_to_SI to use scipy and units only.

Issue being addressed:
Added a function write_CORSIKA7() that allows you to convert .nur file to a hdf5 file.
Added 2 helper functions convert_obs_from_nuradio_efield and convert_obs_positions_from_nuradio_on_ground.
Updated the conversion_fieldstrength_cgs_to_SI to only use scipy and .units.

PR #1095, opened 2026-05-22
Rename and update readCoREASShower.py

fschlueter:
This PR renames readCoREASShower.py to readCoREAS.py and strips away any functionality with respect to a specific detector. The idea is to have a bare wrapper around `coreas.read_CORSIKA7` which reads __all__ simulated observer and provides them with an event and sim_shower. Any detector specific action, i.e., associating specific observers with a station and channel ID can happen afterwards, and outside of this module. 

@sjoerd-bouma, @PhilippLaub, I thought this is useful to read in air-showers simulated with an explicit SKA layout. E.g., what is produced by `write_CORSIKA7` (PR #966). What is missing is a function which takes the "corsika_event" and splits the electric fields stored in one station into the corresponding stations by e.g., comparing their coordinates. What do you think? 

I think readCoREASShower.py is not really actively used. However we could keep it for a while with a deprecation warning (probably  @sjoerd-bouma  will insist on it). I think that step helps in cleaning up the logic behind the various coreas readers. If you are happy with this and find it useful I will update any dependencies in the examples we might have.


fschlueter:
@sjoerd-bouma and @PhilippLaub what do you think? If you think that is the way to go I will adapt the currently failing tests.

PhilippLaub:
Judging from the complaints I got regarding the old `readCoREASShower` module, I think this is really useful and removes some of the confusion around the modules. Thanks Felix!

fschlueter:
okay, I finished what I wanted to do from my side (if the tests pass). @PhilippLaub can you have a look? I updated the two examples using the reader and implemented an example for a "event_builder" function. Does that make sense to you?

fschlueter:
@PhilippLaub that PR would be ready to be reviewed. The test previously failed because of a desy server timeout.

PhilippLaub (approved):
Looks good! Indeed removes much of the confusion :) 
Thanks Felix!

PR #1098, opened 2026-05-29
Update simulation example scripts

MartinRavn:
The current version of the simulation example scripts have a few problems. Here is a list of what I changed:
1. Scripts didn't run because time logger has changed. This is fixed now.
2. Use propagation and ice model from config instead of overwriting it
3. Removed redundant double application of detector response
4. Cut the traces to the readout window defined by trace start time and number of samples in detector description
5. Resample to the detector sampling rate
6. Only create one plot of the simulation output
7. Run example scripts as unit tests
8. Added documentation manual with reference to the example scripts

To do this I had to add a method to `channelReadoutWindowCutter` that cuts the trace using a user-defined `trace_start_time` instead of a trigger (which hasn't been simulated in this case)


NuRadioMC/simulation/examples/A02calculate_channel.py
```diff
@@ -74,25 +72,26 @@ def detector_simulation_filter_amp(evt, station, det):
     channelBandPassFilter.run(evt, station, det, passband=[0, 500 * units.MHz],
                                 filter_type='butter', order=10)
 
-# applies the detector response to the electric fields (the antennas are defined
-# in the json detector description file)
-sim.apply_det_response_sim(sim_station, det, cfg, detector_simulation_filter_amp,
```
cg-laser, line 79: why is this call removed? The user often wants to have the sim channels available. Please leave this in, but add a comment that a call to this function is optional if the SImChannel Objects are not needed.

NuRadioMC/simulation/examples/A02calculate_channel.py
```diff
@@ -102,21 +101,11 @@ def detector_simulation_filter_amp(evt, station, det):
     ax2.plot(channel.get_frequencies()/units.MHz, np.abs(channel.get_frequency_spectrum()/units.V*units.MHz),
              f"-C{i}",label=f'channel id {channel.get_id()}')
 
-    # alternatively, we can sum the traces of the sim channels and we should get equivalent results
```
cg-laser, line 105: I would leave this in. Isn't this useful information ?

cg-laser (commented):
thanks a lot Martin

NuRadioMC/simulation/examples/A02calculate_channel.py
MartinRavn: I added this again and a comment explaining why we do it

NuRadioMC/simulation/examples/A02calculate_channel.py
MartinRavn: I added this again and labels to the plots

MartinRavn:
I also made cutting the traces to a readout window optional. By default, the script returns the full simulated traces, but if the `cut_traces` variable is changed, either a simple trigger is run or a user-defined  trace start time is used.

cg-laser (approved):
looks great, thanks Martin

PR #1101, opened 2026-06-09
Fix: Add missing get_filter() function to ARA hardwareResponseIncorporator.py

nmanic15:
# The Problem

I have been attempting to simulate using hardware response from the ARA detector using:

```
class mySimulation(simulation.simulation):

    def __init__(self, threshold=1.0, trig_chan=0, **kwargs):
        # init code
    def _detector_simulation_filter_amp(self, evt, station, det):
        hardware_response.run(evt, station, det, sim_to_data=True)

    def _detector_simulation_trigger(self, evt, station, det):
        # trigger code

if __name__ == "__main__":
    sim = mySimulation(detectorfile='path_to_json_file', kwargs*)
    sim.run()
```
Yet, when observing the printout statements during simulation, I noticed that there was no amplification being applied to the noise VRMS calculations. This lead me to believe that something was wrong with how hardware response was applied by the ARA detector. Here in the printout you can see that the max amplification is one. This was giving me much smaller noise VRMS values than I anticipated.

image

I first looked to simulation.py to see how the integrated channel response is computed and found where the main issue started inside of the for loop on line 1302 of simulation.py.

```
for channel_id in self._det.get_channel_ids(station_id):
       ff = np.linspace(0, 0.5 * self._config['sampling_rate'], 10000)
       filt = np.ones_like(ff, dtype=complex)
       for i, (name, instance, kwargs) in enumerate(evt.iter_modules(station_id)):
             if hasattr(instance, "get_filter"):
                 filt *= instance.get_filter(ff, station_id, channel_id, self._det, **kwargs)
```

filt is a complex array of floats of the filter amplitudes that is computed for each channel using the function get_filter(). It is initialized as an array of ones and then if an instance of get_filter() is found, get_filter() is computed. The problem is that the ARA hardwareResponseIncorporator.py file doesn't have a get_filter() function like the other detectors do which causes the code to silently continue in calculating VRMS values without hardware response applied unbeknownst to the user.

# The Solution

For filt to not be an array of ones and to be computed from hardware response, a get_filter() function must be added to ARA's hardwareResponseIncorporator.py file. The ARA hardwareResponseIncorporator.py file already determines the entire system's hardware response using get_system_response() found in ARA's analog_components.py file. The system response is already inclusive of the amplifier response, cable response, and filters, so all I had to do was load the system response and then return the complex response by multiplying the gain and phase from the loaded system response.

```
analog_components.load_system_response()
system_response = analog_components.get_system_response(frequencies)
system_complex_response = system_response['gain'] * system_response['phase']

if sim_to_data:
       return system_complex_response
else:
       return 1. / system_complex_response
```

This is extremely similar to the hardware response incorporators of the other detectors except it doesn't need to compute cable response and amplifier response individually.

The system response from ARA comes directly from interpolating real hardware response gain and phase data of the ARA_Electronics_TotalGain_TwoFilters.txt file and it is not channel-dependent. Simulation.py calls get_filter() with station_id, channel_id, and the detector file, but these are ignored. That is why I decided to set those three arguments to have a default of None in get_filter(). When the get_filter() is ran any one of those arguments, it will throw a warning to the user without halting simulation so that the user knows the hardware response is not channel-dependent. 

```
        if station_id is not None or channel_id is not None or det is not None:
            logger.warning("get_filter() warning: station_id/channel_id/det provided but not used."
                           "ARA system response is not channel-dependent")
```

After making these updates and running the simulation again, I got more rational values for the maximum channel amplification and noise VRMS.

Screenshot 2026-06-09 at 4 41 31 PM


fschlueter:
@nmanic15 welcome to NuRadio. You provided an excellent PR description and solution to the problem. I will approve it as is. However, I have two remarks:

1. You could consider using `get_filter` also in the `run` method of the `hardwareResponseIncorperator` this reduces the risk that what you actually apply (in `hardwareResponseIncorperator`) and how you calculate your noise level (in  `simulation.py`) does not diverge. 
2. Unrelated to this PR but say that the response your are applying contain the entire cable delay (= constant phase shift)? This might make problems down the road. The cable delay is typically separately applied in the eFieldToVoltage converter for some technical reason.

fschlueter (dismissed):

NuRadioReco/modules/ARA/hardwareResponseIncorporator.py
```diff
@@ -25,6 +25,48 @@ def __init__(self):
...
+    -------
+    array of complex floats
+        the complex filter amplitudes
+    """
+
+    if station_id is not None or channel_id is not None or det is not None:
+        logger.warning("get_filter() warning: station_id/channel_id/det provided but not used."
+                       "ARA system response is not channel-dependent")
+
+    analog_components.load_system_response()
```
sjoerd-bouma: What does `load_system_response` do, and why is it needed here but not when applying the response through the `run` method?

sjoerd-bouma (commented):
The radio emitter test currently fails - it would be good if you could check if it was doing something incorrect previously (which is now fixed) or that your change has actually broken something. See https://github.com/nu-radio/NuRadioMC/actions/runs/28867391942/job/85621280820?pr=1101

NuRadioReco/modules/ARA/hardwareResponseIncorporator.py
nmanic15: Good notice. This was an oversight on my end. load_system_response's only purpose is to return the gain and phase values for each frequency from whatever hardware response file you're using. Then, get_system_response interpolates those values. load_system_response isn't needed in get_filter() or run() because it is called inside of analog_components when analog_components is imported and then when get_system_response is called it automatically uses those values. Should I remove line 61?

nmanic15:
@sjoerd-bouma I ran the test without the get_filter() changes I made and it was successful. It's possible that my change broke something but I don't know how? I integrated get_filter() similar to how the ARIANNA detector uses get_filter(). You said that the radio emitter is now fixed; when was it fixed? Is it possible that this branch is behind and doesn't have those changes.

nmanic15:
Also how does the radio emitter test work? what two hdf5 files is it comparing.

nmanic15:
I don’t expect the reference radio emitter output file to match the updated output since the reference was generated before these fixes were implemented. It appears to be failing likely because the code to produce the reference HDF5 output file didn't have working hardware response implementation. I added get_filter() in order to fix that issue.

NuRadioReco/modules/ARA/hardwareResponseIncorporator.py
```diff
@@ -25,6 +25,50 @@ def __init__(self):
...
+        Returns
+        -------
+        array of complex floats
+            the complex filter amplitudes
+        """
+    
+        if station_id is not None or channel_id is not None or det is not None:
+            logger.warning("get_filter() warning: station_id/channel_id/det provided but not used."
+                           "ARA system response is not channel-dependent")
+
```
sjoerd-bouma: This warning is quite 'loud' - `get_filter` is used in a lot of detector-independent code, which will then emit this warning all the time. There's also nothing going 'wrong' when I specify the station and/or channel id; I still get the correct response, it just happens to be (modelled as) the same for all channels. I would suggest changing this to 'info', 'debug' or removing the warning altogether.

NuRadioReco/modules/ARA/hardwareResponseIncorporator.py
```diff
@@ -39,20 +83,14 @@ def run(self, evt, station, det, sim_to_data=False):
             system_response = analog_components.get_system_response(frequencies)
             trace_fft = channel.get_frequency_spectrum()
 
+            filt = self.get_filter(frequencies, station.get_id(), channel.get_id(), det, sim_to_data=sim_to_data)
```
sjoerd-bouma, line 81: ... in fact even here you explicitly raise your own warning by passing the station id and channel id (which I think is the right thing to do, e.g. if in future you change the response to be different depending on the channel this will automatically work).

NuRadioReco/modules/ARA/hardwareResponseIncorporator.py
```diff
@@ -25,6 +25,50 @@ def __init__(self):
...
+                           "ARA system response is not channel-dependent")
+    
+        system_response = analog_components.get_system_response(frequencies)
+        system_complex_response = system_response['gain'] * system_response['phase']
+
+        if sim_to_data:
+            return system_complex_response
+        else:
+            filt = np.zeros_like(system_complex_response)
+            mask = np.abs(gain) > 0
```
sjoerd-bouma: gain is not defined and will raise an error (see also the test output). Consider running [`pre-commit install`](https://nu-radio.github.io/NuRadioMC/Introduction/pages/contributing.html#installing-nuradiomc-for-developers) to catch stuff like this before you push : )

sjoerd-bouma (commented):
See comments. Changing the warning is optional, but right now the code is broken due to an undefined variable. Otherwise we still need to figure out what to do with the existing ARA test, I don't know if someone from ARA could confirm if the test is wrong and needs to be updated? It was implemented by @pgiri4 , I believe.

NuRadioReco/modules/ARA/hardwareResponseIncorporator.py
nmanic15: I agree. I'll just get rid of the warning

NuRadioReco/modules/ARA/hardwareResponseIncorporator.py
nmanic15: simple fix! i'll look into installing that : )

NuRadioMC/examples/05_pulser_calibration_measurement/ARA/runARA02.py
```diff
@@ -30,31 +30,33 @@
...
     def _detector_simulation_trigger(self, evt, station, det):
         # save the amplitudes to output hdf5 file
         # save amplitudes per ray tracing solution to hdf5 data output
         calculateAmplitudePerRaySolution.run(evt, station, det)
         triggerSimulator.run(evt, station, det,
-                           threshold_high=1e-6 * self._Vrms,
-                           threshold_low=-1e-6 * self._Vrms,
+                           threshold_high=1. * self._Vrms,
+                           threshold_low=-1. * self._Vrms,
                            high_low_window=50 * units.ns,
```
sjoerd-bouma, line 47: This was the reason for the failing test - for some reason 1/100000 of the Vrms was used as the trigger threshold, and the automatic Vrms calculation changed thanks to the addition of the missing `get_filter` method of the `hardwareResponseIncorporator`. 

I've made the Vpol trigger threshold slightly closer to being sensible and kept the Hpol one ridiculous, because there is essentially no Hpol component to trigger on. Both of these should now trigger on the actual pulse (instead of numerical noise at the start of the trace).

NuRadioMC/test/examples/radio_emitter_output_reference.hdf5
sjoerd-bouma, line 1: Updated the reference file; only the trigger times should be affected (previously, triggered at the start of the traces due to numerical noise / very low trigger thresholds, now triggers on the actual pulses).

sjoerd-bouma (approved):
Thanks @nmanic15! I had another look at the failing test and figured out what was going on (see comment); I've updated the test and the reference file slightly, so they should pass now.

Feel free to merge this (ideally after waiting one day in case anyone else wants to comment).

sjoerd-bouma:
CI errors are due to #1116 and can be ignored (none of the failing tests are affected by commits since [631edda](https://github.com/nu-radio/NuRadioMC/pull/1101/commits/631eddaec3a825c0285550a0ba391177be6989df)).

NuRadioMC/examples/05_pulser_calibration_measurement/ARA/runARA02.py
nmanic15: Great thanks! no wonder it wasn't working

nmanic15:
I updated the branch and there aren't any conflicts for merging but I still cannot merge it. It says that 2 workflows are awaiting approval and that a maintainer needs to approve it.

sjoerd-bouma (approved):

nmanic15:
For some reason I still don't see the merge button at the bottom of my screen. Do I need to have certain developer permissions?

sjoerd-bouma:
> For some reason I still don't see the merge button at the bottom of my screen. Do I need to have certain developer permissions?

Hmm, then maybe people you can't merge because you do not have write permission? Would make sense. I'll merge it for you in any case.

PR #1104, opened 2026-06-22
[CI] Improve caching

sjoerd-bouma:
It seems DESY has become a bit stricter on how many downloads it's providing to a single user, leading to our CI failing because of downloads timing out. There is a discussion going on there to try to find a solution, but in the meantime, this should improve our caching (only one job had properly cached downloads previously).

Something else I realised while looking into this is that if we ever modify the workflow, e.g. by adding additional tests that require additional files, these will never be included in the cache automatically (because the old cache is not overwritten on a cache hit). [This](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache) would be a possible solution, but seems somewhat inelegant to me because it will rapidly fill the cache quota for no reason. The alternative is manually (or automatically) clearing the cache whenever the github workflow changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,17 @@
 !*reference*.nur
 !*reference*.hdf5
 
+# other binary storage files
+*.gz
+*.xz
+*.zip
+*.bz
+*.lzma
+*.tar
+*.npz
+*.npy
+*.root
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/.github/actions/cache-ci-data/action.yml
+++ b/.github/actions/cache-ci-data/action.yml
@@ -16,4 +16,5 @@ runs:
           NuRadioReco/examples/Interactive/data
           NuRadioReco/test/data
           NuRadioReco/detector/AntennaModels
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -10,6 +10,17 @@
 !*reference*.nur
 !*reference*.hdf5
 
+# other binary storage files
```
fschlueter, line 13: was that change intended?

fschlueter (dismissed):
Trusting you on this one.

.gitignore
sjoerd-bouma: I wanted to re-run the failing tests (to see if the caching worked) but weirdly github kept skipping them instead of re-running the failed tests, so I just added this semi-useful harmless commit to make it re-run the tests instead.

sjoerd-bouma:
Added the missing cross section file, sorry @fschlueter. Should be good to go now (hopefully).

The one remaining broken test is due to a mattak error.

fschlueter (approved):

fschlueter:
btw the mattak bug should be fixed

PR #1105, opened 2026-06-22
[hotfix] add missing precalculated parameter for white noise

sjoerd-bouma:
Only one real change - added `amplitude` to the `precalculated_parameters`. The `bandlimited_noise_from_precalculated_parameters` would otherwise raise a KeyError if generating `perfect_white` noise. Found by @EnriqueHuesca / Claude. The rest is just some minor tidying up of docstrings, plus a pending deprecation of the `fftnoise_fullfft` method which is not used internally and unlikely to be used anywhere else.

I'm not convinced that the whole precalculation routine is used much or indeed useful for much, so I would also be open to just removing it altogether, but I haven't gone through the trouble of benchmarking it to check if this feeling is correct.

### DIFF
--- a/NuRadioReco/modules/channelGenericNoiseAdder.py
+++ b/NuRadioReco/modules/channelGenericNoiseAdder.py
@@ -25,7 +25,7 @@ class channelGenericNoiseAdder:
         n_samples_time_domain: int
             number of samples in the time domain to differentiate between odd and even number of samples
         """
-        amps = np.asarray(amps, dtype='complex')
+        amps = np.array(amps, dtype='complex')
         Np = (n_samples_time_domain - 1) // 2
         phases = self.__random_generator.random(Np) * 2 * np.pi
         phases = np.cos(phases) + 1j * np.sin(phases)
@@ -167,7 +167,8 @@ class channelGenericNoiseAdder:
         """
         Precalculate parameters for bandlimited noise.
 
-        Precalculate some parameters to use 
+        Precalculate some parameters to use to generate noise using the 
+        `bandlimited_noise_from_precalculated_parameters`.
 
         Parameters
         ----------
@@ -255,7 +256,6 @@ class channelGenericNoiseAdder:
         Generate noise using previously set parameters
         
         Generates noise using parameters pre-set using `precalculate_bandlimited_noise_parameters`.
-        Only the noise type can 
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


fschlueter:
I am also wondering if we could delete this... Other question, we are we supporting `white` or `perfect_white` noise generation, isn't that unrealistic? I.e., where would one want to add this?

sjoerd-bouma:
> I am also wondering if we could delete this... Other question, we are we supporting `white` or `perfect_white` noise generation, isn't that unrealistic? I.e., where would one want to add this?

Here's the benchmark:
```
from NuRadioReco.modules.channelGenericNoiseAdder import channelGenericNoiseAdder
noiseadder = channelGenericNoiseAdder()

parameters = dict(min_freq=0.05, max_freq=1.2, n_samples=2048, sampling_rate=2.4, amplitude=0.016)
noisetype = 'rayleigh'
time_domain = False

%timeit noiseadder.bandlimited_noise(**parameters, type=noisetype, time_domain=time_domain)
noiseadder.precalculate_bandlimited_noise_parameters(**parameters)
%timeit noiseadder.bandlimited_noise_from_precalculated_parameters(type=noisetype, time_domain=time_domain)
```
giving
```
93.9 µs ± 10.8 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
64.7 µs ± 266 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
so a speedup of ~30%. Kind of at the borderline of what still makes sense, and I would say this is close to an optimal case (if we include the FFT by setting `time_domain=True` the difference decreases). I could be persuaded to deprecate this.

I agree that 'perfect_white' noise makes little sense, we should probably at the very least move to `rayleigh` by default? Maybe we can briefly bring this up on Thursday

fschlueter:
is that pre-calculation thing something we could replace by one-the fly caching? I would probably deprecated is as I think nobody uses it.

EnriqueHuesca:
If 'perfect_white' produces a flat frequency spectrum, isn't that the equivalent of producing a narrow sinc pulse, almost delta-like? I could see a niche use case where you want to generate a pulse as your "signal", to test your analysis/hardware response chain (I've done it "manually" in the past to test noise simulations).

I think that, in any case, this probably does not belong into the noise module and definitely should not be any kind of default.

sjoerd-bouma:
> If 'perfect_white' produces a flat frequency spectrum, isn't that the equivalent of producing a narrow sinc pulse, almost delta-like? I could see a niche use case where you want to generate a pulse as your "signal", to test your analysis/hardware response chain (I've done it "manually" in the past to test noise simulations).
> 
> I think that, in any case, this probably does not belong into the noise module and definitely should not be any kind of default.

It's not a pulse, because the phases are random. The difference is that in rayleigh also the amplitude is random (so the noise is only 'perfectly' white on average, not in individual realizations)

PR #1106, opened 2026-06-23
Set and get parameter covariances

MartinRavn:
I implemented functions to set and get parameter covariances (of two parameters) in the base `ParameterStorage` class since we don't already have this. Before, only errors/variances for each parameter could be stored.

I verified that the covariances are correctly saved and read from .nur files.


MartinRavn:
One small problem with this implementation is that we can not store covariances between x, y, and z of the vertex of a shower, because the parameters are stored as one vector in `showerParameters.vertex`. However, I still think it is useful for other parameters, such as zenith and azimuth.

fschlueter (approved):
Thanks Martin

fschlueter:
> One small problem with this implementation is that we can not store covariances between x, y, and z of the vertex of a shower, because the parameters are stored as one vector in `showerParameters.vertex`. However, I still think it is useful for other parameters, such as zenith and azimuth.

I guess for this purpose we could introduce new parameters (x, y, and z) only be used to store the covariances. But it aint pretty.

sjoerd-bouma:
> One small problem with this implementation is that we can not store covariances between x, y, and z of the vertex of a shower, because the parameters are stored as one vector in `showerParameters.vertex`. However, I still think it is useful for other parameters, such as zenith and azimuth.

I don't think there's any limitation on the number of dimensions of the stored value? So it would be fine to store the [x,y,z,zenith] matrix in there, for example; of course, in that case, you need a transpose in `_parameter_covariance[key1, key2] = _parameter_covariance[key2, key1]`.

MartinRavn:
Yes, I also thought that larger matrices can be saved, and I guess it works for [x,y,z,zenith]. However, if someone stores a 3x3 matrix with  `set_parameter_covariance(shp.vertex, shp.vertex, value)` then `get_parameter_error(shp.vertex)` returns an array, which may be confusing.
But I don't think this is a huge concern. We can find a solution when someone needs it.

fschlueter:
ping

PR #1109, opened 2026-06-25
Replaced deprecated bbox argument in matplotlib.savefig

Schpr:
In the savefig command of matplotlib the bbox argument is deprecated and replaced by box_inches, so I replaced all 3 occurrences of it.


fschlueter (approved):
Was there actually a time where bbox was a valid argument for savefig?

Schpr:
> Was there actually a time where bbox was a valid argument for savefig?

I think back in the days `bbox='tight'` worked, although `bbox_inches` always was the proper way for that. But it already isn't working anymore in matplotlib 2.2, which is the oldest LTS version still maintained, so I can't confirm that...

Issue #1046, opened 2026-02-18
Single event test fails occasionally


sjoerd-bouma:
The single event test (South pole) in the CI occasionally fails with a (very) small error (e.g. [here](https://github.com/nu-radio/NuRadioMC/actions/runs/21951549262/job/63403540142)):
```
AssertionError: 
Arrays are not almost equal to 7 decimals
Trace start times are not equal (maximum difference: 5.373189378587995e-05)
Mismatched elements: 8 / 136 (5.88%)
Max absolute difference among violations: 5.37318938e-05
Max relative difference among violations: 3.74218954e-09
 ACTUAL: array([11747.1574341, 11747.1574341, 11747.1574341, 11747.1574341,
       11747.1574341, 11747.1574341, 11747.1574341, 11747.1574341,
       11317.0540238, 11317.0540238, 11317.0540238, 11317.0540238,...
 DESIRED: array([11747.1574341, 11747.1574341, 11747.1574341, 11747.1574341,
       11747.1574341, 11747.1574341, 11747.1574341, 11747.1574341,
       11317.0540238, 11317.0540238, 11317.0540238, 11317.0540238,...
The arrays are different, difference in traces: [-2.20244074e-08 -4.77256867e-08 -3.70661372e-08 ...  3.03108360e-11
  2.16694573e-12  1.15218299e-11]
Maximum difference between traces [mV] 0.0001003322411050846
```
It seems to do this more or less at random (it passes most of the time). The test itself is seeded so it _shouldn't_ do this. The difference it fails with is very small, so it may be a machine precision issue that results in very slightly different raytracing solutions (e.g.) on different architectures, if something like that is possible? I don't currently have a better guess.

fschlueter:
I think this got fixed through a series of recent PRs

PR #1111, opened 2026-07-01
Break inf while loop and raise warning if desired trace length can no…

fschlueter:
…t be reached

see code/commit msg...

### DIFF
--- a/NuRadioReco/modules/efieldToVoltageConverter.py
+++ b/NuRadioReco/modules/efieldToVoltageConverter.py
@@ -158,6 +158,14 @@ class efieldToVoltageConverter():
 
         # Add post_pulse_time as long as we reach the minimum required trace length
         while times_max - times_min < max_channel_trace_length:
+            if self.__post_pulse_time <= 0:
+                logger.warning(
+                    f"The channel trace length as requested by the detector description ({max_channel_trace_length} ns) "
+                    f"is not reached ({times_max - times_min} ns). To extend it we iterative add `post_pulse_time`. "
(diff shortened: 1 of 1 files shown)


fschlueter:
Happy 11.11 everyone :)

galaxyvita:
Looks good, and fixed my error so the message is clear!

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -158,6 +158,14 @@ def run(self, evt, station, det, channel_ids=None):
 
         # Add post_pulse_time as long as we reach the minimum required trace length
         while times_max - times_min < max_channel_trace_length:
+            if self.__post_pulse_time <= 0:
+                logger.warning(
+                    f"The channel trace length as requested by the detector description ({max_channel_trace_length} ns) "
+                    f"is not reached ({times_max - times_min} ns). To extend it we iterative add `post_pulse_time`. "
```
galaxyvita: To extend it, we iteratively add `post_pulse_time`.

Typo

fschlueter: can you fix it?

fschlueter:
@sjoerd-bouma we still have issues connecting to the desy server. Do you know something about it. We apparently also have an issue with a test checking a reference file, this should not be due to the changes applied here but I think we need to track it down?

sjoerd-bouma:
> @sjoerd-bouma we still have issues connecting to the desy server. Do you know something about it. We apparently also have an issue with a test checking a reference file, this should not be due to the changes applied here but I think we need to track it down?

See #1116 and #1046. The former should be mostly mitigated since #1104.

PR #1112, opened 2026-07-01
Weight parameter for time-dependent template response in the signal chain

zsselcuk:
Added the forgotten weight for the time-dependent template response, so that the weight is correctly assigned for the get_signal_chain_response() function in the rnog_detector.py (See PR #1043 for more)


fschlueter:
Dont we have in the detector the call `dict.get("weight", 1)` ?

fschlueter (approved):
This seems to not make much harm. I thought we handle missing weights in the detector but this is also fine.

PR #1087, opened 2026-05-04
Refactoring ray-tracing classes

fschlueter:
This PR aims at modularizing the (analytic) ray-tracing code to improve readability. It separates standalone functions from class methods (for birefringence methods) and splits the implementation of the two classes `ray_tracing` and `ray_tracing_2D` into two files. Reason for this is that  `ray_tracing` functions as a base_class/wrapper class for analytic raytracer such as the `ray_tracing_2D`, the cpp ray-tracer, and the future multi-layer-2d-raytracer currently developed in #1077.

This PR might make it more difficult to track changes from before (not sure how git handles that moving around of code) and will cause merge conflicts for all open branches/pr on the code in question.

PR #1114, opened 2026-07-06
Improve time logging

fschlueter:
This PR improves the time logging used in `simulation.py` and does two things:

- Reduces the amount of unaccounted time in the time logging from >30% down to < 10% (these numbers are based on the used simulation setting and will change depending on the signal model or details of your detector simulation - e.g. trigger).
- Reduces the overall computing time by reducing overhead in timer_logger.py: This actually also accounts for the biggest drop in unaccounted time. By replacing astropy.time with time it saves quite some time (with my setting -41%)

Few other things:
- Better lazy string formatting for debug messages
- more time logger categories.

### DIFF
--- a/NuRadioMC/SignalProp/analyticraytracing.py
+++ b/NuRadioMC/SignalProp/analyticraytracing.py
@@ -2082,12 +2082,12 @@ class ray_tracing(ray_tracing_base):
         self._R = np.array(((c, -s, 0), (s, c, 0), (0, 0, 1)))
         X1r = self._X1
         X2r = np.dot(self._R, self._X2 - self._X1) + self._X1
-        self.__logger.debug("X1 = {}, X2 = {}".format(self._X1, self._X2))
-        self.__logger.debug('dphi = {:.1f}'.format(self._dPhi / units.deg))
-        self.__logger.debug("X2 - X1 = {}, X1r = {}, X2r = {}".format(self._X2 - self._X1, X1r, X2r))
+        self.__logger.debug("X1 = %s, X2 = %s", self._X1, self._X2)
+        self.__logger.debug("dphi = %.1f", self._dPhi / units.deg)
+        self.__logger.debug("X2 - X1 = %s, X1r = %s, X2r = %s", dX, X1r, X2r)
         self._x1 = np.array([X1r[0], X1r[2]])
         self._x2 = np.array([X2r[0], X2r[2]])
-        self.__logger.debug("2D points {} {}".format(self._x1, self._x2))
+        self.__logger.debug("2D points %s %s", self._x1, self._x2)
 
     def set_solution(self, raytracing_results):
         """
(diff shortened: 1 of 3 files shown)


NuRadioMC/SignalProp/analyticraytracing.py
```diff
@@ -1136,12 +1136,12 @@ def get_path_segments(self, x1, x2, C_0, reflection=0, reflection_case=1):
             # that will produce a downward going ray through x1
             y_turn = get_y_turn(C_0, x1, self.medium.n_ice, self.__b, self.medium.delta_n, self.medium.z_0)
             dy = y_turn - x1[0]
-            self.__logger.debug("relaction case 2: shifting x1 {} to {}".format(x1, x1[0] - 2 * dy))
+            self.__logger.debug("relaction case 2: shifting x1 %s to %s", x1, x1[0] - 2 * dy)
```
sjoerd-bouma: what's relaction (not your fault obviously)

sjoerd-bouma (dismissed):
Thanks Felix. Definitely astropy was not the right module to use there.

I am wondering why for the time logger a completely separate interface is being used, when we simultaneously always log module run times through the `register_run` decorator. But this would involve significant additional refactoring (and not work out of the box for the ray tracing), so I'm not suggesting you fix that in this PR.

I do always get confused by lazy string formatting - if I recall correctly there is an impact only if log levels are (or aren't?) set 'correctly', so I'm not sure if what you've done here actually helps. But I have no objections in any case (other than you're probably introducing merge conflicts with #1087 and #1077 ?)

fschlueter:
> Thanks Felix. Definitely astropy was not the right module to use there.
> 
> I am wondering why for the time logger a completely separate interface is being used, when we simultaneously always log module run times through the `register_run` decorator. But this would involve significant additional refactoring (and not work out of the box for the ray tracing), so I'm not suggesting you fix that in this PR.
> 
> I do always get confused by lazy string formatting - if I recall correctly there is an impact only if log levels are (or aren't?) set 'correctly', so I'm not sure if what you've done here actually helps. But I have no objections in any case (other than you're probably introducing merge conflicts with #1087 and #1077 ?)

a little test:
```
In [1]: import logging

In [2]: logger = logging.getLogger()

In [3]: a = 1

In [4]: %timeit logger.debug("Hello: %s", a)
58.5 ns ± 0.178 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit logger.debug("Hello: %s".format(a))
80.3 ns ± 1.01 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: %timeit logger.debug("Hello: %d".format(a))
79.8 ns ± 0.555 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [7]: a = 2.4

In [8]: %timeit logger.debug("Hello: %.1f".format(a))
81.2 ns ± 0.928 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [9]: %timeit logger.debug("Hello: %.1f", a)
59.6 ns ± 0.494 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [10]: %timeit logger.debug(f"Hello: {a:.1f}")
143 ns ± 0.515 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [11]:
```

fschlueter:
by-passing rule because already approved and already passing CI

fschlueter:
maybe we should advertise this in the changelog...

sjoerd-bouma:
> maybe we should advertise this in the changelog...

Feel free to, I'm happy either way.

PR #1113, opened 2026-07-06
Improve likelihood reconstruction documentation

MartinRavn:
The documentation for the recently implemented likelihood reconstruction code needs some small improvements

PR #1115, opened 2026-07-07
[CI] Raise errors from deprecation warnings

sjoerd-bouma:
In bad practice, this PR does several things:

- Replaces the deprecated `matplotlib.cm.get_cmap` in the `LOFAR.pipelineVisualizer_LOFAR` module (raises errors on newer matplotlib versions). Minor improvements to docstrings.
- Moves a CI test file to `NuRadioReco/test/data` to reduce clutter and make caching simpler
- main change: set `DeprecationWarning`s to raise errors (only in the CI) by default.

This will hopefully help us catch things breaking from deprecations _before_ they happen, and also make it easier to ensure we replace internally deprecated code immediately.

Finally, it turns out 'invalid escape sequences' in Python strings raise a `DeprecationWarning`. The solution is to make sure any string that contains a `\` is marked as a raw string with `r`. I am in two minds about whether we should actively enforce this (it can be tricky to debug, because any developers will not see this error locally unless they also set `DeprecationWarnings` to raise errors), so I've gone ahead and fixed all cases that were present currently, but have turned this particular error off for the CI going forward. For reviewers: all of these changes are contained in commit 658465e97a267701e1a94a7c58d4d352e6539ba8


sjoerd-bouma:
The one `DeprecationWarning` now raising an error in the tests serves as an example and is also already addressed in #1113 which is why I have not tried to fix it here.

fschlueter (dismissed):

fschlueter:
That error seems like the intend of this PR. Shall we fix it here or in a dedicated PR

fschlueter (approved):
Looks like the deprecation error is already being addressed in #1113. Hence this could be merged.

PR #1119, opened 2026-07-09
Loop over sorted channel ids to calcualte noise vrms. Potentially cha…

fschlueter:
…nges what is stored in HDF5 under Vrms.

The commit message already says it. This might break the validation as a different channels Vrms is stored in the generic field `Vrms` in the HDF5 files. For the channel specific field I think we already sorted it in a PR a while ago. 

This is an actual bug because you do not really know for which channel `Vrms` steams. It depends on what `self._det.get_channel_ids(station_id))` returns with no guarantee that it is sorted.

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -1307,7 +1307,7 @@ class simulation:
             self._integrated_channel_response_normalization[station_id] = {}
             self._max_amplification_per_channel[station_id] = {}
 
-            for channel_id in self._det.get_channel_ids(station_id):
+            for channel_id in np.sort(self._det.get_channel_ids(station_id)):
                 ff = np.linspace(0, 0.5 * self._config['sampling_rate'], 10000)
                 filt = np.ones_like(ff, dtype=complex)
                 for i, (name, instance, kwargs) in enumerate(evt.iter_modules(station_id)):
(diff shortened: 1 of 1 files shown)


fschlueter:
Interesting. I thought this would fail the CI validation. I tested it locally and the hdf5 files were different in vrms. It might depend on the detector and whether it returns (by conicidence) an sorted list of the channel ids.

@cg-laser, @sjoerd-bouma any suggestion how to proceed? I am confident that this PR *can* change use simulations ...

as context the problematic lines are:
```
        self._bandwidth = next(iter(next(iter(self._integrated_channel_response.values())).values()))  # get value of first station/channel key pair
        self._Vrms = next(iter(next(iter(self._Vrms_per_channel.values())).values()))
        self._Vrms_efield = next(iter(next(iter(self._Vrms_efield_per_channel.values())).values()))
```
so depending on how `self._det.get_channel_ids(station_id))` sorts things a different channels Vrms and bandwidth are stored in the HDF5 attributes.

fschlueter:
Possible other solution - enforce that `self._det.get_channel_ids(station_id))` returns sorted channel ids (but we would need that to be checked for every detector class...

sjoerd-bouma:
> Possible other solution - enforce that `self._det.get_channel_ids(station_id))` returns sorted channel ids (but we would need that to be checked for every detector class...

The base detector class (and therefore also the generic detector) returns sorted channel ids. I haven't checked the RNO-G detector but it would be sensible if it did the same?

sjoerd-bouma (approved):
I guess this doesn't harm, except if some future detector intentionally returns specific channel ids before others and this does not correspond to a simple increasing order. This seems unlikely. On the other hand, I am not aware of any ambiguity / bugs that this solves as mentioned in my previous comment. Feel free to either merge or close.

fschlueter:
The RNO-G detector obviously does not return sorted channel ids. This could indeed be done there... We already sort the channels when adding the vrms per channel to the station object attributes...

PR #1120, opened 2026-07-09
Add caching to the raytracing class to avoid repeating the raytracer …

fschlueter:
…calculation for a set of consecutive identical source/receiver pairs

This PR implements caching into the raytracer class. This caching is only "active" if consecutive calls (`propagator.set_start_and_end_point(x1, x2)`, `propagator.find_solutions()`) have the exact same `x1` and `x2`.  This can happen for example for `nu_e` cc interactions. We are not only caching the result of `find_solutions` but also the attenuation and focusing calculation. In particular the latter is relevant as currently we are calculating it twice per shower (!) once in `apply_propagation_effects` and once in `get_raytracing_output` (so for a `nu_e`-cc we are calculating the same thing 4 times). 

Note: The raytracing solution of a `nu_e`-cc event is only equal for both showers since we are **not** currently taking the shower Xmax in raytracing into account. If that changes this feature needs to be revisited.

### DIFF
--- a/NuRadioMC/simulation/simulation.py
+++ b/NuRadioMC/simulation/simulation.py
@@ -154,8 +154,12 @@ def calculate_sim_efield(
         x1 = shower.get_parameter(shp.vertex)
         if distance_cut is not None:
             time_logger.start_time('distance cut')
+            # Aggregate energy of showers which are close to the shower in question
             mask_shower_sum = np.abs(vertex_distances - vertex_distances[iSh]) < config['speedup']['distance_cut_sum_length']
             shower_energy_sum = np.sum(shower_energies[mask_shower_sum])
+
+            # Check with the aggregated energy whether the shower could be detected.
+            # (This does not skip showers which are to close to a other shower!)
             if np.linalg.norm(x1 - x2) > distance_cut(shower_energy_sum):
                 time_logger.stop_time('distance cut')
                 continue
@@ -168,14 +172,16 @@ def calculate_sim_efield(
         n_index = medium.get_index_of_refraction(x1)
         cherenkov_angle = np.arccos(1. / n_index)
 
+        if config['speedup']['redo_raytracing']:
+            # force a recalculation even if solutions for the same geometry (e.g., from a previous
+            # shower at the same position) are still stored in the propagator
+            propagator.reset_solutions()
+
         propagator.set_start_and_end_point(x1, x2)
         propagator.use_optional_function('set_shower_axis', shower_direction)
-        if config['speedup']['redo_raytracing']:  # check if raytracing was already performed
-            pass
-            # TODO: initiatlize ray tracer with existing results if available
-
         propagator.find_solutions()
         time_logger.stop_time('ray tracing')
+
         if not propagator.has_solution():
             logger.debug("shower %d and station %d, channel %d from %s to %s does not have any ray tracing solution",
                          shower.get_id(), station_id, channel_id, x1, x2)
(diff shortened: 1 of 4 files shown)


fschlueter:
On a set of 100000 `nu_e` (cc and nc) events and ice-air raytracing this improved the speed by +40% - I believe the effect will be even bigger for in-ice air showers. (Using alvarez2009 and a simple trigger)

NuRadioMC/simulation/simulation.py
```diff
@@ -168,14 +172,16 @@ def calculate_sim_efield(
         n_index = medium.get_index_of_refraction(x1)
         cherenkov_angle = np.arccos(1. / n_index)
 
+        if config['speedup']['redo_raytracing']:
+            # force a recalculation even if solutions for the same geometry (e.g., from a previous
+            # shower at the same position) are still stored in the propagator
+            propagator.reset_solutions()
+
```
sjoerd-bouma, line 179: Looks like you're reusing a config parameter for something different than it was originally intended for (even though what it was intended to do was never implemented). I have mixed feelings about this; I feel like your use case is useful for debugging the caching feature, but not actually useful from a user perspective (c.f. previously where one might not want to run raytracing if re-running a simulation with the same raytracer but want to redo the raytracing if e.g. the ice model has changed).

As it concerns a not-implemented feature previously I can live with keeping things this way, at least until someone implements the originally intended feature, in which case we may have to change what this does again : )

sjoerd-bouma (dismissed):
This is a really nice improvement, thanks Felix! I guess both cases (cached and non-cached solutions) are already sufficiently covered by the existing CI, so I think this can safely be merged.

NuRadioMC/simulation/simulation.py
fschlueter: oh - I was honestly not aware of this. I thought the missing implementation was exactly for this (disable caching once it got implemented. I am happy to change it - your call (but please say something so we can get this done)

fschlueter:
I want to mention that @philippwindischhofer wrote me privately and wondered if the caching for the attenuation and focusing with only `C0` and not `x1` and `x2` is sufficient. I think in practice yes but I am also happy to change it if people think differently!

NuRadioMC/simulation/simulation.py
sjoerd-bouma: I think this may have been implemented in the previous version of simulation.py: if re-simulating from an output HDF5 file (which already contains the ray-tracing solutions), it would enable re-using the already calculated ray-tracing solutions. A user might want to turn this off if the second simulation uses a slightly different raytracer / ice model.

In contrast, I can't think of any reason to ever _not_ use the caching you've implemented, so I would be inclined to remove this option.

sjoerd-bouma:
> I want to mention that @philippwindischhofer wrote me privately and wondered if the caching for the attenuation and focusing with only `C0` and not `x1` and `x2` is sufficient. I think in practice yes but I am also happy to change it if people think differently!

In practice, yes; there is a purely theoretical chance for a different geometry to have the same C0 (but different path lengths and attenuation etc), e.g. if `x3` lies on the path between `x1` and `x2`, so using `x1` and `x2` for caching is technically more correct (which as we all know is the best kind of correct).

philippwindischhofer:
> In practice, yes; there is a purely theoretical chance for a different geometry to have the same C0 (but different path lengths and attenuation etc), e.g. if x3 lies on the path between x1 and x2, so using x1 and x2 for caching is technically more correct (which as we all know is the best kind of correct).

Yeah, my point was that a caching mechanism should not affect the results. If an unsuspecting someone were to calculate attenuation as a function of distance along a ray, they would get very confused.

All that is to say that I'd sleep better if `x1` and `x2` were used to index the cache.

fschlueter:
Given that this PR was already approved. I am merging it.

PR #1123, opened 2026-07-09
Reorganize neutrino simulation strategy.

fschlueter:
Instead of running apply_det_response_sim before the trigger decision, we now run it afterwards to save time. However, this means that we have to use the efield time + cable delay stored in the detector to split events. This ignores possible group delays. In all cases where the reponse has no meaning full phase delay - which it should not! - this should not change anything


cg-laser:
this is a dangerous change as it introduces a very specific approximation that might work now but can easily lead to problems later. It's also against the NuRadio philosophy. The time delays are defined by the modules that are being executed. In this PR, a single parameter from the detector description is picked and assumed that this determines the time delay. 
I also doubt that this really saves substantial computing time. Did you benchmark it? 

Just to give one example, with the new all digital DAQs, we might introduce large digital delays.

fschlueter:
Running `NuRadioMC/examples/RNO_G_trigger_simulation/simulate.py -it nc -n 10000 --seed 2004 --station_id 23 -e 1e19 -i 10`: I got the following run times (reported from the time_logger, ignores setup time):

ARZ2020:
- develop: 8m35s
- this branch: 8m6s (-6%)
Alvarez2009:
- develop: 2m3s
- this branch: 1m32s (-25%)

fschlueter:
I added a calculation of the group delay in the init to warn if the differences across channels is large `self._pulse_based_group_delay_per_channel`. We could take it and added to the times in the event splitter function. Obviously this group delay is only calcualted for a fixed zenith angle which might not be well suited for every channel.

fschlueter:
> Just to give one example, with the new all digital DAQs, we might introduce large digital delays.

If those delays are the same for each channel they cancel. 

I acknowledge that this does "cut a corner" - but given how many corners we cut I thought the performance gain and generally cleaned structure is worth it. My end goal would be to not have to "treat" trigger and non-trigger channels differently (i.e., using the same code for their simulation) but this requires some more work.

fschlueter:
Closing the PR for now. Maybe we want to revisit this idea later in the future.

PR #1125, opened 2026-07-13
ParameterStorage.get_parameter now returns a copy of mutable values

fschlueter:
## Summary

`ParameterStorage.get_parameter` (and the `[]` operator, which is built on it) returned mutable parameter values (`list`, `dict`, `numpy.ndarray`, ...) by reference. Any in-place mutation of the returned value — `station[stnp.dirty_fft_channels].append(...)`, `efield[efp.max_amp_antenna][ch_id] = x`, etc. — silently mutated the stored parameter too, with no `set_parameter` call anywhere in sight. `get_parameters()` (plural) already guards against this by unconditionally `copy.deepcopy`-ing; the singular getter never got the same treatment.

- `get_parameter(self, key, copy=True)` now deep-copies mutable values before returning them. Values of immutable types (`int`, `float`, `complex`, `bool`, `str`, `bytes`, `None`) are returned as-is, since they can't be mutated in place and copying them would just be overhead for no safety benefit.
- `copy=False` is an explicit opt-out for read-only, performance-sensitive call sites that want the stored object itself.
- `set_parameter` is intentionally untouched — it still stores whatever reference it's given, same as `dict.__setitem__`. Symmetric copy-on-set is a separate discussion.

### Is this a behavior change or a bug fix?

The old reference-sharing behavior could be read as intentional, since one module (`calculateAmplitudePerRaySolution.py`) actively depended on `get_parameter`/`[]` handing back the *live* stored object so it could mutate it in place — see below. But that's one caller leaning on an implicit, undocumented aliasing contract, not the norm. Every other call site in the codebase treats the return value as a read-only snapshot, which is what most users of a `get_x()`-style accessor would expect by default, and is exactly the contract `get_parameters()` (plural) already documents.

So this PR treats the old aliasing as a latent bug one module happened to depend on, not a feature being taken away — and the old zero-copy behavior remains available, just opt-in via `get_parameter(key, copy=False)`, for anyone who genuinely wants it (e.g. a hot loop mutating a stored container in place on purpose).

### Fallout found while auditing the codebase

Grepped for code that fetches a mutable parameter and mutates it in place without an explicit `set_parameter` afterwards — that pattern would silently break once `get_parameter` stops returning a live reference.

Found one real instance: `NuRadioReco/modules/custom/deltaT/calculateAmplitudePerRaySolution.py` initialized `efp.max_amp_antenna`/`efp.max_amp_antenna_envelope` to `{}` once, then did `efield[efp.max_amp_antenna][channel_id] = maximum` on every subsequent channel, relying on the returned dict being the actual stored object. Fixed to fetch, mutate, and write back explicitly.

Everywhere else this pattern appears (`stationRFIFilter.py`, `planeWaveDirectionFitter_LOFAR.py`, `channelTemplateCorrelation.py`, ...) the code already calls `set_parameter` again after mutating the fetched value, so those were already correct and are unaffected by this change.

## Test plan

- [x] Added `NuRadioReco/test/framework/test_parameter_storage.py`: list/dict/numpy-array parameters are copied on get, nested mutable structures are deep- not shallow-copied, the `[]` operator also copies, scalar parameters are returned by identity (no copy overhead), `copy=False` returns the exact stored object, and the get→mutate→set round trip used elsewhere in the codebase still works. Wired into CI (`other-tests` job).
- [x] Ran `NuRadioReco/test/likelihood_reconstruction/test_efield_reconstruction.py` (exercises `get_parameter`/`[]` on `ElectricField` extensively, checked against fixed reference values) — passes unchanged.
- [x] Ran `NuRadioReco/test/check_modules.py -r` — no new breakage.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/NuRadioReco/framework/parameter_storage.py
+++ b/NuRadioReco/framework/parameter_storage.py
@@ -13,6 +13,15 @@ import itertools
 import logging
 logger = logging.getLogger('NuRadioReco.framework.parameter_storage')
 
+# Module-level alias so `get_parameter` can use a `copy` keyword argument
+# without shadowing the `copy` module inside its own body.
+_deepcopy = copy.deepcopy
+
+# Types that can never be mutated in place, so returning them without copying
+# is always safe. Everything else (list, dict, set, numpy.ndarray, custom
+# objects, ...) is deep-copied by `get_parameter` unless `copy=False`.
+_IMMUTABLE_SCALAR_TYPES = (int, float, complex, bool, str, bytes, type(None))
+
 
 class ParameterStorage:
     """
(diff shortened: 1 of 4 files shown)


fschlueter:
This PR could be merged, the failing CI is not related and will be addressed in the linked PR.

NuRadioReco/framework/parameter_storage.py
```diff
@@ -13,6 +13,15 @@
 import logging
 logger = logging.getLogger('NuRadioReco.framework.parameter_storage')
 
+# Module-level alias so `get_parameter` can use a `copy` keyword argument
+# without shadowing the `copy` module inside its own body.
+_deepcopy = copy.deepcopy
+
```
sjoerd-bouma, line 19: Why not just `from copy import deepcopy`

sjoerd-bouma (approved):
Thanks, good spot Felix/Claude!

fschlueter:
> Thanks, good spot Felix/Claude!

I take credit for the idea (not the implementation :D)

NuRadioReco/framework/parameter_storage.py
fschlueter: I guess to keep prevent deepcopy to be overwritten in the future or when parameter_storage is imported elsewhere - but honestly would have probably be a good solution also

PR #1126, opened 2026-07-13
Use absolute tolerance for timing observables in reference-comparison tests

fschlueter:
## Summary

`T05validate_nur_file.py` (part of the "Single event simulation" CI job) intermittently fails on the trace-start-time check, unrelated to whatever change is actually under test. Confirmed by comparing two CI runs:

- PR #1125 (`fix/parameter-storage-copy`, unrelated to raytracing): failed with `Trace start times are not equal (maximum difference: 5.373189378587995e-05)`, `8/136` elements mismatched.
- `develop` itself, run [29078118764](https://github.com/nu-radio/NuRadioMC/actions/runs/29078118764) on 2026-07-10, no PR involved: failed with the **exact same numbers to the last digit** (same max diff, same mismatch count, same actual/desired arrays).

Since two unrelated commits produce byte-identical "failures", this isn't noise from either change — it's a deterministic mismatch between the checked-in reference `.nur` file and whatever GSL version the `ubuntu-latest` runner happens to have at a given time. It shows up as "intermittent" only because GitHub periodically rolls the runner image (and its bundled GSL) forward.

## Root cause

Trace start times derive from the raytraced travel time, computed in `get_travel_time()` (`NuRadioMC/SignalProp/CPPAnalyticRayTracing/analytic_raytracing.cpp:255-270`):

```cpp
gsl_integration_qags(&F, pos[1], x2_mirrored[1], 0, 10.e-7, 1000, w, &result, &error);
```

`epsrel = 1e-6`, fixed, unchanged since 2018. On an ~11,747 ns travel time, a 1e-6 relative tolerance permits ~12 ps of residual error that GSL's adaptive Gauss-Kronrod quadrature considers "converged". Exactly where within that budget it stops depends on GSL's internal implementation, which differs subtly across GSL versions/platforms — the observed ~5e-5 ns (4e-9 relative) drift sits comfortably inside that budget.

## Two ways to fix the root problem

1. **Tighten the C++ side**: lower `epsrel` in `get_travel_time` (e.g. to `1e-9`) so the integrator converges far past where cross-version GSL noise lives. Shrinks the noise floor itself, at the cost of slightly slower travel-time integration and a C++ rebuild + reference regeneration. Left as a follow-up.
2. **Loosen the test (this PR)**: check timing observables with an absolute tolerance instead of relative.

## What changed

Beyond the original trace-start-time fix, this PR generalizes the same reasoning to every golden-reference comparison script that checks timing observables (`travel_times`, `trigger_times`, `trigger_times_per_event`) with a purely relative tolerance:

- `NuRadioMC/test/SingleEvents/T05validate_nur_file.py` — trace-start-time check decoupled from the trace-amplitude check (they shared one `decimal=precision`), given its own `decimal=3` (~1.5 ps absolute).
- `NuRadioMC/test/SingleEvents/T04validate_allmost_equal.py` (South Pole + Moore's Bay single-event CI tests) — `travel_times`, `trigger_times`, `trigger_times_per_event` split out into absolute-tolerance checks against `time_accuracy = 1e-3 ns`.
- `NuRadioMC/test/examples/validate_radio_emitter_allmost_equal.py` — same split for `travel_times`, `trigger_times`.
- `NuRadioReco/test/trigger_tests/compare_to_reference.py` — `trigger_time` now uses `assert_allclose(..., atol=1e-3 ns, rtol=0)` instead of the default `rtol=1e-7, atol=0`.

A relative tolerance is physically the wrong tool for these: the noise floor is an absolute quantity set by the integrator's convergence criterion, independent of the (arbitrarily large) travel/trigger time itself. `1e-3 ns` gives ~10-20x headroom over the largest drift observed so far (~9e-5 ns), while still being ~1000x smaller than a single sample bin at typical GHz sampling rates — tight enough to catch a real regression, loose enough to absorb GSL-version noise.

**Follow-up commit — de-duplicated the comparison helpers.** `T04validate_allmost_equal.py` and `validate_radio_emitter_allmost_equal.py` had grown near-identical copies of the same "compare two HDF5 files for (almost) equality" helper functions, which is exactly what caused the duplication when the absolute-tolerance variants above were added to both independently. Extracted all of them (equal/almost-equal × relative/absolute × file-level/per-station) into `NuRadioMC/utilities/test_helpers.py` as plain functions, and both scripts now import from there instead of defining their own copies. In the process, folded a minor inconsistency: the emitter script's zero-entry equality check didn't treat NaN and 0 as equivalent "no solution" sentinels the way the single-events script's did (for `max_amp_shower_and_ray` specifically) — the shared helper now does this for all keys, a strict relaxation that's verified not to change either script's pass/fail outcome. `T05validate_nur_file.py` and `compare_to_reference.py` weren't touched by this — their comparison logic is structurally different (raw trace/JSON diffing, not HDF5-key iteration) and didn't share code with the other two to begin with.

**Deliberately left unchanged:**
- `T02test_analytic_D_T.py` / `T04MooresBay.py` — these compare the numeric integrator against the analytic formula *within a single run* (no stored reference involved), so their partial-relative tolerance reflects genuine, deliberate accumulation of approximation error between the two methods as path length grows, not cross-platform GSL noise. Converting them to a pure absolute tolerance would make them more fragile for no real benefit.
- `NuRadioMC/test/SingleEvents/T03validate.py` — already uses an absolute-dominant tolerance for timing keys (`rtol=1e-9, atol=12 ps`) and needs no change, though note it isn't currently wired into any CI workflow (`validate.sh` is unused by `.github/workflows`).

## Test plan
- [x] Reproduced the exact CI failure locally by diffing test output between `develop` and an unrelated PR branch, confirming byte-identical numbers.
- [x] Re-ran `T02RunSimulation.py` + `T05validate_nur_file.py` for the South Pole single-event test with the new tolerance — passes.
- [x] Re-ran `T04validate_allmost_equal.py` for the South Pole single-event test — timing keys now pass (the one remaining local mismatch, `max_amp_shower_and_ray` on event 17, is a separate pre-existing amplitude-noise flake, unrelated to this change and not present in the actual CI run for PR #1125).
- [x] Re-ran the ARA radio-emitter pulser example end-to-end against `validate_radio_emitter_allmost_equal.py` — passes.
- [x] Re-ran the full trigger-test pipeline against `compare_to_reference.py` — passes.
- [x] Re-verified `T04validate_allmost_equal.py` and `validate_radio_emitter_allmost_equal.py` against real data after extracting the shared helpers — identical pass/fail outcome to before the refactor.
- [x] CI run on this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)


fschlueter:
Thinking about it - probably it makes sense to change all CI to use only absolute tolerance for time based variables ...

fschlueter:
> Thinking about it - probably it makes sense to change all CI to use only absolute tolerance for time based variables ...

done

NuRadioMC/utilities/_test_helpers.py
sjoerd-bouma, line 1: This shouldn't be a public module, I think. Maybe it could even live in `NuRadioMC/test` to minimize clutter, but otherwise this should at most be a private module in `utilities`.

fschlueter: done

sjoerd-bouma (dismissed):
Still need to fix a merge conflict but otherwise I think this is fine. I really hope I (and you) haven't missed an unintended factor of 1e6 increase in the test tolerance somewhere but I didn't spot one.

sjoerd-bouma:
I also probably would prefer a direct comparison between two test runs / branches where one fails and one succeeds rather than two which both fail in the same way - I agree it's probably platform-dependent, but it would have been interesting to know if it's due to e.g. GSL or something else (from a brief bit of archaeology it looks like the GSL versions on failed and successful test runs are the same, so it probably is something else, but I still haven't been able to pin it down).

sjoerd-bouma (approved):

PR #1133, opened 2026-07-16
Refactor (reorganize) analytic raytracer code

fschlueter:
This PR aims at replacing #1087. Currently based  to #1028. Will swap that once 1028 is merged. I specifically asked the AI to use git mv and keep the git history intakt as much as possible. While this PR may not show it but `git blame/git log --follow` on `single_layer_analytic_raytracer` should track the changes from the original file.


fschlueter:
This PR is good to go in my opinion.

NuRadioMC/utilities/medium_base.py
```diff
@@ -380,17 +380,13 @@ class IceModelBirefringence(IceModelSimple):
...
+        Function to load a birefringent ice model with an interpolation of the data as described in:
         https://link.springer.com/article/10.1140/epjc/s10052-023-11238-y
-
-        Parameters
-        ----------
-        bire_model: string
-            choose the interpolation to fit the measured refractive index data
-            options include (A, B, C, D, E) description can be found under: NuRadioMC/NuRadioMC/utilities/birefringence_models/model_description
-
         """
```
sjoerd-bouma, line 403: Please fix the docstrings, adding back a (ideally correct) parameter description. I'm also not sure I see the point of having a public `load_birefringence_model` if this is always called by `__init__` and shares the same parameters, I think this should probably be private at best.

NuRadioMC/utilities/birefringence.py
```diff
@@ -0,0 +1,218 @@
+"""
+This module contains helper functions to calculate the effective refractive
+indices and the polarization vectors for a birefringent medium.
+The calculations are described here: https://link.springer.com/article/10.1140/epjc/s10052-023-11238-y
+"""
```
sjoerd-bouma, line 6: 👍 for the module docstring, could still be improved by using the 
```
Helper functions to calculate the effective refractive index and polarization vectors for a birefringent medium

The calculations...
```
format (single line, empty line, longer description).

sjoerd-bouma (commented):
I don't think `AnalyticRayTracingImpl` is a great name for a (sub)package, and also am not sure it should be public; my feeling is private would be better (I can't think of a reason why users would want to call the 2D raytracer code directly instead of just using the 3D wrapper). Thoughts?

NuRadioMC/utilities/medium_base.py
fschlueter: somewhere in the code I saw a `obj.__init__(...)` so they missused the init function because there was not public interface to reload a model

fschlueter:
> I don't think `AnalyticRayTracingImpl` is a great name for a (sub)package, and also am not sure it should be public; my feeling is private would be better (I can't think of a reason why users would want to call the 2D raytracer code directly instead of just using the 3D wrapper). Thoughts?

I mean I chose the name (so I am not unbiased). I thought about making in private and decided against it. Since even if users go through the wrapper I think exposing the class makes sense for users.. In principle I do not mind renaming - although its a bit of an effort.

fschlueter:
@sjoerd-bouma thanks for your feedback. Apparently nobody else has thoughts... would you like to discuss that more? Maybe @philippwindischhofer and @hanneswarnhofer can also wave in. Otherwise I would prefer to get that merge asap

NuRadioMC/utilities/medium_base.py
sjoerd-bouma: I 'fixed' the docstrings, let me know if this is okay for you.

sjoerd-bouma:
> > I don't think `AnalyticRayTracingImpl` is a great name for a (sub)package, and also am not sure it should be public; my feeling is private would be better (I can't think of a reason why users would want to call the 2D raytracer code directly instead of just using the 3D wrapper). Thoughts?
> 
> I mean I chose the name (so I am not unbiased). I thought about making in private and decided against it. Since even if users go through the wrapper I think exposing the class makes sense for users.. In principle I do not mind renaming - although its a bit of an effort.

How about just `ray_tracing_2D` or `backends` or `implementations` or similar? Shorter, more to the point (I hope). It feels like you're trying to cram too many words in the module description, from the parent path it is already clear that this is a ray tracer, I think.

But we can also postpone this discussion to #1077 and change it again there.

documentation/source/NuRadioMC/pages/Manuals/signal_propagation.rst
```diff
@@ -104,7 +104,8 @@ The following code shows how to perform a analytic ray tracing and extract infor
   .. code-block:: Python
 
     from NuRadioMC.SignalProp import propagation
-    from NuRadioMC.SignalProp.analyticraytracing import solution_types, ray_tracing_2D
+    from NuRadioMC.SignalProp.propagation import solution_types
+    from NuRadioMC.SignalProp.analyticraytracing import ray_tracing_2D
```
sjoerd-bouma, line 108: I guess this is not really where this lives anymore, though this import will still work because it is also imported in `analyticraytracing`.

sjoerd-bouma (approved):
one other option would be to make `Analytic...Impl` private but explicitly expose the `ray_tracing_2D` class in `analyticraytracing` using the `__all__` trick. But seeing as we'll need to have the private/public discussion once more for the multilayer raytracer, let's postpone to #1077 so this can get merged and we aren't constantly fixing merge conflicts in that PR.

I'm trusting my memory that I reviewed this entire PR, because other than the above I think all my comments have now been addressed.

documentation/source/NuRadioMC/pages/Manuals/signal_propagation.rst
fschlueter: shall we fix it?

NuRadioMC/utilities/medium_base.py
fschlueter: thanks!

fschlueter:
Skipping the usual 24h period to keep going ..

PR #1141, opened 2026-07-22
Rnog det and reponse

fschlueter:
Mostly trying to make life for people who import the RNO-G detector by file the life a bit better. Stumbled across issues with the window functions in NuRadio - try to make it more robust ...

### DIFF
--- a/NuRadioReco/detector/RNO_G/rnog_detector.py
+++ b/NuRadioReco/detector/RNO_G/rnog_detector.py
@@ -476,7 +476,7 @@ class Detector():
 
         self.__detector_time = time
 
-        if not self._det_imported_from_file: # 
+        if not self._det_imported_from_file:
             if self.__db is None:
                 self.logger.error("Database is None.")
                 raise ValueError("Database is None.")
--- a/NuRadioReco/detector/RNO_G/rnog_detector.py
+++ b/NuRadioReco/detector/RNO_G/rnog_detector.py
@@ -160,6 +160,7 @@ class Detector():
             self.logger.debug(
                 "Collect time periods of station commission/decommission ...")
             self._time_periods_per_station = self.__db.query_modification_timestamps_per_station()
+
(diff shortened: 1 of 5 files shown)


fschlueter:
@philippwindischhofer or @zsselcuk would you like me to elaborate on the changes?

- It allows to work with a detector file and while a station might be dropped because it became outdate it does not raise an error unless it was the last station.
- Make the window function, which be default is used in the __call__ operator of the response class, more robust against failure. does not fix the underlying issue which is really only fixed if the original response is longer
- improves logging in the detector class

NuRadioReco/detector/RNO_G/rnog_detector.py
```diff
@@ -489,25 +528,36 @@ def update(self, time):
...
-            raise ValueError(
-                "You have imported the detector description from a pickle/json file but it is not valid anymore. Full stop!")
+            for station_id, need_update in update_buffer_for_station.items():
+                if need_update:
+                    self.logger.warning(
+                        f"Station {station_id} is not valid anymore at {_format_t(time)} but the detector description "
+                        "was imported from a pickle/json file and can not be updated. Dropping this "
+                        "station's description entirely.")
+                    # remove everything (could be handled smarter ...)
+                    self.__buffered_stations[station_id] = {}
```
philippwindischhofer, line 544: What's the advantage of doing this vs `del self.__buffered_stations[station_id]`?

philippwindischhofer (commented):
Hi Felix, thanks for cleaning this up! Looks good to me modulo the one question I posted.

NuRadioReco/detector/RNO_G/rnog_detector.py
fschlueter: No idea, did not thought about it...

fschlueter:
Taking philipps comment as approval.

PR #1142, opened 2026-07-22
Add script which helpts to visualize what is configured in the RNO-G …

fschlueter:
…detector database

### DIFF
--- a/NuRadioReco/detector/RNO_G/db_mongo_read.py
+++ b/NuRadioReco/detector/RNO_G/db_mongo_read.py
@@ -1247,10 +1247,12 @@ class Database(object):
         -------
 
         station_data: dict(dict(list))
-            Returns for each station (key = station.id) a dictionary with three entries:
-            "modification_timestamps", "station_commission_timestamps", "station_decommission_timestamps"
-            each containing a list of timestamps. The former combines the latter two + channel (de)comission
-            timestamps.
+            Returns for each station (key = station.id) a dictionary with six entries:
+            "modification_timestamps", "station_commission_timestamps", "station_decommission_timestamps",
+            "channel_modification_timestamps", "calibration_modification_timestamps"
+            each containing a list of timestamps. "modification_timestamps" combines all the others.
+            "channel_modification_details" maps each channel modification timestamp to a list of
+            `(channel_id, "commissioned" | "decommissioned")` tuples describing which channel(s) changed.
         """
         # get distinct set of stations:
         if isinstance(station_ids, int):
@@ -1265,11 +1267,32 @@ class Database(object):
             station_times_comm = self.db[self.__station_collection].distinct("commission_time", {"id": station_id})
             station_times_decomm = self.db[self.__station_collection].distinct("decommission_time", {"id": station_id})
 
-            # get set of (de)commission times for channels
-            channel_times_comm = self.db[self.__station_collection].distinct("channels.commission_time", {"id": station_id})
-            channel_times_decomm = self.db[self.__station_collection].distinct("channels.decommission_time", {"id": station_id})
+            # get set of (de)commission times for channels, keeping track of which channel(s)
+            # changed at each timestamp
+            channel_docs = self.db[self.__station_collection].aggregate([
+                {"$match": {"id": station_id}},
+                {"$unwind": "$channels"},
+                {"$project": {
+                    "_id": 0,
+                    "channel_id": "$channels.id",
+                    "commission_time": "$channels.commission_time",
+                    "decommission_time": "$channels.decommission_time",
+                }},
+            ])
 
-            # # get set of (de)commission times for calibrations
+            channel_modification_details = {}
+            for doc in channel_docs:
+                channel_modification_details.setdefault(doc["commission_time"], []).append(
+                    (doc["channel_id"], "commissioned"))
+                channel_modification_details.setdefault(doc["decommission_time"], []).append(
+                    (doc["channel_id"], "decommissioned"))
+
+            for changes in channel_modification_details.values():
+                changes.sort()
+
+            channel_times_comm_decomm = list(channel_modification_details.keys())
+
+            # get set of (de)commission times for calibrations
             available_ids = self.db[self.__gain_calibration_collection].distinct("id")
             for id in available_ids:
                 if not id.startswith("sta"):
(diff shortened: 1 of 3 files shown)


fschlueter:
probably we should add a unit test for this script - on the other hand this adds another query to a desy server..

fschlueter:
@zsselcuk would you be able to review this PR please.

NuRadioReco/detector/RNO_G/db_mongo_read.py
```diff
@@ -1247,10 +1247,12 @@ def query_modification_timestamps_per_station(self, station_ids=None):
         -------
 
         station_data: dict(dict(list))
-            Returns for each station (key = station.id) a dictionary with three entries:
-            "modification_timestamps", "station_commission_timestamps", "station_decommission_timestamps"
-            each containing a list of timestamps. The former combines the latter two + channel (de)comission
-            timestamps.
+            Returns for each station (key = station.id) a dictionary with six entries:
+            "modification_timestamps", "station_commission_timestamps", "station_decommission_timestamps",
+            "channel_modification_timestamps", "calibration_modification_timestamps"
```
zsselcuk, line 1252: `"channel_modification_details"` is missing

NuRadioReco/detector/RNO_G/db_mongo_read.py
```diff
@@ -1247,10 +1247,12 @@ def query_modification_timestamps_per_station(self, station_ids=None):
...
+            each containing a list of timestamps. "modification_timestamps" combines all the others.
```
zsselcuk, line 1253: Maybe we should add "except for `"channel_modification_details"`"?

NuRadioReco/detector/RNO_G/db_mongo_read.py
```diff
@@ -1265,11 +1267,32 @@ def query_modification_timestamps_per_station(self, station_ids=None):
...
+                    "_id": 0,
+                    "channel_id": "$channels.id",
+                    "commission_time": "$channels.commission_time",
+                    "decommission_time": "$channels.decommission_time",
+                }},
+            ])
+
+            channel_modification_details = {}
+            for doc in channel_docs:
+                channel_modification_details.setdefault(doc["commission_time"], []).append(
```
zsselcuk, line 1285: I think making a dict with channel ids as keys and a list of lists with commission & decommission dates could be more understandable, maybe something like this:
```
{'0': [[datetime.datetime(2022, 7, 16, 0, 0),
   datetime.datetime(2024, 1, 1, 0, 0)],
  [datetime.datetime(2024, 1, 1, 0, 0), datetime.datetime(2024, 6, 23, 0, 0)],
  [datetime.datetime(2024, 6, 25, 0, 0), datetime.datetime(2080, 1, 1, 0, 0)]]
  ...
}
```
This could also handle cases where the decommission date of the earlier deployment coincides with the commission date of the later deployment. An example from station 11 using the current version :
```
datetime.datetime(2024, 1, 1, 0, 0): [(0, 'commissioned'),
   (0, 'decommissioned'),
   (1, 'commissioned'),
   (1, 'decommissioned'),
   (2, 'commissioned'),
   (2, 'decommissioned'),
   ...
  ]
```

NuRadioReco/detector/RNO_G/db_mongo_read.py
```diff
@@ -1286,16 +1309,25 @@ def query_modification_timestamps_per_station(self, station_ids=None):
...
+                channel_times_comm_decomm +
                 calibration_times_comm + calibration_times_decomm).tolist()
+
             mod_set.sort()
             station_times_comm.sort()
             station_times_decomm.sort()
 
+            channel_mod_set = sorted(channel_times_comm_decomm)
+
+            calibration_mod_set = np.unique(calibration_times_comm + calibration_times_decomm).tolist()
```
zsselcuk, line 1321: The `calibration_mod_set` returns all timestamps (commission & decommission times). I think also for this, there should be a `calibration_modification_details` dict similar to one I suggested. In this dict, we can also add the version name as another entry/key/item.

NuRadioReco/detector/RNO_G/plot_detector_configuration_periods.py
```diff
@@ -0,0 +1,274 @@
...
+        Line2D([0], [0], color=COLOR_CHANNEL_CHANGE, lw=1.2, ls="--", label="channel change"),
+    ]
+    if show_calibration_changes:
+        legend_elements.append(
+            Line2D([0], [0], color=COLOR_CALIBRATION_CHANGE, lw=1.2, ls=":", label="calibration change"))
+    legend_elements.append(mpatches.Patch(facecolor="0.6", edgecolor="white", hatch="///", label="ongoing period"))
+    legend = ax.legend(handles=legend_elements, loc="upper left", bbox_to_anchor=(1.01, 1),
+                        frameon=False, fontsize=9)
+
+    # Place the period table (identical to `print_period_table`'s stdout output)
```
zsselcuk, line 227: This looks very small in the plot and the text it hard to read, might help to make it bigger

zsselcuk:
I think it would also be nice to add another dictionary, which reports on the primary_times of the calibrated responses. So that the user knows to which date they should set the database time to get the correct calibration version.

NuRadioReco/detector/RNO_G/plot_detector_configuration_periods.py
```diff
@@ -0,0 +1,274 @@
...
+        ongoing = decomm > now
+        end = min(decomm, now)
+
+        # modification timestamps strictly inside this commissioned interval
+        # split it into sub-periods (caused by a channel/calibration change)
+        boundaries = sorted(set([comm] + [t for t in mod_ts if comm < t < decomm] + [end]))
+
+        for i in range(len(boundaries) - 1):
+            is_last = i == len(boundaries) - 2
+            change_type = _classify_change(boundaries[i], station_data) if i > 0 else None
```
zsselcuk, line 114: Why `if i > 0 else None`?

NuRadioReco/detector/RNO_G/db_mongo_read.py
fschlueter: hm, I guess that depends a bit how we want to use it. Whether we have the time as info or the channel / station id. With that we could also start a smarter updating scheme ... 

However, can we refer this to future PR?

fschlueter:
> I think it would also be nice to add another dictionary, which reports on the primary_times of the calibrated responses. So that the user knows to which date they should set the database time to get the correct calibration version.

I agree, feel free to add one!

NuRadioReco/detector/RNO_G/plot_detector_configuration_periods.py
fschlueter: I guess for the first boundary (commissioning) you do not need a reason for the change

NuRadioReco/detector/RNO_G/plot_detector_configuration_periods.py
fschlueter: Well, but in future the entries will be more - would not fine tune it here.

NuRadioReco/detector/RNO_G/db_mongo_read.py
fschlueter: Again I would defer that to a future PR / once we need the extra information

PR #1143, opened 2026-08-05
Fix sign how readout delays are added to trace start time

fschlueter:
@zsselcuk has an analysis in place to check it. 

Readout delays should be added. And we did (to `time_offset` which we than subtract...)


fschlueter:
why did the unit-test job run as one monolithic job? Any idea @sjoerd-bouma

fschlueter:
> why did the unit-test job run as one monolithic job? Any idea @sjoerd-bouma

I think zeynep fixed it, maybe I checked out from main ...

fschlueter:
I guess we should add a log in the changelog..

fschlueter:
@zsselcuk can you approve?

zsselcuk (approved):

PR #1146, opened 2026-08-14
Correct the string labeling for station 14 & adjust the channel/station/device positions in the json

zsselcuk:
The current positions use the incorrect string labels. After the PR:

- New power string (station) position == old helper string B position
- New helper string B position == old helper string C position
- New helper string C position == old power string (station) position

The helper string C should be shorter; new depths were recalculated using the bott shack measurement at 85.10 m deep and channel depths (can be found in [handwritten deployment notes](https://radio.uchicago.edu/wiki/images/7/7c/RNO-G-2024-station-retrofit-and-installation-notes.pdf))


NuRadioReco/detector/RNO_G/RNO_season_2024.json
```diff
@@ -3860,13 +3860,13 @@
       "ant_orientation_theta": 0.0,
       "ant_rotation_phi": 90.0,
       "ant_rotation_theta": 90.0,
-      "ant_position_x": 34.01333993178059,
-      "ant_position_y": -45.178181287965344,
+      "ant_position_x": 13.375785235156172,
+      "ant_position_y": -9.730687616134674,
```
philippwindischhofer: How did you determine this position? The deployment notes say "[...] installed in the hole near northernmost helper string"—if I understand correctly, your placement of the surface Tx is different from that.

philippwindischhofer (commented):
Thanks a lot @zsselcuk!

philippwindischhofer:
@fschlueter let me take this opportunity to bring up a related point: for how much longer do we want to support the use of these `json` files over the DB? In my opinion, we should not keep storing the survey results in two places, as it is redundant and prone to accumulating inconsistencies. 

I think for the 2026 stations we should give preference to the DB and not produce the `json` files anymore ... but do we want to actively _remove_ them to discourage their use for the legacy stations?

NuRadioReco/detector/RNO_G/RNO_season_2024.json
zsselcuk: I first calculated the absolute position by adding the old station position (old power string location). And to get thet the new relative position, I subtracted the new station location (old helper string B) from that. But I didn't calculate the absoulute position myself using the deployment notes, I exported it from the json also from DB. Do you think the surface pulser location is incorrect as well?

philippwindischhofer: Yes, the way it is now it is not consistent with the deployment notes, correct?

zsselcuk: I am checking the positions and I think they were incorrect from the beginning

philippwindischhofer (approved):
Thanks for correcting this Zeynep!

zsselcuk:
The DB positions are now fixed as well

philippwindischhofer:
Thanks Zeynep! Attached is the timing closure using the new DB positions "out of the box".

[station_14_constraints.pdf](https://github.com/user-attachments/files/31374071/station_14_constraints.pdf)

zsselcuk:
Thanks! This looks great

PR #1147, opened 2026-08-16
Fix date formatting and reduce overhead in logging

fschlueter:
# One formatter per level, built once: constructing them in format() would
            # drop `datefmt` and allocate a Formatter per record.


philippwindischhofer (approved):
Looks good! I assume you tested it to make sure it works (I didn't)

PR #1148, opened 2026-08-20
Hotfix antennapattern; extend test suit

fschlueter:
This fixes an annoying bug which was included since the beginning. 

To find for e.g. a value of `theta` the nodes in the stored antenna pattern which bracket `theta` the code assumes an equidistant spacing of the `theta_nodes`. However, it is never checked that the nodes are equidistant spaced... 

This PR introduces a generic function to find the backeting indices. But since this is a bit slower, we fall back to the original code assuming equidistant nodes if the nodes are actually equidistant. 

Also adding some unit tests to catch these bugs in the future. 

Implementation is mostly made by Claude... 

Big thanks to @CassandraMcGinley for finding this bug and already providing information about the cause!

### DIFF
--- a/.github/workflows/run_tests.yaml
+++ b/.github/workflows/run_tests.yaml
@@ -230,6 +230,11 @@ jobs:
       run: |
         python NuRadioReco/detector/test/test_response.py
 
+    - name: "detector.antennapattern"
+      if: always()
+      run: |
+        python NuRadioReco/detector/test/test_antennapattern.py
+
     - name: "Test unfolding with voltageToEfieldConverter"
       if: always()
       run: |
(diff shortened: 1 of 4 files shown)


fschlueter:
> ## Coverage report
>  
> 
> Click to see where and how coverage changed

I tried to rerun that, that seems to not have worked or not have cause another post... anyway this pr should if all increase the coverage ...

fschlueter:
@philippwindischhofer would be awesome if we can get this merged asap. So that Cassie can continue with the fixed class. The second PR in this stack has lower priority

philippwindischhofer (approved):

PR #1149, opened 2026-08-20
Refactor antennapattern class

fschlueter:
This PR aims at make the class more readable and maintainable, faster, and more secure.

It fixes bugs on the way...

Based on PR #1148 

# Summary:

## Readability:

- store the VEL as a (n_freqs, n_phi, n_theta) cube instead of a flat array with a
  hand-rolled index formula, which replaces the 12 hand-written interpolate_linear
  calls by three collapses of the interpolation cell
- merge interpolate_linear and interpolate_linear_vectorized, they only differed in
  how they handled coincident data points (the latter name is kept as an alias)
- collapse the three identical orientation blocks of AntennaPatternAnalytic into a
  table of per-model defaults, and factor the shared gain -> VEL conversion out of
  its three model branches
- turn the 123k iteration python loop checking the ordering of the antenna file into
  three array comparisons, and document what it does (and does not) check
- fill in the empty docstrings, drop dead debug code

## Speed (antenna response lookup, 2 - 6x faster depending on the number of frequencies):

- build only the on-sky rotation instead of a full radiotools cstrafo, which
  constructs (and inverts) five transformation matrices of which one is used
- buffer the antenna rotation matrix and its inverse, they only depend on the
  orientation of the antenna, of which a detector has very few
- combine the three coordinate rotations into one and apply only the eTheta/ePhi
  block of it, the eR component of the VEL is zero

## Bugfixes:

- AntennaPatternProvider.__init__ reset the buffer of loaded antenna models on every
  instantiation, so a second AntennaPatternProvider() forced a reload from disk
- the 'magphase' interpolation unwrapped the phase along the frequency axis instead
  of between the two data points being interpolated
- an unknown analytic antenna model now raises instead of returning a half
  initialized object

Verified against the previous implementation for 7 antenna models, 3 orientations and
486 directions each (incl. out of band frequencies and out of range directions): the
response agrees to 1e-14 relative.

### DIFF
--- a/NuRadioReco/detector/antennapattern.py
+++ b/NuRadioReco/detector/antennapattern.py
@@ -1740,7 +1740,7 @@ class AntennaPatternAnalytic(AntennaPatternBase):
         logger.error("phase type {} not implemented".format(phase_type))
         raise NotImplementedError("phase type {} not implemented".format(phase_type))
 
-    def _gain_to_vel(self, freq, gain, low_frequency_cutoff=True):
+    def _gain_to_vel(self, freq, gain):
         """
         Convert a gain into a vector effective length, apply the low frequency cutoff
         and normalize to the maximum VEL of the model
@@ -1751,8 +1751,6 @@ class AntennaPatternAnalytic(AntennaPatternBase):
             frequencies at which to evaluate the response
         gain: array of floats
             the gain at those frequencies
-        low_frequency_cutoff: bool
-            if True, suppress the response below ``cutoff_freq`` with a half Hann window
 
         Returns
         -------
@@ -1763,9 +1761,8 @@ class AntennaPatternAnalytic(AntennaPatternBase):
         VEL = np.zeros_like(freq)
         VEL[in_band] = np.sqrt(gain[in_band]) / freq[in_band]
 
-        if low_frequency_cutoff:
-            i_cutoff = np.argmax(freq > self._cutoff_freq)
-            VEL[:i_cutoff] *= hann(2 * i_cutoff)[:i_cutoff]
+        i_cutoff = np.argmax(freq > self._cutoff_freq)
+        VEL[:i_cutoff] *= hann(2 * i_cutoff)[:i_cutoff]
 
         VEL[in_band] *= self._max_VEL / np.max(VEL[in_band])
         return VEL
--- a/NuRadioReco/detector/antennapattern.py
+++ b/NuRadioReco/detector/antennapattern.py
@@ -1565,24 +1565,23 @@ class AntennaPattern(AntennaPatternBase):
             logger.debug("phi bounds {}, {}, {}".format(self.phi_lower_bound, phi, self.phi_upper_bound))
             return np.zeros((2, len(freq)), dtype=complex)
 
+        in_band = ((freq >= self.frequency_lower_bound) & (freq <= self.frequency_upper_bound))
+        if not np.any(in_band):
+            return np.zeros((2, len(freq)), dtype=complex)
+
         # the grids are not necessarily equally spaced (e.g. the theta grid of
         # RNOG_vpol_4inch_center_n1.73), hence look up the bracketing nodes instead of
         # calculating their indices from the grid boundaries
-        i_freq = np.array(get_bracketing_indices(freq, self.frequencies, self._equidistant_freqs))
         i_theta = np.array(get_bracketing_indices(theta, self.theta_angles, self._equidistant_theta))
         i_phi = np.array(get_bracketing_indices(phi, self.phi_angles, self._equidistant_phi))
 
-        # out of band frequencies are interpolated across the whole band and zeroed below
-        out_of_band = (freq < self.frequency_lower_bound) | (freq > self.frequency_upper_bound)
-        i_freq[0, out_of_band] = 0
-        i_freq[1, out_of_band] = self.n_freqs - 1
+        # Collapse the two angular axes first, over the frequency nodes that the request
+        # spans: theta and phi have a single bracket while every requested frequency has its
+        # own, so this interpolates a few hundred grid points instead of 8 per frequency bin.
+        first = max(int(np.searchsorted(self.frequencies, freq[in_band].min(), "right")) - 1, 0)
+        last = min(int(np.searchsorted(self.frequencies, freq[in_band].max(), "left")) + 1, self.n_freqs)
 
         method = self._interpolation_method
-
-        # Collapse the two angular axes first, over the frequency nodes that are actually
-        # needed: theta and phi have a single bracket while every requested frequency has its
-        # own, so this interpolates a few hundred grid points instead of 8 per frequency bin.
-        first, last = int(i_freq.min()), int(i_freq.max()) + 1
         VEL = self.VEL[:, first:last][:, :, i_phi[:, None], i_theta[None, :]]
         VEL = interpolate_linear(
             phi, *self.phi_angles[i_phi], VEL[..., 0, :], VEL[..., 1, :], method)
@@ -1590,12 +1589,36 @@ class AntennaPattern(AntennaPatternBase):
             theta, *self.theta_angles[i_theta], VEL[..., 0], VEL[..., 1], method)
 
         # ... and only then interpolate along the frequency axis, shape (component, n_freq)
-        i_freq -= first
-        VEL = interpolate_linear(
-            freq, *self.frequencies[i_freq + first], VEL[:, i_freq[0]], VEL[:, i_freq[1]], method)
+        VEL = self._interpolate_frequency(freq, self.frequencies[first:last], VEL)
 
-        VEL[:, out_of_band] = 0
+        VEL[:, ~in_band] = 0
         return VEL[0], VEL[1]
+
+    def _interpolate_frequency(self, freq, nodes, VEL):
+        """
+        Interpolate the vector effective length at the frequency nodes onto ``freq``
+
+        Parameters
+        ----------
+        freq: array of floats
+            the requested frequencies
+        nodes: array of floats
+            the frequency nodes VEL is given at
+        VEL: array of complex floats
+            the vector effective length, shape (component, len(nodes))
+
+        Returns
+        -------
+        VEL: array of complex floats
+            the interpolated vector effective length, shape (component, len(freq))
+        """
+        if self._interpolation_method == 'complex':
+            # np.interp does the bracketing internally, in C and for the whole vector at once
+            return np.array([np.interp(freq, nodes, VEL[0]), np.interp(freq, nodes, VEL[1])])
+
+        i_freq = np.array(get_bracketing_indices(freq, nodes, is_equidistant(nodes)))
+        return interpolate_linear(
+            freq, *nodes[i_freq], VEL[:, i_freq[0]], VEL[:, i_freq[1]], self._interpolation_method)
 
 
 class AntennaPatternAnalytic(AntennaPatternBase):
--- a/NuRadioReco/detector/antennapattern.py
+++ b/NuRadioReco/detector/antennapattern.py
@@ -1450,9 +1450,9 @@ class AntennaPatternBase:
 
         This function does three things:
 
-        * `_get_theta_and_phi`: transform the requested signal direction into the coordinate system
+        * ``_get_theta_and_phi``: transform the requested signal direction into the coordinate system
           of the antenna simulation, taking the orientation and rotation of the antenna into account
-        * `_get_antenna_response_vectorized_raw`: interpolate the stored antenna pattern at that
+        * ``_get_antenna_response_vectorized_raw``: interpolate the stored antenna pattern at that
           requested direction (properly transformed) and frequencies (in the native simulated coordinate system)
         * rotate the interpolated vector effective length, so that its two components refer to the
           on-sky unit vectors of the NuRadio coordinate system
--- a/NuRadioReco/detector/antennapattern.py
+++ b/NuRadioReco/detector/antennapattern.py
@@ -1215,7 +1215,7 @@ def get_onsky_rotation(zenith, azimuth):
     """
     Rotation matrix from the ground (x, y, z) into the on-sky (eR, eTheta, ePhi) system
 
-    Equivalent to `radiotools.coordinatesystems.cstrafo` but constructs only the
+    Equivalent to ``radiotools.coordinatesystems.cstrafo`` but constructs only the
     transformation needed here. The matrix is orthonormal, i.e. its transpose describes
     the inverse transformation.
 
--- a/NuRadioReco/detector/antennapattern.py
+++ b/NuRadioReco/detector/antennapattern.py
@@ -1,5 +1,5 @@
 from NuRadioReco.utilities import units, io_utilities
-from radiotools import helper as hp, coordinatesystems as cs
+from radiotools import helper as hp
 
 from scipy import constants
 from scipy.signal.windows import hann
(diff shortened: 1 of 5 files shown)


fschlueter:
@MartinRavn this PR also cleans up a few things you implemented in the analytic antenna pattern. Please have a look.

NuRadioReco/detector/antennapattern.py
```diff
@@ -1817,11 +1775,18 @@ def __new__(cls, *args, **kwargs):
...
+        log_level: int
+            the log level of the antenna pattern logger
         """
         logger.setLevel(log_level)
+
+        # the class is a singleton, only set up the buffer on the very first instantiation
+        # (otherwise every `AntennaPatternProvider()` would discard the loaded patterns)
+        if hasattr(self, "_open_antenna_patterns"):
+            return
+
```
fschlueter, line 1960: MWE to see why that was a bug... 

```
class AntennaPatternProvider(object):
    __instance = None

    def __new__(cls, *args, **kwargs):
        print("Call new")
        if AntennaPatternProvider.__instance is None:
            print("Create new")
            AntennaPatternProvider.__instance = object.__new__(cls)
        return AntennaPatternProvider.__instance

    def __init__(self):
        print("call init")


a = AntennaPatternProvider()
b = AntennaPatternProvider()
```

fschlueter:
Hey @acorstanje I added you as reviewer. I know that you were not happy with the speed of our antenna model class. This PR does increase its speed quite a bit! Furthermore you have experiences in advanced slicing from your interpolator (lol!) :)

acorstanje:
> Hey @acorstanje I added you as reviewer. I know that you were not happy with the speed of our antenna model class. This PR does increase its speed quite a bit! Furthermore you have experiences in advanced slicing from your interpolator (lol!) :)

Hi, looks like a substantial refactor.
It's supposed to go into `hotfix/antennapattern`, so not yet in `develop` or higher. Does that mean it's not usable yet in existing code (with EfieldtoVoltageConverter etc), interface changes or something?
One thing I wonder, I see an explicit interpolation function and computation of nearest neighbor grid points. This seems cumbersome.
Was there a reason not to use e.g. `scipy.interpolate.RegularGridInterpolator`?
It natively works for (ND-)arrays of queries, and the grid points need not be equidistant.
I don't know if the speed would be the same (or better) for this particular use case. It would reduce the code size, relying on maintained libraries.

NuRadioReco/detector/antennapattern.py
```diff
@@ -1232,65 +1211,130 @@ def preprocess_FEKO_mat(path, polarization='X', downscale_freq=1, downscale_zeni
...
+
 class AntennaPatternBase:
     """
     base class of utility class that handles access and buffering to antenna pattern
     """
 
     def _get_antenna_rotation(self, orientation_theta, orientation_phi, rotation_theta, rotation_phi):
         """
+        Rotation from the coordinate system of the antenna simulation into the one of
```
MartinRavn, line 1374: Does this give the Cartesian rotation matrix between the two coordinate systems? I think mentioning this here will make the places where the function is used clearer.

NuRadioReco/detector/antennapattern.py
```diff
@@ -1324,37 +1368,30 @@ def get_antenna_response_vectorized(self, freq, zenith, azimuth, orientation_the
...
-        # As the theta and phi angles are differently defined in WIPLD and NuRadio, also the orientation of the
-        # eTheta and ePhi unit vectors are different.
-        cstrans = cs.cstrafo(zenith=theta, azimuth=phi)
-        V_xyz_raw = cstrans.transform_from_onsky_to_ground(
-            np.array([np.zeros(Vtheta_raw.shape[0]), Vtheta_raw, Vphi_raw]))
+        VEL_theta_raw, VEL_phi_raw = self._get_antenna_response_vectorized_raw(freq, theta, phi)
 
-        rot = self._get_antenna_rotation(
+        # The eTheta and ePhi unit vectors of the antenna simulation and of NuRadio point in
+        # different directions, so rotate the VEL from the one on-sky system into the other
```
MartinRavn: Can you write "from the simulated on-sky system into the NuRadio system"?

NuRadioReco/detector/antennapattern.py
```diff
@@ -1422,205 +1462,131 @@ def __init__(self, antenna_model, path=path_to_antennamodels,
...
+        i_theta = np.array(get_bracketing_indices(theta, self.theta_angles, self._equidistant_theta))
+        i_phi = np.array(get_bracketing_indices(phi, self.phi_angles, self._equidistant_phi))
+
+        # out of band frequencies are interpolated across the whole band and zeroed below
+        out_of_band = (freq < self.frequency_lower_bound) | (freq > self.frequency_upper_bound)
+        i_freq[0, out_of_band] = 0
+        i_freq[1, out_of_band] = self.n_freqs - 1
+
+        # the 8 grid points surrounding the request, shape (component, freq, phi, theta, n_freq)
+        cell = (i_freq[:, None, None, :], i_phi[None, :, None, None], i_theta[None, None, :, None])
```
MartinRavn: Okay, I'm losing track of what is going on here. Is the shape (component, freq, phi, theta, n_freq) correct? what is n_freq? How is i_freq indexed with two ":" and i_phi and i_theta only with one ":"?

NuRadioReco/detector/antennapattern.py
```diff
@@ -1645,169 +1611,161 @@ def __init__(self, antenna_model, cutoff_freq=None, max_VEL=None):
...
-            gain = np.ones_like(freq)
-            gain[fmask] /= np.sqrt(freq[fmask])  # frequency dependent gain fall-off
+        Parameters
+        ----------
+        freq: array of floats
+            frequencies at which to evaluate the response
+        gain: array of floats
+            the gain at those frequencies
+        low_frequency_cutoff: bool
+            if True, suppress the response below ``cutoff_freq`` with a half Hann window
```
MartinRavn: I am not sure it ever makes sense to have `low_frequency_cutoff = False` when we do the normalization after dividing by the frequencies. Then VEL of the 1st frequency bin explodes, which we afterwards normalize to `self._max_VEL`.

NuRadioReco/detector/antennapattern.py
```diff
@@ -1844,17 +1809,33 @@ def load_antenna_pattern(self, name, **kwargs):
             documentation of this class for further information)
```
MartinRavn, line 1980: Can you add a "Returns" section here? Does the function return an instance of "AntennaPatternBase"?

MartinRavn (commented):
I think these changes are great! I finally understand what all the rotations do, and I also noticed that the antenna patterns were surprisingly slow. The refactoring of the analytic antenna models also look good.

fschlueter:
> > Hey @acorstanje I added you as reviewer. I know that you were not happy with the speed of our antenna model class. This PR does increase its speed quite a bit! Furthermore you have experiences in advanced slicing from your interpolator (lol!) :)
> 
> Hi, looks like a substantial refactor. It's supposed to go into `hotfix/antennapattern`, so not yet in `develop` or higher. Does that mean it's not usable yet in existing code (with EfieldtoVoltageConverter etc), interface changes or something? One thing I wonder, I see an explicit interpolation function and computation of nearest neighbor grid points. This seems cumbersome. Was there a reason not to use e.g. `scipy.interpolate.RegularGridInterpolator`? It natively works for (ND-)arrays of queries, and the grid points need not be equidistant. I don't know if the speed would be the same (or better) for this particular use case. It would reduce the code size, relying on maintained libraries.

This PR builds on top of `hotfix/antennapattern`, so first merging `hotfix/antennapattern` (i.e. the related PR) and than this one is the normal order of things, but this PR is "standalone" (including the fix from the other PR) and should not change the API at all. 

I dont think there is a particular reason not to use scipy functions. I know this code if from a very old project of Christian + 10 years probably. So probably historically grown to be not using scipy dependencies... 

Its a good idea to change it if the performance is better/the same.

acorstanje:
> > > Hey @acorstanje I added you as reviewer. I know that you were not happy with the speed of our antenna model class. This PR does increase its speed quite a bit! Furthermore you have experiences in advanced slicing from your interpolator (lol!) :)
> > 
> > 
> > Hi, looks like a substantial refactor. It's supposed to go into `hotfix/antennapattern`, so not yet in `develop` or higher. Does that mean it's not usable yet in existing code (with EfieldtoVoltageConverter etc), interface changes or something? One thing I wonder, I see an explicit interpolation function and computation of nearest neighbor grid points. This seems cumbersome. Was there a reason not to use e.g. `scipy.interpolate.RegularGridInterpolator`? It natively works for (ND-)arrays of queries, and the grid points need not be equidistant. I don't know if the speed would be the same (or better) for this particular use case. It would reduce the code size, relying on maintained libraries.
> 
> This PR builds on top of `hotfix/antennapattern`, so first merging `hotfix/antennapattern` (i.e. the related PR) and than this one is the normal order of things, but this PR is "standalone" (including the fix from the other PR) and should not change the API at all.
> 
> I dont think there is a particular reason not to use scipy functions. I know this code if from a very old project of Christian + 10 years probably. So probably historically grown to be not using scipy dependencies...
> 
> Its a good idea to change it if the performance is better/the same.

I don't think there is an urgent reason to change it to scipy. The vectorizations implemented now should be the main improvement. If there is time it may be helpful.
For now, what is important is if this will run with existing analysis code and produce the same results.
Is there anything in hotfix/antennapattern that would prevent that / require API-related changes?
Then I can test it, and approve if it works (otherwise I could also just take your word, but that's a bit beside the point of reviewing).

fschlueter:
> For now, what is important is if this will run with existing analysis code and produce the same results.
> Is there anything in hotfix/antennapattern that would prevent that / require API-related changes?
> Then I can test it, and approve if it works (otherwise I could also just take your word, but that's a bit beside the point of reviewing).

No! You should be able to run it with your analysis code and should get the same response (modulo numerical errors maybe). Unless your antenna pattern used non-equidistant binning, than the results will change.

NuRadioReco/detector/antennapattern.py
fschlueter: I might be changing this part of the code again but the reason we need the extra n_freq dimension has to do with the fact that we are linearly interpolating between the bracketing indexs of theta, phi, and freq. For for each point in theta or phi we have to indices, for freq which is a vector we have 2 * n_freq indices which is why we need this extra dimension.

NuRadioReco/detector/antennapattern.py
fschlueter: so shall I remove the option all together? That would in principle break the api

MartinRavn: Wasn't `low_frequency_cutoff` introduced in this PR? It is probably best to remove it, since `low_frequency_cutoff = True` will likely lead to an unintended normalization.

NuRadioReco/detector/antennapattern.py
MartinRavn: Okay, I see. So the dimensions are, e.g., (3, 2, 2, 2, 1024), i.e., 8 points (2 * 2 * 2) in 3-dimensional space (freq, theta, phi) for n frequencies. Maybe the confusing part is that i_freq is a 2D array, but i_phi and i_theta only have 2 entries each.
But I don't think there is a need to change this if it works and there isn't an obvious way to make it clearer.

NuRadioReco/detector/antennapattern.py
fschlueter: You are absolutely right - my fold sorry...

fschlueter:
@MartinRavn and @acorstanje thanks for your comments so far. So I continued to (probably over) optimise the class. However, I paid more attention to the readability of the code. Each step has now some in-code doc-strings explaining what happens and the main methods `_get_antenna_response_vectorized_raw` and `get_antenna_response_vectorized` also got some better doc strings. 

As you probably realized are the changes made by heavily relying on AI and when I first pushed that PR I had not gone through all the changes as carefully as I should have. This has now happened! Apologies for that, wont happen again.

In the meantime I also added some reference-value-validation with #1151 which are based on the bug-fixed but non refactored version of the class from #1148. I already merge this branch in this PR to see if the CI is still happy and locally the tests pass! So this gives me confidence that next to my own review the code is correct and does what it should. 

Regarding speed up: The class is now a factor 7-10 times faster depending on the antenna model and orientation (for the interpolation method compex, for magphase that improvement drops by a bit). This is probably not super critical for neutrino simulations but will be a very nice"feature" for SKA I hope.

fschlueter:
The test fails because the used antenna pattern is not stored with perfectly perpendicular orientation and rotation vectors. The new code detects that and corrects for this which does change minimally the antenna pattern model:
```
WARNING - 14:31:52 - NuRadioReco.antennapattern - the orientation stored in RNOG_vpol_v1_n1.73 is 0.0456 deg off perpendicular, correcting it
```
Why is the stored pattern not perfectly perpendicular? Probably storing the real pi/2 rotation angle with too little precision... 

The failed reference value is very little different: 8.25448 (new) to 8.23497 (ref). This is a bug fix rather than a change in the class. I will update the reference!

fschlueter:
Not sure why the coverage is only 85 %, by now I feel quite good about the testing we are doing for that module...

NuRadioReco/detector/antennapattern.py
```diff
@@ -1645,169 +1785,158 @@ def __init__(self, antenna_model, cutoff_freq=None, max_VEL=None):
...
+        return vel_theta, vel_phi
 
 
 class AntennaPatternProvider(object):
+    """
+    Provider class for antenna pattern. The usage of antenna pattern through this class ensures
+    that an antenna pattern is loaded only once into memory which takes a significant time and
+    occupies a significant amount of memory.
+    """
     __instance = None
```
sjoerd-bouma, line 1940: Why not use the existing singleton metaclass?

changelog.txt
```diff
@@ -20,13 +20,43 @@ from the NOAA ETOPO1 dataset via www.opentopodata.org.
...
+  are now part of the buffer key and loading a second variant of the same model is logged.
+- Fixing the 'magphase' interpolation of the antenna response, which unwrapped the phase along the frequency axis instead
+  of between the two data points that are interpolated. The default 'complex' interpolation is unaffected.
 - Fixing the interpolation of antenna models which are not sampled on an equally spaced grid. The node indices were calculated
   from the grid boundaries, which silently picked the wrong nodes and extrapolated. Only RNOG_vpol_4inch_center_n1.73 (used in all
   RNO_G detector descriptions) is affected, where the vector effective length was off by up to 70% at some zenith angles.
+- The AntennaPatternProvider buffer is no longer emptied every time the (singleton) provider is instantiated. Antenna models
+  were reloaded from disk whenever another module created its own provider.
+- The AntennaPatternProvider no longer ignores the keyword arguments (e.g. interpolation_method, max_VEL) when the antenna
+  model is already buffered. They are now part of the buffer key.
```
sjoerd-bouma, line 58: Seems like you've included some changes twice?

fschlueter:
@acorstanje and @MartinRavn can I merge this?

acorstanje:
> @acorstanje and @MartinRavn can I merge this?

I just tried it with my code scripts.
It fails on importing `NuRadioReco.modules.io.coreas.readCoREASShower`
In the directory I see the file is gone in this branch. Is this intentional?

fschlueter:
> > @acorstanje and @MartinRavn can I merge this?
> 
> I just tried it with my code scripts. It fails on importing `NuRadioReco.modules.io.coreas.readCoREASShower` In the directory I see the file is gone in this branch. Is this intentional?

Yes, I renamed (and stripped down) this module to be more general and simpler. The idea is that it returns a coresika_event with all observers in one station/sim_station pair and one needs to implement a specific function to build the desired event/station - structure. An example is added how to do it for the Auger-style. 

https://github.com/nu-radio/NuRadioMC/pull/1095

acorstanje:
> > I just tried it with my code scripts. It fails on importing `NuRadioReco.modules.io.coreas.readCoREASShower` In the directory I see the file is gone in this branch. Is this intentional?
> 
> Yes, I renamed (and stripped down) this module to be more general and simpler. The idea is that it returns a coresika_event with all observers in one station/sim_station pair and one needs to implement a specific function to build the desired event/station - structure. An example is added how to do it for the Auger-style.
> #1095

OK, for this test I don't really need it yet.
However another (minor?) issue is that 
`det.update(time)` no longer works. It says the Detector object has no attribute `update`. 
How should a time be assigned to a detector now?
I guess this was changed earlier and I missed it.

EDIT: this issue arises from using `NuRadioReco.detector.SKA.detector`.
For some reason this class does not inherit from `detector_base` which does have the `update` function.
This is confusing IMO but there is probably a reason.
So this is a different issue (where should the time be set for galactic noise etc?), I'll try and work around now.

acorstanje (approved):
I have run this with some of my analysis scripts using the SKALA4 antenna model, and the results are the same (up to at least the 6 printed digits). So, as far as I'm concerned, it is allright to merge, though of course it is not an exhaustive test.

PR #1151, opened 2026-08-24
Add reference value validation to test_antennapattern

fschlueter:
We might not merge this PR, Its based of #1148 but also contains a commit from #1149. The purpose of this PR is to implement a reference value validation before the larger refactoring of the class in #1149 (to catch eventual issues). 

This PR also adds 1.2 GB of antenna models as CI "dependency". These models are not regularly downloaded as we cache them. However, cached (and zipped reducing their size to ~ 600 MB) they still take space again our 10 GB quota.

We might be hitting this quota with this PR which means that we will need to get rid of some cached entries (github does that for us but this is not efficient). Maybe python versions + linux env for 3.7 and 3.8? @sjoerd-bouma


fschlueter:
If #1149 is merged this one does not need to be merged...

PR #1154, opened 2026-08-28
Lofar lba pipeline

kwat0308:
Issue being addressed:

New construction of the LOFAR LBA Pipeline. 

The pipeline consists of taking raw data / raw CoREAS simulations, transforming them to processed events, which will then be passed into a reconstruction module.

This is a rather big PR, but most changes are self-contained in "LOFAR" specific directories. New changes to general modules / structure outside of LOFAR consists of the following:

- efieldToVoltageConverter : generate Channel objects with group IDs, if the detector object has them
- a new `pipeline/` directory is generated to place pipeline scripts. This can be divided by experiments. 
- a new `LOFAR` directory in `utilities`, which are used for the pipeline and specific modules
- Allowing hybrid showers to be passed through `readCoREASDetector.begin()`, which adds properties of hybrid showers (e.g. a simulated LORA detector) into the Event object.
- new optional installation for the IFT method that is now also implemented (preliminary). with the [niftyre] tag, `jax`, `jaxlib`, and `nifty` will be installed. Unfortunately `jax` and `jaxlib` requires `numpy >=2.0`, which is not compatible with the pip-installable version of radiotools, which uses `np.infty` as opposed to `np.inf`. 

Status:
- data pipeline works from end to end (including IFT reconstruction)
- simulation pipeline works until base pipeline, without IFT reco. Arrival time plots still need to be fixed.
- LORA simulator included (basic, using average angular and core uncertainty)
- LOFAR measured noise adder included (basic, uses average noise trace from many traces)
- other reconstruction methods still need to be added
- the pipeline should be written better, so that it can be used in e.g. a cluster-system on coma

If there is any suggestions for restructuring from the developers, I will be more than happy to hear them :)


kwat0308:
I have also found a bug in the channelMeasuredNoiseAdder (see latest commit), where the traces with the noise were never added into the channel object. I have now fixed this to include such trace.

NuRadioReco/modules/LOFAR/event_gen/CoREASEventGenerator.py
```diff
@@ -0,0 +1,508 @@
...
+        LOGGER.info(f"Processing event {sim_event.get_id()} from CoREAS file {coreas_hdf5_file}")
+        for evt in self.coreas_reader.run(self.detector, None, selected_station_channel_ids=self.selected_station_channel_ids):
+            
+            LOGGER.info(f"Converting electric field to voltage for event {evt.get_id()} at time {START_TIME}")
+            for station in evt.get_stations():
+
+                # set the station time to the start time defined in macros.py
+                # TODO: should this not be random / based on a given observation time? 
+                # it should be read based on LORA triggered event time.
+                station.set_station_time(START_TIME)
```
tjibbe-g, line 361: Probable cause for the weird arrival time plots. As all simulated traces have the same length, setting all start times to the same value will generate plots where each station has the same relative arrival times.

NuRadioReco/modules/LOFAR/event_gen/CoREASEventGenerator.py
```diff
@@ -0,0 +1,508 @@
...
+                if save_debug_plots:
+                    self._save_trace_snapshot(evt, station, output_dir=event_debug_dir, stage="02_efieldToVoltage")
+
+                # resample the trace to that of LOFAR (uses the default if not provided)
+                self.channelResampler.run(evt, station, self.detector)
+                if save_debug_plots:
+                    self._save_trace_snapshot(evt, station, output_dir=event_debug_dir, stage="03_resample")
+
+                # Apply bandpass filter (in principle not needed since bandpass already applied in interpolator)
+                self.channelBandPassFilter.run(evt, station, self.detector, **filter_settings)
```
tjibbe-g, line 378: Maybe I am just out of the loop on this one. But for data events two differing bandpass filters are used. If we want to replicate data as close as possible, why wouldn't we do the same for the simulated event?

NuRadioReco/modules/LOFAR/event_gen/CoREASEventGenerator.py
```diff
@@ -0,0 +1,508 @@
...
+                if save_debug_plots:
+                    self._save_trace_snapshot(evt, station, output_dir=event_debug_dir, stage="04_bandpass")
+
+                # # Add measured noise from LOFAR
+                self.channelMeasuredNoiseAdder.run(evt, station, self.detector)
```
tjibbe-g, line 383: Is it expected that noise is already pre-filtered and therefore it is okay to add it after filtering the trace? Or is it better to filter the combined noise+signal trace? Does that make a difference (I don't know)? Probably best to document the expectations on the noise very well, esp. for the differences when using the noise library vs when not using it. As (correct my if I'm wrong) the noise library is pre-filtered and cleaned of RFI while when just using `channelMeasuredNoiseAdder` this does not have to be the case right?

NuRadioReco/modules/LOFAR/event_gen/CoREASEventGenerator.py
kwat0308: Aha! This makes sense. I am not sure which module / functionality allows the global time to be set, but perhaps omitting this line might mitigate the issue.

NuRadioReco/modules/LOFAR/event_gen/CoREASEventGenerator.py
kwat0308: Yea this makes sense. I can implement the one in data as well.

NuRadioReco/modules/LOFAR/event_gen/CoREASEventGenerator.py
kwat0308: So I didn't generate the noise library so I am not entirely sure if its pre-filtered and RFI cleaned. But it looks like it is already so. I dont think it has to be the case, since the noise adder just adds the noise trace to each channel. However it does resample the trace if the sampling rate from the noise != sampling rate of channels (which I assume that the noise sampling rate == 200 MHz, i.e. already pre-filtered). 

Since @karenterveer you generated the noise library maybe you can provide more insight into this?  If we are not sure, I am sure we can regenerate the noise library, confirming and using all the relevant steps.

karenterveer: The noise library is RFI cleaned and filtered. You could of course consider using non-RFI cleaned noise and running the simulated traces through the RFI cleaner as well to make the simulations even more realistic. I guess since we build this from scratch we might as well do it as realistic as possible.

NuRadioReco/utilities/LOFAR/macros.py
tjibbe-g, line 1: I would rename this to constants as this file contains LOFAR-specific constants, not macros. This is then also in line with the capital naming scheme used only for constants that is widely accepted and is outlined in PEP-8.

sjoerd-bouma: I probably won't have time to review this PR properly until after the RNO-G collaboration meeting (i.e. 2 weeks from now), but from looking at this file these are mostly settings that could be simply default parameter values in the relevant classes, or a 'config' for which e.g. json makes more sense than a python file, in my opinion.

NuRadioReco/pipeline/LOFAR/README.md
```diff
@@ -0,0 +1,91 @@
...
+### General TODOS
+- [x] include basic simulation pipeline
+- [x] verify basic IFT reconstruction pipeline -> verified that it runs, but final reconstruction not checked
+- [ ] include calculation of signal and electric field fluence to sim
+- [ ] include requirements & documentation for using the pipeline -> added optional dependencies for niftyre for now, works with latest versions
+- [ ] include CoREAS fluence reconstruction pipeline
+- [ ] verify that round trip test works
+- [ ] include validation tests -> should be place in `tests/LOFAR`
+- [ ] simple CI/CD pipeline (soft, not enforced, but can be ran to verify installation works)
+- [ ] better data structure to be better integrated with NuRadio format? (e.g. where to place the pipeline, now its in a not-so-good location)
```
tjibbe-g, line 62: Maybe put it in the `modules.LOFAR` folder? It is then also close to all the reconstruction modules. Especially when more modules are implemented, having the pipeline using the reconstruction modules near those same reconstruction modules feels logical to me at least. But maybe a better question for one of the core devs.

NuRadioReco/modules/measured_noise/LOFAR/noiseLibraryConverter.py
```diff
@@ -0,0 +1,128 @@
...
+        if nur_filename == None:
+            temp_dir = tempfile.mkdtemp()
+            self.__nur_filepath = os.path.join(temp_dir, "LOFAR_noise_trace.nur")
+        else:
+            self.__nur_filepath = nur_filename
+        logger.info(f"Creating the noise .nur file into directory {self.__nur_filepath}")
+        
+
+    @register_run()
+    def run(self, event, det):
```
tjibbe-g, line 89: Generating only a single station with a single trace will add the exact same noise to all channels in the simulation (also seen in one of the plots in the slides you sent (attached)). This loses part of the randomness of the noise and will render beamforming methods impossible as noise will add up coherently. 

A better method could be to create a `.nur` file containing all applicable traces where each trace has a random order/part of the noise (blocks) contained in Karen's library. This way you have reproducible noise for events but still keep (more of) the randomness of the noise intact. But other (better) methods probably are also possible. 

Image

tjibbe-g: Just a thought, but could this also propagate into the fluences and cause the weird things seen in those plots? I can imagine that when constructing efields this causes problems as the distinction between signal and noise is less clear. Added the plot for reference

Image

NuRadioReco/pipeline/LOFAR/README.md
kwat0308: I would actually separate it, since the pipeline, which **uses** the modules, are different from modules themselves. So the pipeline is not a module, and thus personally I do not think it should live here. But its a better question for the core devs indeed.

NuRadioReco/modules/measured_noise/LOFAR/noiseLibraryConverter.py
kwat0308: I agree, currently I was just trying to implement it. I think the logical step would be to add a random part of the trace as you say to incur some randomness, and in the most ideal case we should actually get the noise traces per event (i.e. varying per event per station per channel), however this makes the pipeline slower (we can of course pre-generate and store the .nur files though).

NuRadioReco/utilities/LOFAR/macros.py
```diff
@@ -0,0 +1,37 @@
...
+
+# some hard coded values in the base code, can be directly modified here
+DATA_TRACE_LENGTH = 65536
+RFI_CLEANING_TRACE_LENGTH = 8192
+CR_SNR = 6.5
+PASS_BAND = (30, 80)
+START_TIME = datetime.datetime(2012, 10, 1, 0, 0)
+# NRR channel IDs that are always dropped at read-in, regardless of the flagging
+# done by the reader. 3002019 is a permanently broken dipole of CS003.
+ALWAYS_REMOVED_CHANNEL_IDS = (3002019,)
```
karenterveer: For now I would remove this specific channel from the always_removed_channel_ids, I think it only broke after some time, so there are a bunch of events in which it is still intact. At least the IFT method is pretty good at ignoring it when it is broken, and I think the specific reconstruction methods should decide whether a channel like this needs to be removed or not. Also, I know it was me who originally added this to the always removed channels, my bad

NuRadioReco/modules/LOFAR/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .event_gen.dataEventGenerator import dataEventGenerator
```
karenterveer, line 1: so, maybe not a huge deal, but I am noticing that the data pipeline, like this, cannot run without the cr-pulse-interpolator installed. I guess everyone who runs this likely has it installed anyways, but technically the data pipeline can and should be able to run without it. Could the two pipelines maybe import their modules explicitly instead?

NuRadioReco/modules/LOFAR/event_gen/dataEventGenerator.py
```diff
@@ -0,0 +1,305 @@
...
+        output_directory: str
+            The directory where the output files, such as the debug plots and resulting nur files, will be saved. If None, no output files will be saved.
+        """
+
+        self.tbb_directory = tbb_directory
+        self.json_directory = json_directory
+        self.meta_data_directory = meta_data_directory
+        self.block_number_file = block_number_file
+        self.output_directory = output_directory
+        os.makedirs(self.output_directory, exist_ok=True)
```
karenterveer, line 81: Currently the default is "None", but os.makedirs(None) raises an error.

karenterveer:
Another issue I found is that currently the event created in the simulation pipeline has no shower object. This does not affect the reconstruction but it means that no MC truths are stored in the event, which we might wanna do

tjibbe-g:
> Another issue I found is that currently the event created in the simulation pipeline has no shower object. This does not affect the reconstruction but it means that no MC truths are stored in the event, which we might wanna do

Aren't they stored as a `SimShower` object in the event as the simulation pipeline uses `read_CORSIKA7()` under the hood? And then the `Shower` object is used for reconstructed values?

karenterveer:
We also need to be careful about the GDAS atmosphere tool. I could never get it to work whenever I e.g. sent a job array to the cluster as most of the compute nodes could not access the server to fetch per-event atmospheres (did not test for coma, though). Currently, the reconstruction pipeline then falls back to US standard atmosphere, which will cause a pretty serious Xmax bias (for LOFAR simulations using US standard gives an Xmax bias of -10g/cm^2). We need to ensure that we handle this properly with warnings. Ideally I think we should generate our own fallback atmosphere, which should be a mean LOFAR site atmosphere. I have this on my to-do list, but wanted to comment so it's on peoples radar.

karenterveer:
> > Another issue I found is that currently the event created in the simulation pipeline has no shower object. This does not affect the reconstruction but it means that no MC truths are stored in the event, which we might wanna do
> 
> Aren't they stored as a `SimShower` object in the event as the simulation pipeline uses `read_CORSIKA7()` under the hood? And then the `Shower` object is used for reconstructed values?

Oops, my bad, yeah, I got some error and thought that was the reason, but it turned out that I was the reason!

kwat0308:
Its true that the MC parameters are stored as SimShowers, but the Radio-relevant quantities are not stored as a RadioShower (as far as I remember), which is so in the data pipeline.

So this needs to be included at some point if we want absolute identically running pipelines.

karenterveer:
Ok, I have now updated the IFT reconstruction to match my analysis. I tested it locally and it reproduces my results. I also fixed some things in the simulation pipeline, like the noise adding (now it does not add the same noise to all antennas any more, and the core guess is an actual core guess on top of the randomized injected true core). @kwat0308 could you verify everything works on your side? From my side this is now ready to be merged into the original pipeline branch, unless someone finds something.

karenterveer:
I also updated our interactive example script. In the end it now uses the pipeline module directly and includes a quick example reconstruction. This will make testing easier too

NuRadioReco/modules/measured_noise/LOFAR/noiseLibraryConverter.py
kwat0308: With @karenterveer 's commit the different windows of the noise, randomly drawn, is added to each channel trace. 
So for now we can resolve this comment, unless there is a need to have an event-specific noise.

NuRadioReco/modules/LOFAR/event_gen/CoREASEventGenerator.py
kwat0308: Adding / disabling this actually had no effect in the overall plotting routine of the pipeline visualiser. However the arrival times seem to be set properly since the timing model works fine in the IFT reconstruction. 

So I think its a matter of communicating the proper arrival time used in the pipeline visualiser (I think. It still warrants a proper investigation). 

Since its most likely a visualisation issue, I will resolve this now and we will reinvestigate this after a proper PR to the develop branch.

NuRadioReco/modules/LOFAR/event_gen/CoREASEventGenerator.py
kwat0308: I think for now its alright that we can resolve the issue and move this for the general PR. I will add it to the existing TODO list.

NuRadioReco/modules/LOFAR/__init__.py
kwat0308: Okay! We can detach them, since indeed there need not be such a dependency if one does not need it.

NuRadioReco/utilities/LOFAR/constants.py
kwat0308: I think I will leave it as a python file, but rename it to constants.py for now. The point for this is to have a single file that contains all relevant pipeline-related parameters, so that in this way one does not need to search for different classes / source codes to find the default values. But making it a json file requires better I/O management of the constants, which is more tedious.

kwat0308:
I have now pushed all major changes that need to be pushed, including:

- renaming macros.py -> constants.py
- adding the bandpass as applied with the data 
- importing the event processors with the whole module path

 I also verified that both data & simulation pipeline that Karen implemented works :D 

Here is an example of the result of the data pipeline:
reco_plot_134522987

And here is an example of the result of the simulation pipeline:
reco_plot_134522987-2

I think we can safely merge this into the `lofar-lba-pipeline` branch for now, and make a new PR with the `develop` branch.

PR #1157, opened 2026-09-12
fix #1156: wrong reference channel if station does not contain all channels in detector description

### LINKED ISSUE #1156: voltageToEfieldConverters reference channel is taken from detector, not station
Lines 67+ of the voltageToEfieldConverterPerChannelGroup, and also in voltageToEfieldConverterPerChannel (line 59+):
```
use_channels = det.get_channel_ids(station.get_id())
frequencies = station.get_channel(use_channels[0]).get_frequencies()  # assuming that all channels have the  same sampling rate and length
```

If the channel [0] of the detector does not exist in the event (this happens for example in LOFAR events when that channel was flagged during read-in and therefore removed) a KeyError is raised.

My suggested fix is taking the reference channel from the station directly to make sure it exists.

sjoerd-bouma:
Closes #1156


karenterveer (approved):
that was quick!

PR #1160, opened 2026-09-16
Lofar lba pipeline

### LINKED ISSUE #1127: stationGalacticCalibrator applying relative calibration to wrong channel orientations
While cross-checking the LOFAR Galactic calibration against sky-noise sims, I found that the per-polarisation Fourier coefficients of the relative calibration are apparently currently applied to the wrong dipole: the phi=225° dipoles get the curve fitted for phi=135° and vice versa. Calibrating with only the absolute calibration curve C(v) (relative X/Y orientation dependent step skipped), each dipole set's observed sidereal modulation matches the Fourier curve the pipeline applies to the other set.

After fixing it and re-exporting, data and simulated noise agree at data/sim ≈ 0.9–1.1 in both dipoles (before it was ~0.7 for the phi=135° dipoles and ~1.3 for phi=225°).

The fix is swapping two numbers in `__get_channel_polarisation()` in `NuRadioReco/modules/LOFAR/stationGalacticCalibrator.py`:

```
if orientation == 225.0:
    channel_polarisation = 1   # -->  0
elif orientation == 135.0:
    channel_polarisation = 0   # -->  1
```

The plot I attached shows this. I triple checked whether this mix-up of dipoles happened anywhere else in my galactic noise simulation pipeline but since it is all done with NuRadio (all based on the LOFAR.json) it is self-consistent. I am not sure whether this is something that happened during the port of the LOFAR pipeline to NuRadio or already before, but personally I have no way of checking. I am happy to start a PR to fix this but it might be good if someone independently checks this first.

Image

kwat0308:
Issue being addressed:

New construction of the LOFAR LBA Pipeline.

The pipeline consists of taking raw data / raw CoREAS simulations, transforming them to processed events, which will then be passed into a reconstruction module.

This is a rather big PR, but most changes are self-contained in "LOFAR" specific directories. New changes to general modules / structure outside of LOFAR consists of the following:

- efieldToVoltageConverter : generate Channel objects with group IDs, if the detector object has them
- a new pipeline/ directory is generated to place pipeline scripts. This can be divided by experiments.
- a new LOFAR directory in utilities, which are used for the pipeline and specific modules
- Allowing hybrid showers to be passed through readCoREASDetector.begin(), which adds properties of hybrid showers (e.g. a simulated LORA detector) into the Event object.
- new optional installation for the IFT method that is now also implemented (preliminary). with the [niftyre] tag, jax, jaxlib, and nifty will be installed. Unfortunately jax and jaxlib requires numpy >=2.0, which is not compatible with the pip-installable version of radiotools, which uses np.infty as opposed to np.inf.
- fixed a bug for the channelMeasuredNoiseAdder where the noise trace was not saved to the channel object

There are still some major TODOs left before we plan to merge:
- [ ] use pre-RFI cleaned noise instead of RFI cleaned noise for adding noise
- [ ] Include a timing offset adder, u̶s̶i̶n̶g̶ t̶h̶e̶ s̶t̶a̶t̶i̶o̶n̶ c̶l̶o̶c̶k̶ d̶e̶l̶a̶y̶s̶ i̶n̶ `̶i̶o̶.L̶O̶F̶A̶R̶.r̶e̶a̶d̶L̶O̶F̶A̶R̶D̶a̶t̶a̶`̶ i̶f̶ a̶v̶a̶i̶l̶a̶b̶l̶e̶ o̶r̶  scatter based on the timing precision between stations and antenna pairs as given in https://arxiv.org/abs/1603.08354 . 
- [ ] Include fluence plots & simulated truths for the pipeline visualiser
- [ ] Include validation tests to verify the data and simulation pipeline, using existing data / simulation in the desy server
- [x] Include bugfix of stationGalacticCalibrator (#1127 ) (already done)
- [ ] Add description of new modules and pipeline into the documentation.

If there is any suggestions for restructuring from the developers, I will be more than happy to hear them :)


karenterveer:
Note on adding the clock delays (@tjibbe-g ) - Adding the ones from the reader does not really make sense on second thought. The reader removes known station offsets, so it would not make sense to ADD them in the simulation pipeline. We should just stick to adding random offsets according to the known precisions from the paper.

